PR #53, opened 2021-02-04
Feature/spatial types

darrellwarde:
Spatial types have been added to `@neo4j/graphql`. They allow for the query and mutation of nodes containing spatial values, and querying with the following advanced filters which compare a specified point and distance to points stored in the database:

- `LT`
- `LTE`
- `GT`
- `GTE`
- `DISTANCE` (exact distance match)

In addition to the introduction of spatial types, this branch also addresses the following:

- The `IN` and `NOT_IN` filters now behave differently based on whether the field being queried is or isn't an array. For a non-array, the filter will accept an array argument and check for the existence of the field within it. For an array field, the filter will accept a single value and check whether it exists within the field in the database.
- Sorting arguments are no longer added for array fields. Fixed a resultant bug where schema generation would fail if a type was added with just a single array field.
- Pruned which advanced filters are available for which type. Enum comparisons are now done against enums as opposed to strings, DateTime has had a couple of operators added.
- Throw an error during schema generation is creating a 2-dimensional or greater array is attempted. This is not support behaviour.

There are likely additional features and bug fixed which I have forgotten, and I shall update the above list as I review my own code!

Apologies that the commits don't make an awful lot of sense, this will be hard to review on a commit-by-commit basis.

### DIFF
--- a/packages/graphql/tests/integration/point.int.test.ts
+++ b/packages/graphql/tests/integration/point.int.test.ts
@@ -22,7 +22,7 @@ describe("Point", () => {
                 location: Point!
             }
         `;
-        const neoSchema = makeAugmentedSchema({ typeDefs, debug: true });
+        const neoSchema = makeAugmentedSchema({ typeDefs });
         server = constructTestServer(neoSchema, driver);
     });
 
--- a/packages/graphql/tests/integration/points-cartesian.int.test.ts
+++ b/packages/graphql/tests/integration/points-cartesian.int.test.ts
@@ -20,7 +20,7 @@ describe("[CartesianPoint]", () => {
                 locations: [CartesianPoint!]!
             }
         `;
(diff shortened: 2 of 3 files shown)


nodkz:
https://github.com/neo4j/graphql/blob/da7df043cf7cc345afa46cd06ca0657fb4af8e52/packages/graphql/src/schema/make-augmented-schema.ts#L313-L318

@darrellwarde, could you please open an issue with a small reproducible case of this issue?
Or even more open PR with a broken test case in this folder https://github.com/graphql-compose/graphql-compose/tree/master/src/__tests__/github_issues 

A good template for your test case can be this suite https://github.com/graphql-compose/graphql-compose/blob/master/src/__tests__/github_issues/312-test.ts

Tnx.

PR #134, opened 2021-04-08
Test/add-docker-for-testing

danstarns:
Adds docker file and a compose for testing. Make sure you turn off your local neo4j instance and run; 

```
docker-compose -f ./docker-compose.test.yml up --build --abort-on-container-exit --exit-code-from test
```

Did it work?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,0 +1,20 @@
+FROM node:14.16
+
+WORKDIR /app
+
+RUN npm install -g --force yarn
+
+COPY ./package.json ./package.json
+
+COPY ./ ./
+
+WORKDIR /app/examples/neo-push/server
+
+# Invalid ELF header, 
+# Ensures we install bcrypt based on this machine
+RUN yarn add bcrypt
+
+WORKDIR /app
+RUN yarn install
+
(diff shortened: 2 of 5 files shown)


darrellwarde:
No 🤣  Let's not rush something like this in - I had visions of automating the Docker up/down as part of the setup/teardown scripts of Jest, would be nice to have the familiarity of `yarn test` with lots of fanciness happening under the hood.

danstarns:
@darrellwarde Did you get an error?

darrellwarde:
> @darrellwarde Did you get an error?

I did indeed:

```
Unexpected API error for graphql_neo4j_1 (HTTP code 502)
Response body:
Bad response from Docker engine
```

danstarns:
@darrellwarde  Seems to work on windows just fine 😂 Be keen to see if you get any more error than just that? With your fancy visions would everyone need docker installed?

darrellwarde:
> @darrellwarde Seems to work on windows just fine 😂 Be keen to see if you get any more error than just that? With your fancy visions would everyone need docker installed?

Sadly that's literally all I get... 😞  That is true regarding everyone needing Docker installed, I guess I just have dreams of consistent environments everywhere!

danstarns:
@darrellwarde We should all develop on a raspberry PI job sorted! It would be nice to offer docker for contributors... after all docker is designed to be the consistent environment. Keen to see if it works for @oskarhane

PR #193, opened 2021-05-06
RFC - Relationship Properties

darrellwarde:
# Description

This branch contains the up-to-date plan for introducing relationship properties into the Neo4j GraphQL Library.

# Checklist


docs/rfcs/00-relationship-properties/relationship-properties.md
```diff
@@ -0,0 +1,341 @@
...
+
+For the movie "Forrest Gump", match all actors whose name begins with "Tom", and return their name and screenTime, all ordered by their name ascending.
+
+```graphql
+query {
+    movies(where: { title: "Forrest Gump" }) {
+        title
+        actorsConnection(
+            where: { node: { name_STARTS_WITH: "Tom" } }
+            sort: { node: { name: ASC } }
```
colinskow, line 248: Should we not use `options` here to keep it consistent with filtering elsewhere? Also when pagination is added then it won't need to be a breaking change.

darrellwarde: Compared to the page-based pagination on the current relationship fields, the arguments for cursor-based pagination (when we eventually implement it) will be based on the GraphQL Cursor Connections Specification, which has the pagination arguments at the root. 

Because of this, the `options` argument would then only contain `sort`, hence why I moved it to the top-level. However, for consistencies sake as you mention, it is probably worthwhile reintroducing the `options` argument here, and it will also allow for additional options in future if we desire.

docs/rfcs/00-relationship-properties/relationship-properties.md
```diff
@@ -0,0 +1,341 @@
...
+
+input ActedInSort {
+    screenTime: SortDirection
+}
+```
+
+And for composites of both node and relationship properties:
+
+```graphql
+input ActorMovieWhere {
```
smkhalsa, line 103: How would the auto-generated input names work if there are multiple distinct relationships between the same types, but each has a different set of properties?

For example:

```graphql
type Person {
    name: String!
    friends: [Person!]!
        @relationship(type: "HAS_FRIEND", properties: "HasFriend", direction: IN)
    enemies: [Person!]!
        @relationship(type: "HAS_ENEMY", properties: "HasEnemy", direction: IN)
}

interface HasFriend {
    since: DateTime!
    secretHandshake: String
}

interface HasEnemy {
    since: DateTime!
}
```

In this case, it's not clear if `PersonPersonWhere.relationship` should be `HasFriend` or `HasEnemy`.

darrellwarde: Very small change made to use the field name in that type instead of the type name, so in this case there would be `PersonFriendsWhere` and `PersonEnemiesWhere` containing the same "node where" but different "relationship where".

smkhalsa: I agree that this probably won't be an issue for the vast majority of users, and I think your updated approach is a reasonable default.

That said, since this package is generating schema to be merged with user-defined schema, I'd take it as a given that naming collisions will come up (whether caused by this issue or another).

For example:

```graphql
type User {
    name: String!
    status: UserStatus!
        @relationship(type: "HAS_STATUS", properties: "HasStatus", direction: IN)
}

interface HasStatus {
    since: DateTime!
}

type UserStatus {
    online: Boolean!
    message: String
}
```

This will cause a collision on the `UserStatusWhere` type.

I'm not sure it's possible to eliminate all possible collisions, so I'd suggest simply implementing a consistent and predictable way of handling them.

darrellwarde: Seems you're one step ahead of me every step of the way here! 😉 

These are so edge-casey, but you're absolutely right that we should be doing our absolute best to deconflict well in our solution.

Perhaps we need to be classifying `Where` input types by what they apply to, so in the above example:

- `UserWhere` and `UserStatusWhere` - no need to break these for everyone
- `HasStatusRelationshipWhere` - for matching on relationship properties
- `UserStatusConnectionWhere` - composing the above two together

Yes, slightly longer input type names, but should massively reduce naming collisions.

smkhalsa: > Perhaps we need to be classifying Where input types by what they apply to

Yes, I think that makes sense.

I might go with just `HasStatusWhere` instead of `HasStatusRelationshipWhere` though.

smkhalsa: Nevermind, I see why that doesn't work

smkhalsa: Actually, it does. Sorry.

With the change to `UserStatusConnectionWhere`, we should be able to do `HasStatusWhere` instead of `HasStatusRelationshipWhere`

darrellwarde: Ahaha, it's a bit of a mindwarp right, but you're right, changing to `*ConnectionWhere` automatically deconflicts anything which would have clashed with `*RelationshipWhere`.

Did just double check my sanity to see if you can name an interface the same as a type and indeed you can't, so `HasStatusWhere` and `UserStatusConnectionWhere` it is!

Thanks for working this through with me, will update the RFC either over the weekend or on Monday. 🙂

danstarns (approved):

docs/rfcs/00-relationship-properties/relationship-properties.md
```diff
@@ -0,0 +1,341 @@
...
+
+### Example Queries
+
+#### With `where` and `sort` arguments
+
+For the movie "Forrest Gump", match all actors whose name begins with "Tom", and return their name and screenTime, all ordered by their name ascending.
+
+```graphql
+query {
+    movies(where: { title: "Forrest Gump" }) {
```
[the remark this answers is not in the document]
darrellwarde: In my opinion, no - I think you should only be able to filter on what can be returned from the field... Leaving this alone also means less disruption for users who don't care so much about relationship properties.

udaykor:
Will this PR help me use a type definition as declared below

```
type ActorsMovieActedIn @relation(name: "ACTED_IN", direction: OUT) @auth(rules: [{ operations: [CREATE, READ, UPDATE, DELETE], isAuthenticated: true }]) {
    from: Actors
    to: Movies
    actedIn: Boolean
    createdAt: DateTime! @timestamp(operations: [CREATE])
    updatedAt: DateTime! @timestamp
  }
```

darrellwarde:
> Will this PR help me use a type definition as declared below
> 
> ```
> type ActorsMovieActedIn @relation(name: "ACTED_IN", direction: OUT) @auth(rules: [{ operations: [CREATE, READ, UPDATE, DELETE], isAuthenticated: true }]) {
>     from: Actors
>     to: Movies
>     actedIn: Boolean
>     createdAt: DateTime! @timestamp(operations: [CREATE])
>     updatedAt: DateTime! @timestamp
>   }
> ```

Yes and no - it will help you achieve the same functionality, but the syntax will be different! 🙂

udaykor:
> > Will this PR help me use a type definition as declared below
> > ```
> > type ActorsMovieActedIn @relation(name: "ACTED_IN", direction: OUT) @auth(rules: [{ operations: [CREATE, READ, UPDATE, DELETE], isAuthenticated: true }]) {
> >     from: Actors
> >     to: Movies
> >     actedIn: Boolean
> >     createdAt: DateTime! @timestamp(operations: [CREATE])
> >     updatedAt: DateTime! @timestamp
> >   }
> > ```
> 
> Yes and no - it will help you achieve the same functionality, but the syntax will be different! 🙂

I do not mind the syntax change, I care about the relationship, thanks mate :) I am waiting for this to be available to make full migration complete

darrellwarde:
> > > Will this PR help me use a type definition as declared below
> > > ```
> > > type ActorsMovieActedIn @relation(name: "ACTED_IN", direction: OUT) @auth(rules: [{ operations: [CREATE, READ, UPDATE, DELETE], isAuthenticated: true }]) {
> > >     from: Actors
> > >     to: Movies
> > >     actedIn: Boolean
> > >     createdAt: DateTime! @timestamp(operations: [CREATE])
> > >     updatedAt: DateTime! @timestamp
> > >   }
> > > ```
> > 
> > 
> > Yes and no - it will help you achieve the same functionality, but the syntax will be different! 🙂
> 
> I do not mind the syntax change, I care about the relationship, thanks mate :) I am waiting for this to be available to make full my migration to @noe4j/graphql complete. I think you are referring to the `@relation` to be `@relationship`

As per this RFC, there will be a few more changes than just that:

```graphql
type Actor {
  movies: [Movie!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: OUT)
}

type Movie {
  actors: [Actor!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: IN)
}

interface ActedIn {
  actedIn: Boolean
  createdAt: DateTime! @timestamp(operations: [CREATE])
  updatedAt: DateTime! @timestamp
}
```

The above is roughly what the type you described will end up looking like. The work is well underway, and we will be considering prereleases for users to try out and give feedback on soon. 🙂

udaykor:
I love this new library overall, So much cleaner and the neo4j docs on migration steps were just perfect. I love this interface approach. This is going to clean my schema by approximately 14%

darrellwarde:
Hey everyone! You may or may not have watched my talk at NODES 2021, "Neo4j and GraphQL: The Past, The Present and The Future", but I an exciting announcement!

Version 2.0.0-alpha.1 of The Neo4j GraphQL Library is now available for you to try out, with support for relationship properties!

You can install it using:

```bash
npm install @neo4j/graphql@next

# OGM
npm install @neo4j/graphql-ogm@next
```

You can find a migration guide to upgrade to the new version at https://neo4j.com/docs/graphql-manual/2.0/guides/rel-migration/, with most of the changes needed being on the client-side.

Please, try it out, let us know how it's working for you by raising any issues for bugs or feature requests! Please note that this is an alpha release, so some things will be broken and it may not be feature complete quite yet.

Issue #190, opened 2021-05-05
Why ALL is used instead of ANY when converting where query filters to cypher?


morningcloud:
**Describe the bug**
I have a use case where one node (e.g. User) is linked to multiple nodes of the same label (e.g. User Demographics), each User can be linked to one or more different types of demographics depending on the client the users are under. I wanted to write a query to fetch users that have specific demographic data. But the results I am getting for the related queries is not expected. Below explains the use case and the result I am getting and what I actually expected.

**Type definitions**
```
type User {
    client_id: String
    uid: String
    demographics: [UserDemographics] @relationship(type: "HAS_DEMOGRAPHIC", direction: OUT)
}

type UserDemographics {
    client_id: String,
    type: String,
    value: String,
    users: [User] @relationship(type: "HAS_DEMOGRAPHIC", direction: IN)
    }
```
![image](https://user-images.githubusercontent.com/17576651/117159192-6e15f600-ae03-11eb-82a1-7913e01a9bcb.png)

**To Reproduce**

1. Start neo4j & graphql server
2. Execute the following Cypher queries to populate neo4j with sample data:
```
Create (:User {uid: "user1"}),(:User {uid: "user2"}),(:UserDemographics{type:"Gender",value:"Female"}),(:UserDemographics{type:"Gender",value:"Male"}),(:UserDemographics{type:"Age",value:"50+"}),(:UserDemographics{type:"State",value:"VIC"})
MATCH (u:User {uid: 'user1'}) MATCH (d:UserDemographics {type: 'Gender', value:'Female'}) create (u)-[:HAS_DEMOGRAPHIC]->(d)
MATCH (u:User {uid: 'user2'}) MATCH (d:UserDemographics {type: 'Gender', value:'Male'}) create (u)-[:HAS_DEMOGRAPHIC]->(d)
MATCH (u:User {uid: 'user1'}) MATCH (d:UserDemographics {type: 'Age', value:'50+'}) create (u)-[:HAS_DEMOGRAPHIC]->(d)
MATCH (u:User {uid: 'user2'}) MATCH (d:UserDemographics {type: 'Age', value:'50+'}) create (u)-[:HAS_DEMOGRAPHIC]->(d)
MATCH (u:User {uid: 'user1'}) MATCH (d:UserDemographics {type: 'State', value:'VIC'}) create (u)-[:HAS_DEMOGRAPHIC]->(d)
[...]

PR #196, opened 2021-05-07
Replace ALL with ANY in whilst querying relationship equality

darrellwarde:
# Description

Fixes Cypher generation when checking for equality of fields on related nodes, replacing `ALL` with `ANY` to return expected results. See linked issue for example.

`ALL` still used for `_IN` and `_NOT_IN` filters, which _I believe_ is the correct behaviour.

# Issue

#190 

# Checklist

### DIFF
--- a/packages/graphql/src/translate/create-where-and-params.ts
+++ b/packages/graphql/src/translate/create-where-and-params.ts
@@ -286,7 +286,7 @@ function createWhereAndParams({
 
             let resultStr = [
                 `EXISTS((${varName})${inStr}${relTypeStr}${outStr}(:${equalityRelation.typeMeta.name}))`,
-                `AND ALL(${param} IN [(${varName})${inStr}${relTypeStr}${outStr}(${param}:${equalityRelation.typeMeta.name}) | ${param}] INNER_WHERE `,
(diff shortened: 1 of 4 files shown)


johnymontana:
I think the `ANY` is the right approach here. Using the movie example from the tests - when filtering for movies that are in the "Action" genre, you will have movies connected to multiple genre nodes and you only want to apply a predicate that says the movie has at least one relationship to the "Action" genre node, not that all genre relationships are to the "Action" genre node. 

An alternative would be to create additional where filtering inputs such as `genres_some`, `genres_every`, `genres_single`, etc that map to the [various Cypher list predicate functions.](https://neo4j.com/docs/cypher-manual/4.2/functions/predicate/) I don't think exposing each of these list predicates is the approach to take here though as I think the vast majority of use cases users are wanting the `ANY` logic applied and the additional list predicate options add confusion and leak Cypher specific concepts to the GraphQL layer.

oskarhane (approved):
Looking at the integration tests, this seems like the right thing to do.
I think Wills point on creating more fine grained filters if requested sounds sensible, while keeping this as a default.

🌻

Issue #208, opened 2021-05-16
LIMIT: Invalid input. '10.0' is not a valid value. Must be a non-negative integer.",


aspectgfg:
**Describe the bug**
Simple query on a model with a limit or skip parameter in the gql query

```
query Patients($where: PatientWhere!, $options: PatientOptions) {
  patients(where: $where, options: $options) {
    id
    firstName
    middleName
    lastName
  }
}
```

```
{
	"where": {},
	"options": {
                 "limit": 10,
		"sort": [
			{
				"lastName": "ASC"
			}
		]
	}
}
```

**Expected behavior**
Skip and limit to work 

**System (please complete the following information):**
"@neo4j/graphql": "^1.0.1"
"@neo4j/graphql-ogm": "^1.0.1"
"neo4j-driver": "^4.2.3",
Neo4J: Version 1.4.3 (1.4.3.49)

**Additional context**
After editing this line https://github.com/neo4j/graphql/blob/eb4d0c11a183f01b18f652a99319d6701c3e69f3/packages/graphql/src/translate/translate-read.ts#L113
to be 
```
cypherParams[`${varName}_limit`] = neo4j.int(optionsInput.limit)
```
The query successfully runs

darrellwarde:
This is sadly not just a single-fix solution - this is linked to both #186 and #167. As such, sadly it is not such a simple solution, but we are on it and will be addressing this ASAP!

oskarhane:
@aspectgfg Hi, to make sure we cover all the cases and setups when looking into this, can I ask you to create a repo or example setup where you see this behavior?
Out of the box without any other dependencies this is not an issue, that's why it's important to us to see what the different setups where this happens looks like.

darrellwarde:
A fix (hopefully!) for this has now been merged, and will be released shortly (aiming for tomorrow). Please do test and feel free to reopen this ticket if it hasn't addressed your problems! 🙂

darrellwarde:
After much back and forth of thinking we had fixed this bug, it should finally be _fully_ fixed by #297. 🤞  Apologies it took a while to get to this stage, it was a tricky bug to find but then very clear once we had found it!

PR #209, opened 2021-05-16
Use neo4j int for limit and skip

aspectgfg:
# Description

Use neo4j's int for limit and skip

# Issue

https://github.com/neo4j/graphql/issues/208


darrellwarde:
Hey @aspectgfg, thanks for the contribution. As much as I wish this bug were this simple, it is not. As you can see from our Cypher TCK test for pagination, this is already happening for the most part: https://github.com/neo4j/graphql/blob/master/packages/graphql/tests/tck/tck-test-files/cypher/pagination.md

For some reason, however, our custom scalar which is meant to do this for us is sometimes not working in certain scenarios. We do have this on our to-do list for this dev phase, where we are going to take a more holistic look at this (not just `skip` and `limit`), so going to close this based on that fact.

For future pull requests, can you please leave the checklist from the template in? It is meant to help both contributors and reviewers monitor state of a pull request. For instance, you need to sign the Neo4j CLA to contribute to this repo: https://neo4j.com/developer/cla/

PR #226, opened 2021-05-27
Fix temporal type issues with neo4j-driver@4.3.0

oskarhane:
The latest release of neo4j-driver has changes file paths and the code in this library relied on a file being in a certain place.
This PR removes that and imports from the root path only.

I'm not 100% sure how we're going to test this with all versions of the driver we allow as peer dependencies yet. It's late and something for tomorrow.


darrellwarde (approved):
Great, LGTM! 🌟

PR #229, opened 2021-06-02
Update neo-push-server to enforce nodemon uses local ts-node install

ev-dev:
# Description

Minor update to the neo-push demo's start script to explicitly call the local `ts-node` when using `nodemon`

# Issue

When attempting to run the neo-push demo for the first time after setup, I encountered an error within the runtime logs showing that `ts-node` could not be found. 

This was because the `start` script in neo-push-server's package.json used nodemon to execute the ts-node process, but had not set the `--exec` flag to a relative/absolute path, meaning nodemon was implicitly expecting to find the provided argument on a user's $PATH (and I did not have a global ts-node installed). Since ts-node is a dev dependency, the script should be using the locally installed version.

Note: There's a subtle difference in the `test` script, which also calls ts-node without a relative path, because the yarn/npm execution environment is different from the nested nodemon process, and so yarn finds the ts-node executable from within `node_modules/.bin` without issue.

# Checklist

### DIFF
--- a/examples/neo-push/server/package.json
+++ b/examples/neo-push/server/package.json
@@ -4,7 +4,7 @@
     "description": "",
     "main": "./dist/src/index.js",
     "scripts": {
-        "start": "nodemon --watch './src/**/*.ts' --exec ts-node -r tsconfig-paths/register --project ./src/tsconfig.json ./src/index.ts",
+        "start": "nodemon --watch './src/**/*.ts' --exec ./node_modules/.bin/ts-node -r tsconfig-paths/register --project ./src/tsconfig.json ./src/index.ts",
         "build": "tsc  --project src/",
         "seed": "ts-node -r tsconfig-paths/register --project ./src/tsconfig.json ./src/seeder.ts",
         "test": "jest"


ev-dev:
Not sure why the cla check didn't pass again, I tried force pushing an empty commit after I got the cla reply email and the check did seem to re-run but just failed again. 

Think when the first cla check ran, it was probably too soon for my info to be populated, has that ever caused similar problems for first-time contributors befor?

darrellwarde:
> Not sure why the cla check didn't pass again, I tried force pushing an empty commit after I got the cla reply email and the check did seem to re-run but just failed again.
> 
> Think when the first cla check ran, it was probably too soon for my info to be populated, has that ever caused similar problems for first-time contributors befor?

Hey, I'm not too sure what's going on here either, as your name isn't on the allowlist - need to chase up internally as something could be wrong with that pipeline.

darrellwarde (approved):
This looks good for now! It will need changing to pull from the hoisted root level `node_modules` if we ever install ts-node elsewhere in the repo, as Yarn will hoist it to a common root level module. 🙂

Issue #213, opened 2021-05-17
Narrowing interface fields in an implementation fails


lirbank:
Consider the following example:

```ts
import { Neo4jGraphQL } from '@neo4j/graphql';

new Neo4jGraphQL({
  typeDefs: /* GraphQL */ `
    interface Vehicle {
      id: ID!
      color: String # NOTE: 'color' is optional on the interface
    }

    type Car implements Vehicle {
      id: ID!
      color: String! # NOTE: 'color' is mandatory on the type, which should be okay
    }

    type Query {
      cars: [Vehicle!]!
    }
  `,
});
```
Here, `color` is defined as an optional string in the Vehicle interface. In the Car implementation, the `color` field is mandatory. In GraphQL an optional type in an interface may be implemented as mandatory. Trying to do this with Neo4jGraphQL fails with:

```
Error: type Car does not implement interface Vehicle correctly
```

**System:**

"@neo4j/graphql": "1.0.1",

danstarns:
Hi. The error you are seeing is correct, `Vehicle.color` is of type String whereas `Car.color` is of type String NonNull. Change `Car.color` to non-null and your issue should go.

darrellwarde:
I initially thought @danstarns was on point here, but after experimenting it does seem valid to do what you have described here. It's on our books to look at schema validation, so we will look into this particular case whilst we do that work. 🙂

danstarns:
Copverage in PR #200

darrellwarde:
This has been fixed by #300, which will be released in version 2.0.0.

darrellwarde:
In an effort to clean our issue backlog a bit, going to close this one off as the fix is available in any 2.0.0 release whenever you decide to upgrade, be it during prerelease or after stable.

PR #232, opened 2021-06-02
Fix/validate-document

danstarns:
# Description

Validates the typeDefs correctly with Point support. 

# Issue

https://github.com/neo4j/graphql/issues/213

# Checklist


packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -763,9 +783,30 @@ function makeAugmentedSchema(
         generatedResolvers[union.name.value] = { __resolveType: (root) => root.__resolveType };
     });
 
+    const seen = {};
+    let parsedDoc = parse(generatedTypeDefs);
+    parsedDoc = {
```
darrellwarde, line 788: Trying to get my head around this but can't quite figure out what it's doing... Any chance of a quick explanation?

packages/graphql/src/schema/make-augmented-schema.test.ts
```diff
@@ -104,24 +104,6 @@ describe("makeAugmentedSchema", () => {
         expect(() => makeAugmentedSchema({ typeDefs })).toThrow("cannot have interface on relationship");
     });
 
-    test("should throw type X does not implement interface X correctly", () => {
```
darrellwarde, line 107: We should probably leave a negative test for interface implementation in, perhaps moved into validate-document.test.ts instead.

packages/graphql/src/schema/make-augmented-schema.test.ts
```diff
@@ -152,29 +134,6 @@ describe("makeAugmentedSchema", () => {
         expect(() => makeAugmentedSchema({ typeDefs })).toThrow("cannot auto-generate an array");
     });
 
-    /*
-        Removal of validateTypeDefs function
-    */
-    // test("should throw timestamp operations must be an array", () => {
```
darrellwarde, line 158: I would say both of these test cases are valid and should be left in! The more coverage the better!

darrellwarde (changes requested):
I think a bit more negative testing and this should be good to go!

packages/graphql/src/schema/make-augmented-schema.ts
danstarns: It filters out the seen definitions in the document, without would result in an error `Cannot have two types with the name of Int`.

packages/graphql/src/schema/make-augmented-schema.test.ts
danstarns: These tests don't throw anymore.

packages/graphql/src/schema/make-augmented-schema.test.ts
danstarns: There is one https://github.com/neo4j/graphql/pull/232/files#diff-1332c2df9bfa483fbae8bd4af0e5b04092e34ad090756e1b895845c04e2d4c2dR192

darrellwarde: I mean there should be a test to ensure that an error _is_ thrown if a type doesn't implement an interface correctly.

packages/graphql/src/schema/make-augmented-schema.test.ts
darrellwarde: If that's the case, are we sure that this is validating against our directive definitions correctly?

packages/graphql/src/schema/make-augmented-schema.test.ts
danstarns: Since we have removed our own custom logic to do with the matter I don't think it is necessary to test. The problem is now outside our domain and shouldn't be our concern.

packages/graphql/src/schema/make-augmented-schema.test.ts
danstarns: Yeah here https://github.com/neo4j/graphql/pull/232/files#diff-1332c2df9bfa483fbae8bd4af0e5b04092e34ad090756e1b895845c04e2d4c2dR28

darrellwarde: What I'm trying to get at is, if these tests don't throw any more, that means that the type of the `operations` field of the `@timestamp` directive isn't being validated against, and it's letting any values through. Unless I'm misunderstanding?

packages/graphql/src/schema/make-augmented-schema.test.ts
darrellwarde: I think that given we're using a non-default export from the GraphQL library, we should have a few common positive and negative test cases just to prove functionality. Lot's of good "should not throw" tests here, but not as many "should throw" tests (without which, we have little confidence that the function will not always "not throw").

packages/graphql/src/schema/make-augmented-schema.test.ts
danstarns: It works fine and the GraphQL validate function will take care of it. As you can see I have asserted that we are hitting the GraphQL validation with my attached test and I don't think it's maintainable or necessary to assert the validation of each and every scalar, directive and type applied to the schema.

darrellwarde:
I have branched and continued work on this, now superseded by #300.

PR #239, opened 2021-06-08
Feat/composite-where-on-delete-and-disconnect

danstarns:
# Description

A branch from pair programming with @darrellwarde. The code in here is a horrible and inconsistent mess though, let's get it into our working 2.0 branch so we can refactor.

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,6 +1,0 @@
-{
-  "hooks": {
-      "pre-commit": "lint-staged",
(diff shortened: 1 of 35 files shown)


darrellwarde:
Looks like these tests are failing because 2.0.0 doesn't yet have the recent changes from master. Let's merge this first despite the failing tests, and then get those changes from master merged in?

Issue #207, opened 2021-05-14
__resolveType is not used when working with graphQL interface type


Pitchlab:
**Describe the bug**
When working with an interface type named `Content` with a few types , e.g. `Content_A` and `Content_B` (see below), using these in a query on type `Foo` (see below) gives the following error: 

```
"Abstract type \"Content\" must resolve to an Object type at runtime for field \"ContentMatch.content\". Either the \"Content\" type should provide a \"resolveType\" function or each possible type should provide an \"isTypeOf\" function."
```

Trying to remedy this by adding `__resolveType` to a custom resolver does not resolve the issue. Digging into past issues tells me that on the old `neo4j-graphql-js` code, the `makeAugmentedSchema` does **not** call the `__resolveType` code from the custom resolver. I highly suspect that this is the caase also in the new `neo4j-graphql` code. 

**Type definitions**

```
interface Content {
    uri: ID!
}

type Content_A implements Content {
   uri: ID!
   a: String
}

type Content_B implements Content {
   uri: ID!
   b: String
}

type Foo {
   content: [Content] 
}

query FooQuery {
  Foo {
    content {
      uri
      ...on Content_A {
        a
      }
     ...on Content_B {
        b
      }
    }
  }
}
```
**To Reproduce**

See above. 

**Expected behavior**
The `__resolveType` code is called and the error is gone, and we can happily use interfaces. 

**System**
-   OS: MacOS 10.15.1
-   Version: ^1.0.0-beta.2
-   Node.js version: 14.15.1

iyosayi:
Currently experiencing this issue. This has brought the entire project to a halt, is there any workaround for this? Because we rely on using Unions/Interfaces a lot.

darrellwarde:
Hi both, sorry for the radio silence on this one, and sorry to hear you've had to revert back to the old library until this is looked at. 🙁 

We already had our hands full with key bug fixes and features when this was raised, but this certainly looks like a candidate for one of the next major bugs we address. I'll keep you posted!

DomVinyard:
Yeah, running into this same error too, with Unions. We are not seeing this with autogenerated types, only with custom resolver.

Schema:
```
union U = A | B

type A {
  id: ID
}
type B {
  id: ID
}
type User {
  test: U
}
type Query {
   Me: User # custom OGM resolver, returns user 1
}

```

Two queries:
:white_check_mark: Succeeds
```
query {
  Me {
    test {
      ...on A {
        id
      }
    }
  }
}
[...]

PR #260, opened 2021-06-15
Prevent non null union resolvers from being overwritten.

gerarddj:
Partial fix for #207 . Prevents non null union resolvers from being overwritten with a __resolveType.


darrellwarde:
Morning @gerarddj! I've seen that you've raised this as well as #260, but subsequently closed them both. I would love to know why, as this looks like a good fix to me!

I hope you weren't discouraged by the failing CLA check - we need to manually add to the allowlist (which I've now done for you), and this won't fail if it runs for you again.

darrellwarde (approved):
I'm going to look at the rest of the `__resolveType` problems today, but this looks like a good start, thanks! 🙂

PR #262, opened 2021-06-16
Use upperFirst from graphql-compose in all cases

darrellwarde:
# Description

No point in using another library or having our own util function for it.

# Checklist


packages/ogm/package.json
```diff
@@ -29,8 +29,7 @@
         "@graphql-tools/merge": "^6.2.13",
         "@neo4j/graphql": "^1.0.2",
         "camelcase": "^6.2.0",
-        "pluralize": "^8.0.0",
-        "upper-case-first": "^2.0.2"
+        "pluralize": "^8.0.0"
```
danstarns, line 32: We should do the same as upper first with the OGM: WDYT? 

```
import { Neo4jGraphQL, upperFirst } from "@neo4j/graphql";
```

packages/graphql/src/index.ts
```diff
@@ -17,5 +17,8 @@
  * limitations under the License.
  */
 
+import { upperFirst } from "graphql-compose";
```
danstarns, line 20: IMO it's not GraphQL's composes responsibility and kinda confusing. But I get why we have done this to reduce dependencies.

danstarns (approved):

packages/ogm/package.json
darrellwarde: `graphql-compose` isn't a dependency of the OGM (despite the fact it will always be there due to dependency on `@neo4j/graphql`) and ESLint rightfully complains, so the solution here for the OGM is to get around that fact.

packages/graphql/src/index.ts
darrellwarde: I get it! But definitely no point in having the `upper-case-first` dependency, and after our own `upperCaseFirst` function had a bug in it, I would rather use one from a "trusted source".

In `utils/index.ts` we could do:

```ts
import { upperFirst } from "graphql-compose";
export { upperFirst }
```

Because then you could do `import { upperFirst } from "./utils"` which feels a _bit_ nicer to use, if you like?

PR #271, opened 2021-06-18
fix: use package json for useragent name and version

danstarns:
# Description

Uses `package.json` to grab the version and package name vs using the environment variable. Changes to ts-config can be explained here: https://stackoverflow.com/a/61467483/10687857

# Checklist

### DIFF
--- a/packages/graphql/src/utils/execute.ts
+++ b/packages/graphql/src/utils/execute.ts
@@ -53,7 +53,7 @@ async function execute(input: {
     }
 
     // @ts-ignore: Required to set connection user agent
-    input.context.driver._userAgent = `${environment.NPM_PACKAGE_VERSION}/${environment.NPM_PACKAGE_NAME}`; // eslint-disable-line no-underscore-dangle
+    input.context.driver._userAgent = `${environment.NPM_PACKAGE_NAME}/${environment.NPM_PACKAGE_VERSION}`; // eslint-disable-line no-underscore-dangle
 
     const session = input.context.driver.session(sessionParams);
 
--- a/packages/graphql/src/utils/execute.ts
+++ b/packages/graphql/src/utils/execute.ts
@@ -52,8 +52,16 @@ async function execute(input: {
         }
     }
 
-    // @ts-ignore: Required to set connection user agent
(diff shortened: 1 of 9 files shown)


darrellwarde:
It seems you need to sign the CLA! 😉

darrellwarde:
Have you tested this by checking the query logs in a Desktop instance or something similar?

darrellwarde:
> It seems you need to sign the CLA! 😉

Rerunning that stage now...

packages/graphql/src/utils/execute.ts
```diff
@@ -53,7 +53,7 @@ async function execute(input: {
     }
 
     // @ts-ignore: Required to set connection user agent
-    input.context.driver._userAgent = `${environment.NPM_PACKAGE_VERSION}/${environment.NPM_PACKAGE_NAME}`; // eslint-disable-line no-underscore-dangle
+    input.context.driver._userAgent = `${environment.NPM_PACKAGE_NAME}/${environment.NPM_PACKAGE_VERSION}`; // eslint-disable-line no-underscore-dangle
```
oskarhane, line 56: These are internal API:s and I don't think the driver supports updating the config once the driver is instantiated.
To get this to work with driver `4.3.0` I have to change this to

```js
input.context.driver._config.userAgent = `${environment.NPM_PACKAGE_NAME}/${environment.NPM_PACKAGE_VERSION}`;
```

which implies that it could have been a change in the internal API:s for the 4.3.0 version of the driver.

oskarhane (commented):
I think the right approach here is to either:

* ask the driver to expose a public API to update the user agent after the driver has been created
* or we can send transaction metadata with every query - which I think is the correct approach here.
* 
The user agent _is_ the driver, but the code generating queries is the library.
That's how browser, bloom etc are doing it.

https://neo4j.com/docs/api/javascript-driver/current/typedef/index.html

https://github.com/neo4j/neo4j-browser/blob/master/src/shared/services/bolt/transactions.ts#L96
https://github.com/neo4j/neo4j-browser/blob/master/src/shared/modules/commands/cypher.test.ts#L91

packages/graphql/src/utils/execute.ts
danstarns: Cool, this works for me now. I added both support for 4.2 and 4.3

oskarhane (approved):
Let's merge this into `master` but keep it out of the.`2.0.0` branch for now. We'll add a more robust solution there.

PR #277, opened 2021-06-21
config: remove codeowners

danstarns:
# Description

Removes code owners so team members don't get spam on holidays. As the team grows I would say this is unnecessary for Darrell and me to **both** review everything going in.


oskarhane (approved):
LGTM

Issue #280, opened 2021-06-26
Auth cypher param dropped if no Authorization header causing exception instead of unauthorized error


mathix420:
**The simplified schema used**
```graphql
type User {
    id: ID! @id
    publishedVideos: [Video!]! @relationship(type: "PUBLISHER", direction: OUT)
}

type Video {
    id: ID! @id
    publisher: User! @relationship(type: "PUBLISHER", direction: IN)
}

extend type Video @auth(rules: [{ where: { publisher: { id: "$jwt.sub" } } }])
```

**Query used**
```graphql
query getVideos {
    videos {
        id
    }
}
```

**Behavior**
- When passing an `Authorization` header with the JWT, everything is working fine.
- Once the `Authorization` is removed, got the following error:
```
Expected parameter(s): this_auth_where0_publisher_id
```

**DEBUG data**
*Authenticated test:*
```js
@neo4j/graphql:execute About to execute Cypher:

Cypher:
    [HIDDEN]
Params:
    {
      "this_auth_where0_publisher_id": "719dd44f-c7c6-422e-8bf8-41127c8a306e",
      "auth": {
        "isAuthenticated": true,
        "roles": [],
        "jwt": {
          [HIDDEN]
          "sub": "719dd44f-c7c6-422e-8bf8-41127c8a306e"
        }
      }
    } +0ms
```
*Non-authenticated test:*
```js
@neo4j/graphql:execute About to execute Cypher:

Cypher:
    [HIDDEN]
Params:
{
  "auth": {
    "isAuthenticated": false,
    "roles": []
  }
} +10s
```

**Expected behavior**
*Non-authenticated:*
```js
@neo4j/graphql:execute About to execute Cypher:

Cypher:
    [HIDDEN]
Params:
{
  "this_auth_where0_publisher_id": null,
  "auth": {
    "isAuthenticated": false,
    "roles": []
  }
} +10s
```

important-security-account:
@mathix420 GitHub has notified us that your OpenVSX API key still could be active. Please revoke the key to ensure it cannot be used by unauthorized parties. It can be found in your pull request.

PR #281, opened 2021-06-26
fix: losing params while creating Auth Predicate

mathix420:
# Description

Avoid `createAuthPredicate` to lose params when no values attached to them.
In the case of a non authed request for example `$jwt.sub` return undefined.

To fix this I just check `paramValue` before setting it to the `params` object, if it's `undefined` set it to `null`.

# Issue

Fix: https://github.com/neo4j/graphql/issues/280

# Checklist

- [x] ~~Documentation has been updated~~ (Not applicable)
- [x] ~~TCK tests have been updated~~ (Not applicable)
- [x] ~~Example applications have been updated~~ (Not applicable)
- [x] ~~New files have copyright header~~ (Not applicable)

### DIFF
--- a/packages/graphql/src/translate/create-auth-and-params.test.ts
+++ b/packages/graphql/src/translate/create-auth-and-params.test.ts
@@ -548,4 +548,146 @@ describe("createAuthAndParams", () => {
             });
         });
     });
+
+    describe("params", () => {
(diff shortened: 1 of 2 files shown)


darrellwarde (approved):
LGTM! Will be a good stopgap until we look at throwing unauthenticated errors before even hitting the database. 🙂

Issue #206, opened 2021-05-14
Add @Unique Directive Back


fordneild:
**Is your feature request related to a problem? Please describe.**
I would like to enforce the uniqueness of certain fields, which I was able to do using neoj4-graphql-js.
**Describe the solution you'd like**
@unique directive should give the nodes property a unique constraint
**Describe alternatives you've considered**
Otherwise, I have to rely on developers not writing bugging code

**Additional context**

manonthemat:
100% agree. I'm utterly confused why it's not in the library.

lfritsche:
I'm using apoc.schema.assert right after the connection has been verified (`driver.verifyConnectivity()`).
e.g.: `CALL apoc.schema.assert(null,{Product:['id']})` to prevent duplicate product IDs. 
This works quiete well and gets executed every time I redeploy/restart the ApolloServer.

Would `@Unique` work in a similar way? Does this approach conflict with your internal index management or future `@Unique` implementation?

PR #285, opened 2021-06-29
Global Node Resolution (Relay Specification)

litewarp:
# Description

Part 2 of the Relay Spec Implementation started in #282.

# Issue

# Checklist


darrellwarde:
So, time to start thinking about how to go about this extremely simple but also quite complex part of the Relay spec. A few options of how to deal with this and my initial thoughts about them:

1. Have users add `implements Node` to all of their types (I believe what is proposed here) - personally not a fan of this approach because it means a lot of boilerplate for users, and the onus is on them to make sure _all_ of their nodes implement the interface
2. Automatically add `implements Node` to each of the users types, probably enabled by a configuration option - at this point, my preferred approach, of which really has two options within it:
a. Automatically add an `id` field - could result in conflicts, and a lot is happening under the hood without users' "consent"
b. Put the onus on users to fulfil the `Node` contract and add the `id` field themselves, which they can then decide which property to match it to and whether it should be automatically created, and an error is thrown if any types do not have an `id` field - out of this list, my preferred option, as users are aware of what Relay spec compliance means and are in control if they desire to be

And then also @litewarp, you posted a very detailed suggestion in the other PR:

> So I think the solution to this is to use id encoding. All we would need from users is an identification of a unique field property. Then you can encode the TypeName and the uniqueValue into a string that you use for the id. This way the id is globally unique amongst types, you can auto-generate it based on identified unique keys, etc.
> 
> Using the [graphql-relay-js](https://github.com/graphql/graphql-relay-js/blob/fa978fbfaa367b79b4616d091471be87e84f3dbe/src/node/node.js#L87) library again as example, there are two helpers:
> 
> ```ts
> /**
>  * Takes a type name and an ID specific to that type name, and returns a
>  * "global ID" that is unique among all types.
>  */
> export function toGlobalId(type: string, id: string | number): string {
>   return base64([type, GraphQLID.serialize(id)].join(':'));
> }
> 
> /**
>  * Takes the "global ID" created by toGlobalID, and returns the type name and ID
>  * used to create it.
>  */
> export function fromGlobalId(globalId: string): ResolvedGlobalId {
>   const unbasedGlobalId = unbase64(globalId);
>   const delimiterPos = unbasedGlobalId.indexOf(':');
>   return {
>     type: unbasedGlobalId.substring(0, delimiterPos),
>     id: unbasedGlobalId.substring(delimiterPos + 1),
>   };
> }
> ```
> 
> So long as the id value is unique across the Node, by encoding it with the typeName, it will be unique across all types.
> 
> For users that don't have an id field, or want to use another field, you can autogenerate it for them. All you'll need them to do is identify the field with the @id directive?
> 
> ```graphql
> # using uuid
> type Movie {
>   id: ID!
> 
>   _id: ID @id
> }
> 
> # using other field
> type Actor {
>   id: ID!
>  
>   imdbNo: String! @id(autogenerate: false) 
> }
> ```
> 
> This would also make node resolution easier since all you'll have to do is decode the id, so two database trips aren't required.

A few thoughts to get the ball rolling on this one, please feel free to chime in @danstarns and @oskarhane!

litewarp:
So some thoughts after reading @darrellwarde's comments above.

1. Is making every type a "Node" a good thing?  I'm not sure I want to opt into every type I generate being a "node" in my schema. For example, if I have a nested return type that is not itself a label in the database. So if there is automatic opt-in, there will need to be some mechanism for opting out. Perhaps that could be another argument added to the existing `@exclude` directive? But I suppose the point I'm making is that adding "implements Node" to the types that are actually "nodes" might be better since it requires affirmative action on the part of the user to state that this type is in fact a node per the relay rules. And thus, by being forced to opt-in, they have to pay mind to the constraints of that decision, especially if it requires users to define a unique field or id which they have not already.  It also allows for incremental adoption, although if I am honest, I am not sure what that use case is or how it would work except in theory.

2.  Do you want to implement a primary key or uniqueness validation?  There's an outstanding PR (#206) for this.  If you do plan on implementing that directive, should we add an argument indicating its the primary key? Not sure if this would have side-effects apart from the relay Id choice.  I'm not sure how you will be able to enforce global resolution without some uniqueness within the type. It may be worth considering adding a incrementing id as a default in the absence of a user-defined unique key.

3.  In terms of global resolution, I am coming to the conclusion that id encoding will be the only feasible way without forcing each "node" to have a uuid.

darrellwarde:
> 1. Is making every type a "Node" a good thing?  I'm not sure I want to opt into every type I generate being a "node" in my schema. For example, if I have a nested return type that is not itself a label in the database. So if there is automatic opt-in, there will need to be some mechanism for opting out. Perhaps that could be another argument added to the existing `@exclude` directive? But I suppose the point I'm making is that adding "implements Node" to the types that are actually "nodes" might be better since it requires affirmative action on the part of the user to state that this type is in fact a node per the relay rules. And thus, by being forced to opt-in, they have to pay mind to the constraints of that decision, especially if it requires users to define a unique field or id which they do have not already.  It also allows for incremental adoption, although I would need to think more about how that might work.

For sure, there would have to be something akin to the `@exclude` or `@ignore` directive to allow certain types to not require the interface. However, you are perhaps right, and on further delving into the spec I found the following note (https://relay.dev/assets/files/connections-932f4f2cdffd79724ac76373deb30dc8.htm#note-e0232):

> The naming echoes that of the “Node” interface and “node” root field as described in a later section of this spec. Spec-compliant clients can perform certain optimizations if this field returns an object that implements Node, however, this is not a strict requirement for conforming.

That final phrase perhaps changes the game a little bit, and maybe a per-type approach is the way forward to enable users to incrementally adopt.

> 2. Do you want to implement a primary key or uniqueness validation?  There's an outstanding PR ([Add @Unique Directive Back #206](https://github.com/neo4j/graphql/issues/206)) for this.  If you do plan on implementing that directive, should we add an argument indicating its the primary key? Not sure if this would have side-effects apart from the relay Id choice.  I'm not sure how you will be able to enforce global resolution without some uniqueness within the type. It may be worth considering adding a incrementing id as a default in the absence of a user-defined unique key.

Our problem here is that in order to implement this "properly" (as this "product" should, as opposed to the previous experimental Neo4j Labs project), we really need a migration tool as creating constraints is a one-off activity, which doesn't really fit well into the library as it is.

It's perhaps a naive approach, but if users use the current `@id` directive, we can be _fairly_ confident that IDs won't get reused, purely because the odds of that happening are extremely slim when generating UUIDs. If users choose to generate IDs themselves, then for the time-being we can probably just say that the onus is on them to enforce uniqueness.

> 3. In terms of global resolution, I am coming to the conclusion that id encoding will be the only feasible way without forcing each "node" to have a uuid.

I'm still unsure about this!

Taking your previous example of:

```graphql
# using uuid
type Movie {
  id: ID!

  _id: ID @id
}

# using other field
type Actor {
  id: ID!
 
  imdbNo: String! @id(autogenerate: false) 
}
```

My issue with the above is that the `id` field is really just syntactic sugar, and it's not clear where the value has come from. If we were to take this approach, my preference would be:

```graphql
# using uuid
type Movie {
  id: ID! @property(name: "_id")

  _id: ID @id
}

# using other field
type Actor {
  id: ID! @property(name: "imdbNo")
 
  imdbNo: String! @id(autogenerate: false) 
}
```

Then it is much more apparent where that value is coming from, rather than it magically happening under the hood.

We actually had a chat about options here this morning, so I will add another comment below outlining what we discussed, and also add the ID encoding option.

darrellwarde:
So, @danstarns, @oskarhane and I had a chat about the options this morning, so want to write up those options and give examples of how they might operate. Going to split this into two comments as there are actually a couple of choices to be made here.

Firstly, we chatted about a configuration option which would enforce node IDs on every type which I was going to write up, but following the discovery of the following note (https://relay.dev/assets/files/connections-932f4f2cdffd79724ac76373deb30dc8.htm#note-e0232), will perhaps discount this for brevity's sake:

> The naming echoes that of the “Node” interface and “node” root field as described in a later section of this spec. Spec-compliant clients can perform certain optimizations if this field returns an object that implements Node, however, this is not a strict requirement for conforming.

With that in mind, we then move onto options for deciding _where_ to implement the `Node` interface:

1. Add `implements Node` to the desired types, which is a very "what you see is what you get approach":

```graphql
type Actor implements Node {
  # Must contain an `id` field
  ...actorFields
}

type Movie {
  ...movieFields
}
```

2. A suggestion by @danstarns, is to add a new `@node` directive. This would be a more brief way of specifying implementing nodes, and in the future would allow the setting of "node options" (such as "sortable", "filterable", etc.):

```graphql
type Actor @node {
  # Must contain an `id` field
  ...actorFields
}

type Movie {
  ...movieFields
}
```

darrellwarde:
Once we have specified which types implement the `Node` interface, we then move onto _how_ we implement the interface (whether `implements Node` or `@node` is used in the following examples is unimportant:

1. The explicit addition of an `id` field of type `ID!`, and decorated with the `@id` directive. :

```graphql
type Actor implements Node {
  id: ID! @id
}
```

2. The suggestion of @litewarp is to allow for "ID encoding" as detailed in their previous comment:

> So I think the solution to this is to use id encoding. All we would need from users is an identification of a unique field property. Then you can encode the TypeName and the uniqueValue into a string that you use for the id. This way the id is globally unique amongst types, you can auto-generate it based on identified unique keys, etc.
> 
> Using the [graphql-relay-js](https://github.com/graphql/graphql-relay-js/blob/fa978fbfaa367b79b4616d091471be87e84f3dbe/src/node/node.js#L87) library again as example, there are two helpers:
> 
> ```ts
> /**
>  * Takes a type name and an ID specific to that type name, and returns a
>  * "global ID" that is unique among all types.
>  */
> export function toGlobalId(type: string, id: string | number): string {
>   return base64([type, GraphQLID.serialize(id)].join(':'));
> }
> 
> /**
>  * Takes the "global ID" created by toGlobalID, and returns the type name and ID
>  * used to create it.
>  */
> export function fromGlobalId(globalId: string): ResolvedGlobalId {
>   const unbasedGlobalId = unbase64(globalId);
>   const delimiterPos = unbasedGlobalId.indexOf(':');
>   return {
>     type: unbasedGlobalId.substring(0, delimiterPos),
>     id: unbasedGlobalId.substring(delimiterPos + 1),
>   };
> }
> ```
> 
> So long as the id value is unique across the Node, by encoding it with the typeName, it will be unique across all types.
> 
> For users that don't have an id field, or want to use another field, you can autogenerate it for them. All you'll need them to do is identify the field with the @id directive?
> 
> ```graphql
> # using uuid
> type Movie {
>   id: ID!
> 
>   _id: ID @id
> }
> 
> # using other field
> type Actor {
>   id: ID!
>  
>   imdbNo: String! @id(autogenerate: false) 
> }
> ```
> 
> This would also make node resolution easier since all you'll have to do is decode the id, so two database trips aren't required.

And an extension of this which I proposed which would make this a little more explicit:

> My issue with the above is that the `id` field is really just syntactic sugar, and it's not clear where the value has come from. If we were to take this approach, my preference would be:
> 
> ```graphql
> # using uuid
> type Movie {
>   id: ID! @property(name: "_id")
> 
>   _id: ID @id
> }
> 
> # using other field
> type Actor {
>   id: ID! @property(name: "imdbNo")
>  
>   imdbNo: String! @id(autogenerate: false) 
> }
> ```
> 
> Then it is much more apparent where that value is coming from, rather than it magically happening under the hood.

Perhaps for this part of the decision, we don't have to choose between the two but could easily enough allow for both approaches?

I think whichever way we go, a directive to allow for the mapping to a different database property (such as the hypothetical `@property` directive I used above) will be necessary for finer control.

litewarp:
Just a random thought but could you combine the @node directive with the definition of the id property? Would save you an additional directive. You could even let users omit the id field entirely. 

```graphql 
# id field required
type Actor @node(id: "_id") {
  id: ID!

  _id: ID! @id 
}

# id field auto generated
type Movie @node(id: "imdbNo") {
  # id will be in the resulting schema

  # validation would throw if id is defined or no @id property is present

  imdbNo: String! @id 
}


```

darrellwarde:
> Just a random thought but could you combine the @node directive with the definition of the id property? Would save you an additional directive. You could even let users omit the id field entirely.
> 
> ```graphql
> # id field required
> type Actor @node(id: "_id") {
>   id: ID!
> 
>   _id: ID! @id 
> }
> 
> # id field auto generated
> type Movie @node(id: "imdbNo") {
>   # id will be in the resulting schema
> 
>   # validation would throw if id is defined or no @id property is present
> 
>   imdbNo: String! @id 
> }
> ```

I also had small thoughts about this in my head whilst writing up the options, it could be an interesting idea! Perhaps the options could be as follows...

```graphql
# Just a normal node, nothing to do with Relay or the Node interface
type Movie {
  title: String
}

# Will implement the Node interface, and the user has chosen to define the field themselves
type Movie @node {
  id: ID! @id
  title: String
}

# Will implement the Node interface, and the id field will be generated off the Movie title (we probably need to worry about uniqueness)
type Movie @node(id: "title") {
  title: String!
}

# An Error will be thrown - there is a risk that the desired identifier will be null
type Movie @node(id: "title") {
  title: String
}

# An Error will be thrown - the id field has not been specified
type Movie @node {
  title: String
}

# An Error will be thrown - an id field has been defined but of the wrong type
type Movie @node {
  id: String!
  title: String
}
```

Just as a start to thinking about how this might work in practice. @danstarns, what do you think of the above?

darrellwarde:
Closing in favour if discussion in #588, with the implementation to be done in a different branch. Thanks @litewarp!

Issue #192, opened 2021-05-06
Relationship properties


danstarns:
**Is your feature request related to a problem? Please describe.**
Ability to add properties on relationships

**Additional context**
Came from here https://github.com/neo4j/graphql/issues/179

colinskow:
Great feature! I believe this is the biggest block to being able to migrate from `neo4j-graphql-js`.

For reference here is the [previous documentation](https://github.com/neo4j-graphql/neo4j-graphql-js/blob/master/docs/graphql-relationship-types.md) on relationship properties.

The RFC is more complex than the previous implementation, but the extra power seems worth it!

litewarp:
Great proposal. Any chance you could (a) add a total count field to the automatically generated connection properties, or (b) allow users to augment the connection properties?

darrellwarde:
> Great feature! I believe this is the biggest block to being able to migrate from `neo4j-graphql-js`.
> 
> For reference here is the [previous documentation](https://github.com/neo4j-graphql/neo4j-graphql-js/blob/master/docs/graphql-relationship-types.md) on relationship properties.
> 
> The RFC is more complex than the previous implementation, but the extra power seems worth it!

We really do recognise it is likely the biggest blocker to migration, hence why we're jumping onto this so quickly. 🙂 

More complex, and sadly a couple of unavoidable breaking changes, but this will also form the foundation for the much asked for cursor-based pagination in future, which we're very excited for!

darrellwarde:
> Great proposal. Any chance you could (a) add a total count field to the automatically generated connection properties, or (b) allow users to augment the connection properties?

Thanks for the kind words and suggestions! 🙂  

The returned connection will be based on the filters provided, so it might be a confusing behaviour to add a total count field - is that the total count of all edges, or just the edges in this particular connection? If the latter, you can just get the length of the edges array.

If the former, we have another piece of work on our roadmap which is to introduce aggregation queries which will allow you to query aggregations such as count. We want to get relationship properties in first, but that is hopefully going to follow not too far behind.

Regarding customising the connection types... _Maaaaaaybe_, but that would be well down the line, as we want to build a stable platform for relationship properties first.

danstarns:
Overall this is great! Introducing relationship properties will help unblock a lot of users and better position us for things such as Relay. I do have a few concerns that I will share below.
 
**Breaking changes that will bump the library up to V 2.0.** Depending on when we implement this, that seems imminent, we could aggravate a lot of existing users.

[...]

PR #289, opened 2021-07-02
Only return union members which have fragments in the selection set

darrellwarde:
# Description

When querying a union relationship, only members of that union that have fragments in the selection set will be projected.

# Issue

#192 

# Checklist


danstarns:
I noticed you only changed the connections union projection however, what about the normal projection?

darrellwarde:
> I noticed you only changed the connections union projection however, what about the normal projection?

I believe that already works (but differently), and this doesn't because of subquery unions... But correct me if I'm wrong!

danstarns (approved):

Issue #283, opened 2021-06-27
Datetime returns object (year, month, day etc) instead of string


MarianAlexandruAlecu:
`Datetime` type is returning an object instead of a string

<img width="1440" alt="Screenshot 2021-06-27 at 18 29 30" src="https://user-images.githubusercontent.com/5826632/123550347-c3bcae00-d775-11eb-8b5a-c768f14f4cc4.png">

```
scalar Point
scalar DateTime
scalar PointInput

type Mutation {
  createPosts(input: PostCreateInput!): Post!
    @cypher(
      statement: """
      MATCH (user:User { id: $auth.jwt.sub })
      CREATE (post:Post)-[:HAS_USER]->(user)
      SET
        post = $input,
        post.datetime = datetime(),
        post.id = randomUUID()
      RETURN post
      """
    )
}

type Post {
  id: ID! @id
  title: String!
  description: String!
  datetime: DateTime @readonly
  user: User @readonly @relationship(type: "HAS_USER", direction: OUT)
}
```

This is how `datetime` field looks in Neo4J Browser: `"2021-06-27T15:34:15.146000000Z"`

If I comment `scalar DateTime` while node is running, it will start to return a string (_sometimes_)

I would expect to simply return `datetime: 2021-06-27T15:34:15.146000000Z` so that I can easily parse this in frontend. Maybe I'm missing something obvious in my setup? 🤔 

-   OS: macOS 11.4 M1
-   Version: @neo4j/graphql@1.0.2 & 1.0.3
-   Node.js version: v14.17.0

darrellwarde:
Assuming that the type definitions you have provided are what you're using for your application, you need to remove _all_ of those scalars - they are all automatically generated and included in your schema as soon as you use them anywhere, so you don't need to include them yourself. `Point` and `PointInput` aren't even scalars, but object and input types respectively, so this would massively break your schema.

Out of interest, why are you using custom Cypher for your `createPosts` Mutation? That all looks like stuff that works out of the box with the library.

darrellwarde:
@MarianAlexandruAlecu, any update on this based on the information above? Will assume fixed and will close in the next few days if no further activity.

MarianAlexandruAlecu:
@darrellwarde I found the issue. Here is the full code so you can reproduce it.
```js
// index.ts
import fs from "fs";
import path from "path";
import { Neo4jGraphQL } from "@neo4j/graphql";
import { ApolloServer } from "apollo-server";
[...]

Issue #278, opened 2021-06-22
Point type fails in mutation response and query


keanugrieves:
**Describe the bug**
I've had the same issue as #220 beginning with 1.0.2 and I just assumed it would magically fix itself when I got closer to production and was ready to deploy a new DB, but this time, I was able to reproduce it with a fresh DB instance and a fresh install of the latest neo4j releases. It creates the node just fine, but it returns a strange error message anytime you query for a Point.

**Type definitions**
```
type User {
  id: ID! @readonly
  meta_enabled: Boolean @default(value: true)
  meta_createdAt: DateTime! @timestamp(operations: [CREATE])
  meta_updatedAt: DateTime! @timestamp(operations: [CREATE, UPDATE])
  meta_robotsUpdatedAt: DateTime @readonly
  auth_email: String! @readonly
  auth_lastLoginFrom: Point!
  auth_password: String! @private
  auth_role: UserAuthRole! @readonly
  name_first: String!
  name_last: String!
}
```

**Mutation:**
```
mutation {
  createUsers(
    input: [
      {
        id: "1234567890123456"
        auth_email: "user@test.com"
        auth_lastLoginFrom: {
          longitude: -123.1381874
          latitude: 44.0341663
        }
        auth_role: USER
        name_first: "Test"
        name_last: "User"
      }
    ]
  ) {
    users {
      id
      auth_lastLoginFrom {
        latitude
        longitude
      }
    }
  }
}
```

**Response:**
```
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Point.latitude.",
      "locations": [
        {
          "line": 8,
          "column": 9
        }
      ],
      "path": [
        "createUsers",
        "users",
        0,
        "auth_lastLoginFrom",
        "latitude"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "Error: Cannot return null for non-nullable field Point.latitude.",
            "    at completeValue (...\node_modules\\graphql\\execution\\execute.js:559:13)",
            "    at resolveField (...\node_modules\\graphql\\execution\\execute.js:472:19)",
            "    at executeFields (...\node_modules\\graphql\\execution\\execute.js:292:18)",
            "    at collectAndExecuteSubfields (...\node_modules\\graphql\\execution\\execute.js:748:10)",
            "    at completeObjectValue (...\node_modules\\graphql\\execution\\execute.js:738:10)",
[...]

PR #297, opened 2021-07-05
Fix for loss of scalar and field level resolvers

darrellwarde:
# Description

Whilst debugging #283, it became apparent that the use of custom resolvers was triggering the call of `wrapCustomResolvers`, which was full of bugs and completely mauling the resolvers map. The purpose of this function was to inject auth information into the context, and this is now done as a schema level resolvers, and custom resolvers are added on top of our generated schema to ensure they go untouched.

# Issue

This work should address several issues:

- [x] #283 
- [x] #278 
- [x] #207 

As well as _fully_ fix the issues with `Int` and `Float` values.

This PR will be moved out of draft when fully tested against all of these issues.

# Checklist


danstarns (approved):
Nice, it's good to get rid of this function `wrapResolvers`.

Issue #296, opened 2021-07-05
Autogeneration for relationships not working


YizYah:
**Describe the bug**
I do not see CRUD functions generating for relationships in the way that your documentation implies.

**Type definitions**
type Person {
    name: String!
    born: Int!
    actedInMovies: [Movie!]! @relationship(type: "ACTED_IN", direction: OUT)
    directedMovies: [Movie!]! @relationship(type: "DIRECTED", direction: OUT)
}

type Movie {
    title: String!
    released: Int!
    actors: [Person!]! @relationship(type: "ACTED_IN", direction: IN)
    director: Person! @relationship(type: "DIRECTED", direction: IN)
}

**To Reproduce**

1. Follow the code in your [Getting Started](https://neo4j.com/docs/graphql-manual/current/getting-started/) document to create a local server.
2. Replace the Type Definitions with those shown in your [Relationship Type Definitions Example](https://neo4j.com/docs/graphql-manual/current/type-definitions/relationships/#_type_definitions).
3. Execute the Mutations shown in your [Inserting Data Example](https://neo4j.com/docs/graphql-manual/current/type-definitions/relationships/#_inserting_data)
4. See error for the `CreateActor` mutation
```
{
  "error": {
    "errors": [
      {
        "message": "Cannot query field \"createActors\" on type \"Mutation\". Did you mean \"createMovies\"?",
        "locations": [
          {
            "line": 2,
            "column": 3
          }
        ],
        "extensions": {
          "code": "GRAPHQL_VALIDATION_FAILED",
          "exception": {
            "stacktrace": [
              "GraphQLError: Cannot query field \"createActors\" on type \"Mutation\". Did you mean \"createMovies\"?",
              "    at Object.Field (/home/yisroel/projects/neo4jGraphql/node_modules/graphql/validation/rules/FieldsOnCorrectTypeRule.js:48:31)",
              "    at Object.enter (/home/yisroel/projects/neo4jGraphql/node_modules/graphql/language/visitor.js:323:29)",
              "    at Object.enter (/home/yisroel/projects/neo4jGraphql/node_modules/graphql/utilities/TypeInfo.js:370:25)",
              "    at visit (/home/yisroel/projects/neo4jGraphql/node_modules/graphql/language/visitor.js:243:26)",
              "    at Object.validate (/home/yisroel/projects/neo4jGraphql/node_modules/graphql/validation/validate.js:69:24)",
[...]

Issue #252, opened 2021-06-14
Allow User to Pass in Decoded JWT to Context


litewarp:
**Is your feature request related to a problem? Please describe.**
The current auth JWT functionality does not permit manual verification of tokens through third-party apps or using an RSA key. This makes it impossible to use with Firebase Auth, or any endpoint using a JWKS authentication schema (e.g., Auth0, Azure AD/B2C) since there is no simple secret to pass into the neo4j JWT config. 

**Describe the solution you'd like**
The ability to manually set the JWT claims in the context, either by overriding the JWT object or with a flag that indicates a second location for the JWT object.

**Describe alternatives you've considered**
The current workaround is to have the backend re-encode the already-decoded JWT with a secret that's then passed to the neo4j JWT config. It introduces two extra steps in front of the auth functionality for every request, one to reencode the token and the neo4j library's decoding of that token so it is not ideal. 

**Additional context**
Last communication from @darrellwarde in Discord:

> So the reason this didn't work for you is because of https://github.com/neo4j/graphql/blob/master/packages/graphql/src/classes/Neo4jGraphQL.ts#L120

> Because of that reassignment, anything you pass into the context yourself will get overridden if a JWT can't be found...

> There's obviously some use cases we haven't catered for here, which is to be expected as auth is a deep and complex subject area! I think the conclusion you've collectively come to, of setting the Authorization header, is the best you're going to get at the moment.

I'm not sure there's a security benefit here to keeping JWT's locked to only those than can decode simple secrets. Third-party libraries like Firebase Auth or the JWKS scheme use rotating keys and tokens which are far more preferable than a single secret. Plus people are already working around it by decoding these objects and then re-encoding them.

tlester:
This is an important RFE.   Everything is great if you are paying for Auth0, fails, when you need to decode your own.

jacob-buscher:
@litewarp 

I'm currently making firebase auth work by setting JWT to no verify and then verifying the token in the context function with the firebase admin library.

Not sure if that helps or not, here's my implementation:
```typescript
const neoSchema = new Neo4jGraphQL({
  typeDefs,
  driver,
  config: {
    jwt: {
      noVerify: true,
    } as any,
  },
});

const getAuth = async (req) => {
  const authHeader = req.headers.authorization;
  if (!authHeader) {
    return null;
  }
  const idToken = authHeader ? authHeader.split("Bearer ")[1] : req.body.token;
[...]

PR #303, opened 2021-07-07
Allows users to pass in decoded JWT

darrellwarde:
# Description

Accept a decoded JWT payload in the key `jwt` in the context. 

# Issue

#252 

(Also addresses #296, a small documentation change)

# Checklist


danstarns (approved):
Looks good to me.

packages/graphql/src/translate/create-auth-and-params.ts
```diff
@@ -67,6 +67,10 @@ function createAuthPredicate({
     }
     const { jwt } = context;
 
+    if (!jwt) {
+        throw new Error("Can't generate auth predicate - no JWT in context");
```
mathix420, line 71: This condition is making the whole graphql schema unusable for a non-authed user if `jwt: {}` is not set in the context.
It may be good to update docs accordingly.

danstarns: Hey @mathix420. This doesn't sound like a feature! Could you provide something reproducible?

mathix420: Hello @danstarns! I will try to give you that today, but i'm quite busy right now so it might be for this weekend.

mathix420: Eventually work can wait 10 minutes :sweat_smile:
```typescript
const { Neo4jGraphQL } = require("@neo4j/graphql");
const { ApolloServer } = require("apollo-server");
const neo4j = require("neo4j-driver");
require('dotenv').config();

const AURA_ENDPOINT = process.env.NEO4J_URL;
const PASSWORD = process.env.NEO4J_PASS;
const USERNAME = 'neo4j';

const typeDefs = `
    type Video {
        id: ID! @id
        title: String!
        publisher: String!
    }

    extend type Video @auth(rules: [
        { where: { publisher: "$jwt.sub" } },
    ]),
`;

const driver = neo4j.driver(AURA_ENDPOINT, neo4j.auth.basic(USERNAME, PASSWORD),{
      disableLosslessIntegers: true,
});

const neoSchema = new Neo4jGraphQL({ typeDefs, driver });

    schema: neoSchema.schema,
    context: ({ req }) => ({ req }),
});

server.listen(4000).then(() => console.log("Online"));
```

```graphql
query listVideos {
    videos {
        id
        title
    }
}
```

```json
{
  "errors": [
    {
      "message": "Can't generate auth predicate - no JWT in context",
      ...
    }
  ],
  "data": null
}
```

I think the issue is comming from the `where` rule.

danstarns: @mathix420 Thank you 🙏 we will look into this.

mathix420: **EDIT:** The "Quick workaround" is actually not working at all :stuck_out_tongue: 
It just avoid to fall into the `!jwt` condition, but never checks the jwt, and do not load jwt from cookies/headers.

PR #309, opened 2021-07-12
Merge changes from master into 2.0.0

darrellwarde:
# Description

As per title.


packages/graphql/package.json
```diff
@@ -41,7 +41,7 @@
         "@types/faker": "5.1.7",
         "@types/is-uuid": "1.0.0",
         "@types/jest": "26.0.8",
-        "@types/jsonwebtoken": "8.5.0",
+        "@types/jsonwebtoken": "^8.5.4",
```
danstarns, line 44: This is the only one that's variable 🤔

darrellwarde: We're a bit all over the place with this, our actual `dependencies` have it, but our `devDependencies` are locked - these could probably all also be prefixed with `^`!

danstarns: Let's not change anything seeming it works just fine, I fear adding a ^ could introduce unforeseen issues.

danstarns (approved):
MERGE it 🥳

PR #310, opened 2021-07-12
Docs media

danstarns:
# Description
Adds some links to content about the lib on the main readme.


darrellwarde (approved):
I just capitalised "NODES" (conference name), but other than that, sure, why not, looks good!

PR #317, opened 2021-07-12
refactor: change union where on connections

danstarns:
# Description

Separate where for each union with node and relationship arguments.


packages/graphql/tests/integration/connections/unions.int.test.ts
```diff
@@ -141,7 +141,7 @@ describe("Connections -> Unions", () => {
         }
     });
 
-    test("With `where` argument", async () => {
+    test("With `where` argument on node", async () => {
```
darrellwarde, line 144: We could probably do with a test validating that a node + relationship filter works as expected?

packages/graphql/tests/tck/tck-test-files/schema/connections/unions.md
```diff
@@ -38,636 +38,607 @@ interface Wrote {
```
darrellwarde, line 38: Any chance you could format the schema TCK tests with the old `gql` tag trick so that these are easier to review?

darrellwarde (changes requested):
Looks good! Could maybe just do with one more test and TCK tests formatting?

packages/graphql/tests/integration/connections/unions.int.test.ts
danstarns: Sure

packages/graphql/tests/tck/tck-test-files/schema/connections/unions.md
danstarns: Yea

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -1031,9 +1035,33 @@ function makeAugmentedSchema(
                 const relatedNodes = nodes.filter((n) => connectionField.relationship.union?.nodes?.includes(n.name));
 
                 relatedNodes.forEach((n) => {
+                    const unionWhereName = `${connectionField.typeMeta.name}${n.name}Where`;
```
darrellwarde, line 1038: Thanks for fixing the formatting on the TCK tests - allowed me to notice this.

Normally we would have like `MovieSearchGenreConnectionBlah`, but for this we have `MovieSearchConnectionGenreBlah` - like `MovieSearchConnectionGenreWhere`.

Really not a big deal and I can see how this was caused, and will approve anyway, just figured I would point it out in case you'd like to change it.

darrellwarde (approved):
This looks good!

Just one comment, but it's really not a blocker at all, just an observation.

packages/graphql/src/schema/make-augmented-schema.ts
danstarns: I'm going to leave this, due to the fact we would have to split the `typeMeta.name` apart from the manually appended `connection` word and then rejoin it together - unnecessary logic IMO.

Issue #315, opened 2021-07-12
1.1.0 breaks custom cypher.


tlester:
**Describe the bug**
1.0.3 with double quotes works fine.
1.1.0 with escaped double quotes breaks.
1.1.0 with non-escaped single quotes breaks.
1.0.3 with non-escaped single quotes works fine.

```graphql
type Query {
    getContent(userID: ID): [Content]
        @cypher(
            statement: """
            MATCH (myUser:User {id: $userID})
            OPTIONAL MATCH (myUser)<-[:HAS_USER]-(:Profile)-[:HAS_FRIEND]->(x)<-[:HAS_FRIEND]-(:Profile)-[:HAS_USER]->(myFriends:User)
            CALL {
                WITH myUser, myFriends
                MATCH (myFriends)-[:HAS_POST]->(post:Post)
                RETURN post
                UNION
                WITH myUser, myFriends
                MATCH (myFriends)-[:HAS_COMMENT]->(:Comment)<-[:HAS_COMMENT]-(post:Post)
                RETURN post
                UNION
                WITH myUser, myFriends
                MATCH (myFriends)-[:HAS_REACTION]->(:Reaction)<-[:HAS_REACTION]-(post:Post)
                RETURN post
                UNION
                WITH myUser, myFriends
                MATCH (myUser)<-[:HAS_USER]-(myProfile:Profile)
                MATCH (myProfile)-[:HAS_USER]->(myUsers:User)
                MATCH (myUsers)-[:HAS_POST]->(post:Post)
                RETURN post
                UNION
                WITH myUser, myFriends
                MATCH (myUser)<-[:HAS_USER]-(myProfile:Profile)
                MATCH (myProfile)-[:HAS_USER]->(myUsers:User)
                MATCH (myUsers)-[:HAS_COMMENT]->(:Comment)<-[:HAS_COMMENT]-(post:Post)
                RETURN post
                UNION
                WITH myUser, myFriends
                MATCH (myUser)<-[:HAS_USER]-(myProfile:Profile)
                MATCH (myProfile)-[:HAS_USER]->(myUsers:User)
                MATCH (myUsers)-[:HAS_REACTION]->(:Reaction)<-[:HAS_REACTION]-(post:Post)
                RETURN post
            }
            RETURN DISTINCT apoc.map.merge(properties(post), { __resolveType: 'Post' }) AS result ORDER BY result.modifiedDate DESC
            """
        )
}
```

[...]

PR #318, opened 2021-07-13
Custom Cypher and Union return type - preserve array return

darrellwarde:
# Description

The recent changes to custom Cypher made it so that `expectMultipleReturns` of `apoc.cypher.runFirstColumn` is only set to `true` if a node match was found for the return type. Union return types did not meet this requirement, so `expectMultipleReturns` was set to `false` and only 1 record would ever be returned.

It should be clarified that Union returns for custom Cypher will not have relationships automatically projected, and this will only change when we work on querying Unions/Interfaces from the root Query type.

# Issue

#315 

# Checklist


danstarns (approved):
Nice fix.

PR #321, opened 2021-07-13
Jwt type

danstarns:
# Description

 Previously a string and a little bit strange to work within typescript.


danstarns:
> To 2.0.0 instead? At what point do we stop merging to master? Just thinking about merges, they're becoming a bit cumbersome.

Sure

darrellwarde (approved):

PR #330, opened 2021-07-17
Fix errors on unauthenticated requests

mathix420:
# Description

Force `get-jwt.ts` to always return an object.

# Issue

Fix the issue described here: https://github.com/neo4j/graphql/pull/303#discussion_r671148932

# Checklist

- [x] ~Documentation has been updated~ (not applicable)
- [x] ~Example applications have been updated~ (not applicable)
- [x] ~New files have copyright header~ (not applicable)

### DIFF
--- a/packages/graphql/src/auth/get-jwt.ts
+++ b/packages/graphql/src/auth/get-jwt.ts
@@ -27,7 +27,7 @@ const debug = Debug(DEBUG_AUTH);
 
 function getJWT(context: Context): any {
     const jwtConfig = context.neoSchema.config?.jwt;
-    let result;
+    let result = {};
 
     if (!jwtConfig) {
         debug("JWT not configured");
@@ -67,7 +67,7 @@ function getJWT(context: Context): any {
         if (jwtConfig.noVerify) {
             debug("Skipping verifying JWT as noVerify is not set");
 
-            result = jsonwebtoken.decode(token);
+            result = jsonwebtoken.decode(token) || {};
         } else {
             debug("Verifying JWT");
 


darrellwarde:
This really needs to be merged into `master`, but changing the base branch results in a whole load of merge conflicts!

However, I'm really trying to understand what this is trying to achieve. I followed through the conversation at https://github.com/neo4j/graphql/pull/303/files?file-filters%5B%5D=.adoc&file-filters%5B%5D=.ts#r671148932, and I'm trying to understand what results you would expect to see here. An empty array? All of the videos?

My assumption when I added the check in, was that if you have a field or type with an `@auth` directive, surely there needs to be some JWT available to check against in the context?

mathix420:
Hi @darrellwarde,

Ok, I'm going to open a new MR on `master`.

I think `where` and `bind` rules within `@auth` should not raise any error in case of an unauthenticated request.
In the example from my comment, I was expecting an empty array.

Here is my full use-case:
```graphql
type Video @exclude(operations: [CREATE, DELETE]) {
    // ...
}
extend type Video @auth(rules: [
    { where: { OR: [{ publisher: { id: "$jwt.sub" } }, { processing: "published" }] } },
    { operations: [UPDATE], allow: { publisher: { id: "$jwt.sub" } } },
])
```
So anyone can see a published video, and only its publisher can update it or see it even unpublished. So I only expect an `Unauthenticated` error if someone tries to update a video without being its publisher.

I use a custom directive to raise an exception to block unauthenticated requests on specific fields.
`@requireAuth`:
```typescript
class RequireAuthDirective extends SchemaDirectiveVisitor<any, any> {
  public visitFieldDefinition(field: GraphQLField<any, any>) {
    const { resolve } = field;

    field.resolve = async function(...args) {
      const [, , context] = args;
      if (!context?.jwt?.logedIn) throw new AuthenticationError("Bad authentication")
      return resolve.apply(this, args);
    }
  }
}
```

darrellwarde:
> Hi @darrellwarde,
> 
> Ok, I'm going to open a new MR on `master`.
> 
> I think `where` and `bind` rules within `@auth` should not raise any error in case of an unauthenticated request.
> In the example from my comment, I was expecting an empty array.
> 
> Here is my full use-case:
> 
> ```graphql
> type Video @exclude(operations: [CREATE, DELETE]) {
>     // ...
> }
> extend type Video @auth(rules: [
>     { where: { OR: [{ publisher: { id: "$jwt.sub" } }, { processing: "published" }] } },
>     { operations: [UPDATE], allow: { publisher: { id: "$jwt.sub" } } },
> ])
> ```
> 
> So anyone can see a published video, and only its publisher can update it or see it even unpublished. So I only expect an `Unauthenticated` error if someone tries to update a video without being its publisher.
> 
> I use a custom directive to raise an exception to block unauthenticated requests on specific fields.
> `@requireAuth`:
> 
> ```ts
> class RequireAuthDirective extends SchemaDirectiveVisitor<any, any> {
>   public visitFieldDefinition(field: GraphQLField<any, any>) {
>     const { resolve } = field;
> 
>     field.resolve = async function(...args) {
>       const [, , context] = args;
>       if (!context?.jwt?.logedIn) throw new AuthenticationError("Bad authentication")
>       return resolve.apply(this, args);
>     }
>   }
> }
> ```

I guess what I'm trying to get at is, what does the Cypher look like? Like for the Video example, with the where clause:

```
MATCH (this:Video)
WHERE this.id = ???????????????
RETURN this
```

How can we generate the Cypher if there is no JWT to put in it? Maybe I'm misunderstanding the problem, @danstarns has a much better understanding of auth than I do but sadly he's out of the office today.

mathix420:
Yes I got your point, I'm not sure what would be best to handle this use case.

- Bypass the query and return null/empty/default value? I don't even know if it's feasible.
- Use `null` in the Cypher query, but it would be an undefined behavior.

I'm relatively new to the Neo4J ecosystem, so I'm not sure I can help you this way. But I do think that this is a pretty common use case and this should not throw any error.

danstarns:
Just linking this in as I think it has relevance https://github.com/neo4j/graphql/pull/281/

darrellwarde:
Superseded by #342, let's discuss there!

Issue #332, opened 2021-07-18
Error thrown when totalCount is only field on connection field in 2.0.0-alpha.3


dmoree:
**Describe the bug**
When querying `totalCount` on a connection field an error `"Cannot read property 'fieldsByTypeName' of undefined"` is thrown if it is the only field.

**Type definitions**
Based on `neo-push` example

**To Reproduce**

1. Clone issue repository
```
git clone https://github.com/dmoree/neo4j-graphql-issue-total-count.git
```
2. Seed database and start server
```
cd neo4j-graphql-issue-total-count && yarn install && yarn seed && yarn start
```
3. run following `query` in GraphiQL at `http://localhost:4000/graphql` (or as defined in `.env`)
```graphql
query TotalCountOfComments {
  posts {
    commentsConnection {
      totalCount
    }
  }
}

```

**Expected behavior**
Would expect `totalCount: 3` as seen in the following `query` (see graph below):
```graphql
query TotalCountOfCommentsWithEdges {
  posts {
    commentsConnection {
      totalCount
      edges {
        node {
          id
        }
      }
    }
  }
}
```

**Screenshots**
![Screen Shot 2021-07-18 at 12 33 21 AM](https://user-images.githubusercontent.com/10360334/126059461-b73cd607-598a-441b-a9a8-5440b4aea7f9.png)

**System (please complete the following information):**

-   OS: macOS 11.4
-   Version: `@neo4j/graphql@2.0.0-alpha.3`
-   Node.js version: `14.16.0`

litewarp:
So I just submitted a PR, but for some context, I never anticipated that someone would only want the totalCount when calling a connection, so the cypher was not written with that option in mind. 

An easy work-around is in the PR, but if the decision is made to rewrite the cypher so that the database only returns totalCount (instead of returning edges + totalCount to the server), it may require help from someone with more cypher experience.

litewarp:
Nevermind. I think I fixed the cypher so #333 should now fully resolve the issue.

darrellwarde:
The fix for this will be released in 2.0.0-alpha.4 shortly, please let us know if it works for you! 🙂

PR #333, opened 2021-07-18
Return totalCount if no edges are requested (Fix 332)

litewarp:
~This is a naive fix for #332. Specifically, it returns an empty object for "edges" in the `create-connection-and-params.ts` function instead of undefined. This allows the rest of the function to continue without erring but it also means that the database is still returning the edges to the server; the server just ignores it.~

~A full fix would be to rewrite the cypher so that only the totalCount is returned. But that's trickier, so I'll defer to what @darrellwarde and @danstarns want to do.~

**Update**: Cypher should now request only totalCount if that's the only field requested in the tree.

**But** see [this comment](https://github.com/neo4j/graphql/issues/332#issuecomment-882124477) about a possible error in the projection of edges on a connection field that resolves to a union


darrellwarde:
Hey @litewarp, thanks for this, but I was working on this at the tail-end of last week and #334 covers this as well as returning an accurate count for union types. Can we continue discussions there? 🙂

Issue #327, opened 2021-07-16
Undefined cursor if no connection is found (v2.0.0-alpha.3)


kristjanmik:
**Describe the bug**
After the #282 merge, I'm getting the error: Cannot read property 'cursor' of undefined.
This only happens when no nodes are connected with a relationship, but the error is not raised if relationships and nodes are found.

**Type definitions**
```gql
type Query {
  activity(id: ID!): Activity
}

interface ActivityHasMediaProperties {
  timestamp: DateTime
}

type Activity {
  media: [Media]!
    @relationship(
      type: "HAS_MEDIA"
      properties: "ActivityHasMediaProperties"
      direction: OUT
    )
}
```

**To Reproduce**

**Querying for activity1 works. This is expected as at least one relationship exists.**
![Screenshot 2021-07-16 at 12 18 26](https://user-images.githubusercontent.com/911107/125946277-910c262d-be49-46d9-b1ad-984661007fd4.png)

```gql

query Activity {
  activity(id: "activity1") {
    id
    mediaConnection{
      edges{
        timestamp
      }
    }
  }
}
```

Results
```json
{
  "data": {
    "activity": {
      "id": "activity1",
      "mediaConnection": {
        "edges": [
          {
            "timestamp": "2021-04-14T00:00:00.000Z"
          }
        ]
      }
    }
  }
}
```

**Querying for activity3 results in the above error. This is not expected, should return an empty array.**
![Screenshot 2021-07-16 at 12 18 37](https://user-images.githubusercontent.com/911107/125946370-d2ab2050-fe00-4138-b042-dc1b7ccb035d.png)

```gql
query Activity {
  activity(id: "activity3") {
    id
    mediaConnection{
      edges{
        timestamp
      }
    }
  }
}
```

Results
```json
{
  "errors": [
    {
      "message": "Cannot read property 'cursor' of undefined",
      "locations": [
        {
          "line": 151,
          "column": 3
        }
      ],
      "path": [
        "activity"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "Error: Cannot read property 'cursor' of undefined",
            "    at Model.<anonymous> (/...masked.../node_modules/@neo4j/graphql-ogm/src/classes/Model.ts:105:19)",
            "    at Generator.next (<anonymous>)",
            "    at fulfilled (/...masked.../node_modules/@neo4j/graphql-ogm/dist/classes/Model.js:23:58)",
[...]

PR #334, opened 2021-07-19
Return totalCount if no edges, and prevent errors for accessing undefined properties

darrellwarde:
# Description

Addresses some bugs introduced with cursor-based pagination, and adds more test cases.

- No longer assume that `edges` in the resolve tree - this was safe to assume before, but now `totalCount` or `pageInfo` can be present instead
- Project `__resolveType` for unions to allow for calculation of `totalCount`, but filter out before return to match fragments in the selection set
- Don't return `edges` in Cypher unless it's necessary - this is when either `edges` or `pageInfo` have been asked for

Problems still to discuss and address:

- ~~`startCursor` and `endCursor` - nullable or non-nullable? The spec says that they are non-nullable (https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo.Fields), but there is a longstanding PR to change this (https://github.com/facebook/relay/pull/2655), so do we go with what is currently accurate, or what it "should" be?~~
- Pagination on union connection fields - currently the `first` and `after` arguments are not added to union fields, and if they are, what should the behaviour be?

# Issue

- #327
- #332 

# Checklist

### DIFF
--- a/packages/graphql/src/schema/make-augmented-schema.ts
+++ b/packages/graphql/src/schema/make-augmented-schema.ts
@@ -121,8 +121,8 @@ function makeAugmentedSchema(
         fields: {
             hasNextPage: "Boolean!",
             hasPreviousPage: "Boolean!",
-            startCursor: "String!",
-            endCursor: "String!",
+            startCursor: "String",
+            endCursor: "String",
         },
     });
 
--- a/packages/graphql/src/schema/make-augmented-schema.ts
+++ b/packages/graphql/src/schema/make-augmented-schema.ts
@@ -1130,9 +1130,20 @@ function makeAugmentedSchema(
 
                         const totalCount = isInt(count) ? count.toNumber() : count;
 
+                        const unionEdges = edges?.filter((edge) => {
+                            if (
+                                Object.keys(edge.node).length === 1 &&
+                                Object.prototype.hasOwnProperty.call(edge.node, "__resolveType")
+                            ) {
+                                return false;
+                            }
+
+                            return true;
+                        });
+
                         return {
                             totalCount,
-                            ...createConnectionWithEdgeProperties(edges, args, totalCount),
+                            ...createConnectionWithEdgeProperties(unionEdges, args, totalCount),
                         };
                     },
                 },
--- a/packages/graphql/src/schema/pagination.ts
+++ b/packages/graphql/src/schema/pagination.ts
@@ -24,7 +24,7 @@ import { Integer, isInt } from "neo4j-driver";
  * Adapted from graphql-relay-js ConnectionFromArraySlice
  */
 export function createConnectionWithEdgeProperties(
-    arraySlice: { node: Record<string, any>; [key: string]: any }[],
+    arraySlice: { node: Record<string, any>; [key: string]: any }[] = [],
     args: { after?: string; first?: number } = {},
     totalCount: number
 ) {
@@ -54,8 +54,8 @@ export function createConnectionWithEdgeProperties(
     return {
         edges,
         pageInfo: {
-            startCursor: firstEdge.cursor,
-            endCursor: lastEdge.cursor,
+            startCursor: firstEdge?.cursor,
+            endCursor: lastEdge?.cursor,
             hasPreviousPage: lastEdgeCursor > 0,
             hasNextPage: typeof first === "number" ? sliceEnd < totalCount : false,
         },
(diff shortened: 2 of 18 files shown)


litewarp:
Just putting my 2 cents in here about the nullability of the start and end cursors. While the official spec says they are non null, I have yet to see an implementation in the wild that doesn't null them out when no edges are present. See, e.g., the graphql-relay package. 

I also don't know what the alternative would be - an empty string?

litewarp:
Also the *official* [relay example](https://relay.dev/docs/guides/graphql-server-specification/#schema) has the start and end cursors as nullable

darrellwarde:
> Also the _official_ [relay example](https://relay.dev/docs/guides/graphql-server-specification/#schema) has the start and end cursors as nullable

Wow, it really must just be that the spec is well out of date, good find! Will update the schema now and add some tests.

darrellwarde:
Another interesting difference in there is that `node` is nullable:

```graphql
type ShipEdge {
  cursor: String!
  node: Ship
}
```

This would make projecting unions in the way we like a lot easier, so tempted to take that from there as well...

litewarp:
For union fields, I'm not sure what the issue is with doing pagination on the union fields. If the cypher "collects" the results of the union, shouldn't pagination perform the same as a non-union field?

darrellwarde:
> For union fields, I'm not sure what the issue is with doing pagination. If the cypher "collects" the results of the union, shouldn't you be able to slice the collection in the same manner as a non-union field?

Well at the moment, the pagination arguments aren't even applied to union connection fields, so that would be a start! 😝 

But the complexity comes in the decision of what we paginate on. For unions, we only return members which have fragments in the selection set, so for a union `union Production = Movie | Series`, the following would apply:

```graphql
# Returns only Movies
{
  actors {
    productionsConnection {
      edges {
        node {
          ... on Movie {
            title
          }
        }
      }
    }
  }
}

# Returns only Series
{
  actors {
    productionsConnection {
      edges {
        node {
          ... on Series {
            title
          }
        }
      }
    }
  }
}

# Returns Movies and Series
{
  actors {
    productionsConnection {
      edges {
        node {
          ... on Movie {
            title
          }
          ... on Series {
            title
          }
        }
      }
    }
  }
}
```

This is quite an opinionated take on unions, as normally you would receive an empty object, even if there isn't a fragment for that type. 

At the moment, `totalCount` is calculated on _all_ members of the union, whether they're in the selection set or not. So it's a bit of work to align the lot and come up with a good pagination solution!

Probably not one for this particular PR, but worth thinking about.

litewarp:
Where does the filtering take place? Is it embedded in the cypher query?

If it's in the GraphQL layer, you could possibly calculate the cursors before applying the filters. The only issue in that case is that you might return less than the limit since if you pull 20 items from the db but 3 are empty, you'll only return 17. So while you can pick up where you left off easily (thanks cursor pagination!) I still think the numbers would get out of sync if you were trying to do page-based pagination (versus infinite-scroll, e.g.)

So yeah the only way I can see how to do it consistently is to return an edge with a possibly null node.

litewarp:
Ok, three ways I think you can do this:

1. Filter out the empty nodes prior to collection in the Cypher command. 

I'm not a cypher expert, but couldn't you run a WHERE filter on the returned edges to filter out the edges before returning it to the GraphQL layer? For example:

```
CALL {
    WITH this
    CALL {
        WITH this
        OPTIONAL MATCH (this)-[this_wrote:WROTE]->(this_Book:Book)
        WITH { words: this_wrote.words, node: { __resolveType: "Book", title: this_Book.title } } AS edge
        RETURN edge
    UNION
        WITH this
        OPTIONAL MATCH (this)-[this_wrote:WROTE]->(this_Journal:Journal)
        WITH { words: this_wrote.words, node: { __resolveType: "Journal", subject: this_Journal.subject } } AS edge
        RETURN edge
    }
    ## <------ FILTER OUT EMPTY NODES HERE?
    WITH collect(edge) as edges, count(edge) as totalCount
    RETURN { edges: edges, totalCount: totalCount } AS publicationsConnection
}
```

2. Drop the opinionated implementation on unions and return empty nodes. 

I'm in favor of this one. It keeps the count consistent, and the worst case scenario is that you get an edge like the below, which is still helpful as it indicates the resolveType and existence of a node, even if no properties are returned. 

```
// empty edge

{
  cursor: "1234579348",
  node: {
    __resolveType: "Movie"
  }

```

3. Keep the opinionated implementation and do fuzzy math to keep the cursors consistent.

You can swap the order of the union filter and the createConnection functions so that you generate the connection based on all the returned values but then filter out the empty ones before returning to the client. The forward cursor-based pagination will still work since the cursors will be calculated in order, but it means that the number of items requested by the user might be different than what is returned. 

There's also some wonky math with the totalCount you'd have to keep track of, but I suppose it's possible.

darrellwarde:
Hey @litewarp, I've referenced your comment above in new issue #344 so that we discuss the approach to unions there, and get this PR reviewed and merged! 🙂

danstarns (approved):
Looks good to me, nice coverage.

PR #342, opened 2021-07-21
Throw on missing auth param

danstarns:
# Description

Related: https://github.com/neo4j/graphql/pull/330
Related: https://github.com/neo4j/graphql/pull/303#discussion_r671148932

In this PR I have done two things, the first is to remove the error of throwing if the JWT is not present - Before we were throwing an error too early and not allowing users to use the `$context` syntax. Not only were we throwing an error too early, but the error being thrown was, in my opinion, [giving away too much information](https://owasp.org/www-community/Improper_Error_Handling). We have the `isAuthenticated` flag that one should use to enforce that the JWT can be decoded correctly. 

Secondly, in this PR, I have added an error that is thrown when either the `$jwt.xyz` or `$context.xyz` value(s) are `undefined` implying that they aren't present on the payload. My decision to throw an error when the values are `undefined`, over setting them to `null`, is due to two factors; passing `null` through could return particular pieces of data where a property is `null` and thus potentially leaking information. Secondly, specifying rules in your auth directive such as the three following: 

```gql
@auth(rules: [{ where: { publisher: "$jwt.sub" } }])
```

```gql
@auth(rules: [{ allow: { publisher: "$jwt.sub" } }])
```

```gql
@auth(rules: [{ bind: { publisher: "$jwt.sub" } }])
```

It saying "A request should have `$jwt.sub` in the payload" and if the value is not present... This PR makes the assumption that the request is `Unauthenticated` based on the rules and requirements defined in the directive. This PR will throw the `Unauthentacted` error to be inconspicuous towards any potential attack. 

Thank you @mathix420 for the work on this, and even working on a fix, I apologise if this decision doesn't work for your use case and I'm willing to listen and try to understand your requirements better. 

# Checklist


mathix420:
It's great you have fixed these issues!

The only issue I have with these changes is that this particular rule will not work correctly as the first condition will raise an exception.
```graphql
{ where: { OR: [{ publisher: "$jwt.sub" }, { processing: "published" }] } }
```

Maybe a parameter similar to `isAuthenticated` would be a good idea, like `allowUnauthenticated`. This way we can avoid raising any error if no auth is present. The dev need to explicitly set this setting, so he will be aware of probable security issues.
```graphql
{ where: { OR: [{ publisher: "$jwt.sub" }, { processing: "published" }] }, allowUnauthenticated: true }
```

I don't know if you judge this use-case important enough for this to be implemented, if not I will try to solve this with a quick workaround. In any case nice work!

darrellwarde:
> It's great you have fixed these issues!
> 
> The only issue I have with these changes is that this particular rule will not work correctly as the first condition will raise an exception.
> 
> ```graphql
> { where: { OR: [{ publisher: "$jwt.sub" }, { processing: "published" }] } }
> ```
> 
> Maybe a parameter similar to `isAuthenticated` would be a good idea, like `allowUnauthenticated`. This way we can avoid raising any error if no auth is present. The dev need to explicitly set this setting, so he will be aware of probable security issues.
> 
> ```graphql
> { where: { OR: [{ publisher: "$jwt.sub" }, { processing: "published" }] }, allowUnauthenticated: true }
> ```
> 
> I don't know if you judge this use-case important enough for this to be implemented, if not I will try to solve this with a quick workaround. In any case nice work!

I finally understand what you mean! 

I think the solution which @danstarns has presented here is the correct solution for the moment, but perhaps this new feature you're suggesting could be a good solution. I'll raise a separate issue off your comment!

Honestly, your use case is slightly odd to us, we never imagined that people would compare to "static" values in the `@auth` directive, which is why we found it so confusing. It's really not a behaviour we would even expect to work really!

darrellwarde (commented):
A much better solution than what I contributed! 😄  Worth having some integration tests?

danstarns:
> A much better solution than what I contributed! 😄 Worth having some integration tests?

Sure added them in.

darrellwarde (approved):
This looks good! Hopefully there's not any other edge cases we've missed.

Issue #345, opened 2021-07-21
allowUnauthenticated @auth option to disable error throwing


darrellwarde:
It's great you have fixed these issues!

The only issue I have with these changes is that this particular rule will not work correctly as the first condition will raise an exception.
```graphql
{ where: { OR: [{ publisher: "$jwt.sub" }, { processing: "published" }] } }
```

Maybe a parameter similar to `isAuthenticated` would be a good idea, like `allowUnauthenticated`. This way we can avoid raising any error if no auth is present. The dev need to explicitly set this setting, so he will be aware of probable security issues.
```graphql
{ where: { OR: [{ publisher: "$jwt.sub" }, { processing: "published" }] }, allowUnauthenticated: true }
```

I don't know if you judge this use-case important enough for this to be implemented, if not I will try to solve this with a quick workaround. In any case nice work!

_Originally posted by @mathix420 in https://github.com/neo4j/graphql/issues/342#issuecomment-884061188_

darrellwarde:
Done in #355!

PR #355, opened 2021-07-23
feat(graphql): add allowUnauthenticated parameter to auth rules

mathix420:
# Description

Add allowUnauthenticated parameter in auth rules to avoid throwing exceptions if no auth is provided.

**Changes:**
- Update documentation `auth/authentication`.
- If `paramValue` is `undefined` inject `false` instead of the previous condition `key = $param`, as it should always be falsy
- Fix a small issue: if a `paramValue` is `null`, `apoc.validate` raise an exception if the check is done like this `value = null`. So I just replaced the condition for it to be `value IS null` ([more infos](https://community.neo4j.com/t/neo-clienterror-procedure-procedurecallfailed-with-apoc-do-when/14017/12))
- Fix the 4.3.0 Neo4J deprecation of `EXISTS()` with recommanded `IS NOT NULL`. ([more infos](https://neo4j.com/docs/cypher-manual/current/deprecations-additions-removals-compatibility/#:~:text=ASSERT%20exists(node.property)))

# Issue

https://github.com/neo4j/graphql/issues/345

# Checklist

### DIFF
--- a/docs/asciidoc/auth/authentication.adoc
+++ b/docs/asciidoc/auth/authentication.adoc
@@ -96,7 +96,7 @@ extend type Todo @auth(rules: [{ isAuthenticated: true }])
 == `allowUnauthenticated`
 In some cases, you may want to allow unauthenticated requests while also having auth-based rules. You can use the `allowUnauthenticated` parameter to avoid throwing an exception if no auth is present in the context.
 
-In the example below, only the publisher can see his blog posts even if it is not published yet. Once the blog post is published, anyone can see it.
+In the example below, only the publisher can see his blog posts if it is not published yet. Once the blog post is published, anyone can see it.
 
 [source, graphql]
 ----
--- a/docs/asciidoc/auth/authentication.adoc
+++ b/docs/asciidoc/auth/authentication.adoc
@@ -92,3 +92,26 @@ type Todo {
 
 extend type Todo @auth(rules: [{ isAuthenticated: true }])
 ----
+
+== `allowUnauthenticated`
+In some cases, you may want to allow unauthenticated requests while also having auth-based rules. You can use the `allowUnauthenticated` parameter to avoid throwing an exception if no auth is present in the context.
+
+In the example below, only the publisher can see his blog posts even if it is not published yet. Once the blog post is published, anyone can see it.
+
+[source, graphql]
+----
+type BlogPost {
+    id: ID!
+    publisher: User!
+    published: Boolean!
+}
+
+extend type BlogPost {
+    password: String! @auth(rules: [
+        { operations: [READ], where: { OR: [
+            { publisher: "$jwt.sub" },
+            { published: true },
+        ] }, allowUnauthenticated: true }
+    ])
+}
+----
--- a/docs/asciidoc/auth/authentication.adoc
+++ b/docs/asciidoc/auth/authentication.adoc
@@ -98,22 +98,20 @@ In some cases, you may want to allow unauthenticated requests while also having 
 
 In the example below, only the publisher can see his blog posts even if it is not published yet. Once the blog post is published, anyone can see it.
 
-WARNING: In case of unauthenticated requests, auth parameters will be replaced by `null` in the cypher query. In this case, `publisher: "$jwt.sub"` will become `publisher: null`; This can lead to unwanted results if the field used for querying is nullable in Neo4J.
-
 [source, graphql]
 ----
-type BlogPost {
+type BlogPost
+    @auth(
+        rules: [
+            {
+                operations: [READ]
+                where: { OR: [{ publisher: "$jwt.sub" }, { published: true }] }
+                allowUnauthenticated: true
+            }
+        ]
+    ) {
     id: ID!
-    publisher: User!
+    publisher: String!
     published: Boolean!
-}
-
-extend type BlogPost {
-    password: String! @auth(rules: [
-        { operations: [READ], where: { OR: [
-            { publisher: "$jwt.sub" },
-            { published: true },
-        ] }, allowUnauthenticated: true }
-    ])
 }
 ----
--- a/docs/asciidoc/auth/authentication.adoc
+++ b/docs/asciidoc/auth/authentication.adoc
@@ -98,6 +98,8 @@ In some cases, you may want to allow unauthenticated requests while also having 
 
 In the example below, only the publisher can see his blog posts even if it is not published yet. Once the blog post is published, anyone can see it.
 
+WARNING: In case of unauthenticated requests, auth parameters will be replaced by `null` in the cypher query. In this case, `publisher: "$jwt.sub"` will become `publisher: null`; This can lead to unwanted results if the field used for querying is nullable in Neo4J.
+
 [source, graphql]
 ----
 type BlogPost {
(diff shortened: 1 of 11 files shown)


darrellwarde:
Hey @mathix420, thanks for the contribution!

This really needs some integration tests to prove that unexpected errors won't be thrown on execution of the Cypher.

mathix420:
@darrellwarde Should I open 2 separate issues for explaining more in details the `EXISTS()` change and `apoc.validate` behavior on null comparisons?

darrellwarde:
> @darrellwarde Should I open 2 separate issues to explain more in details the `EXISTS()` change and `apoc.validate` behavior on null comparisons?

Hey @mathix420, sorry for the slow reply here! No need to open 2 PRs, happy to review as a batch - was just waiting on some integration tests but realise now I missed your commit!

darrellwarde:
It does look like you have some unit test failures though which will need reviewing.

mathix420:
> It does look like you have some unit test failures though which will need reviewing.

Yup, sorry dumb mistake I was focalized on `create-auth-and-params` tests :sweat_smile: 
Everything should be fix now.

I'm not sure if I should replace all others occurences of `EXISTS()` in this PR, should I?

danstarns:
> > It does look like you have some unit test failures though which will need reviewing.
> 
> Yup, sorry dumb mistake I was focalized on `create-auth-and-params` tests 😅
> Everything should be fix now.
> 
> I'm not sure if I should replace all others occurences of `EXISTS()` in this PR, should I?

Hey, no need to replace all other `EXISTS` we can do this in another PR. Thanks for pointing out its deprecation.

mathix420:
@danstarns great, have a nice day!

packages/graphql/src/schema/get-auth.ts
```diff
@@ -21,7 +21,13 @@ import { DirectiveNode, valueFromASTUntyped } from "graphql";
...
 
 const validOperations: AuthOperations[] = ["CREATE", "READ", "UPDATE", "DELETE", "CONNECT", "DISCONNECT"];
-const validFields = ["operations", "AND", "OR", "allow", "where", "bind", "isAuthenticated", "roles"];
+const validFields = [
+    "operations",
+    "AND", "OR",
+    "allow", "where", "bind",
+    "isAuthenticated", "allowUnauthenticated",
+    "roles"
+];
```
darrellwarde, line 30: I don't like this array formatting, surely ESLint/Prettier didn't format it like this?

docs/asciidoc/auth/authentication.adoc
```diff
@@ -92,3 +92,28 @@ type Todo {
...
+}
+
+extend type BlogPost {
+    password: String! @auth(rules: [
+        { operations: [READ], where: { OR: [
+            { publisher: "$jwt.sub" },
+            { published: true },
+        ] }, allowUnauthenticated: true }
+    ])
+}
```
darrellwarde, line 118: These type definitions don't match the description in the paragraph above, at the moment they just have auth on the password field. Additionally, `{ publisher: "$jwt.sub" }` wouldn't work because `publisher` is an object type and not a string type. So it should probably look something like the following:

```suggestion
type BlogPost
    @auth(
        rules: [
            {
                operations: [READ]
                where: { OR: [{ publisher: "$jwt.sub" }, { published: true }] }
                allowUnauthenticated: true
            }
        ]
    ) {
    id: ID!
    publisher: String!
    published: Boolean!
}
```

I must add, that the only reason that the `{ published: true }` is by complete and utter fluke - I'm almost tempted to say that it's a bad idea to put this example in the docs because of this, because it might not work in all cases. I imagine it certainly wouldn't work if there was any recursion down relationship fields.

packages/graphql/src/translate/create-auth-and-params.test.ts
```diff
@@ -685,5 +685,141 @@ describe("createAuthAndParams", () => {
...
+            const context: Context = { neoSchema, jwt: {} };
+
+            const result = createAuthAndParams({
+                context,
+                entity: node,
+                operation: "READ",
+                allow: { parentNode: node, varName: "this" },
+            });
+
+            expect(result[1]).toEqual({});
```
darrellwarde, line 754: Could we assert what we expect `result[0]` to be?

packages/graphql/src/translate/create-auth-and-params.test.ts
```diff
@@ -685,5 +685,141 @@ describe("createAuthAndParams", () => {
...
```
darrellwarde, line 822: Could we assert what we expect `result[0]` to be?

packages/graphql/src/translate/create-auth-and-params.ts
```diff
@@ -102,13 +106,19 @@ function createAuthPredicate({
...
                     throw new Neo4jGraphQLAuthenticationError("Unauthenticated");
                 }
 
-                const param = `${chainStr}_${key}`;
-                res.params[param] = paramValue;
-                res.strs.push(`EXISTS(${varName}.${key}) AND ${varName}.${key} = $${param}`);
+                if (paramValue === undefined) {
+                    res.strs.push("false");
+                } else if (paramValue === null) {
+                    res.strs.push(`${varName}.${key} IS null`);
```
darrellwarde, line 116: Real nitpicky, but could we capitalise `null` here?

darrellwarde (changes requested):
Thanks for PR, always grateful for contributions! 🙂 

Please see code comments, and also, I would like to see some integration tests for more complex cases. For example, what happens with `allow` and `allowUnauthenticated` if there is one post which is published and one that isn't?

packages/graphql/src/translate/create-auth-and-params.ts
mathix420: Sure, I have replaced that.
This specific condition is required as an issue appear if we perform a `null` comparison with `=`, the result would be `null` and not a boolean. This caused APOC.validate to raise an exception. https://community.neo4j.com/t/neo-clienterror-procedure-procedurecallfailed-with-apoc-do-when/14017/12

packages/graphql/src/schema/get-auth.ts
mathix420: Sure, I just put back the previous formatting.

docs/asciidoc/auth/authentication.adoc
mathix420: I'm not sure to fully understand what could cause an issue in a recursive relationship, I assume this should work the same as the classic `where` auth rule should work. But I'm probably misunderstanding something, can you please give me an example of where you think this shouldn't work.

darrellwarde: I can't give a specific example, it's more just the fact that the only reason this kind of thing works at all is completely by accident, so it's completely untested and we have no actual confidence that it works. Based on that, there's a higher chance that there will be some edge cases where it doesn't work, but we can face them if and when they arrive!

darrellwarde (changes requested):
Literally just a single word documentation change, which I think can just be applied from my suggestion once I've finished this review!

darrellwarde (approved):
As expected, I was able to just apply my word change suggestion.

In which case, this all looks good! 🚀  Thanks for the improvement, and working through changes so quickly!

danstarns:
Thanks @mathix420

danstarns (approved):

Issue #350, opened 2021-07-22
Attributes get lost when calling a field multiple times using aliases


acmeguy:
**Describe the bug**
Only the last set of parameters is used when the same field is called multiple time using aliases.

**Type definitions**
type AnyThing {
    value (some_value: String!): String @cypher(statement: """RETURN $some_value""")
}

**To Reproduce**
Run the following query:
```
query {
  AnyThing(label: "Country") {
    A_val: value(some_value: "A")
    B_val: value(some_value: "B")
    C_val: value(some_value: "C")
  }
}
```
See error in the returned set:
```
  "data": {
    "AnyThing": [
      {
        "A_val": "C",
        "B_val": "C",
        "C_val": "C"
      },
]}
```

**Expected behavior**
Should return this:
```
  "data": {
    "AnyThing": [
      {
        "A_val": "A",
        "B_val": "B",
        "C_val": "C"
      },
]}
```
-   OS: [e.g. Linux]
-   Version: ^1.2.2
-   Node.js version: 14.16.0

dmoree:
@danstarns While the example above is using the `@cypher` directive, this issue persists even with `@relationship` fields. This may ultimately be related to how `@neo4j/graphql` deals with aliased fields.

Below is a reproduction based on the `neo-push` example that adds a `flagged` field to a `Comment`.

#### To Reproduce

1. Clone repository
```
git clone https://github.com/dmoree/neo4j-graphql-issue-350.git
```
2. Seed and start server
```
cd neo4j-graphql-issue-350 && yarn install && yarn seed && yarn start
```
3. Run each `query` in GraphiQL at `http://localhost:4000/graphql` (or as defined in `.env`)
##### Flagged
```graphql
query PostsWithFlaggedComments {
  posts {
    comments(where: { flagged: true }) {
      id
      flagged
    }
  }
}
```
```
{
  "data": {
    "posts": [
      {
        "comments": [
          {
            "id": "163a5bb0-ee16-4cf2-807d-84586f926c2d",
            "flagged": true
          }
        ]
      }
    ]
  }
}
```
##### Unflagged
```graphql
query PostsWithUnflaggedComments {
  posts {
    comments(where: { flagged: false }) {
      id
      flagged
    }
  }
}
```
```
{
  "data": {
    "posts": [
      {
        "comments": [
          {
            "id": "11ceb53d-0075-4f95-b3f3-cceb2d66aa0d",
            "flagged": false
          },
          {
[...]

PR #357, opened 2021-07-23
Fix/multiple cypher direcive on field using aliases

danstarns:
# Description

Just contains recreation tests at the moment. 

Will eventually fix: #350


danstarns:
https://github.com/neo4j/graphql/pull/359

PR #358, opened 2021-07-25
feat: Resolve union type in custom cypher resolver

john2094:
allows solving cypher with union type results, in the current behavior graphql crashes.

An example of this is the following:

                 type Movie {
                        title: String!
                        actors: [Actor] @relationship(type: "ACTED_IN", direction: IN)
                        directors: [Actor] @relationship(type: "DIRECTED", direction: IN)
                    }
                    type Director {
                        name: String!
                        movies: [Movie] @relationship(type: "DIRECTED", direction: OUT)
                    }
                    type Actor {
                        name: String!
                        movies: [Movie] @relationship(type: "ACTED_IN", direction: OUT)
                    }
                    type Query {

                        personInMovie(title: String!): [Person] @cypher(statement: """
                            MATCH (m:Movie {title: $title})<-[:ACTED_IN|DIRECTED]-(p)
                            RETURN p
                        """)
                    }

### DIFF
--- a/packages/graphql/src/schema/resolvers/cypher.ts
+++ b/packages/graphql/src/schema/resolvers/cypher.ts
@@ -18,6 +18,7 @@
  */
 
 import { isInt } from "neo4j-driver";
+import { UnionTypeDefinitionNode } from "graphql/language/ast";
 import { execute } from "../../utils";
 import { BaseField, Context } from "../../types";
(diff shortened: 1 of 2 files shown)


john2094:
I realized that it does not work correctly for certain cases, I will correct them and make a new pull request

PR #359, opened 2021-07-26
Fix for field with multiple aliases

dmoree:
# Description

This pull request adds the ability to alias fields multiple times in queries. Currently when translating a graqhql request into a cypher query, the field name is used as opposed to the alias provided. Since the default field resolver also uses the field name this went unnoticed as long as any field was aliased at most one time during a query. However, as noted in #350 and described in detail in https://github.com/neo4j/graphql/issues/350#issuecomment-885957460; if a field is aliased multiple times, the use of the field name in the projection causes the last defined aliased field to override the others.

Aliasing of fields also requires to modify the returned object from the `create` and `update` resolver by making the key the aliased field.

# Issue

- #350 

# Checklist

### DIFF
--- a/packages/graphql/src/schema/resolvers/create.ts
+++ b/packages/graphql/src/schema/resolvers/create.ts
@@ -19,13 +19,14 @@
 
 import camelCase from "camelcase";
 import pluralize from "pluralize";
+import { FieldNode, GraphQLResolveInfo } from "graphql";
 import { execute } from "../../utils";
 import { translateCreate } from "../../translate";
 import { Node } from "../../classes";
 import { Context } from "../../types";
 
 export default function createResolver({ node }: { node: Node }) {
-    async function resolve(_root: any, _args: any, _context: unknown) {
+    async function resolve(_root: any, _args: any, _context: unknown, _info: GraphQLResolveInfo) {
         const context = _context as Context;
         const [cypher, params] = translateCreate({ context, node });
 
@@ -36,8 +37,14 @@ export default function createResolver({ node }: { node: Node }) {
             context,
         });
 
+        const responseField = _info.fieldNodes[0].selectionSet?.selections.find(
+            (selection) => selection.kind === "Field" && selection.name.value === pluralize(camelCase(node.name))
+        ) as FieldNode; // Field exist by construction and must be selected as it is the only field.
+
+        const responseKey = responseField.alias ? responseField.alias.value : responseField.name.value;
+
         return {
-            [pluralize(camelCase(node.name))]: Object.values(result[0] || {}),
+            [responseKey]: Object.values(result[0] || {}),
         };
     }
 
--- a/packages/graphql/src/schema/resolvers/create.ts
+++ b/packages/graphql/src/schema/resolvers/create.ts
@@ -26,7 +26,7 @@ import { Node } from "../../classes";
 import { Context } from "../../types";
 
 export default function createResolver({ node }: { node: Node }) {
-    async function resolve(_root: any, _args: any, _context: unknown, _info: GraphQLResolveInfo) {
+    async function resolve(_root: any, _args: any, _context: unknown, info: GraphQLResolveInfo) {
         const context = _context as Context;
         const [cypher, params] = translateCreate({ context, node });
 
@@ -37,7 +37,7 @@ export default function createResolver({ node }: { node: Node }) {
             context,
         });
 
-        const responseField = _info.fieldNodes[0].selectionSet?.selections.find(
+        const responseField = info.fieldNodes[0].selectionSet?.selections.find(
             (selection) => selection.kind === "Field" && selection.name.value === pluralize(camelCase(node.name))
         ) as FieldNode; // Field exist by construction and must be selected as it is the only field.
 
--- a/packages/graphql/src/translate/create-projection-and-params.ts
+++ b/packages/graphql/src/translate/create-projection-and-params.ts
@@ -139,14 +139,7 @@ function createProjectionAndParams({
     chainStr?: string;
     varName: string;
 }): [string, any, ProjectionMeta?] {
-    function reducer(res: Res, [k, field]: [string, any]): Res {
-        let key = k;
-        const alias: string | undefined = field.alias !== field.name ? field.alias : undefined;
-
-        if (alias) {
-            key = field.name as string;
-        }
-
+    function reducer(res: Res, [key, field]: [string, any]): Res {
         let param = "";
         if (chainStr) {
             param = `${chainStr}_${key}`;
@@ -157,11 +150,11 @@ function createProjectionAndParams({
         const whereInput = field.args.where as GraphQLWhereArg;
         const optionsInput = field.args.options as GraphQLOptionsArg;
         const fieldFields = (field.fieldsByTypeName as unknown) as FieldsByTypeName;
-        const cypherField = node.cypherFields.find((x) => x.fieldName === key);
-        const relationField = node.relationFields.find((x) => x.fieldName === key);
-        const pointField = node.pointFields.find((x) => x.fieldName === key);
-        const dateTimeField = node.dateTimeFields.find((x) => x.fieldName === key);
-        const authableField = node.authableFields.find((x) => x.fieldName === key);
+        const cypherField = node.cypherFields.find((x) => x.fieldName === field.name);
+        const relationField = node.relationFields.find((x) => x.fieldName === field.name);
+        const pointField = node.pointFields.find((x) => x.fieldName === field.name);
+        const dateTimeField = node.dateTimeFields.find((x) => x.fieldName === field.name);
+        const authableField = node.authableFields.find((x) => x.fieldName === field.name);
 
         if (authableField) {
             if (authableField.auth) {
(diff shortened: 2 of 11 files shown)


packages/graphql/tests/integration/issues/350.int.test.ts
```diff
@@ -0,0 +1,149 @@
...
+import { Neo4jGraphQL } from "../../../src/classes";
+
+describe("https://github.com/neo4j/graphql/issues/350", () => {
+    let driver: Driver;
+
+    beforeAll(async () => {
+        driver = await neo4j();
+        const session = driver.session();
+        try {
+            await session.run(`
```
darrellwarde, line 34: We haven't done this elsewhere... But honestly, I'm not against it, our integration tests leave the database a real mess.

darrellwarde (changes requested):
Thank for for this @dmoree, and for the code comments explaining each step - this PR should save us a lot of time.

I've just tested this out, however, and it's still missing some cases, namely for `DateTime` and `Point` fields (and maybe more). For example, for the simple schema:

```graphql
type Movie {
    datetime: DateTime
    point: Point
}
```

If I run the query:

```graphql
{
  movies {
    d1: datetime
    d2: datetime
    p1: point {
      latitude
      longitude
    }
    p2: point {
      latitude
      longitude
    }
  }
}
```

That produces the Cypher:

```cypher
MATCH (this:Movie)
RETURN this { d1: apoc.date.convertFormat(toString(this.d1), "iso_zoned_date_time", "iso_offset_date_time"), d2: apoc.date.convertFormat(toString(this.d2), "iso_zoned_date_time", "iso_offset_date_time"), p1: { point: this.p1 }, p2: { point: this.p2 } } as this
```

Note `this.d1`, `this.p1`, etc.

This should be:

```cypher
MATCH (this:Movie)
RETURN this { d1: apoc.date.convertFormat(toString(this.datetime), "iso_zoned_date_time", "iso_offset_date_time"), d2: apoc.date.convertFormat(toString(this.datetime), "iso_zoned_date_time", "iso_offset_date_time"), p1: { point: this.point }, p2: { point: this.point } } as this
```

packages/graphql/tests/integration/issues/350.int.test.ts
danstarns: I'm not against it either, something for another time but I wonder if we can add this as a global post hook that runs once after all the tests are complete.

packages/graphql/src/schema/resolvers/create.ts
```diff
@@ -19,13 +19,14 @@
...
 import pluralize from "pluralize";
+import { FieldNode, GraphQLResolveInfo } from "graphql";
 import { execute } from "../../utils";
 import { translateCreate } from "../../translate";
 import { Node } from "../../classes";
 import { Context } from "../../types";
 
 export default function createResolver({ node }: { node: Node }) {
-    async function resolve(_root: any, _args: any, _context: unknown) {
+    async function resolve(_root: any, _args: any, _context: unknown, _info: GraphQLResolveInfo) {
```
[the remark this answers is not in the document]
dmoree: Of course, I'm not sure what I was thinking here.

dmoree:
I had only written tests specifically for the issue raised, so there was an oversight with regards to a `datetime` and `point` field; sorry for that. It should use `field.name` anytime `varName` is being accessed. Fixed in https://github.com/neo4j/graphql/pull/359/commits/3ea67c768f8e1945a30b08033cbb2987cb7406f4.

I also added the basic tests provided above for `datetime` and `point` fields.

packages/graphql/tests/integration/issues/350.int.test.ts
dmoree: I originally added this in a `beforeEach` but moved it to `beforeAll` since I only had one test. I wanted this to be able to use the `toHaveLength` method to ensure I was getting back the correct number of records. In order to ensure this I had to clear the database.

It would be nice to have a utility function that clears the database if needed. Something like:
```typescript
const clearDatabase = async (driver: Driver) => {
  const session = driver.session()
  try {
    await session.run(`
      MATCH (n)
      DETACH DELETE n
    `)
  } finally {
    await session.close()
  }
}
```

You could then call this in a `beforeEach` hook when writing tests if needed and in a global `afterAll` hook to clean up the database when done.

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -422,7 +415,11 @@ function createProjectionAndParams({
                     : `${key}: apoc.date.convertFormat(toString(${varName}.${key}), "iso_zoned_date_time", "iso_offset_date_time")`
```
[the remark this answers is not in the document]
dmoree: You are correct. Anytime `varName` is accessed it should use `field.name`. Fixed in https://github.com/neo4j/graphql/commit/3ea67c768f8e1945a30b08033cbb2987cb7406f4.

packages/graphql/tests/integration/issues/350.int.test.ts
darrellwarde: So, I just checked out this branch and I just cannot get the tests (in other suites) to pass. When I remove this clear-down, all of the tests pass except for this one. 

I know it's not right and our integration tests shouldn't be so co-dependent, but for the sake of getting this merged, can you change this test so that it can work without clearing the database down? 

For example, you could give the types unique names such as "Post350" and "Comment350", and then do `MATCH (p:Post350) DETACH DELETE (p)` (and the same for comments) before the tests, instead of clearing the whole database.

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -403,26 +396,30 @@ function createProjectionAndParams({
...
-                    ? `${key}: [p in ${varName}.${key} | { ${fields.join(", ")} }]`
+                    ? `${key}: [p in ${varName}.${field.name} | { ${fields.join(", ")} }]`
                     : `${key}: { ${fields.join(", ")} }`
             );
         } else if (dateTimeField) {
             res.projection.push(
                 dateTimeField.typeMeta.array
-                    ? `${key}: [ dt in ${varName}.${key} | apoc.date.convertFormat(toString(dt), "iso_zoned_date_time", "iso_offset_date_time") ]`
-                    : `${key}: apoc.date.convertFormat(toString(${varName}.${key}), "iso_zoned_date_time", "iso_offset_date_time")`
+                    ? `${key}: [ dt in ${varName}.${field.name} | apoc.date.convertFormat(toString(dt), "iso_zoned_date_time", "iso_offset_date_time") ]`
```
darrellwarde, line 414: I'm assuming `key` is equivalent to `field.alias`, hence why they're interchangeable?

darrellwarde (changes requested):
Thanks for making those changes!

I'm just going to have to ask you to change the integration test slightly as per my comment. It's 100% on us to tidy up the test framework, but for now I'd appreciate it if you could adjust as suggested so we can move forward quickly.

packages/graphql/src/translate/create-projection-and-params.ts
dmoree: That is correct. The key is whatever the field that is being requested is called. So if it has an alias then the key is equivalent to the alias. The projection produced a valid cypher request before due to these [lines](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/create-projection-and-params.ts#L143-L148) where it checks to see if there is an alias on the field and reassigns the key to be the field name.

packages/graphql/tests/integration/issues/350.int.test.ts
dmoree: Sorry; fixed in https://github.com/neo4j/graphql/pull/359/commits/a8ca8f7b4cfb7404ecde4cd0838560f4056013c5. I was having trouble with `toHaveLength`. I ultimately abandoned it, but it would be nice to have cleaner tests.

packages/graphql/tests/integration/issues/350.int.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+        const comment2Id = generate({
+            charset: "alphabetic",
+        });
+
+        const comment2Content = "comment 2 content";
+
+        const query = `
+            query {
+                posts {
```
danstarns, line 84:
```suggestion
                posts(where: { id: "${postId}" }) {
```

danstarns (changes requested):
Hey, could you add a where to this test, it should help with the failing. Here is the failing test for your info: 

```
✕ Retain attributes when aliasing the same field multiple times in a single query (200 ms)
  ● https://github.com/neo4j/graphql/issues/350 › Retain attributes when aliasing the same field multiple times in a single query
    expect(received).toContainEqual(expected) // deep equality
    Expected value: {"content": "comment 1 content", "flagged": true}
    Received array: []
      122 |             });
      123 |             expect(result.errors).toBeFalsy();
    > 124 |             expect(result?.data?.posts[0].flaggedComments).toContainEqual({
          |                                                            ^
      125 |                 content: comment1Content,
      126 |                 flagged: true,
      127 |             });
      at packages/graphql/tests/integration/issues/350.int.test.ts:124:60
      at fulfilled (packages/graphql/tests/integration/issues/350.int.test.ts:23:58)
          at runMicrotasks (<anonymous>)
```

danstarns (approved):
Thank you <3

PR #361, opened 2021-07-27
Fix/empty and or

### LINKED ISSUE #360: Missing optional variable in AND statement generates invalid Cypher query
**Describe the bug**
When building a GraphQL query using WHERE... AND with variables inside the AND statement, omitting an optional variable results in a Cypher error. 

By contrast, it is possible to omit optional variables that are used outside the AND statement. In that case the query runs and the whole term is ignored in the resulting Cypher query.

**Type definitions**
```graphql
type Event {
    id: ID!
    name: String
    start: DateTime
    end: DateTime
    activity: String
}
```

**To Reproduce**
1. Use a server running Neo4j GraphQL 1.2.3 and a schema containing the above type definition 
2. Use a client such as Apollo Client that can pass a GraphQL query with variables to the server
3. Run the query:

```graphql
query ($rangeStart: DateTime, $rangeEnd: DateTime, $activity: String) {
  events (
    where: { 
      AND: [
        { start_GTE: $rangeStart }
        { start_LTE: $rangeEnd }
        { activity: $activity }
      ]
    }
  ) 
  {
    start
    activity
  }  
}            
```

with the variables
```json
{
  "rangeStart": "2021-07-18T00:00:00+0100",
  "rangeEnd": "2021-07-18T23:59:59+0100"
}
```

... that is, omitting the optional variable `$activity`

**Expected behavior**
I would ideally like Event nodes with any value in the activity field to be included in the result

**Actual result**
The error message is returned to the client:
```json
{
  "errors": [
    {
      "message": "Invalid input ')': expected whitespace, comment or an expression (line 2, column 85 (offset: 103))\n\"WHERE (this.start >= $this_AND_start_GTE AND this.start <= $this_AND1_start_LTE AND )\"\n                                                                                     ^",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "events"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "code": "Neo.ClientError.Statement.SyntaxError",
          "name": "Neo4jError"
        }
      }
    }
  ],
  "data": null
}
```

- Windows
- @neo4j/graphql@1.2.3
- node 14.16.1

danstarns:
# Description

Fixes #360

### DIFF
--- a/packages/graphql/tests/integration/issues/360.int.test.ts
+++ b/packages/graphql/tests/integration/issues/360.int.test.ts
@@ -1,0 +1,204 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Driver } from "neo4j-driver";
+import { graphql } from "graphql";
+import { generate } from "randomstring";
+import camelcase from "camelcase";
+import pluralize from "pluralize";
+import { Neo4jGraphQL } from "../../../src/classes";
+import neo4j from "../neo4j";
+
+describe("360", () => {
+    let driver: Driver;
+
+    beforeAll(async () => {
+        driver = await neo4j();
+    });
+
+    afterAll(async () => {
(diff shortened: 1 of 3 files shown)


packages/graphql/tests/integration/issues/360.int.test.ts
```diff
@@ -0,0 +1,204 @@
...
+                    CREATE (:${type} {id: randomUUID(), name: randomUUID(), start: datetime(), end: datetime()})
+                `,
+                { rangeStart, rangeEnd }
+            );
+
+            const gqlResult = await graphql({
+                schema: neoSchema.schema,
+                source: query,
+                contextValue: { driver },
+            });
```
darrellwarde, line 196: If looking to reproduce the scenario in #360, this is missing `variableValues`, and the expected results will need changing.

```suggestion
            const gqlResult = await graphql({
                schema: neoSchema.schema,
                source: query,
                contextValue: { driver },
                variableValues: { rangeStart, rangeEnd },
            });
```

darrellwarde (changes requested):
Nice fix! One of the tests just needs updating. 🙂

darrellwarde (approved):
LGTM!

PR #362, opened 2021-07-27
A new approach for deciding which union members to return!

darrellwarde:
# Description

Excerpt from documentation which explains the change in detail:

Which union members are returned by a Query are dictated by the `where` filter applied.

For example, the following will return all user content, and you will specifically get the title of each blog.

```graphql
query GetUsersWithBlogs {
    users {
        name
        content {
            ... on Blog {
                title
            }
        }
    }
}
```

Whilst the query below will only return blogs. We could for instance use a filter to check that the title is not null to essentially return all blogs:

```graphql
query GetUsersWithAllContent {
    users {
        name
        content(where: { Blog: { title_NOT: null }}) {
            ... on Blog {
                title
            }
        }
    }
}
```

Conceptually, this maps to the `WHERE` clauses of the subquery unions in Cypher. Going back to the first example with no `where` argument, each subquery has a similar structure:

```cypher
CALL {
    WITH this
    OPTIONAL MATCH (this)-[has_content:HAS_CONTENT]->(blog:Blog)
    RETURN { __resolveType: "Blog", title: blog.title }
UNION
    WITH this
    OPTIONAL MATCH (this)-[has_content:HAS_CONTENT]->(journal:Post)
    RETURN { __resolveType: "Post" }
}
```

Now if we were to leave both subqueries and add a `WHERE` clause for blogs, it would look like this:

```cypher
CALL {
    WITH this
    OPTIONAL MATCH (this)-[has_content:HAS_CONTENT]->(blog:Blog)
    WHERE blog.title IS NOT NULL
    RETURN { __resolveType: "Blog", title: blog.title }
UNION
    WITH this
    OPTIONAL MATCH (this)-[has_content:HAS_CONTENT]->(journal:Post)
    RETURN { __resolveType: "Post" }
}
```

As you can see, the subqueries are now "unbalanced", which could result in massive overfetching of `Post` nodes.

So, when a `where` argument is passed in, we only include union members which are in the `where` object, so it is essentially acting as a logical OR gate, different from the rest of our `where` arguments:

```cypher
CALL {
    WITH this
    OPTIONAL MATCH (this)-[has_content:HAS_CONTENT]->(blog:Blog)
    WHERE blog.title IS NOT NULL
    RETURN { __resolveType: "Blog", title: blog.title }
}
```

# Issue

This is a new approach for the following PRs, which don't make so much sense for the new count queries and connection totalCount:

- #289 
- #312

# Checklist


packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -1137,20 +1137,20 @@ function makeAugmentedSchema(
...
                         const totalCount = isInt(count) ? count.toNumber() : count;
 
-                        const unionEdges = edges?.filter((edge) => {
-                            if (
-                                Object.keys(edge.node).length === 1 &&
-                                Object.prototype.hasOwnProperty.call(edge.node, "__resolveType")
-                            ) {
-                                return false;
-                            }
+                        // const unionEdges = edges?.filter((edge) => {
```
danstarns, line 1140: Can we remove this commented out code?

packages/graphql/src/translate/connection/create-connection-and-params.ts
```diff
@@ -89,143 +89,148 @@ function createConnectionAndParams({
...
+                const relatedNodeVariable = `${nodeVariable}_${n.name}`;
+                const nodeOutStr = `(${relatedNodeVariable}:${n.name})`;
+
+                const unionSubquery: string[] = [];
+                const unionSubqueryElementsToCollect = [...elementsToCollect];
+
+                const nestedSubqueries: string[] = [];
+
+                if (node) {
+                    // Doing this for unions isn't necessary, but this would also work for interfaces if we decided to take that direction
```
danstarns, line 102: I don't understand this, if it's not necessary for unions why are we doing it? Could this comment be misleading.

danstarns (approved):
All good, I just added two comments on the comments(meta).

packages/graphql/src/translate/connection/create-connection-and-params.ts
darrellwarde: I actually think it is necessary, so comment removed... 🤫

PR #368, opened 2021-07-29
Alias connections

danstarns:
# Description

Adds the ability to alias a connection and the selections inside of it.

### DIFF
--- a/packages/graphql/src/schema/make-augmented-schema.ts
+++ b/packages/graphql/src/schema/make-augmented-schema.ts
@@ -1150,12 +1150,12 @@ function makeAugmentedSchema(
                 [connectionField.fieldName]: {
                     type: connection.NonNull,
                     args: composeNodeArgs,
-                    resolve: (rootValue, args: ConnectionQueryArgs, ctx, resolveInfo: GraphQLResolveInfo) => {
+                    resolve: (source, args: ConnectionQueryArgs, ctx, info: GraphQLResolveInfo) => {
                         return connectionFieldResolver({
                             connectionField,
                             args,
-                            resolveInfo,
-                            rootValue,
+                            info,
+                            source,
                         });
                     },
                 },
--- a/packages/graphql/src/schema/make-augmented-schema.ts
+++ b/packages/graphql/src/schema/make-augmented-schema.ts
@@ -27,6 +27,7 @@ import {
     EnumTypeDefinitionNode,
     GraphQLInt,
     GraphQLNonNull,
+    GraphQLResolveInfo,
     GraphQLSchema,
     InputObjectTypeDefinitionNode,
     InterfaceTypeDefinitionNode,
@@ -69,7 +70,7 @@ import getFieldTypeMeta from "./get-field-type-meta";
 import Relationship, { RelationshipField } from "../classes/Relationship";
 import getRelationshipFieldMeta from "./get-relationship-field-meta";
 import getWhereFields from "./get-where-fields";
-import { createConnectionWithEdgeProperties } from "./pagination";
+import { connectionFieldResolver } from "./pagination";
 import { validateDocument } from "./validation";
 
 function makeAugmentedSchema(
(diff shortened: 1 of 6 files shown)


packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -389,9 +389,15 @@ function createProjectionAndParams({
...
         if (connectionField) {
             if (!inRelationshipProjection) {
-                if (!res.meta.connectionFields) res.meta.connectionFields = [];
-                res.meta.connectionFields.push(field as ResolveTree);
-                res.projection.push(literalElements ? `${field.name}: ${field.name}` : `${field.name}`);
+                if (!res.meta.connectionFields) {
+                    res.meta.connectionFields = [];
+                }
+
+                const f = field as ResolveTree;
```
darrellwarde, line 396: It feels to me like instead of this reassignment, `field` should be of type `ResolveTree` in the reducer arguments instead of `any`.

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -1149,15 +1150,13 @@ function makeAugmentedSchema(
...
-                    resolve: (source, args: ConnectionQueryArgs) => {
-                        const { totalCount: count, edges } = source[connectionField.fieldName];
-
-                        const totalCount = isInt(count) ? count.toNumber() : count;
-
-                        return {
-                            totalCount,
-                            ...createConnectionWithEdgeProperties(edges, args, totalCount),
-                        };
+                    resolve: (rootValue, args: ConnectionQueryArgs, ctx, resolveInfo: GraphQLResolveInfo) => {
```
darrellwarde, line 1153: Can we change `rootValue` back to `source`? We're quite inconsistent with this in our resolver function signatures, and [`source` is the name used in the official codebase](https://github.com/graphql/graphql-js/issues/2842#issuecomment-725499214).

Ideally the same for `ctx` to `context` and `resolveInfo` to `info` - they're all only stylistic changes, but we should be aiming to have a consistent codebase.

packages/graphql/tests/integration/connections/alias.int.test.ts
```diff
@@ -0,0 +1,769 @@
+/*
```
darrellwarde, line 1: I can't believe with all of these tests you've written that I'm to ask for another...

But would be good to make sure you can alias a combination of nested fields - `edges`, `node`, `pageInfo` and `startCursor` for instance.

packages/graphql/src/schema/pagination.ts
```diff
@@ -17,17 +17,65 @@
...
 
 /**
  * Adapted from graphql-relay-js ConnectionFromArraySlice
  */
-export function createConnectionWithEdgeProperties(
-    arraySlice: { node: Record<string, any>; [key: string]: any }[] = [],
-    args: { after?: string; first?: number } = {},
-    totalCount: number
-) {
+export function createConnectionWithEdgeProperties({
```
darrellwarde, line 68: I can't really argue with this as it obviously works... But it seems a "shame" that this is no longer really "Adapted from graphql-relay-js ConnectionFromArraySlice" as per the function comment.

Very open-ended question, was there no other way this could have been done whilst maintaining the original function? I'm just intrigued as to how the function from the `graphql-relay` package achieves aliasing, because it must do I imagine?

darrellwarde (changes requested):
Looks great! I have just a few observations and comments, if you wouldn't mind humouring me? 🙂

packages/graphql/src/schema/make-augmented-schema.ts
danstarns: I must say I do find this odd https://github.com/graphql/graphql-js/blob/c589c3d285cb1ec44b09bf0b50ec041ec083760c/src/graphql.ts#L62

darrellwarde (approved):
Thanks for making those changes, looks good!

Issue #369, opened 2021-07-30
Relationship properties - connection variable not defined


aitor3ml:
**Describe the bug**
Variable not defined error when trying to request relationship properties (using interface).

**Type definitions**
```
type Dato {
  uuid: ID
  dependeTo : [Dato] @relationship(type: "DEPENDE", direction: OUT, properties: "Depende")
  dependeFrom : [Dato] @relationship(type: "DEPENDE", direction: IN, properties: "Depende")
}

interface Depende {
  uuid: ID
  variable: String
}

type Query {
  getDato(uuid: String): Dato
    @cypher(
      statement: """
        MATCH (d:Dato {uuid: $uuid}) RETURN d
        """
    )
}
```

**To Reproduce**
1. Run a server with the previous graphql schema
2. Execute the following Query
This will work correctly.
```
{
  getDato(uuid:"4ebd5854-9caf-403f-bc96-f23f1a39e83d"){
    uuid
    dependeTo {
        uuid
      dependeFrom {
        uuid
      }      
    }
  }
}
```

3. Then run the following Query
This will return error.
```
{
  getDato(uuid:"4ebd5854-9caf-403f-bc96-f23f1a39e83d"){
    uuid
    dependeToConnection {
      edges {
        uuid
        variable
      }
    }
  }
}
```

```
{
  "errors": [
    {
      "message": "Variable `dependeToConnection` not defined (line 6, column 22 (offset: 223))\n\"RETURN this { .uuid, dependeToConnection } AS this\"\n                      ^",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "getDato"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "code": "Neo.ClientError.Statement.SyntaxError",
          "name": "Neo4jError",
          "stacktrace": [
            "Neo4jError: Variable `dependeToConnection` not defined (line 6, column 22 (offset: 223))",
            "\"RETURN this { .uuid, dependeToConnection } AS this\"",
            "                      ^",
            ": ",
            "    at captureStacktrace (/home/amartin/projects/lynx/gits/lynx-backend/node_modules/neo4j-driver-core/lib/result.js:239:17)",
            "    at new Result (/home/amartin/projects/lynx/gits/lynx-backend/node_modules/neo4j-driver-core/lib/result.js:59:23)",
[...]

PR #370, opened 2021-07-30
Fix/issue 369

danstarns:
# Description

Pulls out connection meta from projection and applies it to the cypher used in the cypher directive(top-level). Fixes: #369


darrellwarde (approved):
Looks good, nice fix!

PR #376, opened 2021-08-02
feat: Resolve union type in custom cypher

john2094:
It allows to solve custom cypher with results of type union,  currently this query crashes. 

An example of this is the following:

                type Movie {
                    title: String!
                    actors: [Actor] @relationship(type: "ACTED_IN", direction: IN)
                    directors: [Actor] @relationship(type: "DIRECTED", direction: IN)
                }
                type Director {
                    name: String!
                    movies: [Movie] @relationship(type: "DIRECTED", direction: OUT)
                }
                type Actor {
                    name: String!
                    movies: [Movie] @relationship(type: "ACTED_IN", direction: OUT)
                }
                type Query {

                    personInMovie(title: String!): [Person] @cypher(statement: """
                        MATCH (m:Movie {title: $title})<-[:ACTED_IN|DIRECTED]-(p)
                        RETURN p
                    """)
                }

### DIFF
--- a/packages/graphql/src/schema/resolvers/cypher.ts
+++ b/packages/graphql/src/schema/resolvers/cypher.ts
@@ -18,6 +18,8 @@
  */
 
 import { isInt } from "neo4j-driver";
+import { UnionTypeDefinitionNode } from "graphql/language/ast";
+import { ResolveTree } from "graphql-parse-resolve-info";
 import { execute } from "../../utils";
 import { BaseField, ConnectionField, Context } from "../../types";
 import { graphqlArgsToCompose } from "../to-compose";
@@ -26,6 +28,7 @@ import createAuthParam from "../../translate/create-auth-param";
 import { AUTH_FORBIDDEN_ERROR } from "../../constants";
 import createProjectionAndParams from "../../translate/create-projection-and-params";
 import createConnectionAndParams from "../../translate/connection/create-connection-and-params";
+import { Node } from "../../classes";
 
 export default function cypherResolver({
     field,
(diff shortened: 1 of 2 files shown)


darrellwarde:
Morning @john2094, we can't accept contributions from users who haven't signed our CLA (https://neo4j.com/developer/cla/#sign-cla). Could you please do that and then report back?

john2094:
Hello @darrellwarde, I signed the agreement.

darrellwarde:
Hey @john2094, thanks for the contribution and signing our CLA. 🙂 

We're going to leave this one to sit for a little while until we're out of the release candidate stage for 2.0.0 - we're only trying to introduce very small bug fixes at this stage to keep things stable. We'll revisit when 2.0.0 has been released. I hope you understand! 🙂

john2094:
Ok, I undertand.

LapoElisacci:
Hi @darrellwarde, any update on this one here?
Thanks

darrellwarde:
Hey @LapoElisacci, no update, but completely agree this has been sat on the sidelines for too long now! 

@john2094, any chance you could get `master` merged into this and get everything up to date so we can take another look?

rcbevans:
I'm needing this now.  @john2094 is this something you have plans to be able to take a look at?  If not I can take your changes as a starting point and try and get them up to date on latest main branch.

I see @oskarhane merged some pretty deep changes yesterday that seem like they could benefit this change.

rcbevans:
I have a branch with the changes from this branch re-implemented on master: https://github.com/rcbevans/graphql/tree/resolve-cypher-unions

rcbevans:
The same `NodeReference` check on a field type is done in [create-projection-and-params](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/create-projection-and-params.ts#L184) which causes cypher fields on concrete types which return unions to fail to resolve correctly also.

The code looks near enough identical between the [cypher resolver](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/schema/resolvers/cypher.ts) and [create-projection-and-params](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/create-projection-and-params.ts) so it would perhaps be trivial to apply the same change there.

As a general observation, I wonder if the code here could/should be refactored so it can be common to both uses, since it looks pretty much copy-pasted.

john2094:
> I'm needing this now.  @john2094 is this something you have plans to be able to take a look at?  If not I can take your changes as a starting point and try and get them up to date on latest main branch.
> 
> I see @oskarhane merged some pretty deep changes yesterday that seem like they could benefit this change.

Hello i dont have time now to take this, but be free to take my code as an starting point.

john2094:
> The same `NodeReference` check on a field type is done in [create-projection-and-params](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/create-projection-and-params.ts#L184) which causes cypher fields on concrete types which return unions to fail to resolve correctly also.
> 
> The code looks near enough identical between the [cypher resolver](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/schema/resolvers/cypher.ts) and [create-projection-and-params](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/create-projection-and-params.ts) so it would perhaps be trivial to apply the same change there.
> 
> As a general observation, I wonder if the code here could/should be refactored so it can be common to both uses, since it looks pretty much copy-pasted.

Yeah! There Is basically a copy paste of this method with a little modification to get  a cypher query to  each node type that belongs to union definition. After, i build a cypher query with union of all of those subquerys.

rcbevans:
Awesome. I already ported your cypher resolver change on top of latest, I'll see tomorrow about applying the same approach to resolving unions returned from cypher fields in the projection logic.

darrellwarde:
Brilliant, we look forward to receiving a PR @rcbevans! 🙂 Going to close this one off for now if work isn't going to continue directly on this branch.

PR #378, opened 2021-08-02
Change minimum database version to 4.1.5

darrellwarde:
# Description

Following reports that our changes to connect operations no longer work on 4.1.0, I iterated through versions to find the minimum which we now support - 4.1.5 is where all tests start passing.

Also fixes version checking, now uses `semver` package for comparison. For non-semantic APOC version, strips off the trailing number.

Also, now throws all compatibility issues in one error, rather than throwing as soon as it hits each.

Worthwhile adding an integration test pipeline for 4.1.5 exactly (rather than 4.1.x as current, I'm assuming 4.1.9) so we can easily tell if and when we break compatibility? 

# Checklist

### DIFF
--- a/docs/asciidoc/introduction.adoc
+++ b/docs/asciidoc/introduction.adoc
@@ -63,7 +63,7 @@ Additionally, prerelease version numbers may have additional suffixes, for examp
 
 == Requirements
 
-1. https://neo4j.com/[Neo4j Database] 4.1.0+
+1. https://neo4j.com/[Neo4j Database] 4.1.5+
 2. https://neo4j.com/developer/neo4j-apoc/[APOC] 4.1.0+
 3. https://nodejs.org/en/[Node.js] 12+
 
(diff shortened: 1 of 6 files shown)


oskarhane (approved):
Sounds good to add an additional compat test for minimun supported version 👍 
Can you create a card for it (or add the test job in the pipe)?

darrellwarde:
> Sounds good to add an additional compat test for minimun supported version 👍
> Can you create a card for it (or add the test job in the pipe)?

I sure can, will add to the pipeline now! 🙂

danstarns (approved):

darrellwarde:
Made some code changes since approvals - would appreciate a re-review before merging @danstarns @oskarhane. 🙂

oskarhane (approved):

darrellwarde:
Merging because test failure is on Aura Professional build which isn't actually hooked up to the pipeline yet - need to work around a few unexpected failures there.

PR #379, opened 2021-08-02
Error auth relationship properties

danstarns:
# Description

Throws error when `@auth` is used on relationship props. This is not supported and we assume that the parent nodes auth is sufficient enough.


darrellwarde (approved):
Looks good!

Is it worthwhile checking that we definitely honour all of the other directives on relationship properties? 🧐

danstarns:
> Looks good!
> 
> Is it worthwhile checking that we definitely honour all of the other directives on relationship properties? 🧐

No, we don't, some more PRS inbound soon.

PR #381, opened 2021-08-02
Error relationship on relationship properties

danstarns:
# Description

Throws when relationship directive is used on relationship properties. Related: #379


darrellwarde (approved):
Good finds this morning, whoops! 🙈

PR #382, opened 2021-08-02
Adds a simple util file which can be called to teardown database

darrellwarde:
# Description

Will need to be used to avoid hitting the 50,000 node limit when testing against Aura Free (would take a while, but would be a nuisance).

### DIFF
--- a/packages/graphql/tests/integration/teardown.ts
+++ b/packages/graphql/tests/integration/teardown.ts
@@ -1,3 +1,28 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * File not imported in project, and to be used in test pipelines to teardown target database.
+ *
+ *   `ts-node tests/integration/teardown.ts`
+ */
+
 import neo4j from "./neo4j";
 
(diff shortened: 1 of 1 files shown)


packages/graphql/tests/integration/teardown.ts
```diff
@@ -0,0 +1,20 @@
...
+        console.log("Encountered an error whilst clearing down database!");
+        throw e;
+    } finally {
+        await session.close();
+        await driver.close();
+    }
+};
+
+// eslint-disable-next-line no-void
+void teardown().then(() => console.log("Successfully cleared down database."));
```
danstarns, line 20: Why void ?

danstarns (commented):
This file isn't imported and so the function is never called?

darrellwarde:
> This file isn't imported and so the function is never called?

No, it's to literally be called in TeamCity, `ts-node tests/integration/teardown.ts` before running tests against Aura, so we don't go over the 50,000 node limit. It's just a nice way to take advantage of the driver provided in `neo4j.ts`.

packages/graphql/tests/integration/teardown.ts
darrellwarde: 🤣  Whatever you do here, either TypeScript or ESLint complains!!! This required the least comments in code.

danstarns:
> > This file isn't imported and so the function is never called?
> 
> No, it's to literally be called in TeamCity, `ts-node tests/integration/teardown.ts` before running tests against Aura, so we don't go over the 50,000 node limit. It's just a nice way to take advantage of the driver provided in `neo4j.ts`.

Could we add a comment in this file for this ? I fear it is deleted in some random 'cleanup' in the future as someone doesn't know where it's being called from.

darrellwarde:
> > > This file isn't imported and so the function is never called?
> 
> > 
> 
> > No, it's to literally be called in TeamCity, `ts-node tests/integration/teardown.ts` before running tests against Aura, so we don't go over the 50,000 node limit. It's just a nice way to take advantage of the driver provided in `neo4j.ts`.
> 
> 
> 
> Could we add a comment in this file for this ? I fear it is deleted in some random 'cleanup' in the future as someone doesn't know where it's being called from. 

Yeah good call, will do so in the morning! Only a temporary solution anyway, I'm hoping we can do something better around this in the near future. 🙂

oskarhane (approved):
+1 on @danstarns comment on adding a purpose comment to the file.
But other than that 👍

danstarns (commented):
nice

PR #383, opened 2021-08-03
Refactor/unify relationship fields

danstarns:
# Description

This was initial work for allowing the cypher directive to work on relationship properties. After chatting with the team we decided to exclude the cypher directive on relationship properties and [You will find an error being thrown inside this PR](https://github.com/neo4j/graphql/compare/2.0.0...danstarns:refactor/relationship-fields?expand=1#diff-23d81b942bb1369342a8d1c7762d1d20680b0ad436dddb4893d33a997886cc13R270). 

To not waste the work on this... This branch includes a refactor, the refactor was the start of the cypher directive work, as the **existing relationship properties fields were completely different to how nodes work**.  Some moving around needed to happen in order to know what field has a cypher directive on it and what the source of that cypher was. After observing, we already had a function called [`getObjFieldMeta`](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/schema/get-obj-field-meta.ts) and this PR refactors the [duplicated code](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/schema/get-relationship-field-meta.ts) around relationship properties to use that function. What this does mean is that the fields on a relationship are now split up into individual arrays such as: 

```
primitiveFields?: PrimitiveField[];
scalarFields?: CustomScalarField[];
enumFields?: CustomEnumField[];
dateTimeFields?: DateTimeField[];
pointFields?: PointField[];
ignoredFields?: BaseField[];
```

In my option makes; the code more explicit over of doing a string check on the type, for example, [here](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/translate/projection/elements/create-relationship-property-element.ts#L34) and [here](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/translate/projection/elements/create-relationship-property-element.ts#L37), and then I believe it makes the codebase more consistent plus makes sharing functionally between node properties and relationship properties easier. 

Finally, we have a function called [`getWhereFields`](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/schema/get-where-fields.ts) that was used on relationship properties and [not on nodes](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/schema/make-augmented-schema.ts#L477), this PR reuses that function for node where fields.


packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -255,60 +251,49 @@ function makeAugmentedSchema(
...
-            cartesianPointInTypeDefs = relationshipFieldMeta.some((field) => field.typeMeta.name === "CartesianPoint");
-        }
-
-        relationshipFields.set(relationship.name.value, relationshipFieldMeta);
-
-        const fields = {};
-
         relationship.fields?.forEach((field) => {
-            const forbiddenDirectives = ["auth", "relationship"];
+            const forbiddenDirectives = ["auth", "relationship", "cypher"];
```
darrellwarde, line 263: Should there also be a check at the interface level?

```ts
const cypherDirective = (relationship.directives || []).find((x) => x.name.value === "cypher");
if (cypherDirective) {
    throw new Error("Cannot have @cypher directive on relationship properties interface");
}
```

darrellwarde (changes requested):
This looks good, I think sensible to do what you've done with the fields for now.

I do think long-term having polymorphic fields is the way forward, the GraphQL library itself is a prime example of this implemented well with `fields` and `args` - imagine if you had all of the different field and arguments types as separate fields there like we do here!

Just one addition that I think needs adding.

packages/graphql/src/schema/make-augmented-schema.ts
danstarns: The schema validation covers this, but I just did add a test case for it.

danstarns:
> This looks good, I think sensible to do what you've done with the fields for now.
> 
> I do think long-term having polymorphic fields is the way forward, the GraphQL library itself is a prime example of this implemented well with `fields` and `args` - imagine if you had all of the different field and arguments types as separate fields there like we do here!
> 
> Just one addition that I think needs adding.

Yes having an array of fields where each is either an instance of a class or has a `.kind` property would be neater.

packages/graphql/src/schema/make-augmented-schema.ts
darrellwarde: Of course it does, I forgot that there was separate location definition for interface and object types!

darrellwarde (approved):
Looks good to me, nice perseverance getting passed the Docker rate limiting! 🙈

PR #386, opened 2021-08-03
Integration test causal consistency

darrellwarde:
# Description

Integration tests don't work against Aura Professional due to causal clustering. This is an attempt at passing the last bookmark into GraphQL execution so that test data is available.


darrellwarde:
This seems to fix the problems we were having getting our tests to pass against an Aura Professional instance. A bit crude, but perfectly effective.

danstarns (approved):
Looks good, I feel we should aim to create a wrapper around this so we only need to edit it in one place. Happy to get this in now and work on the wrapper in the future.

darrellwarde:
> Looks good, I feel we should aim to create a wrapper around this so we only need to edit it in one place. Happy to get this in now and work on the wrapper in the future.

Couldn't agree more. 🙂

PR #389, opened 2021-08-05
refactor: change schema references from relationship to edge

danstarns:
# Description

Change schema references from relationship to edge


oskarhane (approved):
That is a lot of changes, but well needed.

Maybe put something in the docs that in the typedefs we refer to the `@relationship` to map to neo4j relationships, where as in the schema we use `edge(s)` to be consistent with the general API language used by Relay?

darrellwarde (approved):
I trust that you caught all of them! 😜 

In an effort to try to keep the RFC up-to-date, made this change in there also.

PR #391, opened 2021-08-05
Documentation changes following editorial comments

darrellwarde:
# Description

Addresses the following editorial comments on the new documentation:

> 1. In the chapter "Getting started”, where steps of installing and creating database are described, the following looks a little bit strange: 
      1. Step 1
       2. Step 2
       3. Step 3
I like structured text, but it seems here that one number is redundant. It would be better to omit first numbers, wouldn’t it? To leave only word “Step” and its number. Or do the opposite - to omit “Step and its number”.

> 2. I really like your style, but recently I found out that we don’t use word “whilst” and avoid using “we, us”.  I know it’s difficult to be aware of such things in the absence of Style Guide. But if you have time and it’s possible to replace “we” by “you” without any change of meaning or loss of clarity I think it would be wonderful to do it (especially in the chapter "Getting started”). It’s recommended	to use second person or imperative mood.


oskarhane (approved):
Good changes 👍

NataliaIvakina (commented):
Good work!

PR #394, opened 2021-08-06
feat: export auth errors from index

danstarns:
# Description

Exports the auth error classes so one can override the messages, reference https://discord.com/channels/787399249741479977/818578492723036210/872962949445193748


oskarhane (approved):
LGTM

PR #396, opened 2021-08-06
Reserved properties

danstarns:
# Description

Throws error when the user uses reserved names on types, interfaces and interface fields. This is mainly to stop any conflicts with Relay.

### DIFF
--- a/packages/graphql/src/constants.ts
+++ b/packages/graphql/src/constants.ts
@@ -34,3 +34,27 @@ export const REQUIRED_APOC_PROCEDURES = ["apoc.util.validate", "apoc.do.when", "
 export const DEBUG_AUTH = `${DEBUG_PREFIX}:auth`;
 export const DEBUG_GRAPHQL = `${DEBUG_PREFIX}:graphql`;
 export const DEBUG_EXECUTE = `${DEBUG_PREFIX}:execute`;
+
+// [0]Property [1]Error
+export const RESERVED_NODE_LABELS = [
+    [
+        "PageInfo",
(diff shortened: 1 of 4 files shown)


packages/graphql/src/constants.ts
```diff
@@ -34,3 +34,27 @@ export const REQUIRED_APOC_PROCEDURES = ["apoc.util.validate", "apoc.do.when", "
...
+    ["Node", "Type name 'Node' reserved to support relay See https://relay.dev/graphql/"],
+];
+
+// [0]Property [1]Error
+export const RESERVED_INTERFACE_NAMES = [
+    ["Node", "Interface name 'Node' reserved to support relay See https://relay.dev/graphql/"],
+];
+
+// [0]Property [1]Error
+export const RESERVED_INTERFACE_PROPERTIES = [
```
darrellwarde, line 57: `RESERVED_FIELD_NAMES`? To avoid the leakage of "properties". Like, you could have a property "node" when we introduce aliasing, it's just GraphQL field names that are the problem.

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -254,12 +259,34 @@ function makeAugmentedSchema(
...
+            }
+
+            if (relationship.name.value === label) {
+                toThrowError = true;
+            }
+
+            if (toThrowError) {
+                throw new Error(message);
+            }
+        });
```
darrellwarde, line 276: This could probably be put into a function (exact duplicate of above), but not necessary now. 🙂

darrellwarde (approved):
Looks good! 😄

PR #400, opened 2021-08-09
Remove surplus _IN and _NOT_IN filters for relationships

darrellwarde:
# Description

Let's take the opportunity of this breaking release to remove some surplus requirements, `_IN` and `_NOT_IN` for one-to-many and one-to-one relationships. We don't currently treat these relationship cardinalities differently anywhere else, the filters are completely untested, and the same filters can be achieved through other means (see migration guide for example).

# Checklist

### DIFF
--- a/docs/asciidoc/guides/2.0.0-migration/miscellaneous.adoc
+++ b/docs/asciidoc/guides/2.0.0-migration/miscellaneous.adoc
@@ -46,3 +46,44 @@ const neoSchema = new Neo4jGraphQL({
 ----
 
 If you need to do this, please report the scenario as an issue on GitHub.
+
+== `_IN` and `_NOT_IN` filters on relationships removed
+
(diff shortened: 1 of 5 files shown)


danstarns (changes requested):
If we are going to remove them, Let's remove them for good. You have forgotten about the code behind these [here](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/translate/create-where-and-params.ts#L71) and [here](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/translate/create-where-and-params.ts#L190). I'm not sure I completely understand the reasoning for removing them and if they work just fine why are we getting rid?

darrellwarde:
> If we are going to remove them, Let's remove them for good. You have forgotten about the code behind these [here](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/translate/create-where-and-params.ts#L71) and [here](https://github.com/neo4j/graphql/blob/2.0.0/packages/graphql/src/translate/create-where-and-params.ts#L190). I'm not sure I completely understand the reasoning for removing them and if they work just fine why are we getting rid?

Honestly, I don't think the Cypher for them is even correct (wrong predicate), and we really shouldn't have any untested paths in our code. I was able to remove these two filters and only 1 (!) TCK test broke - I don't think that's really the level of quality we should be aiming for.

oskarhane (approved):
LGTM

danstarns (approved):

Issue #388, opened 2021-08-04
1.2.4 breaks aliased ID's in Unions


tlester:
Error:  

```
Cannot return null for non-nullable field Post.id.
    at new ApolloError (index.ts:49)
```

Union:
```
 query GetContent($userID: ID, $limit: Int, $skip: Int) {
        getContent(userID: $userID, limit: $limit, skip: $skip) {
            ... on Post {
                postID: id
                createdDate
                modifiedDate
                __typename
            }
            ... on Image {
                imageID: id
                createdDate
                modifiedDate
                __typename
            }
        }
        getContentCount(userID: $userID)
    }
```

id is non-nullable.... and I suspect that the aliasing isn't respecting that.

darrellwarde:
@tlester, can we please have the type definitions? Especially the definition for `getContent`.

tlester:
```
 getContentCount(userID: ID): Int
            @cypher(
                statement: """
                match (myUser:User {id: $userID})
                optional match (myUser)<-[:HAS_USER]-(:Profile)-[:HAS_FRIEND]->(x)<-[:HAS_FRIEND]-(:Profile)-[:HAS_USER]->(myFriends:User)
                    call {
                       with myUser, myFriends
                       match (myFriends)-[:HAS_POST]->(post:Post)
                       return post
                       union
                       with myUser, myFriends
                       match (myFriends)-[:HAS_COMMENT]->(:Comment)<-[:HAS_COMMENT]-(post:Post)
                       return post
                       union
                       with myUser, myFriends
                       match (myFriends)-[:HAS_REACTION]->(:Reaction)<-[:HAS_REACTION]-(post:Post)
                       return post
                       union
                       with myUser, myFriends
                       match (myUser)<-[:HAS_USER]-(myProfile:Profile)
                       match (myProfile)-[:HAS_USER]->(myUsers:User)
                       match (myUsers)-[:HAS_POST]->(post:Post)
                       return post
                       union
                       with myUser, myFriends
                       match (myUser)<-[:HAS_USER]-(myProfile:Profile)
                       match (myProfile)-[:HAS_USER]->(myUsers:User)
                       match (myUsers)-[:HAS_COMMENT]->(:Comment)<-[:HAS_COMMENT]-(post:Post)
                       return post
                       union
                       with myUser, myFriends
                       match (myUser)<-[:HAS_USER]-(myProfile:Profile)
                       match (myProfile)-[:HAS_USER]->(myUsers:User)
[...]

PR #423, opened 2021-08-13
Aliased fields from custom Cypher

darrellwarde:
# Description

Given a custom Cypher field like the following:

```graphql
type Query {
    getContent(userID: ID): [Content]
        @cypher(
            statement: """
            MATCH (myUser:User {id: $userID})
            OPTIONAL MATCH (myUser)-[:HAS_FRIEND]->(myFriends:User)
            CALL {
                WITH myUser, myFriends
                MATCH (myUser)-[:HAS_POST]->(post:Post)
                RETURN post
                UNION
                WITH myUser, myFriends
                MATCH (myFriends)-[:HAS_POST]->(post:Post)
                RETURN post
            }
            RETURN DISTINCT apoc.map.merge(properties(post), { __resolveType: 'Post' }) AS result ORDER BY result.modifiedDate DESC
            """
        )
}
```

It's impossible/very difficult to account for field aliasing in the return results, for example with a query like:

```graphql
query GetContent($userID: ID) {
    getContent(userID: $userID) {
        __typename
        ... on Post {
            postContent: content
        }
    }
}
```

So instead, in the default field resolver, if we see that a field was requested under an alias:

1. First check whether the field is already in the source in its aliased state
2. Failing the above, assume that it was not possible to alias so return the unaliased field

If the latter is hit, then the GraphQL runtime will deal with aliasing for us.

# Issue

#388 

# Checklist

### DIFF
--- a/packages/graphql/src/schema/resolvers/defaultField.ts
+++ b/packages/graphql/src/schema/resolvers/defaultField.ts
@@ -25,11 +25,13 @@ import { GraphQLResolveInfo } from "graphql";
  */
 // eslint-disable-next-line consistent-return
 export default function defaultFieldResolver(source: any, args: any, context: unknown, info: GraphQLResolveInfo) {
-    const responseKey = info.fieldNodes[0].alias ? info.fieldNodes[0].alias.value : info.fieldNodes[0].name.value;
     if ((typeof source === "object" && source !== null) || typeof source === "function") {
-        const property = source[responseKey];
+        const property =
+            info.fieldNodes[0].alias && Object.prototype.hasOwnProperty.call(source, info.fieldNodes[0].alias.value)
+                ? source[info.fieldNodes[0].alias.value]
+                : source[info.fieldNodes[0].name.value];
         if (typeof property === "function") {
-            return source[responseKey](args, context, info);
+            return property(args, context, info);
         }
         return property;
     }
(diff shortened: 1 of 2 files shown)


oskarhane (approved):
LGTM

packages/graphql/src/schema/resolvers/defaultField.ts
```diff
@@ -25,11 +25,13 @@ import { GraphQLResolveInfo } from "graphql";
  */
 // eslint-disable-next-line consistent-return
 export default function defaultFieldResolver(source: any, args: any, context: unknown, info: GraphQLResolveInfo) {
-    const responseKey = info.fieldNodes[0].alias ? info.fieldNodes[0].alias.value : info.fieldNodes[0].name.value;
     if ((typeof source === "object" && source !== null) || typeof source === "function") {
-        const property = source[responseKey];
+        const property =
+            info.fieldNodes[0].alias && Object.prototype.hasOwnProperty.call(source, info.fieldNodes[0].alias.value)
```
danstarns, line 30: If there is `info.fieldNodes[0].alias` wouldnt there always be `info.fieldNodes[0].alias.value` ? I ask because this turnay operator is so scary too look at 😱 and could we just replace with: 

```js
info.fieldNodes[0].alias && source[info.fieldNodes[0].alias.value] : source[info.fieldNodes[0].name.value]
```

darrellwarde: Reading your comment, I would have thought so, it even matches the TypeScript definition!

But the tests fail when I switch it over, so perhaps not. 😞

Issue #425, opened 2021-08-14
Generated count-Queries don't work (no method ".toNumber")


SimplyComple0x78:
**Describe the bug**
When I generate an API using the Neo4j-Apollo-Stack, also count-Queries (e.g. "animalCount") are generated. Though, they didn't work in my case because of a type-issue.

**Type definitions**
Let's take for example
```
type Person {
  name: String
}
```

**To Reproduce**
To reproduce, take the following script:

```
import { Neo4jGraphQL } from "@neo4j/graphql";
import { ApolloServer, gql } from "apollo-server";
import { driver, auth } from "neo4j-driver";

const typeDefs = gql`
  type Person {
    name: String
  }
`;

const drv = driver("neo4j://localhost", auth.basic("neo4j", "test"));

const neoSchema = new Neo4jGraphQL({ typeDefs, driver: drv });

const server = new ApolloServer({
  schema: neoSchema.schema,
});

server.listen().then(({ url }) => {});
```

When I run it and execute the query
```
query Query {
  peopleCount
}
```

I'm only getting the following error:
```
Error: result.records[0]._fields[0].toNumber is not a function
    at Model.<anonymous> (/project/node_modules/@neo4j/graphql-ogm/dist/classes/Model.js:96:23)
    at Generator.next (<anonymous>)
    at fulfilled (/project/node_modules/@neo4j/graphql-ogm/dist/classes/Model.js:23:58)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

**Expected behavior**
I expect the query to return the number of Person-Entities in the database rather than to raise an exception.

**System (please complete the following information):**
-    OS: Linux (Ubuntu, Bionic Beaver)
-    Version: @2.0.0
-    Node.js version: 12.14.0 and 14.16.0 (tested both)

**Additional context**
I checked out the error and it seems it's calling ".toNumber" on a number which isn't working. Typing isn't helpful there, because the result has type any. I'm gonna send you a PR for that.

PR #426, opened 2021-08-14
Fixed issue #425 (broken count)

SimplyComple0x78:
# Description

I fixed the issue #425 but I don't know if there is another reason for that issue to happen.
Please investigate that in the review.

# Issue

https://github.com/neo4j/graphql/issues/425

# Checklist

### DIFF
--- a/packages/graphql/src/schema/resolvers/count.ts
+++ b/packages/graphql/src/schema/resolvers/count.ts
@@ -16,6 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { isInt, Integer } from "neo4j-driver";
 import { execute } from "../../utils";
 import { translateCount } from "../../translate";
 import { Node } from "../../classes";
@@ -34,12 +35,14 @@ export default function countResolver({ node }: { node: Node }) {
             raw: true,
         });
 
-        const count: number | any = result.records[0]._fields[0];
+        const count = result.records[0]._fields[0];
 
-        if (typeof count === "number") {
-            return count;
+        // @ts-ignore: count is unknown, and to cast to object would be an antipattern
+        if (isInt(count)) {
+            return (count as Integer).toNumber();
         }
-        return count.toNumber();
+
+        return count;
     }
 
     return {
--- a/packages/graphql/src/schema/resolvers/count.ts
+++ b/packages/graphql/src/schema/resolvers/count.ts
@@ -34,7 +34,12 @@ export default function countResolver({ node }: { node: Node }) {
             raw: true,
         });
 
-        return result.records[0]._fields[0].toNumber();
+        const count: number | any = result.records[0]._fields[0];
+
+        if (typeof count === "number") {
+            return count;
+        }
+        return count.toNumber();
     }
 
     return {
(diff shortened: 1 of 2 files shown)


packages/graphql/src/schema/resolvers/count.ts
```diff
@@ -34,7 +34,12 @@ export default function countResolver({ node }: { node: Node }) {
             raw: true,
         });
 
-        return result.records[0]._fields[0].toNumber();
+        const count: number | any = result.records[0]._fields[0];
+
+        if (typeof count === "number") {
+            return count;
+        }
+        return count.toNumber();
```
darrellwarde, line 42: Damn, this is definitely needed here, sorry about that! We actually do this slightly differently elsewhere in the codebase and would like to stick with that for consistency:

```suggestion
        const count = result.records[0]._fields[0];

        // @ts-ignore: count is unknown, and to cast to object would be an antipattern
        if (isInt(count)) {
            return (count as Integer).toNumber();
        }

        return count;
```

Which will also require the following imports at the top of the file:

```ts
import { isInt, Integer } from "neo4j-driver";
```

darrellwarde (changes requested):
Thanks for the contribution!

Before we can do anything with this, can you please [sign our CLA](https://neo4j.com/developer/cla/#sign-cla)?

See my comment regarding changes to be made. Would also be good to see the functionality tested for both a plain number and Neo4j Integer.

packages/graphql/src/schema/resolvers/count.ts
SimplyComple0x78: Agreed, changed that part.

SimplyComple0x78:
Changed the implementation as requested, added unit-tests for both cases and signed the CLA.

Is there anything left?
* Sadly, the tests weren't successful even after my changes, so I only checked the tests my new ones are in.
* I don't know how you're working with mocking - I felt the need to mock `import { execute } from "../../utils";` and did so (I don't know what's happening in there). In case that's the wrong direction, feel free to point me on other options.

darrellwarde:
Morning! There's a little manual step on our side for the CLA, so I've just done that now and will re-run the tests shortly. 🙂 

I actually see that there are no comprehensive unit tests for any of the resolvers... So maybe don't worry so much. Was just taking a very much "belt and braces" approach!!!

Regarding TeamCity, it's something we're looking to address by switching over to GitHub Actions so that open-source contributors can have better access to our pipelines. 🙂

danstarns:
I do find it really odd how and when this would not be an Integer. Seeming we simply return the result of executing [count](https://neo4j.com/docs/cypher-manual/current/functions/aggregating/#functions-count) that happens to return an Integer.

darrellwarde:
> I do find it really odd how and when this would not be an Integer. Seeming we simply return the result of executing [count](https://neo4j.com/docs/cypher-manual/current/functions/aggregating/#functions-count) that happens to return an Integer.

I think it can end up being either a native `number` or driver `Integer`, as we've experienced (painfully) throughout the rest of the project - we just forgot to put the safeguard here!

darrellwarde (approved):
The tests are exactly what I was looking for, thanks for this fix! 🙂

PR #432, opened 2021-08-16
GitHub actions (attempt number 2)

darrellwarde:
# Description

We're back for round 2 with GitHub Actions! Little did I realise that PRs from forked repos run with different permissions to direct branches, and these changes account for that. Changes are:

- Documentation now builds on the fork, archives the build and then passes it to another workflow for publishing
- A separate flow for pull requests without SonarCloud and CodeCov for now, being actively looked at
- Build of master and releases now separated for cleanliness, looking into shared functionality
- CLA check is now a separate build - it's fine to run the tests from anybody's fork because everything is run in an isolated environment with no write-access to the repo or secret access


oskarhane (approved):
LGTM 🌻

PR #436, opened 2021-08-17
GitHub Actions tweaks following release

darrellwarde:
# Description

Introduces fallback workflows which can be used to coax releases out step-by-step if the main workflow doesn't work.

Pulls SonarCloud push out into separate workflow.


oskarhane (approved):
Looks good 🧨

PR #447, opened 2021-08-20
Disable fail-fast for integration tests

darrellwarde:
Disable fail-fast for integration tests - useful for us to be able to see which versions tests fail on


oskarhane (approved):
👍

PR #449, opened 2021-08-20
Feature: Add Time Scalar

dmoree:
# Description

Adds a `Time` scalar. It is passed in as an RFC 3339 compliant time string and stored in the database as a neo4j `Time` instance.

# Checklist

### DIFF
--- a/packages/graphql/src/schema/scalars/Time.ts
+++ b/packages/graphql/src/schema/scalars/Time.ts
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { GraphQLError, GraphQLScalarType, Kind } from "graphql";
+import neo4j from "neo4j-driver";
+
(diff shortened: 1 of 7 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-449.surge.sh

packages/graphql/src/schema/scalars/Time.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    return value;
+};
+
+export const parseTime = (value: any) => {
+    const validatedValue = validateTime(value);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const { hour, minute, second, fraction, offsetDirection, offsetHour, offsetMinute } = RFC_3339_REGEX.exec(
+        validatedValue
+    )!.groups!;
```
darrellwarde, line 42: As far as I know, RegEx matching is quite expensive, so doing it once with `test` in `validateTime` and then again with `exec` here is a bit wasteful. See my suggestion below for potentially getting rid of `validateTime` and performing the whole lot in `parseTime` - it also means one less non-null assertion also.

```suggestion
export const parseTime = (value: any) => {
    if (typeof value !== "string") {
        throw new TypeError(`Value must be of type string: ${value}`);
    }
    
    const match = RFC_3339_REGEX.exec(value);
    
    if (!match) {
        throw new TypeError(`Value must be formatted as Time: ${value}`);
    }
    
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    const { hour, minute, second, fraction, offsetDirection, offsetHour, offsetMinute } = match.groups!;
```

darrellwarde (changes requested):
This looks great - tidy implementation and super thorough testing! Just one comment about the use of RegEx - happy to be countered on it though.

packages/graphql/src/schema/scalars/Time.ts
dmoree: Great point.

darrellwarde:
Hey @dmoree, looks like there's some unresolved references to the removed `validateTime` function, can you take a look? 🙂

dmoree:
> Hey @dmoree, looks like there's some unresolved references to the removed validateTime function

Fixed in https://github.com/neo4j/graphql/pull/449/commits/3bbdfaa7a9218f2e7282ce160d7ab48db7b7d107

darrellwarde (approved):
Looks great, let's get this in! 😄

PR #451, opened 2021-08-22
Feature: Add LocalTime Scalar

dmoree:
# Description

Adds a `LocalTime` scalar. It is passed in as a time string without timezone information and stored in the database as a neo4j `LocalTime` instance.

# Checklist

### DIFF
--- a/packages/graphql/src/schema/scalars/LocalTime.ts
+++ b/packages/graphql/src/schema/scalars/LocalTime.ts
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { GraphQLError, GraphQLScalarType, Kind } from "graphql";
+import neo4j from "neo4j-driver";
+
+export const LOCAL_TIME_REGEX = /^(?<hour>[01]\d|2[0-3]):(?<minute>[0-5]\d):(?<second>[0-5]\d)(\.(?<fraction>\d{1}(?:\d{0,8})))?$/;
+
+export const validateLocalTime = (value: any) => {
+    if (typeof value !== "string") {
+        throw new TypeError(`Value must be of type string: ${value}`);
+    }
+
+    if (!LOCAL_TIME_REGEX.test(value)) {
+        throw new TypeError(`Value must be formatted as LocalTime: ${value}`);
+    }
+
+    return value;
+};
+
+export const parseLocalTime = (value: any) => {
+    const validatedValue = validateLocalTime(value);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const { hour, minute, second, fraction } = LOCAL_TIME_REGEX.exec(validatedValue)!.groups!;
+
+    // Calculate the number of nanoseconds by padding the fraction of seconds with zeroes to nine digits
+    let nanosecond = 0;
+    if (fraction) {
+        nanosecond = +`${fraction}000000000`.substring(0, 9);
+    }
+
+    return {
+        hour: +hour,
+        minute: +minute,
+        second: +second,
+        nanosecond,
+    };
+};
+
+const parse = (value: any) => {
+    const { hour, minute, second, nanosecond } = parseLocalTime(value);
+
+    return new neo4j.types.LocalTime(hour, minute, second, nanosecond);
+};
+
+export default new GraphQLScalarType({
+    name: "LocalTime",
+    description: "A local time, represented as a time string without timezone information",
+    serialize: (value: typeof neo4j.types.LocalTime) => {
+        return validateLocalTime(value.toString());
+    },
+    parseValue: (value) => {
+        return parse(value);
+    },
+    parseLiteral: (ast) => {
+        if (ast.kind !== Kind.STRING) {
+            throw new GraphQLError(`Only strings can be validated as LocalTime, but received: ${ast.kind}`);
+        }
+        return parse(ast.value);
+    },
+});
--- a/packages/graphql/src/schema/scalars/LocalTime.ts
+++ b/packages/graphql/src/schema/scalars/LocalTime.ts
@@ -22,22 +22,19 @@ import neo4j from "neo4j-driver";
 
 export const LOCAL_TIME_REGEX = /^(?<hour>[01]\d|2[0-3]):(?<minute>[0-5]\d):(?<second>[0-5]\d)(\.(?<fraction>\d{1}(?:\d{0,8})))?$/;
 
-export const validateLocalTime = (value: any) => {
+export const parseLocalTime = (value: any) => {
     if (typeof value !== "string") {
         throw new TypeError(`Value must be of type string: ${value}`);
     }
 
-    if (!LOCAL_TIME_REGEX.test(value)) {
+    const match = LOCAL_TIME_REGEX.exec(value);
+
+    if (!match) {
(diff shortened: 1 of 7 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-451.surge.sh

packages/graphql/src/schema/scalars/LocalTime.ts
```diff
@@ -0,0 +1,77 @@
...
+        throw new TypeError(`Value must be formatted as LocalTime: ${value}`);
+    }
+
+    return value;
+};
+
+export const parseLocalTime = (value: any) => {
+    const validatedValue = validateLocalTime(value);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const { hour, minute, second, fraction } = LOCAL_TIME_REGEX.exec(validatedValue)!.groups!;
```
darrellwarde, line 40: Same suggestion as on #449 - happy to discuss!

```suggestion
export const parseLocalTime = (value: any) => {
    if (typeof value !== "string") {
        throw new TypeError(`Value must be of type string: ${value}`);
    }
    
    const match = LOCAL_TIME_REGEX.exec(value);
    
    if (!match) {
        throw new TypeError(`Value must be formatted as LocalTime: ${value}`);
    }
    
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    const { hour, minute, second, fraction } = match.groups!;
```

darrellwarde (changes requested):
Thanks for also doing `LocalTime` - nearly there on all of the temporal types now apart from `Duration`! 😄  The same comment as on #449 left.

danstarns (changes requested):
Hey @dmoree, Thanks for this. I would like to see something like `localTimeFields` [on the Node](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/classes/Node.ts#L84) and Relationship classes. Then these fields can get picked up in particular places such as [specifying the field has auth on it](https://github.com/neo4j/graphql/blob/f16d2a27b4bc8be088e149658619190379b597d4/packages/graphql/src/classes/Node.ts#L143) and [allowing one to coalesce the field](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/where/create-relationship-where-and-params.ts#L59)(open question?).

darrellwarde:
> Hey @dmoree, Thanks for this. I would like to see something like `localTimeFields` [on the Node](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/classes/Node.ts#L84) and Relationship classes. Then these fields can get picked up in particular places such as [specifying the field has auth on it](https://github.com/neo4j/graphql/blob/f16d2a27b4bc8be088e149658619190379b597d4/packages/graphql/src/classes/Node.ts#L143) and [allowing one to coalesce the field](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/where/create-relationship-where-and-params.ts#L59)(open question?).

Could they not all be bundled into `dateTimeFields` (maybe rename it `temporalFields`)? I don't think there's really going to be any behaviour differences between the various types.

danstarns:
> > Hey @dmoree, Thanks for this. I would like to see something like `localTimeFields` [on the Node](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/classes/Node.ts#L84) and Relationship classes. Then these fields can get picked up in particular places such as [specifying the field has auth on it](https://github.com/neo4j/graphql/blob/f16d2a27b4bc8be088e149658619190379b597d4/packages/graphql/src/classes/Node.ts#L143) and [allowing one to coalesce the field](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/where/create-relationship-where-and-params.ts#L59)(open question?).
> 
> Could they not all be bundled into `dateTimeFields` (maybe rename it `temporalFields`)? I don't think there's really going to be any behaviour differences between the various types.
> 
> We could also get this working with `@timestamp`, which would be cool, but not essential!

I'm not sure if this would work as we do some [specific stuff](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/create-projection-and-params.ts#L435) when projecting `dateTimeFields`.

darrellwarde:
> > > Hey @dmoree, Thanks for this. I would like to see something like `localTimeFields` [on the Node](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/classes/Node.ts#L84) and Relationship classes. Then these fields can get picked up in particular places such as [specifying the field has auth on it](https://github.com/neo4j/graphql/blob/f16d2a27b4bc8be088e149658619190379b597d4/packages/graphql/src/classes/Node.ts#L143) and [allowing one to coalesce the field](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/where/create-relationship-where-and-params.ts#L59)(open question?).
> > 
> > 
> > Could they not all be bundled into `dateTimeFields` (maybe rename it `temporalFields`)? I don't think there's really going to be any behaviour differences between the various types.
> > We could also get this working with `@timestamp`, which would be cool, but not essential!
> 
> I'm not sure if this would work as we do some [specific stuff](https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/create-projection-and-params.ts#L435) when projecting `dateTimeFields`.

I actually assumed that `Date` was bundled in there also, but actually, it's "loose" much like `Time` and `LocalTime` are here - we should probably find a way to treat them a bit more uniformly rather than polluting the `Node` and `Relationship` type. There a lot of similarities between all of these types, so maybe a single field on `Node` and `Relationship` with some specific type checks around the codebase is appropriate for any specific behaviours?

dmoree:
@danstarns @darrellwarde I had a similar thought process while creating #449 trying to get `@timestamp` to work with it. I didn't want to change too much and decided to follow a similar pattern with `Date`. If we encompassed them all into `temporalFields`, in order to address @danstarns concern above we could do:
```
...
} else if (temporalField?.typeMeta.name === 'DateTime') {
  res.projection.push(createDatetimeElement({ resolveTree: field, field: temporalField, variable: varName }));
} ...
```

In other words, it is possible to branch off according to `field.typeMeta.name`

Time zones may make it difficult (or impossible) to consistently use `@timestamp` with `Date` or `LocalTime`

dmoree:
> Thanks for also doing LocalTime - nearly there on all of the temporal types now apart from Duration

I could do `Duration` in a similar manner to `Time` and `LocalTime`. I don't work too much with durations, so is there any special considerations, or could I follow the same pattern? My only thought is that perhaps `_GT`, `_LT`, etc. don't really apply here.

darrellwarde:
Hey @dmoree, looks like there's some unresolved references to the removed `validateLocalTime` function, can you take a look? 🙂

darrellwarde:
> @danstarns @darrellwarde I had a similar thought process while creating #449 trying to get `@timestamp` to work with it. I didn't want to change too much and decided to follow a similar pattern with `Date`. If we encompassed them all into `temporalFields`, in order to address @danstarns concern above we could do:
> 
> ```
> ...
> } else if (temporalField?.typeMeta.name === 'DateTime') {
>   res.projection.push(createDatetimeElement({ resolveTree: field, field: temporalField, variable: varName }));
> } ...
> ```
> 
> In other words, it is possible to branch off according to `field.typeMeta.name`
> 
> Time zones may make it difficult (or impossible) to consistently use `@timestamp` with `Date` or `LocalTime`

Exactly, this is how I would go about it - there is enough similarities between them that distinction at this level makes sense.

However, I propose for now that we merge this functionality as it stands (at the end of the day, it's already in-line with the already released `Date` scalar), and then work to bring all 4 temporal types into step as a separate piece of work.

darrellwarde:
> > Thanks for also doing LocalTime - nearly there on all of the temporal types now apart from Duration
> 
> I could do `Duration` in a similar manner to `Time` and `LocalTime`. I don't work too much with durations, so is there any special considerations, or could I follow the same pattern? My only thought is that perhaps `_GT`, `_LT`, etc. don't really apply here.
> 
> Edit: According to the documentation on [Ordering and comparison of values](https://neo4j.com/docs/cypher-manual/current/syntax/operators/#cypher-ordering) `Duration` values cannot be compared and have special considerations for ordering.
> 
> P.S. This thread of the conversation can be moved to another PR.

Definitely a different PR and discussion for `Duration`, and also no rush to get that it, just figured it was worth bringing up! 🙂

dmoree:
> Hey @dmoree, looks like there's some unresolved references to the removed validateLocalTime function

Fixed in https://github.com/neo4j/graphql/pull/451/commits/26567896cd965678e0544a7c84414330149cbfe6.

> Definitely a different PR and discussion for Duration

- #459

darrellwarde (approved):
LGTM, thanks for all of these contributions! 🙂

darrellwarde:
Necessity for something akin to `temporalFields` captured in #461, merging this as basic support for new temporal types right now.

PR #454, opened 2021-08-23
Add Slack notifications to release workflow

darrellwarde:
# Description

Adds Slack notifications to release workflow - message body configured in the Slack workflow itself.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -290,6 +290,21 @@ jobs:
           git commit -am "Bump versions to ${{ steps.get-version.outputs.version }}"
           git push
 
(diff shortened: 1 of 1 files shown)


danstarns:
I must say, I don't want this, I get enough notifications as it is XD. If we are going to add it can we not direct it to our private channels?

danstarns:
> I must say, I don't want this, I get enough notifications as it is XD. If we are going to add it can we not direct it to our private channels?

Looking back at this, I realise it's only going to notify us on a release so it's not the end of the world.

danstarns (approved):

PR #458, opened 2021-08-24
Return bookmarks so that users can read their own writes

darrellwarde:
# Description

Introduce a new `CreateInfo` and `UpdateInfo` type which is added to `MutationResponse` types. These new types contain a `bookmark` field, the value of which users can pass into subsequent queries to ensure they can read their own writes in the case of causal clustering. A `bookmark` field has also been added to the existing `DeleteInfo` type.

Additionally, took the opportunity to pad out responses with more Mutation metadata:

- For create operations:
  - `nodesCreated`
  - `relationshipsCreated`
- For update operations:
  - `nodesCreated`
  - `nodesDeleted`
  - `relationshipsCreated`
  - `relationshipsDeleted`

With a few bugs resulting in more nodes/relationships being created/deleted than expected recently, these new fields might more easily flag when cardinality has gone adrift. Additionally, it means that in our integration tests, we can use this count to validate change quantities, compared to querying using the driver and then counting `records` as we currently do in places.

# Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -3,3 +3,6 @@ graphql:
 
 ogm:
   - packages/ogm/**/*
+  
+documentation:
+  - docs/modules/ROOT/**/*
--- a/.github/workflows/fallback-bump-version.yml
+++ b/.github/workflows/fallback-bump-version.yml
@@ -1,0 +1,38 @@
+name: "[Fallback] Bump Version"
+
+on:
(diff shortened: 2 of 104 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-458.surge.sh

oskarhane (approved):
This change looks really good!

PR #463, opened 2021-08-25
@alias directive - map a GraphQL field to a Neo4j node/relationship property under a different name

oskarhane:
# Description

This PR introduces the `@alias` directive to map GraphQL fields to Neo4j properties under a different name.

Example

```graphql
type User {
    id: ID! @alias(property: "dbId")
    username: String! @alias(property: "dbUserName")
    livesIn: [City] @relationship(direction: OUT, type: "LIVES_IN", properties: "UserLivesInProperties")
}

type City {
    name: String
}

interface UserLivesInProperties @relationshipProperties {
    since: DateTime @alias(property: "moveInDate")
}
```

When the API is queried for the `id` field on the User type, the underlying Cypher query will retrieve the value from the database property `dbId`

Example Cypher

```cypher
MATCH (this:User) RETURN this {id: this.dbId} AS this
```

This works for node properties as well as relationship properties, and should work in combination with other directives like `@id`, `@timestamp`, `@default` etc.

It should also work in all types of queries (regular + relay type connections) and mutations.

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-463.surge.sh

darrellwarde:
Looks like you need to add your new documentation page to https://github.com/oskarhane/graphql/blob/%40alias/docs/modules/ROOT/content-nav.adoc because you can't see it in https://neo4j-graphql-docs-463.surge.sh/graphql-manual/2.0/type-definitions/ 🙂

packages/graphql/src/schema/get-alias-meta.ts
```diff
@@ -0,0 +1,47 @@
...
+        return undefined;
+    }
+
+    const stmtArg = directive.arguments?.find((x) => x.name.value === "property");
+    if (!stmtArg) {
+        throw new Error("@alias property argument is required");
+    }
+    if (stmtArg.value.kind !== "StringValue") {
+        throw new Error("@alias property argument was not a string");
+    }
```
darrellwarde, line 38: These checks are a bit redundant because schema validation covers this for us.

packages/graphql/src/schema/get-alias-meta.test.ts
```diff
@@ -0,0 +1,127 @@
...
+                },
+                {
+                    // @ts-ignore
+                    name: { value: "RANDOM 2" },
+                },
+            ],
+        };
+
+        expect(() => getAliasMeta(field)).toThrow("@alias property argument was not a string");
+    });
```
darrellwarde, line 95: These tests really belong in `packages/graphql/src/schema/validation/validate-document.test.ts`, where you will be able to test using actual type definitions, like the other tests in there.

packages/graphql/src/schema/get-obj-field-meta.ts
```diff
@@ -91,11 +92,13 @@ function getObjFieldMeta({
             const relationshipMeta = getRelationshipMeta(field);
             const cypherMeta = getCypherMeta(field);
             const typeMeta = getFieldTypeMeta(field);
+            const aliasMeta = getAliasMeta(field);
```
darrellwarde, line 95: Shouldn't this be inside the `if (aliasDirective)` block below? The same could be said for the other "get*Meta" functions above - it feels a bit wasteful to call a function if we _know_ it's going to return undefined.

darrellwarde (changes requested):
Broadly looks good! I've left a few comments. 🙂  

One observation that I have is that there seems to be a lot of dynamic calculation of the DB property through calling `mapToDbProperty`. Is there no way that this could be calculated once at schema augmentation time, and saved in the `BaseField` type? Then we could just blindly pluck out `field.propertyName` (for example) and trust that it is correct, whether or not `@alias` was used on the field.

packages/graphql/src/schema/get-alias-meta.ts
oskarhane: This was to make TS happy cause not all ValueNode types has the `.value` property. 
Happy to cast instead if that makes things clearer.

packages/graphql/src/schema/get-obj-field-meta.ts
oskarhane: Usually I'd vote for consistency in the code - but sure, I'll move it.

danstarns (approved):
LGTM, nice work :)

PR #465, opened 2021-08-26
Fix flakey Aura tests

darrellwarde:
# Description

Addresses some of the flakey Aura integration tests by passing around bookmarks in places where they weren't before.

Updates the driver dev dependency to 4.3.x - `session.lastBookmark()` return type changed between versions so being up-to-date makes this easier to handle. This did break TCK tests, for which a fix has been put in place. 

Make sure to update your dependencies after this is merged to ensure everything works.


oskarhane (approved):
Lot's of changed files - the sooner we get this in, the better.

Issue #461, opened 2021-08-25
Add temporalFields to Node and Relationship 


darrellwarde:
Add `temporalFields` to the `Node` and `Relationship` classes so that the following scalar types can be treated quite uniformly in the context of auth and timestamps:

- `DateTime`
- `LocalDateTime`
- `Date`
- `Time`
- `LocalTime`

This is following the work of these PRs which added the scalars but not this functionality:

- #449 
- #451 
- #460 
- #275 

Any differences between each temporal type to be distinguished as suggested by @dmoree: https://github.com/neo4j/graphql/pull/451#issuecomment-903945576

PR #466, opened 2021-08-29
Refactor: dateTimeFields to temporalFields

dmoree:
# Description

- Refactors `dateTimeFields` on `Node` and `Relationship` to `temporalFields`.
- Expand `temporalFields` to {`DateTime`, `Date`, `Time`, `LocalDateTime`, `LocalTime`}
- Expand `@timestamp` to {`DateTime`, `Time`}

# Issue

- #461 

# Checklist


oskarhane (commented):
Thanks for this @dmoree!
This looks good to me, but you'd need to rebase and resolve the conflicts from a merge we did the other day. Sorry about that!

dmoree:
No worries. Resolved by keeping `dbPropertyName`. Thanks for `@alias`.

darrellwarde (changes requested):
So everything that's already here looks good!

The only thing I would like to see is some TCK/integration tests validating that `Time` fields work as expected as timestamps, as I believe that is new functionality which is introduced in this PR. 

Is there any other new functionality that this enables? I remember @danstarns mentioning `@auth`, possibly worth validating that all of the temporal field types work there?

Nice work, excited to get this in!

dmoree:
I've added TCK and integration tests on par with the existing `DateTime` scalar (it may also be time to create a `tests/integration/directives` directory like TCK).

This doesn't change much in the way of `authableFields` as `dateTimeFields` was already there. Perhaps @danstarns can talk about how he sees temporal fields figuring into `@auth`. It may be a failure of imagination on my part, but wouldn't one want to `allow` a temporal value relative to the value at execution time? For example, a user can only see posts of another within 30 days or some other duration. Or perhaps can only edit future events? This would require special attention in `createAuthPredicate`. I'm happy to have a discussion on what that would look like.

darrellwarde (approved):
Brilliant, thanks for adding those timestamp tests in! I think this is a good place to stop for now - if we were to expand this out to changes to the `@auth` directive then I think that this PR would have definitely suffered from scope creep!

The point was more about just testing the existing `@auth` functionality against these "new" types, but we don't have comprehensive type by type coverage for this directive anyway, so this is something we can add later if we want to up our test coverage.

+1 on the integration tests being due a bit of a reorganisation, we'll keep it in mind!

Thanks again. 🙂

danstarns (approved):

Issue #293, opened 2021-07-05
The usage of ALL any ANY with the @auth directive


darrellwarde:
@darrellwarde Hey, a bit late to the party, but I think the same behaviour should apply to `@auth where & bind` in case we are referencing a relationship that is a list

With the following schema:
```graphql
type User {
  id: ID! @id
  name: String!
}
type Chat {
  id: ID! @id
  title: String @readonly
  description: String @readonly
  users: [User] @relationship(type: "HAS_USER", direction: OUT)
}

extend type Chat
  @auth(rules: [
    { operations: [CREATE, READ, UPDATE, DELETE], where: { users: { id: "$jwt.sub" } } }
  ])
```

How can I get all `Chats` for the logged user? Because `where` & `bind` apply `ALL()`, I will be able to get only chats where `$jwt.sub` is the only user.

On the other hand, I think `ALL()` makes sense if it was a singular relationship:
```graphql
type Chat {
  user: User @relationship(type: "HAS_USER", direction: OUT)
}
```

_Originally posted by @MarianAlexandruAlecu in https://github.com/neo4j/graphql/issues/190#issuecomment-873641640_

MarianAlexandruAlecu:
@darrellwarde an even better solution would be allow devs to specify if they want `ALL` or `ANY`.

E.g.
```graphql
extend type Chat
  @auth(rules: [
    { operations: [CREATE, READ, UPDATE, DELETE], where: { users: { id: "$jwt.sub" } }, predicate: ANY }
  ])
```

Is this too cumbersome? What do you think? 🤔

danstarns:
> ```graphql
> extend type Chat
>   @auth(rules: [
>     { operations: [CREATE, READ, UPDATE, DELETE], where: { users: { id: "$jwt.sub" } }, predicate: ANY }
>   ])
> ```

This is interesting, I don't think your suggested location is the best and something like this would be better IMO: 

```gql
extend type Chat
    @auth(rules: [{ operations: [CREATE, READ, UPDATE, DELETE], where: { ANY: { users: { id: "$jwt.sub" } } } }])
```

MarianAlexandruAlecu:
@danstarns Your version is perfect! I'm looking forward to use it in my code 😺

rcbevans:
@danstarns  Is this something being worked on?

[...]

PR #469, opened 2021-09-01
Feat: #293 Add support for ANY and ALL in auth rules

rcbevans:
This change is a contribution to resolve an immediate need, and #293.

Add the ability to explicitly specify ANY or ALL in auth rules immediately preceding a relation field to allow overriding the ANY or ALL which is generated in the output cypher statement.

The ANY or ALL only overrides the relation immediately following.

For example given the schema:

```graphql
type Group {
    id: ID
    name: String
    members: [User] @relationship(type: "MEMBER_OF", direction: IN)
}

extend type Group
    @auth(
        rules: [
            {
                operations: [READ, UPDATE, DELETE, CONNECT, DISCONNECT]
                where: { ANY: { members: { id: "$jwt.sub" } } }
            }
        ]
    )

type User {
    id: ID
    name: String
    groups: [Group] @relationship(type: "MEMBER_OF", direction: OUT)
    posts: [Post] @relationship(type: "HAS_POST", direction: OUT)
    content: [Search] @relationship(type: "HAS_POST", direction: OUT) # something to test unions
}
```

the query

```graphql
{
    groups {
        id
        members {
            id
        }
    }
}

```

will generate the query

```cypher
MATCH (this:Group) WHERE EXISTS((this)<-[:MEMBER_OF]-(:User)) AND ANY(members IN [(this)<-[:MEMBER_OF]-(members:User) | members] WHERE members.id IS NOT NULL AND members.id = $this_auth_where0_ANY_members_id) RETURN this { .id, members: [ (this)<-[:MEMBER_OF]-(this_members:User) WHERE this_members.id IS NOT NULL AND this_members.id = $this_members_auth_where0_id | this_members { .id } ] } as this
```

allowing the return of any groups with a member User with id matching the $jwt.sub.

# Description

Please provide a description of the work completed in this branch.

# Issue

# Checklist


danstarns:
Hey @rcbevans, thank you for your contribution. It makes sense to allow users to specify this. 

Please could you add integration tests for this? 

How does this work with nested rules? IE: 

```
where: { ANY: { group: { members: { id: "$jwt.sub" } } } }
```

rcbevans:
> Hey @rcbevans, thank you for your contribution. It makes sense to allow users to specify this.
> 
> Please could you add integration tests for this?
> 
> How does this work with nested rules? IE:
> 
> ```
> where: { ANY: { group: { ANY: { members: { id: "$jwt.sub" } } } } }
> ```

I will take a look and try and figure out integration tests.

In terms of behavior, the ANY/ALL specifier overrides the predicate used for exactly one relational edge immediately following it.

So you need to specify an ANY modifier before any relational edge where you want it.  So in your example, it'll allow the nodes where `any` group has `any` member which match the jwt subject.

If you were to remove the second ANY, it would default to the preexisting predicate based on the auth `kind`, which for `where` is `ALL`.

I wanted to maintain backward compatibility, but allow the additional flexibility for those who need it.  For symmetry, you can also explicitly use `ALL` as well as `ANY`.

rcbevans:
I have updated the unit, tck, and integration tests, as well as documentation.

rcbevans:
It also seems like it would be nice to support `NOT` as a modifier.

dmoree:
@rcbevans `ANY` and `ALL` are list predicates. Could you not achieve a `NOT` modifier by expanding the predicate list to include each of the cypher list predicates `{ANY,ALL,NONE,SINGLE}`. This would be a simple addition to this PR.

`NOT ANY` would translate to `NONE`
`NOT ALL` could translate to `SINGLE` (Perhaps `NOT ALL` is more like "some" than "single")

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-469.surge.sh

darrellwarde:
Hey @rcbevans, we did have a bit of a chat about this internally, and we have slight concerns about how this differs to the `where` arguments of the generated schema (i.e. the ones used in your queries/mutations). Even though we don't expose a method of achieving `ANY`/`ALL` on relationships fields there, we do expose it for scalar list fields using `_INCLUDES` and `_NOT_INCLUDES`, and ideally that is what we would do here to keep things consistent. So for the example which you gave:

```graphql
extend type Group
    @auth(
        rules: [
            {
                operations: [READ, UPDATE, DELETE, CONNECT, DISCONNECT]
                where: { ANY: { members: { id: "$jwt.sub" } } }
            }
        ]
    )
```

This would instead be:

```graphql
extend type Group
    @auth(
        rules: [
            {
                operations: [READ, UPDATE, DELETE, CONNECT, DISCONNECT]
                where: { members_INCLUDES: { id: "$jwt.sub" } } 
            }
        ]
    )
```

This will guarantee a consistent developer experience when both designing your API and using your API, and it will also potentially allow for easier use of common code to generate Cypher clauses in future. Thoughts on this?

rcbevans:
So rather than prefixing with a modifier, instead use a postfix to the name which adjusts the semantics?

Presumably no postfix would carry the current `ALL()` semantics, `_INCLUDES` would be `SOME()`, `_NOT_INCLUDES` would be `NOT(SOME())`.  Would `_NOT` translate to `NOT(ALL())`?

I think that design makes sense, given the stimilarity to other parts of the API, and it's simpler to extend vs modifiers to the recursive projection builder which have to carry additional state.

rcbevans:
As an aside, I could see it being useful to support where clauses checking for NULL/NOT NULL, which I don't think is currently possible, do you have any ideas how that could be supported? It seems it would have to be implemented via a special value rather than key modifier

```
where: { caseSealedDate: NULL }
```

Or

```
where: { publishedDate: ANY }
```

Currently the first part of the generated where check is a check for NOT NULL, so to support this the code would have to look ahead at the value being null and emit the appropriate filter.

darrellwarde:
> So rather than prefixing with a modifier, instead use a postfix to the name which adjusts the semantics?
> 
> Presumably no postfix would carry the current `ALL()` semantics, `_INCLUDES` would be `SOME()`, `_NOT_INCLUDES` would be `NOT(SOME())`. Would `_NOT` translate to `NOT(ALL())`?
> 
> I think that design makes sense, given the stimilarity to other parts of the API, and it's simpler to extend vs modifiers to the recursive projection builder which have to carry additional state.

I'm assuming by `SOME` you mean `ANY`? I _believe_ the mapping should be:

- No postfix = `ALL()`
- `_NOT` = `NOT(ALL())
- `_INCLUDES` = `ANY()`
- `_NOT_INCLUDES` = `NOT(ANY())`

> As an aside, I could see it being useful to support where clauses checking for NULL/NOT NULL, which I don't think is currently possible, do you have any ideas how that could be supported? It seems it would have to be implemented via a special value rather than key modifier
> 
> ```
> where: { caseSealedDate: NULL }
> ```
> 
> Or
> 
> ```
> where: { publishedDate: ANY }
> ```
> 
> Currently the first part of the generated where check is a check for NOT NULL, so to support this the code would have to look ahead at the value being null and emit the appropriate filter.

So in the `where` clauses of the actual generated APIs, the checks you've described above would be achieved with:

```
where: { caseSealedDate: null }
```

And:

```
where: { publishedDate_NOT: null }
```

Noting the lowercase `null` there, as to me capitalisation suggests an enum.

rcbevans:
I think that makes sense.  I'll take a look at how much will be involved to get that implemented.

rcbevans:
Closing this PR as the alternate design has been implemented in #500

PR #470, opened 2021-09-02
Feat/top level aggregations

danstarns:
# Description

Please provide a description of the work completed in this branch.

# Issue

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-470.surge.sh

packages/graphql/src/schema/resolvers/numerical.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    // @ts-ignore: outputValue is unknown
+    if (isInt(value)) {
+        return (value as Integer).toNumber();
+    }
+
+    return value;
+}
+
+export default numerical;
```
oskarhane, line 15: Missing file header

packages/graphql/src/schema/resolvers/id.ts
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    if (typeof value === "number") {
+        return value.toString(10);
+    }
+
+    return value;
+}
+
+export default id;
```
oskarhane, line 19: Missing file header here as well

darrellwarde (changes requested):
Speedy work to get this in so quickly! 🏃  Nice work getting it all working without any use of APOC.

A lot to review here, so just a few questions/observations from an initial once-over.

Why `AggregationSelection` for the type names? Why not just `Aggregate`, for example `moviesAggregate(where: MovieWhere): MovieAggregate!`?

I still find the use of `min` and `max` with `String` and `ID` confusing - aggregates are very much linked to sorting, in that `min` is the first entry in a sorted list, and `max` is the last. This works for the numerical and temporal types, in that if you sort a list of them, they are sorted in descending/ascending order. However, if you sort a list of strings, they are not ordered by length of the string, but how they rank alphabetically. If I was to make assumptions about what these fields do, I would expect `min` and `max` to return the first and last values alphabetically, not by length. I wouldn't be against these having different names - `shortest` and `longest` perhaps, and maybe have `first` and `last` to return the first and last entries alphabetically.

docs/modules/ROOT/pages/queries.adoc
```diff
@@ -92,3 +102,79 @@ query {
...
+== Aggregate
+
+Neo4j GraphQL supports aggregations on fields with type: 
+
+1. `ID`
+2. `String`
+3. `Int` - Numerical
+4. `Float` - Numerical
+5. `BigInt` - Numerical
+6. `DateTime`
```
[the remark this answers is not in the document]
danstarns: Added

danstarns:
> `AggregationSelection`

It's named `AggregationSelection` over `Aggregation` so that when we add input types for "aggregations where" we can call them `AggregationWhere`.

danstarns:
> Speedy work to get this in so quickly! 🏃 Nice work getting it all working without any use of APOC.
> 
> A lot to review here, so just a few questions/observations from an initial once-over.
> 
> Why `AggregationSelection` for the type names? Why not just `Aggregate`, for example `moviesAggregate(where: MovieWhere): MovieAggregate!`?
> 
> I still find the use of `min` and `max` with `String` and `ID` confusing - aggregates are very much linked to sorting, in that `min` is the first entry in a sorted list, and `max` is the last. This works for the numerical and temporal types, in that if you sort a list of them, they are sorted in descending/ascending order. However, if you sort a list of strings, they are not ordered by length of the string, but how they rank alphabetically. If I was to make assumptions about what these fields do, I would expect `min` and `max` to return the first and last values alphabetically, not by length. I wouldn't be against these having different names - `shortest` and `longest` perhaps, and maybe have `first` and `last` to return the first and last entries alphabetically.

It's named AggregationSelection over Aggregation so that when we add input types for "aggregations where" we can call them AggregationWhere.

packages/graphql/tests/integration/aggregations/top_level/bigint.test.ts
```diff
@@ -0,0 +1,288 @@
...
+
+        const neoSchema = new Neo4jGraphQL({ typeDefs });
+
+        try {
+            await session.run(
+                `
+                    CREATE (:Movie {id: $id, imdbRatingBigInt: 1.1})
+                    CREATE (:Movie {id: $id, imdbRatingBigInt: 2.1})
+                    CREATE (:Movie {id: $id, imdbRatingBigInt: 3.1})
+                    CREATE (:Movie {id: $id, imdbRatingBigInt: 4.1})
```
oskarhane, line 122: Just a little confusing / bad practice using a shared `ID` field here.
I understand the intent, but how about using a String field or something instead?

oskarhane (commented):
Looks really good! Just a small comment on using a bad practice in the test files that I think we should avoid.

packages/graphql/tests/integration/aggregations/top_level/bigint.test.ts
danstarns: Good catch I've made this change.

docs/modules/ROOT/pages/queries.adoc
```diff
@@ -73,6 +81,8 @@ query {
 [[queries-count]]
 == Count
 
+Count queries accepts an `where` argument for xref::filtering.adoc[Filtering] data.
```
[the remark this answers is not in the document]
darrellwarde: Doesn't look like this suggestion has been applied?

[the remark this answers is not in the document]
danstarns: https://github.com/neo4j/graphql/pull/470/commits/d877c7b650cfb40420cec3110fc6cc3b484d62ea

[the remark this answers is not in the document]
darrellwarde: Weird - if you look at https://github.com/neo4j/graphql/pull/470/files#diff-28b20e3631f2bc1833e3dfb299ee4276783c62b002b1e482c8607d7b6773a3a7R84 it looks unchanged? 🧐

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -528,6 +624,56 @@ function makeAugmentedSchema(
             },
         });
 
+        composer.createObjectTC({
+            name: `${node.name}AggregateSelection`,
```
darrellwarde, line 628: There is an inconsistency here, the scalars are "AggregationSelection" and object types are "AggregateSelection" - I think they should all be either one or the other. I believe "AggregateSelection" would be accurate.

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -528,6 +624,56 @@ function makeAugmentedSchema(
...
+                        res[field.fieldName] = localDateTimeAggregationSelection.NonNull;
+                    }
+
+                    if (field.typeMeta.name === "LocalTime") {
+                        res[field.fieldName] = localTimeAggregationSelection.NonNull;
+                    }
+
+                    if (field.typeMeta.name === "Time") {
+                        res[field.fieldName] = timeAggregationSelection.NonNull;
+                    }
```
darrellwarde, line 670: This is a bit clunky - is it worthwhile having an object like the following above:

```ts
const aggregateSelections = {
  ID: idNumericalAggregationSelection.NonNull,
  String: stringNumericalAggregationSelection.NonNull,
  ...
};
```

And then referring to it using:

```ts
res[field.fieldName] = aggregateSelections[field.typeMeta.name];
```

Just a suggestion, a bit tidier perhaps?

packages/graphql/tests/integration/aggregations/top_level/alias.int.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    beforeAll(async () => {
+        driver = await neo4j();
+    });
+
+    afterAll(async () => {
+        await driver.close();
+    });
+
+    test("should preform many aggregations while aliasing each field and return correct data", async () => {
```
darrellwarde, line 37: Sorry, I am a pedant! 😞 

```suggestion
    test("should perform many aggregations while aliasing each field and return correct data", async () => {
```

packages/graphql/tests/integration/aggregations/top_level/bigint.test.ts
```diff
@@ -0,0 +1,288 @@
...
+
+        const neoSchema = new Neo4jGraphQL({ typeDefs });
+
+        try {
+            await session.run(
+                `
+                    CREATE (:Movie {testString: $testString, imdbRatingBigInt: 1.1})
+                    CREATE (:Movie {testString: $testString, imdbRatingBigInt: 2.1})
+                    CREATE (:Movie {testString: $testString, imdbRatingBigInt: 3.1})
+                    CREATE (:Movie {testString: $testString, imdbRatingBigInt: 4.1})
```
darrellwarde, line 60: You've created this big integer values in all of these tests as floats - it's probably worth changing these to values over 2147483647 so that we can properly validate this behaviour.

packages/graphql/tests/integration/aggregations/top_level/id.int.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+        const neoSchema = new Neo4jGraphQL({ typeDefs });
+
+        try {
+            await session.run(
+                `
+                    CREATE (:Movie {testId: $id, id: 1})
+                    CREATE (:Movie {testId: $id, id: 2})
+                    CREATE (:Movie {testId: $id, id: 3})
+                    CREATE (:Movie {testId: $id, id: 4})
```
darrellwarde, line 60: When ID values are inserted through our API, they are inserted as strings. Probably worthwhile doing the same in these tests, as well as giving them variable lengths so that we can actually validate the behaviour.

darrellwarde (changes requested):
A few changes to request, mostly in the test files interesting!

I won't pretend to understand why the return types can't just be "Aggregate" instead of "AggregateSelection", but I'll trust your judgement on that one. But I think they should definitely all be aligned as pointed out. 

It's looking good though, sorry it's taken so long for a review! 😞

docs/modules/ROOT/pages/queries.adoc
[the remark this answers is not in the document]
danstarns: There are two references in this file, should be good now.

packages/graphql/tests/integration/aggregations/top_level/bigint.test.ts
danstarns: Yes!

packages/graphql/tests/integration/aggregations/top_level/id.int.test.ts
danstarns: Yes!

packages/graphql/src/schema/make-augmented-schema.ts
danstarns: Good observation, I have made a change now to make this less repetitive.

packages/graphql/src/schema/make-augmented-schema.ts
danstarns: Yes!

danstarns:
> A few changes to request, mostly in the test files interesting!
> 
> I won't pretend to understand why the return types can't just be "Aggregate" instead of "AggregateSelection", but I'll trust your judgement on that one. But I think they should definitely all be aligned as pointed out.
> 
> It's looking good though, sorry it's taken so long for a review! 😞

Thanks for your feedback, the reason it's "AggregateSelection" instead of "Aggregate" is due to the fact that we intend on adding input types for aggregating using the `where` argument and so when they are added we will have(using `Int` as an example) `IntAggregateSelection` & `IntAggregateInput`.

darrellwarde:
> > A few changes to request, mostly in the test files interesting!
> > I won't pretend to understand why the return types can't just be "Aggregate" instead of "AggregateSelection", but I'll trust your judgement on that one. But I think they should definitely all be aligned as pointed out.
> > It's looking good though, sorry it's taken so long for a review! 😞
> 
> Thanks for your feedback, the reason it's "AggregateSelection" instead of "Aggregate" is due to the fact that we intend on adding input types for aggregating using the `where` argument and so when they are added we will have(using `Int` as an example) `IntAggregateSelection` & `IntAggregateInput`.

I think what I'm trying to get at (and what I failed to get at before) is, even with the input types added, there wouldn't be a naming collision if we were to just name the return type `IntAggregate` - the "Selection" at the end just feels surplus to requirement. It's not really a huge deal, it's just for brevity and cleanliness of the output schema.

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -121,6 +124,83 @@ function makeAugmentedSchema(
...
+        ["DateTime"],
+        ["LocalDateTime"],
+        ["LocalTime"],
+        ["Time"],
+        ["Duration"],
+    ];
+
+    const aggregationSelectionTypeNames = aggregationSelectionTypeMatrix.map(([name]) => name);
+
+    const aggregationSelectionTypes = aggregationSelectionTypeMatrix.reduce((res, [name, fields]) => {
```
darrellwarde, line 194: You _should_ be able to get rid of the hacky `as unknown) as ObjectTypeComposer<unknown, unknown>;` from below if you do:

```suggestion
    const aggregationSelectionTypes = aggregationSelectionTypeMatrix.reduce<Record<string, ObjectTypeComposer<unknown, unknown>>>((res, [name, fields]) => {
```

Cleaner to specify your types up front.

darrellwarde (approved):
I think a bit of TypeScript could be improved where I've pointed out, but I guess it works. Either accept my approval or implement the change, your call! 😛  Looking good otherwise!

PR #475, opened 2021-09-08
Specify mapping of GraphQL type -> Neo4j node label using `@node` directive

oskarhane:
# Description

This PR introduces a new `@node` directive with two arguments: `label: String` and `additionLabels: [String!]`.

### `@node(label: "DbLabel")`
In this example typedefs

```graphql
type Movie @node(label: "DbMovieLabel") {
  title: String
}
```

The GraphQL type `Movie` will map to the Neo4j label `DbMovieLabel`.

So the query 
```graphql
{
  movies {
    title
  }
}
```
will result in the Cypher query

```cypher
MATCH (this:DbMovieLabel) RETURN this {.title} AS this
```

### `@node(additionalLabels: ["ExtraDbLabel", "ExtraDbLabel2"])`
In this example typedefs

```graphql
type Movie @node(additionalLabels: ["ExtraDbLabel", "ExtraDbLabel2"]) {
  title: String
}
```

So the query 
```graphql
{
  movies {
    title
  }
}
```
will result in the Cypher query

```cypher
MATCH (this:Movie:ExtraDbLabel:ExtraDbLabel2) RETURN this {.title} AS this
```

Only nodes with ALL the specified `additionalLabels` will be matched.
The main label of the type will still be used as seen in the example above.

### Combined
These two arguments can also be used together.

```graphql 
type Movie @node(label: "DbMovieLabel", additionalLabels: ["ExtraDbLabel", "ExtraDbLabel2"]) {
  title: String
}
``` 

```graphql
{
  movies {
    title
  }
}
```
will result in the Cypher query

```cypher
MATCH (this:DbMovieLabel:ExtraDbLabel:ExtraDbLabel2) RETURN this {.title} AS this
```

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-475.surge.sh

packages/graphql/src/classes/Node.ts
```diff
@@ -53,6 +54,7 @@ export interface NodeConstructor {
     ignoredFields: BaseField[];
     auth?: Auth;
     exclude?: Exclude;
+    nodeDirective?: NodeDirective;
```
angrykoala, line 57: naming conventions are a bit inconsistent here we may want to discuss if is worth it adding ...Directive to all directives

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -216,10 +218,11 @@ function makeAugmentedSchema(
         });
 
         const otherDirectives = (definition.directives || []).filter(
-            (x) => !["auth", "exclude"].includes(x.name.value)
+            (x) => !["auth", "exclude", "node"].includes(x.name.value)
```
angrykoala, line 221: Maybe we should move this to a constant or define a more explicit way of registering directives?

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -216,10 +218,11 @@ function makeAugmentedSchema(
...
         );
         const authDirective = (definition.directives || []).find((x) => x.name.value === "auth");
         const excludeDirective = (definition.directives || []).find((x) => x.name.value === "exclude");
+        const nodeDirectiveDefinition = (definition.directives || []).find((x) => x.name.value === "node");
```
angrykoala, line 225: Again, naming collision here :/

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -251,12 +251,15 @@ function createProjectionAndParams({
...
         if (relationField) {
+            // WARN: In some test this node is undefined. If this is intended, we should update typings to reflect that
             const referenceNode = context.neoSchema.nodes.find((x) => x.name === relationField.typeMeta.name) as Node;
             const nodeMatchStr = `(${chainStr || varName})`;
             const inStr = relationField.direction === "IN" ? "<-" : "-";
             const relTypeStr = `[:${relationField.type}]`;
             const outStr = relationField.direction === "OUT" ? "->" : "-";
-            const nodeOutStr = `(${param}:${referenceNode?.name})`;
+            const labels =
+                referenceNode?.nodeDirective?.getLabelsString(referenceNode.name) || `:${referenceNode?.name}`;
```
danstarns, line 261: This `||` is scattered throughout this change and I fear that it could be brittle and or missed in particular use cases, what do you think about creating a function that does both the `.getLabelsString(referenceNode.name)` and the `|| :${referenceNode?.name}`; ?

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -251,12 +251,15 @@ function createProjectionAndParams({
...
 
         if (relationField) {
+            // WARN: In some test this node is undefined. If this is intended, we should update typings to reflect that
             const referenceNode = context.neoSchema.nodes.find((x) => x.name === relationField.typeMeta.name) as Node;
             const nodeMatchStr = `(${chainStr || varName})`;
             const inStr = relationField.direction === "IN" ? "<-" : "-";
             const relTypeStr = `[:${relationField.type}]`;
             const outStr = relationField.direction === "OUT" ? "->" : "-";
-            const nodeOutStr = `(${param}:${referenceNode?.name})`;
+            const labels =
```
danstarns, line 260:
```suggestion
            const labels = getNodeLabels(referenceNode);
```

danstarns, line 260:
```suggestion
            const labels = referenceNode.getLabels()
```

packages/graphql/src/translate/create-projection-and-params.ts
angrykoala: We were discussing the possibility of this being part of the Node class (something like `referenceNode.getLabels()`)
thoughts?

danstarns:
An open question here - Shall we start using the `@node` directive in all documentation examples? For example: 

```diff
- type User {
+ type User @node {
}
```
Even though an empty `@node` directive won't achieve anything, if and when we decide to make the `@node` directive a requirement, a change, for most users, would be easier. We have done something similar with the `@relationshipProperties` directive.

packages/graphql/src/translate/create-projection-and-params.ts
danstarns: I think that would be better as everything will be in one place, that being said does it need to be a function? Could it be defined once in the `constructor` of each node?

angrykoala: Function, a get function or setting in a constructor, I would say that any of those should be fine if we decide to go for this approach of adding it to Node class

oskarhane:
> An open question here - Shall we start using the `@node` directive in all documentation examples? For example:
> 
> ```diff
> - type User {
> + type User @node {
> }
> ```
> 
> Even though an empty `@node` directive won't achieve anything, if and when we decide to make the `@node` directive a requirement, a change, for most users, would be easier. We have done something similar with the `@relationshipProperties` directive.

Yeah, this is a good point. With the introduction of this directive, I think it'd be good to show the long term intent with it already now.

darrellwarde (changes requested):
Big +1 on @danstarns suggestion of removing the `||` everywhere in favour of an accessor on the `Node` class. I'm wondering whether instead of `nodeDirective` on the `Node` class, should there just be a `label` field given that is all that the `@node` directive provides? Even as we extend it, we should be extracting the value out of the directive, rather than just saving all of the directive fields with another object on the `Node` type. Otherwise we're going to see a load of `node.nodeDirective.somethingElse` everywhere as we extend it, for example `node.nodeDirective.plural` which should probably just be `node.plural`. If that makes sense?

angrykoala:
@darrellwarde In the case of the labels is quite clear that having something in the Node class is needed, however, Node and NodeDirective are still distinct things and I think, as it grows, we will still want to keep it different to avoid Node ending up as a God class :thinking:

docs/modules/ROOT/pages/type-definitions/database-mapping.adoc
```diff
@@ -43,3 +43,94 @@ interface UserLivesInProperties @relationshipProperties {
...
+RETURN this { .title } as this
+----
+
+==== AdditionalLabels
+
+AdditionalLabels lets you define extra Neo4j labels that need to exist in the node for that GraphQL type.
+
+[source, graphql, indent=0]
+----
+type Actor @node(additionalLabels=["Person", "User"]) {
```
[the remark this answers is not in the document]
angrykoala: :facepalm: My dumb brain does not differentiate between GraphQL and TypeScript, good catch!

darrellwarde:
> @darrellwarde In the case of the labels is quite clear that having something in the Node class is needed, however, Node and NodeDirective are still distinct things and I think, as it grows, we will still want to keep it different to avoid Node ending up as a God class 🤔

Personally, I only see a class as a "God class" if it's tying together domains which aren't at all related to each other. I can't think of anything which we would add to the `@node` directive which wouldn't sit naturally in the `Node` domain.

oskarhane:
> Big +1 on @danstarns suggestion of removing the `||` everywhere in favour of an accessor on the `Node` class. I'm wondering whether instead of `nodeDirective` on the `Node` class, should there just be a `label` field given that is all that the `@node` directive provides? Even as we extend it, we should be extracting the value out of the directive, rather than just saving all of the directive fields with another object on the `Node` type. Otherwise we're going to see a load of `node.nodeDirective.somethingElse` everywhere as we extend it, for example `node.nodeDirective.plural` which should probably just be `node.plural`. If that makes sense?

Thanks @danstarns and @darrellwarde, this was exactly the kind of feedback we wanted on this piece of code. 
We discussed it while writing it and landed in "let's commit it like this and see that D&D says about it".

Personally I do think extracting the things from the `@node` directive and place them directly in the node class makes sense. i.e. a `.labels` property in it as Darrell suggests.

angrykoala:
Changes addressed, a small summary of changes for the PR:

* labels are now accessed through the getters labels and labelString in Node. The bulk of the work still happens at the nodeDirective. The reasoning for using a getter instead of setting a parameter is simply to ensure consistency between name+nodeDirective and the labels string, being this a bit cleaner than having another parameter. Can be easily changed if needed.

* NodeBuilder is a simple builder to generate valid Node instances for testing, I think this approach is quite flexible for generating dummy data and simple enough. I would like some input on what do you think, as this can be rolled to more complex object in the future which may help with testing

packages/graphql/src/utils/test/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export { default as NodeBuilder } from "./node-builder";
+export const dummy = null; // To avoid linter prefer-default
```
darrellwarde, line 21: I think better to have an `eslint-disable` comment over this.

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -251,12 +251,14 @@ function createProjectionAndParams({
         }
 
         if (relationField) {
+            // WARN: In some test this node is undefined. If this is intended, we should update typings to reflect that
```
darrellwarde, line 254: Can you explain what you mean by this?

docs/modules/ROOT/pages/type-definitions/database-mapping.adoc
```diff
@@ -43,3 +42,94 @@ interface UserLivesInProperties @relationshipProperties {
...
+== `@node`
+
+The `@node` directive is used to specify the configuration of a GraphQL object type which represents a Neo4j node.
+
+=== Definition
+
+[source, graphql, indent=0]
+----
+"""Indicates the underlying Neo4j labels that this type will be mapped to."""
+directive @node(label: String, additionalLabels: [String]) on OBJECT
```
darrellwarde, line 56: This still needs to be married up to the definition in "packages/graphql/src/schema/validation/directives.ts":

```suggestion
"""Informs @neo4j/graphql of node metadata"""
directive @node(
    """Map the GraphQL type to a custom Neo4j node label"""
    label: String
    """Map the GraphQL type to match additional Neo4j node labels"""
    additionalLabels: [String]
) on OBJECT
```

darrellwarde (changes requested):
Much better without all of that conditional logic! Mainly just repeating a few comments which were missed from my last review here. 🙂

packages/graphql/src/translate/create-projection-and-params.ts
angrykoala: Our typings are casting the variable `referenceNode` as type Node. But in some tests (22) this variable is `undefined`.

Sadly, I'm not sure (due to lack of context) if this is an error on our tests (e.g. A mock not being properly set) or if it is intended and thus, we should update the type to be `Node | undefined` (which, in turn, requires further updates on the code.

I'll update the comment, to try and explain myself better. Or do you think I should remove the comment altogether?

darrellwarde: Honestly, I think that's just laziness in writing tests more than anything - there's no problem with the production code itself as we wouldn't have gotten to the translation step if `referenceNode` was `undefined`.

I think definitely remove the comment, perhaps just keep a note of this in Trello or something instead.

darrellwarde (dismissed):
Looks good to me, nice work! 🏆

danstarns (changes requested):
Since we merged the aggregations, you will need to make a small change in the translateAggregate function to include the labels.

angrykoala:
@danstarns Done. Please review my last commit to make sure the tests I added make sense :sweat_smile: 

Also, a couple of changes needed by our lovely eslint

danstarns (approved):
Nice work 🍻

PR #479, opened 2021-09-10
Move schema TCK tests to run directly in Jest (assert using inline snapshots)

oskarhane:
# Description

Migrate schema tests from markdown into jest/ts and use inline snapshots.
Using inline snapshots makes in easy to see the output and easy to update multiple snapshots when changes are made.

This should take away some pain and fasten the feedback loop a bit.

Next on the agenda: - Do the same with the Cypher tests.

This was obviously not made by hand - let me know if something's missed in the migration.

New commands to run the schema tests `yarn run test:schema` and `yarn run test:schema:watch`.

### Inline snapshots over snapshot files
An alternative to keep the TCK aspect would be to use snapshot files rather than inline snapshots. Why I didn't go for that is because I wonder if there's actually any value in us providing output for TCK at this point and hosing the result inline just makes it's easier to reason about when you're developing.
Inline snapshots doesn't clutter the file system either, which is a plus.

# Checklist


packages/graphql/tests/schema/comments.test.ts
```diff
@@ -0,0 +1,216 @@
...
+              nodesCreated: Int!
+              relationshipsCreated: Int!
+            }
+
+            type CreateMoviesMutationResponse {
+              info: CreateInfo!
+              movies: [Movie!]!
+            }
+
+            \\"\\"\\"A custom scalar.\\"\\"\\"
```
angrykoala, line 77: Are the escape characters needed here? 🤔

oskarhane: yeah, those snapshots are generated - so whatever's there is correct :)

packages/graphql/tests/schema/arrays.test.ts
```diff
@@ -0,0 +1,149 @@
...
+ * limitations under the License.
+ */
+
+import { printSchemaWithDirectives } from "@graphql-tools/utils";
+import { lexicographicSortSchema } from "graphql/utilities";
+import { Neo4jGraphQL } from "../../src";
+
+describe("Arrays", () => {
+    test("Arrays", () => {
+        const typeDefs = `
```
darrellwarde, line 26: Could we get the `gql` tag added to these so that formatting can be applied correctly?

```suggestion
        const typeDefs = gql`
```

darrellwarde:
This is super cool, I think this will help our productivity a lot. Would be great to see the appropriate tag for formatting GraphQL added, and then all of the new files formatted properly.

packages/graphql/tests/schema/arrays.test.ts
oskarhane: Ah, good catch! Updated now.

danstarns (approved):
I think this is great and will help with productivity and also allow us to employ programmatic and dynamic approaches to our schema testing. I think we should get this in a quickly as possible as to make it easier for the work being developed and the open PRS to switch. Do you see something similar for the cypher ones?

PR #481, opened 2021-09-10
Update tck tests on nested disconnect

angrykoala:
# Description

This PR updates the file `disconnect.md` fixing the usage of disconnect instead of connect for the test.
Further testing of disconnect already in update.md

# Checklist


danstarns (approved):
Good catch on this one. Feel free to use whitespace and tabs inside the cypher output - it usually helps with readability however, due to the complexity, the output here is never going to be easy to read.

PR #484, opened 2021-09-12
Feature: Allow Default Options on Relationship Fields

dmoree:
# Description

This PR allows for specifying a default options object on relationship fields. For example,
```graphql
type Movie {
  title: String!
  genres(options: GenreOptions = [{ name: ASC }]): [Genre!]!
    @relationship(type: "HAS_GENRE", direction: OUT)
}

type Genre {
  name: String!
}
```

By specifying such an argument a query like:
```graphql
query {
  movies {
    title
    genres {
      name
    }
}
```

will return a list of movies with genres sorted by name in ascending order. An options argument on the field in a query will override the default. Meaning the query

```graphql
query {
  movies {
    title
    genres(options: { sort: [{ name: DESC }] }) {
      name
    }
}
```
will sort the genres by name in descending order as intended.

In the case of unions the options field is called `QueryOptions` and is restricted to `limit` and `offset`.

# Checklist


darrellwarde:
Thanks for the contribution @dmoree, we'll get around to looking at it over the coming days. 🙂

darrellwarde:
So we have discussed this, and we're not 100% sure that achieving this using default arguments is the best way forward here in terms of usability. There is a lot of boilerplate in `options: GenreOptions = [{ name: ASC }]` which is maybe not too intuitive to use. 

So the main question from me is, does this need to be specific to relationship fields, or is it likely that users are just always going to want to sort genres by their name ascending? Based on the answer to that being "I always want to sort like that", then my suggestion would instead be either:

Introduce a new `@defaultSort` (or similar) to be used on a field:

```graphql
type Genre {
  name: String! @defaultSort(direction: ASC)
}
```

Or use the new `@node` directive to specify this:

```graphql
type Genre @node(defaultSort: { name: ASC }) {
  name: String!
}
```

I think I definitely prefer the former, because we can just throw an error if it's used more than once in any type, we can define a static definition for it, and it's going to be less brittle than adding this to the `@node` directive. Thoughts on this?

dmoree:
I think your approach actually solves the problem better in that what was to be achieved was ultimately to sort on the node.

The second option would perhaps allow for multiple `ORDER BY` statements that can be ordered according to the position in the array. It would also share the same structure of the existing options field.

Frankly, this PR was premature and this would probably be better as an issue. If you agree, feel free to close this and I will move the discussion over there.

darrellwarde:
True, being able to do multiple sorts is more akin to the existing sort functionality elsewhere, and saves the introduction of _another_ directive!

Let's move this over to an issue for now, I've created #499 from my comments, and the rest of the team and others can chime in if they like.

Issue #482, opened 2021-09-10
Auth rules do not compose correctly


rcbevans:
**Describe the bug**
Given the following auth rules:

```
@auth(rules: [
    { 
        operations: [READ], 
        roles: ["""member"""], 
        where: {
            owner: { authId: """$jwt.sub""" },
        }
    },
    { 
        operations: [READ], 
        roles: ["""admin"""],
    },
])
```

I am trying to express that a `member` or an `admin` can perform read operations on the annotated type, but that `members` can *only* see nodes where the connected `owner` node has an `authId` matching their jwt subject.

However, the generated Cypher for the auth, whilst correctly checking for either member or admin roles, is not composing them with other rules in an appropriate way leading to the following Cypher being executed:

```
MATCH (this:AppConn)
WHERE EXISTS((this)<-[:OWNS_CONNECTION]-(:Person)) AND ALL(owner IN [(this)<-[:OWNS_CONNECTION]-(owner:Person) | owner] WHERE owner.authId IS NOT NULL AND owner.authId = $this_auth_where0_owner_authId)
CALL apoc.util.validate(NOT(ANY(r IN ["member"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) OR ANY(r IN ["admin"] WHERE ANY(rr IN $auth.roles WHERE r = rr))), "@neo4j/graphql/FORBIDDEN", [0])
RETURN this { .id, .name, .description, created_time: apoc.date.convertFormat(toString(this.created), "iso_zoned_date_time", "iso_offset_date_time"), updated_time: apoc.date.convertFormat(toString(this.updated), "iso_zoned_date_time", "iso_offset_date_time"), source_app_user: this.authenticatedUser, user_id:  apoc.cypher.runFirstColumn("MATCH (this)<-[:OWNS_CONNECTION]-(p:Person) RETURN p.authId", {this: this, auth: $auth}, false) } as this
SKIP $this_offset
LIMIT $this_limit
```

The Cypher is *always* applying the where filter such that admins cannot see everything, rather than combining it with a stronger requirement like roles to scope the filtering.

**Type definitions**
```
type Person @auth(rules: [
    { 
        operations: [READ], 
        roles: ["""member""", """admin"""] 
    },
    {
        operations: [CREATE, UPDATE, CONNECT, DISCONNECT, DELETE ]
        roles: ["""admin"""] 
    },
]) @exclude(operations: [CREATE, UPDATE]) {
[...]

PR #490, opened 2021-09-16
Fix: Combine where statements with roles as moderators, as with allow, bind

rcbevans:
Given the following auth rules:

```
@auth(rules: [
    {
        operations: [READ],
        roles: ["""member"""],
        where: {
            owner: { authId: """$jwt.sub""" },
        }
    },
    {
        operations: [READ],
        roles: ["""admin"""],
    },
])
```

I am trying to express that a member or an admin can perform read operations on the annotated type, but that members can only see nodes where the connected owner node has an authId matching their jwt subject.

However, the generated Cypher for the auth, whilst correctly checking for either member or admin roles, is not composing them with other rules in an appropriate way leading to the following Cypher being executed:

```
MATCH (this:AppConn)
WHERE EXISTS((this)<-[:OWNS_CONNECTION]-(:Person)) AND ALL(owner IN [(this)<-[:OWNS_CONNECTION]-(owner:Person) | owner] WHERE owner.authId IS NOT NULL AND owner.authId = $this_auth_where0_owner_authId)
CALL apoc.util.validate(NOT(ANY(r IN ["member"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) OR ANY(r IN ["admin"] WHERE ANY(rr IN $auth.roles WHERE r = rr))), "@neo4j/graphql/FORBIDDEN", [0])
RETURN this { .id, .name, .description, created_time: apoc.date.convertFormat(toString(this.created), "iso_zoned_date_time", "iso_offset_date_time"), updated_time: apoc.date.convertFormat(toString(this.updated), "iso_zoned_date_time", "iso_offset_date_time"), source_app_user: this.authenticatedUser, user_id:  apoc.cypher.runFirstColumn("MATCH (this)<-[:OWNS_CONNECTION]-(p:Person) RETURN p.authId", {this: this, auth: $auth}, false) } as this
SKIP $this_offset
LIMIT $this_limit
```

The Cypher is always applying the where filter such that admins cannot see everything, rather than combining it with a stronger requirement like roles to scope the filtering.

This change now moderates `where` rules with their co-existent `roles` rules such that the `where` policy for one role will only apply when the `role` policy applies:

```
MATCH (this:AppConn)
WHERE ANY(r IN ["member"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) AND EXISTS((this)<-[:OWNS_CONNECTION]-(:Person)) AND ALL(owner IN [(this)<-[:OWNS_CONNECTION]-(owner:Person) | owner] WHERE owner.authId IS NOT NULL AND owner.authId = $this_auth_where0_owner_authId) OR ANY(r IN ["admin"] WHERE ANY(rr IN $auth.roles WHERE r = rr))
CALL apoc.util.validate(NOT(ANY(r IN ["member"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) OR ANY(r IN ["admin"] WHERE ANY(rr IN $auth.roles WHERE r = rr))), "@neo4j/graphql/FORBIDDEN", [0])
RETURN this { .id, .name, .description, created_time: apoc.date.convertFormat(toString(this.created), "iso_zoned_date_time", "iso_offset_date_time"), updated_time: apoc.date.convertFormat(toString(this.updated), "iso_zoned_date_time", "iso_offset_date_time"), user_id:  apoc.cypher.runFirstColumn("MATCH (this)<-[:OWNS_CONNECTION]-(p:Person) RETURN p.authId", {this: this, auth: $auth}, false) } as this
SKIP $this_offset
LIMIT $this_limit
```

I still perform a check to skip generation of a `WHERE` statement if no rules contain a `where`.

# Issue

#482 

# Checklist


rcbevans:
@darrellwarde, @danstarns I have updated the tests and documentation for this behavior.  Please let me know if there are any other changes you would like to see made.

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-490.surge.sh

rcbevans:
> What is here looks good, however, it is very hard to tell whether something _actually works_ based off TCK tests. Any chance of some integration tests for this functionality?
> 
> I know it's been a while since you've looked at this, apologies. disappointed

I'll see if I can add some integration tests for this.

darrellwarde:
Right, I've merged recent changes from `master` into this and added a little integration test, should be good to go! As I've now made changes on this branch, I shouldn't really approve, so will wait until the New Year for the return of the rest of the team.

rcbevans:
Thanks Darrell.

With the changes to tck tests, contributions, including bringing changes currently forked off up to speed on latest should be much less time consuming!

darrellwarde (dismissed):
We were a bit worried that this would be a breaking change for users who have auth rules which this would affect, but good to merge now as major release around the corner!

packages/graphql/src/translate/create-auth-and-params.test.ts
```diff
@@ -94,6 +94,92 @@ describe("createAuthAndParams", () => {
...
+                name: "Movie",
+                relationFields: [],
+                cypherFields: [],
+                enumFields: [],
+                scalarFields: [],
+                primitiveFields: [idField],
+                temporalFields: [],
+                interfaceFields: [],
+                objectFields: [],
+                pointFields: [],
```
angrykoala, line 135: Most of these empty array fields are already set by NodeBuilder class, so we should be able to remove all default fields:
```suggestion
                primitiveFields: [idField],
```

packages/graphql/src/translate/create-auth-and-params.test.ts
```diff
@@ -94,6 +94,92 @@ describe("createAuthAndParams", () => {
...
+
+            const sub = generate({
+                charset: "alphabetic",
+            });
+
+            // @ts-ignore
+            const context: Context = { neoSchema };
+            context.jwt = {
+                sub,
+            };
```
angrykoala, line 165: Here, we should try to avoid `@ts-ignore`, for example using `ContextBuilder`

```suggestion
            const context = new ContextBuilder({
                jwt: {
                    sub,
                },
            }).instance();
```

packages/graphql/src/translate/create-auth-and-params.test.ts
```diff
@@ -94,6 +94,92 @@ describe("createAuthAndParams", () => {
...
+
+            const result = createAuthAndParams({
+                context,
+                entity: node,
+                where: { node, varName: "this" },
+            });
+
+            expect(trimmer(result[0])).toEqual(
+                trimmer(`
+                    ((ANY(r IN [\"admin\"] WHERE ANY(rr IN $auth.roles WHERE r = rr))) OR ((ANY(r IN [\"member\"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) AND this.id IS NOT NULL AND this.id = $this_auth_where1_id)))
```
angrykoala, line 175: Eslint will complain due to unnecessary escape characters:
```suggestion
                    ((ANY(r IN ["admin"] WHERE ANY(rr IN $auth.roles WHERE r = rr))) OR ((ANY(r IN ["member"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) AND this.id IS NOT NULL AND this.id = $this_auth_where1_id)))
```

packages/graphql/src/translate/create-auth-and-params.test.ts
```diff
@@ -94,6 +94,92 @@ describe("createAuthAndParams", () => {
...
+                        },
+                    ],
+                    type: "JWT",
+                },
+            }).instance();
+
+            // @ts-ignore
+            const neoSchema: Neo4jGraphQL = {
+                nodes: [node],
+            };
```
angrykoala, line 155: Mmm, I may be missing something, but I think this variable is no longer used in this test case :thinking: 
```suggestion
```

packages/graphql/tests/integration/directives/auth/roles.int.test.ts
```diff
@@ -866,4 +867,91 @@ describe("auth/roles", () => {
...
+                    schema: neoSchema.schema,
+                    source: query,
+                    contextValue: { driver, req: userReq, driverConfig: { bookmarks: session.lastBookmark() } },
+                });
+
+                expect(gqlResultUser.data).toEqual({
+                    [type.plural]: [{ id: userId, name: "User1", password: "password" }],
+                });
+
+                // request with role "admin" - should return all users
```
angrykoala, line 936: As we are trying to test 2 separate flows (`request with role "user"` and `request with role "admin"`) How about splitting this test case into 2 different test cases (using these comments as the test title) and using the top `describe` block for the setup?

packages/graphql/tests/tck/tck-test-files/directives/auth/arguments/roles-where.md
```diff
@@ -0,0 +1,1774 @@
+# Cypher Auth Where
```
angrykoala, line 1: Markdown-based tck tests have been discontinued in favor of jest-based tests, migration should be pretty straightforward (an example can be found in `packages/graphql/tests/tck/tck-test-files/directives/auth/arguments/where/where.test.ts`

angrykoala (changes requested):
This PR looks great, there is a set of markdown-based tests that are no longer supported

Additionally, I added a few proposals to slightly improve typings in some of the tests

packages/graphql/tests/integration/directives/auth/roles.int.test.ts
darrellwarde: Given that this is more of an opinionated comment compared to the Markdown TCK test and ESLint violations, how about we leave this one be in favour of getting this PR merged?

angrykoala: I don't think this is a big deal, so I'm happy to merge as is, but I wouldn't say that this being 2 separate test cases, by definition, is opinionated

darrellwarde: In my opinion, it could be easily seen as one test case, because it is showing that users with different roles can query the same instance of the schema with the same data and receive the expected results.

darrellwarde: I'm going to merge because there is bigger fish to fry and this has been open for **4 months**. 😨

Issue #487, opened 2021-09-16
Fields relational/cypher fields are not resolved for custom queries/mutations returning Interfaces or Unions


rcbevans:
**Describe the bug**
Related nodes, cypher fields are not resolved on interface/union types returned by custom cypher fields/mutations.  Queries/Mutations returning concrete types *do* correctly have requested fields and edges traversed for use in the response.

**Type definitions**
Given a simple schema:

```

type Author {
    name: String!  @relationship(type: "WROTE", direction: OUT)
}

type Director {
    name: String! @relationship(type: "DIRECTED", direction: OUT)
}

interface ThingWithTitleInterface {
    title: String!
}

type Book implements ThingWithTitleInterface {
   title: String!

    author: Author! @relationship(type: "WROTE", direction: IN)
    numberOfPages: Int!
}

type Movie implements ThingWithTitleInterface {
    title: String!

    director: Director! @relationship(type: "DIRECTED", direction: IN)
    runtime: Int!
}

union ThingWithTitle: Book | Movie

type Query {
    getByTitle(title: String!): [ThingWithTitle!] @cypher(statement: "<cypher query>")
    
    **** OR ****

    getByTitle(title: String!): [ThingWithTitleInterface!] @cypher(statement: "<cypher query>")
}
```

I should be able to call with a query like:

```
getByTitle(title: "Jurassic Park") {
    title
    ... on Book {
        author {
           name
       }
        numberOfPages
    }
    ... on Movie {
        director {
           name
       }
        runtime
    }
}
```

however, the call throws an error:

`Cannot return null for non-nullable field Movie.director`

Looking at the cypher being generated:

```
WITH apoc.cypher.runFirstColumn("<cypher query>", {auth: $auth, input: $input}, false) as x
                UNWIND x as this
                WITH this


RETURN this  AS this
```
It is clear that no attempt is being made to resolve connected edges, either director, or author, whichever concrete type is returned by `getByTitle`.

Fields and relationships *are* correctly evaluated *IFF* a concrete type is returned by the custom query or mutations, for example a custom query returning a specific Book or Movie.

**Expected behavior**
The GraphQL Library should correctly resolve and return data as requested by the caller.

[...]

PR #491, opened 2021-09-17
feat: resolve fields on unions returned by top-level custom cypher

rcbevans:
Resolve fields on unions returned from top-level cypher statements.

This is a port of @john2094's change for the same feature onto the latest master branch.

# Description

In the cypher resolver, check if the return from the cypher field is a union, and if so, generate the projection appropriately

# Issue

#487 

# Checklist

### DIFF
--- a/packages/graphql/src/schema/resolvers/cypher.ts
+++ b/packages/graphql/src/schema/resolvers/cypher.ts
@@ -18,8 +18,11 @@
  */
 
 import { isInt } from "neo4j-driver";
-import { execute } from "../../utils";
+import { UnionTypeDefinitionNode } from "graphql/language/ast";
+import { ResolveTree } from "graphql-parse-resolve-info";
 import { BaseField, ConnectionField, Context } from "../../types";
+import { Node } from "../../classes";
+import { execute } from "../../utils";
 import { graphqlArgsToCompose } from "../to-compose";
 import createAuthAndParams from "../../translate/create-auth-and-params";
 import createAuthParam from "../../translate/create-auth-param";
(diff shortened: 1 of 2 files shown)


rcbevans:
This code is lifted and shifted from the previous PR onto master.

Looking at the code though, it seems like this implementation is executing the cypher query once for each type in the union since `addSecurityAndConnection` is called within the foreach loop.

I am also taking a look at the cypherField projection in [create-project-and-params](https://github.com/neo4j/graphql/blob/d51fb4c112d7e6efe96207f1c8155c08b0d2a3f8/packages/graphql/src/translate/create-projection-and-params.ts#L171).  I'm new to Neo4j/Cypher and so I'm not 100% sure what the generated cypher should look like.

darrellwarde:
Thanks @rcbevans, we'll take a look when we're back from the weekend - have a good one! 🙂

rcbevans:
Just looked at the code being executed and it is indeed performing the APOC call twice within an outer union.

```
call {


                WITH apoc.cypher.runFirstColumn("<statement>", {auth: $auth, input: $input}, false) as x
                UNWIND x as this
                WITH this
            

            match(thisnode:Account)

            where thisnode=this or apoc.map.merge(properties(thisnode), { __resolveType: 'Account' })=properties(this)\n" +
            return thisnode { __resolveType: "Account", .id, ... } as this
union

                WITH apoc.cypher.runFirstColumn("<statement>", {auth: $auth, input: $input}, false) as x
                UNWIND x as this
                WITH this
            

            match(thisnode:Page)

            where thisnode=this or apoc.map.merge(properties(thisnode), { __resolveType: 'Page' })=properties(this)\n" +
            return thisnode { __resolveType: "Page", .id, ... } as this
}
with this AS this
            RETURN this
```

rcbevans:
I have an alternate PR which tackles resolving both top level cypher, as well as cypherFields, #495.

That implementation does a single apoc call then uses the same pattern as in relational fields to match the union types and and project.

darrellwarde:
> I have an alternate PR which tackles resolving both top level cypher, as well as cypherFields, #495.
> 
> That implementation does a single apoc call then uses the same pattern as in relational fields to match the union types and and project.

So are you suggesting that we close this and review only #495 instead? 🙂 Thanks for all of this work!

rcbevans:
@darrellwarde yes I think so.  #495 is a more efficient implementation

PR #492, opened 2021-09-17
Custom pluralization

angrykoala:
# Description

Exposes the parameter `plural` to the `@node` directive to define a custom plural for the generated operations.

```graphql
type Tech @node(plural: "Techs") {
    name: String
}
```

# Checklist

- [X] Documentation has been updated

### DIFF
--- a/docs/modules/ROOT/pages/type-definitions/database-mapping.adoc
+++ b/docs/modules/ROOT/pages/type-definitions/database-mapping.adoc
@@ -62,7 +62,7 @@ directive @node(
 ----
 
 === Usage
-`@node` can be used with the optional parameters `label` and `additionalLabels`.
+`@node` can be used with the following optional parameters.
 
 ==== `label`
 The parameter `label` defines the label to be used in Neo4j instead of the GraphQL type name:
@@ -123,7 +123,7 @@ MATCH (this:Actor:Person:User)
 RETURN this { .name } as this
 ----
 
-Note that both parameters can be used at the same time:
+Note that `additionalLabels` can be used along with `label`:
 
 [source, graphql, indent=0]
 ----
@@ -138,3 +138,27 @@ In this case, the resulting Cypher query will use the labels `ActorDB` and `Pers
 MATCH (this:ActorDB:Person)
 RETURN this { .name } as this
 ----
+
+==== `plural`
+The parameter `plural` redefines how to compose the plural of the type for the generated operations. This is particularly
+useful for types that are not correctly pluralized or are non-english words.
+
+[source, graphql, indent=0]
+----
+type Tech @node(plural: "Techs") {
+  name: String
+}
+----
+
+This way, instead of the wrongly generated `teches`, the type is properly written as `techs`:
+
+[source, graphql, indent=0]
+----
+{
+  techs {
+    title
+  }
+}
+----
+
+The same is applied to other operations such as `createTechs`. Note that database labels will not change.
--- a/packages/graphql/src/classes/Node.ts
+++ b/packages/graphql/src/classes/Node.ts
@@ -61,41 +61,23 @@ export interface NodeConstructor {
 
 class Node {
     public name: string;
-
     public relationFields: RelationField[];
-
     public connectionFields: ConnectionField[];
-
     public cypherFields: CypherField[];
-
     public primitiveFields: PrimitiveField[];
-
     public scalarFields: CustomScalarField[];
-
     public enumFields: CustomEnumField[];
-
     public otherDirectives: DirectiveNode[];
-
     public unionFields: UnionField[];
-
     public interfaceFields: InterfaceField[];
-
     public interfaces: NamedTypeNode[];
-
     public objectFields: ObjectField[];
-
     public temporalFields: TemporalField[];
-
     public pointFields: PointField[];
-
     public ignoredFields: BaseField[];
-
     public exclude?: Exclude;
-
     public nodeDirective?: NodeDirective;
-
     public auth?: Auth;
-
     public description?: string;
 
     /*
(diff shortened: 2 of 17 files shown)


packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -193,12 +192,14 @@ function makeAugmentedSchema(
...
-    const aggregationSelectionTypes = aggregationSelectionTypeMatrix.reduce<Record<string, ObjectTypeComposer<unknown, unknown>>>((res, [name, fields]) => {
+    const aggregationSelectionTypes = aggregationSelectionTypeMatrix.reduce<
+        Record<string, ObjectTypeComposer<unknown, unknown>>
+    >((res, [name, fields]) => {
         return {
             ...res,
             [name]: composer.createObjectTC({
                 name: `${name}AggregateSelection`,
-                fields: fields ? fields : { min: `${name}!`, max: `${name}!` },
+                fields: fields || { min: `${name}!`, max: `${name}!` },
```
angrykoala, line 202: eslint made me do it

packages/graphql/src/classes/Node.ts
```diff
@@ -176,6 +160,14 @@ class Node {
     get labels(): string[] {
         return this.nodeDirective?.getLabels(this.name) || [this.name];
     }
+
+    getPlural(camelcase: boolean): string {
```
angrykoala, line 164: As previously discussed, the parameter can be removed by fixing the inconsistencies between queries and mutations in automatic casing. However, that would be a breaking change requiring a major version update

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-492.surge.sh

packages/graphql/src/schema/make-augmented-schema.ts
darrellwarde: I actually think I prefer this over the ternary operator... 😳

oskarhane: Both are easy to read and understand imo. Let's go with what the tooling prefers :)

packages/graphql/src/schema/parse-node-directive.ts
```diff
@@ -17,25 +17,24 @@
...
-
-    const labelValue: string | undefined = label ? valueFromASTUntyped(label.value) : undefined;
-    const additionalLabelsValue: string[] | undefined = additionalLabels
-        ? valueFromASTUntyped(additionalLabels.value)
-        : undefined;
+    return new NodeDirective({
+        label: getArgumentValue<string>(nodeDirective, "label"),
+        additionalLabels: getArgumentValue<string[]>(nodeDirective, "additionalLabels"),
+        plural: getArgumentValue<string>(nodeDirective, "plural"),
+    });
```
oskarhane, line 32: 👍

oskarhane (commented):
Great, the code look a bit cleaner now that the pluralize call isn't used deep in the lib anymore!

I have a few questions (that we might answer with tests):

1. What happens if I use the same plural for multiple `@node`'s, or if there's a conflict with an existing auto-pluralized?
2. Does it work as expected with GraphQL aliases (note, not the `@alias` directive). 
```graphql
query {
  actorsQueryAlias: actors {
    name
  }
}
```

docs/modules/ROOT/pages/type-definitions/database-mapping.adoc
```diff
@@ -138,3 +138,27 @@ In this case, the resulting Cypher query will use the labels `ActorDB` and `Pers
 MATCH (this:ActorDB:Person)
 RETURN this { .name } as this
 ----
+
+==== `plural`
+The parameter `plural` redefines how to compose the plural of the type for the generated operations. This is particularly
+useful for types that are not correctly pluralized or are non-english words.
```
darrellwarde, line 144:
```suggestion
useful for types that are not correctly pluralized or are non-English words.
```

angrykoala:
@oskarhane 

>  What happens if I use the same plural for multiple @node's, or if there's a conflict with an existing auto-pluralized?

At the moment in this, any conflict (same plural, collision with pluralize or plural being the same as a type) only the last type will be taken into account. With all previous collided types completely ignored. Note that this is the same behavior that happens if pluralize collides with a type (e.g `Type User` and `Type Users`). I've written a few tests to validate this behavior (and to test if we want to change it), however, I'm not sure on what would be the ideal behaviour.

> Does it work as expected with GraphQL aliases (note, not the @alias directive).

Working as expected, I added a test to validate this

packages/graphql/src/classes/Node.ts
darrellwarde, line 164: I would personally prefer this to be called using `getPlural({ camelCase: true })`, but it is very much a personal preference thing.

darrellwarde (changes requested):
Looks great! Literally just one letter to capitalise, and then just a general thought about a function signature.

packages/graphql/src/classes/Node.ts
angrykoala: I prefer as well, but it collides with the actual camelCase library :laughing: 
I guess we could be more explicit with that fact that this is an option or flag (`camelCaseFlag`)?

darrellwarde: I was actually talking about the destructuring vs parameter, but funny about the naming conflict!

I think it's just when you see `getPlural(true)` in the codebase, it's unclear what that "magic flag" as you call it is doing. But no real need to change, more of an observation!

angrykoala: nah, I though about that, and you are right (already changed it). I think is better to have explicit flags rather than random booleans.

To avoid name conflict I went with an options parameter instead of destructuring

oskarhane (dismissed):
LGTM, well done! 💯

Issue #493, opened 2021-09-17
Error when filtering on union relationships


dmoree:
#### Describe the bug
When using a union with the `@relationship` directive if the return type is `[{Union}!]!` an error will be thrown as non-matched nodes are returned as null.

#### Type definitions
```graphql
type Director {
  id: ID! @id
  name: String!

  movies: [Movie!]! @relationship(type: "DIRECTED", direction: OUT)
}

type Actor {
  id: ID! @id
  name: String!

  movies: [Movie!]! @relationship(type: "ACTED_IN", direction: OUT)
}

union Person = Director | Actor

type Movie {
  id: ID! @id
  title: String!

  directors: [Director!]! @relationship(type: "DIRECTED", direction: IN)
  actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)

  people: [Person!]! @relationship(type: "HAS_PERSON", direction: OUT)
}
```

#### To Reproduce

1. Clone repository
```
git clone git@github.com:dmoree/neo4j-graphql-issue-filter-union-field.git
```
2. Seed and start server
```
cd neo4j-graphql-issue-filter-union-field && yarn install && yarn seed && yarn start
```
3. Run the query in GraphiQL at `http://localhost:4000/graphql` (or as defined in `.env`)
###### Query
```graphql
query FilterPeopleByActorName {
  movies {
    people(where: {Actor: { name: "Keanu Reeves"}}) {
      ...on Actor {
        name
      }
    }
  }
}
```

###### Response
```
{
  "error": {
    "errors": [
      {
        "message": "Cannot return null for non-nullable field Movie.people.",
        "locations": [
          {
            "line": 3,
            "column": 5
          }
        ],
        "path": [
          "movies",
          0,
          "people",
          0
        ]
      }
    ],
    "data": null
  }
}
```

#### Expected Behavior
Expect to be able to filter union relationships by type.

###### Entire graph
![Screen Shot 2021-09-17 at 10 05 58 AM](https://user-images.githubusercontent.com/10360334/133827059-8c957de6-5658-4cf4-98b9-1ee19f057070.png)
###### Cypher Query
![Screen Shot 2021-09-17 at 10 14 26 AM](https://user-images.githubusercontent.com/10360334/133828047-f931271a-95cb-49ad-88ef-194cf69e2136.png)
#### System (please complete the following information):

-   OS: macOS@11.5.2
-   Version: @neo4j/graphql@2.2.0
-   Node.js version: 14.16.0

[...]

PR #494, opened 2021-09-17
Fix: Filter Null Values In Union Relationship

dmoree:
# Description

Proposed fix for #493. Current behavior returns null values. This PR filters out null values.

# Issue

- #493 

# Checklist

### DIFF
--- a/packages/graphql/src/translate/create-projection-and-params.ts
+++ b/packages/graphql/src/translate/create-projection-and-params.ts
@@ -268,7 +268,7 @@ function createProjectionAndParams({
                 );
 
                 const unionStrs: string[] = [
-                    `${key}: ${!isArray ? "head(" : ""} [(${
+                    `${key}: ${!isArray ? "head(" : ""} [${param} IN [(${
                         chainStr || varName
                     })${inStr}${relTypeStr}${outStr}(${param})`,
                     `WHERE ${referenceNodes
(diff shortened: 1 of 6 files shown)


packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -323,7 +323,7 @@ function createProjectionAndParams({
                     return innerHeadStr.join(" ");
                 });
                 unionStrs.push(headStrs.join(" + "));
-                unionStrs.push(") ]");
+                unionStrs.push(`) ] WHERE ${param} IS NOT NULL]`);
```
rcbevans, line 326: Would it make sense to only do this if the return type is an array of non-nullable unions?

from the integration test below only for 

`search: [Search!]`

and not

`search: [Search]`

since in the latter case, it's permitted that the return array contain null (I'm not sure what cases there could be where this is desirable, but the GraphQL schema permits it).

rcbevans (approved):

packages/graphql/src/translate/create-projection-and-params.ts
dmoree: I can see the logic of it, but I'm not sure if this would be desired at all as `null` has no information to it. I think the expectation would be to return only non-null values from the database even if the type is not nonnull.

I believe the same reasoning is why all the return types of the generated queries and mutations are nonnull lists of nonnull types.

darrellwarde (approved):
Sure, I think this makes sense! Eventually we'd like to refactor to use subqueries everywhere, and I believe these kind of problems go away then.

PR #495, opened 2021-09-18
feat: Resolve fields on unions returned by custom cypher

rcbevans:
Use a similar pattern to the relationField for resolving fields on union types returned by custom cypher.

This resolution is done on both root cypher queries/mutations, and on cypherFields.

Given a schema:

```
type Actor {
    name: String
    movies(title: String): [Movie]
        @cypher(
            statement: """
            MATCH (m:Movie {title: $title})
            RETURN m
            """
        )

    tvShows(title: String): [Movie]
        @cypher(
            statement: """
            MATCH (t:TVShow {title: $title})
            RETURN t
            """
        )

    movieOrTVShow(title: String): [MovieOrTVShow]
        @cypher(
            statement: """
            MATCH (n)
            WHERE (n:TVShow OR n:Movie) AND ($title IS NULL OR n.title = $title)
            RETURN n
            """
        )

    randomNumber: Int
        @cypher(
            statement: """
            RETURN rand()
            """
        )
}

union MovieOrTVShow = Movie | TVShow

type TVShow {
    id: ID
    title: String
    numSeasons: Int
    actors: [Actor]
        @cypher(
            statement: """
            MATCH (a:Actor)
            RETURN a
            """
        )
    topActor: Actor
        @cypher(
            statement: """
            MATCH (a:Actor)
            RETURN a
            """
        )
}

type Movie {
    id: ID
    title: String
    actors: [Actor]
        @cypher(
            statement: """
            MATCH (a:Actor)
            RETURN a
            """
        )
    topActor: Actor
        @cypher(
            statement: """
            MATCH (a:Actor)
            RETURN a
            """
        )
}
```

The GraphQL query:
```
{
    actors {
        movieOrTVShow(title: "some title") {
            ...on Movie {
                id
                title
                topActor {
                    name
                }
            }
            ...on TVShow {
                id
                title
                topActor {
                    name
                }
            }
        }
    }
}
```

Will generate the cypher query:

```
MATCH (this:Actor)
RETURN this { movieOrTVShow: [this_movieOrTVShow IN apoc.cypher.runFirstColumn("MATCH (n) WHERE (n:TVShow OR n:Movie) AND ($title IS NULL OR n.title = $title) RETURN n", {this: this, auth: $auth, title: $this_movieOrTVShow_title}, false) WHERE ("Movie" IN labels(this_movieOrTVShow)) OR ("TVShow" IN labels(this_movieOrTVShow)) | [ this_movieOrTVShow IN [this_movieOrTVShow] WHERE ("Movie" IN labels(this_movieOrTVShow)) | this_movieOrTVShow { __resolveType: "Movie", .id, .title, topActor: head([this_movieOrTVShow_topActor IN apoc.cypher.runFirstColumn("MATCH (a:Actor) RETURN a", {this: this_movieOrTVShow, auth: $auth}, false) | this_movieOrTVShow_topActor { .name }]) } ] + [ this_movieOrTVShow IN [this_movieOrTVShow] WHERE ("TVShow" IN labels(this_movieOrTVShow)) | this_movieOrTVShow { __resolveType: "TVShow", .id, .title, topActor: head([this_movieOrTVShow_topActor IN apoc.cypher.runFirstColumn("MATCH (a:Actor) RETURN a", {this: this_movieOrTVShow, auth: $auth}, false) | this_movieOrTVShow_topActor { .name }]) } ] ] } as this
```

Note, that this pattern of resolving union types makes use of the `labels(<node>)` cypher function, which means that it expects the type being returned by a custom cypher query to be a note, not a map of properties.  This change also no longer requires devs to manually add the `__resolveType` when returning a union.  Because of this, existing code which explicitly returned a map with `__resolveType` will no longer work, because `labels` expects a node.

# Description

Resolve fields on both top level custom cypher, and cypher fields which return unions

# Issue

#493, #494, #487 

# Checklist


rcbevans:
The implementation is an alternative to #491 where it uses the same pattern as on relationFields to match types returned by the apoc statement then projects.  It covers both top level custom cypher, and cypher fields returning unions.

packages/graphql/tests/integration/issues/388.int.test.ts
```diff
@@ -55,7 +55,7 @@ describe("https://github.com/neo4j/graphql/issues/388", () => {
                         MATCH (myFriends)-[:HAS_POST]->(post:Post)
                         RETURN post
                     }
-                    RETURN DISTINCT apoc.map.merge(properties(post), { __resolveType: 'Post' }) AS result ORDER BY result.modifiedDate DESC
```
rcbevans, line 58: This implementation uses `labels(<node>)` to resolve the node to a concrete type.  The `labels` function requires that it be passed a node, *not* a map  (the integration tests required this merging to be done in the cypher query as no attempt was being made to resolve it to a concrete type, but that is no longer the case).

That said, I can see that it may be useful to support custom cypher which returns a map corresponding to a union of types which do not have labels that exist in the database, so handling the case when a map could be useful in that scenario.

Is there a cypher function/utility to check if a variable is a map?

Then the code could be updated to be something like

`WHERE IS_MAP(this) OR ("U1" IN labels(this)) ... etc`

to skip calling labels, or attempting to further resolve fields, if the value returned is a map.

rcbevans:
Can someone from the neo4j -graphql tesam take a look at this and provide feedback

danstarns:
Looks good, thank you for this contribution and sorry it took soo long for one of us to get to this. I merged the TCK tests with the new format and also added a [integration test to cover this](https://github.com/neo4j/graphql/pull/495/commits/e69164d823d6972eba8e1582733e0a13d2c54ee9).

packages/graphql/tests/tck/tck-test-files/directives/cypher.test.ts
```diff
@@ -272,4 +312,56 @@ describe("Cypher directive", () => {
             }"
         `);
     });
+
+    test("Union directive", async () => {
```
danstarns, line 316: @rcbevans This is the new TCK format.

packages/graphql/tests/integration/issues/388.int.test.ts
danstarns: I think how you have done it is fine, the hack users were using with __resolveType was a workaround and this is much better! On your question about function/utility to check if a variable is a map - I cant find any :/

AccsoSG:
Has this feature been released? Today I tried to test this feature with version 2.5.4. I used the same schema and query from @rcbevans. However, I ran into the following errors:

````json
{
  "errors": [
    {
      "message": "Abstract type \"MovieOrTVShow\" must resolve to an Object type at runtime for field \"Actor.movieOrTVShow\". Either the \"MovieOrTVShow\" type should provide a \"resolveType\" function or each possible type should provide an \"isTypeOf\" function.",
...
}]}
````

I also get the following error when I only query the __typename field:
````graphql
{
    actors {
        movieOrTVShow(title: "TestMovie1") {
        	__typename
      }
    }
}
````
````json
{
  "errors": [
    {
      "message": "Invalid input ']': expected \"NOT\" or an expression (line 4, column 85 (offset: 262))\r\n\"RETURN n\", {this: this, auth: $auth, title: $this_movieOrTVShow_title}, false) |    ] } as this\"\r\n                                                                                     ^",
...          
}]}
````

rcbevans:
Is this a cypher field returning a union, or an interface?

I would expect to see the library generating a projection per type in the union, based on `labels(n)`.

I'm not sure what I'd expect to see for an interface type...

rcbevans:
The code for the cypher resolver can be found at https://github.com/neo4j/graphql/blob/master/packages/graphql/src/schema/resolvers/cypher.ts.

Line 121 shows it generating the code to project the result of a cypher field returning a union.  That code has been there a little while so it should definitely be in the latest release.

AccsoSG:
I generated the exact same scheme that you presented above. So movieOrTVShow field returns a union.

AccsoSG:
I added a root cypher query to the schema. This query works as expected:
````graphql
extend type Query {
    moviesOrTVShows: [MovieOrTVShow]
    @cypher(
            statement: """
            MATCH (n)
            WHERE (n:TVShow OR n:Movie)
            RETURN n
            """
        )
}
````
````graphql
query {
  moviesOrTVShows {
    __typename
  	...on Movie {
        id
        title
        topActor {
            name
        }
    }
    ...on TVShow {
        id
        title
        topActor {
            name
        }
    }
  }
}
````
````json
{
  "data": {
    "moviesOrTVShows": [
      {
        "__typename": "Movie",
        "id": "1",
        "title": "TestMovie1",
        "topActor": {
          "name": "ActorA"
        }
      },
      {
        "__typename": "Movie",
        "id": "2",
        "title": "TestMovie2",
        "topActor": {
          "name": "ActorA"
        }
      },
      {
        "__typename": "TVShow",
        "id": "4",
        "title": "TVShow1",
        "topActor": {
          "name": "ActorA"
        }
      }
    ]
  }
}
````

AccsoSG:
Hi @rcbevans I've analyzed the problem in more detail. I think there are three problems.
For the analysis, I have simplified your scheme a little:
````graphql
extend type Query {
    moviesOrTVShows: [MovieOrTVShow]
    @cypher(
            statement: """
            MATCH (n)
            WHERE (n:TVShow OR n:Movie)
            RETURN n
            """
        )
}

type Actor {
    name: String
    movieOrTVShow: [MovieOrTVShow]
        @cypher(
            statement: """
            MATCH (n)
            WHERE (n:TVShow OR n:Movie)
            RETURN n
            """
        )
}

union MovieOrTVShow = Movie | TVShow

type TVShow {
    id: ID
    title: String
    numSeasons: Int
}

type Movie {
    id: ID
    title: String
}
````

I use the following queries:
(1)
````graphql
query {
  moviesOrTVShows {
    __typename
    ... on Movie {
      title
    }
    ... on TVShow {
      title
    }
  }
}
````
(2)
````graphql
query {
  actors {
    name
    movieOrTVShow {
      __typename
      ... on Movie {
        title
      }
      ... on TVShow {
        title
      }
    }
  }
}
````
(1) The root cypher query works as expected. The cypher resolver is used for resolution (https://github.com/neo4j/graphql/blob/master/packages/graphql/src/schema/resolvers/cypher.ts.)

(2) The second query fails with error message. The cypher resolver is not used for resolution. Instead, this resolver is used: https://github.com/neo4j/graphql/blob/master/packages/graphql/src/schema/resolvers/read.ts
````json
"message": "Abstract type \"MovieOrTVShow\" must resolve to an Object type at runtime for field \"Actor.movieOrTVShow\". Either the \"MovieOrTVShow\" type should provide a \"resolveType\" function or each possible type should provide an \"isTypeOf\" function."
````
For the second query, the following cypher query is generated:
````cypher
MATCH (this:Actor)
RETURN this { .name, movieOrTVShow: [this_movieOrTVShow IN apoc.cypher.runFirstColumn("MATCH (n) WHERE (n:TVShow OR n:Movie) RETURN n", {this: this, auth: $auth}, false) WHERE ("Movie" IN labels(this_movieOrTVShow)) OR ("TVShow" IN labels(this_movieOrTVShow))  |   [ this_movieOrTVShow IN [this_movieOrTVShow] WHERE ("Movie" IN labels(this_movieOrTVShow)) | this_movieOrTVShow { __resolveType: "Movie",  .title } ] + [ this_movieOrTVShow IN [this_movieOrTVShow] WHERE ("TVShow" IN labels(this_movieOrTVShow)) | this_movieOrTVShow { __resolveType: "TVShow",  .title } ] ] } as this
````
The query also works and leads to this result:
image

And here you can see two problems. First, only one Movie / TVShow node is returned at a time. This is because the apoc.cypher.runFirstColumn function is incorrectly called with the parameter "false".
Second, you can see that movieOrTVshow is a nested array. Therefore, the __resolverType field cannot be found when resolving.

For the first two problems, I've adapted the following codes for testing purposes. However, I'm not sure if this is really a clean solution.

(A)
https://github.com/neo4j/graphql/blob/edbd30370ceb431e6824132ba885bd2ded55d951/packages/graphql/src/translate/create-projection-and-params.ts#L285
Since referenceNode is always undefined in the query, I ignored it when calculating expectMultipleValues. However, I don't know the reason why referenceNode is even considered.

````typescript
const expectMultipleValues = cypherField.typeMeta.array ? "true" : "false";
````

(B)
https://github.com/neo4j/graphql/blob/edbd30370ceb431e6824132ba885bd2ded55d951/packages/graphql/src/translate/create-projection-and-params.ts#L308
With the function apoc.coll.flatten I flattened the nested array
````typescript
res.projection.push(`${key}: apoc.coll.flatten([${apocStr}])`);
````

After these changes, the second query also worked as expected.
````json
{
  "data": {
    "actors": [
      {
        "name": "ActorA",
        "movieOrTVShow": [
          {
            "__typename": "Movie",
            "title": "TestMovie1"
          },
          {
            "__typename": "Movie",
            "title": "TestMovie2"
          },
          {
            "__typename": "Movie",
            "title": "TestMovie3"
          },
          {
            "__typename": "TVShow",
            "title": "TVShow1"
          },
          {
            "__typename": "TVShow",
            "title": "TVShow2"
          },
          {
            "__typename": "TVShow",
            "title": "TVShow3"
          }
        ]
      },
      {
        "name": "ActorBA",
        "movieOrTVShow": [
          {
            "__typename": "Movie",
            "title": "TestMovie1"
          },
          .......
        ]
      },
      {
        "name": "ActorC",
        "movieOrTVShow": [
          {
            "__typename": "Movie",
            "title": "TestMovie1"
          },
          ........
        ]
      }
    ]
  }
}
````
The third problem is that both queries won't work if you only query the __typename field. However, the error messages are different. I think the cypher query is not generated validly. I don't currently have a solution for the third problem.

Sorry for the very long comment 😀 @rcbevans What do you say? Have you been able to reproduce my problem?

rcbevans:
That all makes sense, I didn't realize you were having issues with a non-root cypher field, else I would have pointed to the code in `create-projections-and-params.ts` where the cypher fields are projected.

That file is a bit of a monster, and currently the code to generated projections for specific types of fields aren't isolated to be reused in multiple places.

It looks like the library is attempting to resolve the union nodes correctly, but seems it was written with an assumption that the provided cypher statement will return a single list, not multiple rows of nodes; does changing the statement to return `COLLECT(n)` fix the issue?

The difference in assumptions between a root cypher field and non-root one does seem confusing, I'm not sure if it was intentional, or is documented anywhere.

rcbevans:
It does seem the use of `referenceNode` on line 285 of `packages/graphql/src/translate/create-projection-and-params.ts` is a typo, given that referenceNode cannot be truthy if the field being resolved is a union field.  I think it could be a typo, and instead should be `referencedNodes` as defined on line 215.

Changing this line will introduce breaking behavior for existing code, as apoc would now told to expect multiple rows of results rather than a single row, and so any code which was returning a single list of results would then likely have broken cypher to be generated.

From looking at the code, I think the expected behavior is that the cypher statement return a list; I expect making that change will resolve your issue.

AccsoSG:
Yes that is true. The code is really not easy to understand.

Do you mean that I should write the cypher query like this?
````cypher
MATCH (n)
WHERE (n:TVShow OR n:Movie)
RETURN COLLECT(n)
````
Unfortunately that does not work. The same error occurs.

---
That seems complicated. In fact, I was already thinking that this could be a typo. 
Perhaps it makes sense to set expectMultipleValues differently when using referenceUnion 🤔

PR #500, opened 2021-09-20
FEAT: extend auth rules with field modifiers

rcbevans:
This pull request adds initial extended support for modifiers on auth clauses.  The API is designed to mimic the interface of the generated GraphQL query/mutation.

For primitives:

Support _NOT, _IN, _NOT_IN

_NOT checks for inequality using <>
_IN checks for membership in a list of values
_NOT_IN checks that the value is not a member of a list of values

For relationships:

Support _NOT, _INCLUDES, _NOT_INCLUDES, _EVERY, _NOT_EVERY

Also add support for a `null` value to check EXISTS()/NOT EXISTS.

_NOT negates the default ANY/ALL, depending on the clause
_ INCLUDES specifies `ANY`
_NOT_INCLUDES specifies `NOT ANY`
_ EVERY specifies `ALL`
_NOT_EVERY specifies `NOT ALL`

Using null as a value for the relationship checks that no edge exists, _NOT: `null` checks that an edge must exist.

# Issue

The implemented design is that which was proposed by @darrellwarde on issue #469 , following discussion on #293.

# Checklist

### DIFF
--- a/docs/modules/ROOT/pages/auth/auth-directive.adoc
+++ b/docs/modules/ROOT/pages/auth/auth-directive.adoc
@@ -1,7 +1,7 @@
 [[auth-directive]]
 = `@auth` directive
 
-The `@auth` directive definition is dynamically generated on runtime based on user type definitions.
+The `@auth` directive definition is dynamically generated at runtime based on user type definitions.
 
 == `rules`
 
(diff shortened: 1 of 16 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-500.surge.sh

darrellwarde:
Super cool, this is really nice! I'm sure @danstarns could tune in, but can you explain about the default `ANY`/`ALL` and the necessity of `EVERY` here? `EVERY` isn't present in the generated API and there has been no real call for its addition so far, so wondering why it is necessary here?

Another detail we failed to discuss is the distinction between `_IN`/`_NOT_IN` and `_INCLUDES`/`_NOT_INCLUDES`, in that the former should be used for _singular_ primitive/relationship fields, and the latter should be used for array primitive/relationship fields. I haven't reviewed the code here yet, but from your description it doesn't sound like this is the case?

Thanks so much for this though, this will really expand out the auth functionality! 😄

rcbevans:
By default, allow would evaluate relationships using ANY, and where/bind evaluated using ALL.

I didn't want to break existing semantics of working code so opted to allow devs to be explicit about ANY/ALL using _INCLUDES, and _EVERY. I also found a comment in queries.md that there is intent to support _EVERY using ALL but that it wasn't there yet, so it seemed like it aligned.

Without EVERY there is no way of expressing ALL semantics on an allow clause, so this closes that gap too.

IN/NOT_IN is only applied to primitives, INCLUDES/NOT_INCLUDES only applies to relationships.

The code right now doesn't distinguish between singular/array primitives or relationships.  All primitives are treated as singular, and all relationships are treated as array.

rcbevans:
I don't know when I'll get around to looking at this again.  I'd already moved onto other things and rolled the current implementation out for use through a forked package.

binnyFriedman:
Hey, I really need this feature.
Are you planning in implementing this in the near future?
I will be happy to help with a PR (just to close the gap between @rcbevans PR and the current state)
Otherwise, is there any workaround? 
Thanks

Issue #212, opened 2021-05-17
Subscription resolvers are returning null.


tlester:
**Describe the bug**
When using subscriptions, the subscription resolver returns null when using @neo4j/graphql.   If you take the same code and run with a straight apollo server without neo4j, it returns the expected data.  

**Type definitions**
```
type Query {
        currentNumber: Int
}
type Subscription {
        numberIncremented: Int
}
```

resolver:

```
Query: {
        currentNumber() {
            return currentNumber
        },
    },
Subscription: {
    numberIncremented: {
        subscribe: () => pubsub.asyncIterator(["NUMBER_INCREMENTED"]),
  },
},
```

Here's the code that is executing the pubsub. publish.

```
let currentNumber = 0
function incrementNumber() {
    currentNumber++
    pubsub.publish("NUMBER_INCREMENTED", { numberIncremented: currentNumber })
    setTimeout(incrementNumber, 1000)
}

incrementNumber()
```

And how I have my server configured:

```
const neoSchema = new Neo4jGraphQL({
    typeDefs,
    resolvers,
    config: {
        jwt: {
            secret: process.env.NEO4J_GRAPHQL_JWT_SECRET,
        },
    },
    debug: true,
})

const NEO4J_USER = process.env.NEO4J_USER
const NEO4J_PASS = process.env.NEO4J_PASS
const NEO4J_URI = process.env.NEO4J_URI

const driver = neo4j.driver(NEO4J_URI, neo4j.auth.basic(NEO4J_USER, NEO4J_PASS))

const init = async (driver) => {
    await initializeDatabase(driver)
}

try {
    init(driver).catch((e) => console.log(e))
    console.log("ne4j initialized...")
} catch (error) {
    console.error(`Failed to property initialize database`, error)
}

const apolloServer = new ApolloServer({
    schema: neoSchema.schema,
    context: ({ req }) => {
        return {
            driver,
            pubsub,
            req,
        }
    },
    subscriptions: {
        onConnect: async (connectionParams, webSocket, context) => {
            console.log("xxx")
            console.log(connectionParams)
        },
        onDisconnect: (websocket, context) => {
            console.log("WS Disconnected!")
        },
        path: "/graph",
    },
    introspection: true,
    playground: true,
})

const httpServer = createServer(app)

apolloServer.applyMiddleware({ app, path: "/graph" })

apolloServer.installSubscriptionHandlers(httpServer)
const port = process.env.PORT || 8080

httpServer.listen({ port }, () => {
[...]

PR #503, opened 2021-09-21
Fix missing subscription payload

angrykoala:
# Description
Adds a return of the obj parameter, this seems to be the expected way of using addSchemaLevelResolver. Its omission was causing the missing results in subscription payload when using neo4j/graphql library

# Issue

#212 

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,7 @@ e2e/fixtures/cache/
 
 test-api.ts
 packages/package-tests/**/package-lock.json
+scratchpad/
 
 .yarn/*
 !.yarn/patches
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -81,6 +81,7 @@ e2e/fixtures/cache/
 
 test-api.ts
 packages/package-tests/**/package-lock.json
+scratchpad/
```
angrykoala, line 84: just a folder for manual testing, I think it is nice to have one. Not sure about naming

angrykoala:
Sadly, the only way of reproducing this was through an Apollo server and websockets. Automated testing of this bug is not trivial. Is it worth building the whole testing infrastructure? Or someone have ideas for a simpler test?

darrellwarde (approved):
Good find, can't believe we missed this!

Issue #505, opened 2021-09-21
Create-auth-and-params does not wrap sub-clauses in parentheses, causing complex rules to evaluate incorrectly


rcbevans:
**Describe the bug**
Complex auth rules, combined with where filters can generate cypher queries with invalid predicate precedence which evaluate incorrectly.

**Type definitions**

simplified for readability

```
type Person {
    id: ID! @id(autogenerate: false)

    authId: String
    
    workspaces: [Workspace] @relationship(type: "MEMBER_OF", direction: OUT)

    adminOf: [Workspace] @relationship(type: "HAS_ADMIN", direction: IN)
}

type Workspace @auth(rules: [
    { 
        operations: [READ],
        roles: ["""member"""],
        where: { 
            OR: [
                { members_INCLUDES: { authId: """$jwt.sub""" } }
                { admins_INCLUDES: { authId: """$jwt.sub""" } }
            ]
        }
    },
    {
        operations: [READ, CREATE, UPDATE, CONNECT, DISCONNECT, DELETE ]
        roles: ["""admin"""]
    },
]) @exclude(operations: [CREATE, UPDATE]) {
    id: ID! @id(autogenerate: false)

    name: String!

    members: [Person!] @relationship(type: "MEMBER_OF", direction: IN)
    admins: [Person!] @relationship(type: "HAS_ADMIN", direction: OUT)

    pages: [Page!] @relationship(type: "HAS_PAGE", direction: OUT)
}

type Page implements AccountOrPageInterface @auth(rules: [
    { 
        operations: [READ], 
        roles: ["""member"""], 
        where: {
            OR: [ 
                { owner: { authId: """$jwt.sub""" } }, 
                { 
                    AND: [ 
                        { shared: true },
                        { workspace: { members_INCLUDES: { authId: """$jwt.sub""" } } },
                    ] 
                } 
            ] 
        } 
    },
    {
        operations: [READ, CREATE, UPDATE, CONNECT, DISCONNECT, DELETE ]
        roles: ["""admin"""] 
    },
]) {
    id: ID! @id(autogenerate: false)

    name: String!
    shared: Boolean!

    owner: Person! @relationship(type: "CREATED_PAGE", direction: IN)

    workspace: Workspace! @relationship(type: "HAS_PAGE", direction: IN)
}
```

**To Reproduce**
execute a GraphQL query

```
query Pages($pagesWhere: PageWhere) {
    pages(where: $pagesWhere) {
        id
        name
    }
}
```

[...]

PR #506, opened 2021-09-22
fix: wrap auth sub clauses in parentheses to preserve precedence

rcbevans:
# Description

Complex auth rules, combined with where filters can generate cypher queries with invalid predicate precedence which evaluate incorrectly.

Type definitions

simplified for readability

```
type Person {
    id: ID! @id(autogenerate: false)

    authId: String
    
    workspaces: [Workspace] @relationship(type: "MEMBER_OF", direction: OUT)

    adminOf: [Workspace] @relationship(type: "HAS_ADMIN", direction: IN)
}

type Workspace @auth(rules: [
    { 
        operations: [READ],
        roles: ["""member"""],
        where: { 
            OR: [
                { members_INCLUDES: { authId: """$jwt.sub""" } }
                { admins_INCLUDES: { authId: """$jwt.sub""" } }
            ]
        }
    },
    {
        operations: [READ, CREATE, UPDATE, CONNECT, DISCONNECT, DELETE ]
        roles: ["""admin"""]
    },
]) @exclude(operations: [CREATE, UPDATE]) {
    id: ID! @id(autogenerate: false)

    name: String!

    members: [Person!] @relationship(type: "MEMBER_OF", direction: IN)
    admins: [Person!] @relationship(type: "HAS_ADMIN", direction: OUT)

    pages: [Page!] @relationship(type: "HAS_PAGE", direction: OUT)
}

type Page implements AccountOrPageInterface @auth(rules: [
    { 
        operations: [READ], 
        roles: ["""member"""], 
        where: {
            OR: [ 
                { owner: { authId: """$jwt.sub""" } }, 
                { 
                    AND: [ 
                        { shared: true },
                        { workspace: { members_INCLUDES: { authId: """$jwt.sub""" } } },
                    ] 
                } 
            ] 
        } 
    },
    {
        operations: [READ, CREATE, UPDATE, CONNECT, DISCONNECT, DELETE ]
        roles: ["""admin"""] 
    },
]) {
    id: ID! @id(autogenerate: false)

    name: String!
    shared: Boolean!

    owner: Person! @relationship(type: "CREATED_PAGE", direction: IN)

    workspace: Workspace! @relationship(type: "HAS_PAGE", direction: IN)
}
```

To Reproduce
execute a GraphQL query

```
query Pages($pagesWhere: PageWhere) {
    pages(where: $pagesWhere) {
        id
        name
    }
}
```

with a where filter on the workspace id

```
{
  "pagesWhere": {
    "workspace": {
      "id": "workspace_id"
    }
  }
}
```

The library generates the cypher:

```
MATCH (this:Page)
WHERE EXISTS((this)<-[:HAS_PAGE]-(:Workspace)) AND ANY(this_workspace IN [(this)<-[:HAS_PAGE]-(this_workspace:Workspace) | this_workspace] WHERE this_workspace.id = $this_workspace_id) AND ANY(r IN ["member"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) AND (EXISTS((this)<-[:CREATED_PAGE]-(:Person)) AND  ALL(owner IN [(this)<-[:CREATED_PAGE]-(owner:Person) | owner] WHERE owner.authId IS NOT NULL AND owner.authId = $this_auth_where0_OR0_owner_authId) OR (this.shared IS NOT NULL AND this.shared = $this_auth_where0_OR1_AND0_shared AND EXISTS((this)<-[:HAS_PAGE]-(:Workspace)) AND  ALL(workspace IN [(this)<-[:HAS_PAGE]-(workspace:Workspace) | workspace] WHERE EXISTS((workspace)<-[:MEMBER_OF]-(:Person)) AND  ANY(members IN [(workspace)<-[:MEMBER_OF]-(members:Person) | members] WHERE members.authId IS NOT NULL AND members.authId = $this_auth_where0_OR1_AND1_workspace_members_INCLUDES_authId)))) OR ANY(r IN ["admin"] WHERE ANY(rr IN $auth.roles WHERE r = rr))
CALL apoc.util.validate(NOT(ANY(r IN ["member"] WHERE ANY(rr IN $auth.roles WHERE r = rr)) OR ANY(r IN ["admin"] WHERE ANY(rr IN $auth.roles WHERE r = rr))), "@neo4j/graphql/FORBIDDEN", [0])
RETURN this { .id, .name } as this
```

Note the lack of parentheses around conjoined AND and OR statements.

In the case of simple rules, where all clauses are AND'ed there is no problem, but when there are rules being OR'ed together, the lack of parentheses causes the predicates to evaluate incorrectly yielding in incorrect results.

Expected behavior
All predicates should be wrapped in parentheses to ensure that precedence is explicit and the cypher evaluates correctly.

Screenshots

The example here is using the latest @pandaai/graphql-fork package which consists of master combined with #501 , #500, #495, and #490 but the issue exists in master.

See create-auth-and-params.ts#L165, create-auth-and-params.ts#L227, create-auth-and-params.ts#L296, create-auth-and-params.ts#L315, create-auth-and-params.ts#L315

Updating these locations to wrap the joined strings in parentheses maintains precedence and allows the generated cypher to produce semantically correct results.

In order to minimize noise in this change, I have specifically targeted `OR` joins which would be evaluated incorrectly.  Auth clauses generated by create-auth-and-params are wrapped in parentheses when joining, and `OR` joins have the predicates they join wrapped in addition.

# Issue

#505

# Checklist

### DIFF
--- a/packages/graphql/src/translate/create-auth-and-params.ts
+++ b/packages/graphql/src/translate/create-auth-and-params.ts
@@ -22,7 +22,7 @@ import { Neo4jGraphQLAuthenticationError, Node } from "../classes";
 import { AuthOperations, BaseField, AuthRule, BaseAuthRule, Context } from "../types";
 import { AUTH_UNAUTHENTICATED_ERROR } from "../constants";
 import mapToDbProperty from "../utils/map-to-db-property";
-import joinPredicates from "../utils/join-predicates";
+import joinPredicates, { isPredicateJoin, PREDICATE_JOINS } from "../utils/join-predicates";
 
 interface Res {
     strs: string[];
@@ -73,7 +73,7 @@ function createAuthPredicate({
 
     const result = Object.entries(rule[kind] as any).reduce(
         (res: Res, [key, value]) => {
-            if (key === "AND" || key === "OR") {
+            if (isPredicateJoin(key)) {
                 const inner: string[] = [];
 
                 (value as any[]).forEach((v, i) => {
@@ -93,7 +93,7 @@ function createAuthPredicate({
                     res.params = { ...res.params, ...authPredicate[1] };
                 });
 
-                res.strs.push(joinPredicates(inner, key as "AND" | "OR"));
+                res.strs.push(joinPredicates(inner, key));
             }
 
             const authableField = node.authableFields.find((field) => field.fieldName === key);
@@ -225,9 +225,7 @@ function createAuthAndParams({
             { strs: [], params: {} }
         );
 
-        const filteredSubPredicates = subPredicates.strs.filter(Boolean);
-
-        return [joinPredicates(filteredSubPredicates, "OR"), subPredicates.params];
+        return [joinPredicates(subPredicates.strs, "OR"), subPredicates.params];
     }
 
     function createSubPredicate({
@@ -269,7 +267,7 @@ function createAuthAndParams({
             }
         }
 
-        ["AND", "OR"].forEach((key) => {
+        PREDICATE_JOINS.forEach((key) => {
             const value = authRule[key] as AuthRule["AND"] | AuthRule["OR"];
 
             if (!value) {
@@ -294,7 +292,7 @@ function createAuthAndParams({
                 predicateParams = { ...predicateParams, ...par };
             });
 
-            thisPredicates.push(joinPredicates(predicates, key as "AND" | "OR"));
+            thisPredicates.push(joinPredicates(predicates, key));
             thisParams = { ...thisParams, ...predicateParams };
         });
 
@@ -328,9 +326,7 @@ function createAuthAndParams({
         { strs: [], params: {} }
     );
 
-    const filteredSubPredicates = subPredicates.strs.filter(Boolean);
-
-    return [joinPredicates(filteredSubPredicates, "OR"), subPredicates.params];
+    return [joinPredicates(subPredicates.strs, "OR"), subPredicates.params];
 }
 
 export default createAuthAndParams;
--- a/packages/graphql/src/translate/create-auth-and-params.ts
+++ b/packages/graphql/src/translate/create-auth-and-params.ts
@@ -22,6 +22,7 @@ import { Neo4jGraphQLAuthenticationError, Node } from "../classes";
 import { AuthOperations, BaseField, AuthRule, BaseAuthRule, Context } from "../types";
 import { AUTH_UNAUTHENTICATED_ERROR } from "../constants";
 import mapToDbProperty from "../utils/map-to-db-property";
+import joinPredicates from "../utils/join-predicates";
 
 interface Res {
     strs: string[];
@@ -92,7 +93,7 @@ function createAuthPredicate({
                     res.params = { ...res.params, ...authPredicate[1] };
                 });
 
-                res.strs.push(`(${inner.join(` ${key} `)})`);
+                res.strs.push(joinPredicates(inner, key as "AND" | "OR"));
             }
 
             const authableField = node.authableFields.find((field) => field.fieldName === key);
@@ -162,7 +163,7 @@ function createAuthPredicate({
         { params: {}, strs: [] }
     );
 
-    return [result.strs.join(" AND "), result.params];
+    return [joinPredicates(result.strs, "AND"), result.params];
 }
 
 function createAuthAndParams({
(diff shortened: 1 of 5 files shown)


packages/graphql/src/translate/create-auth-and-params.ts
```diff
@@ -92,7 +93,7 @@ function createAuthPredicate({
                     res.params = { ...res.params, ...authPredicate[1] };
                 });
 
-                res.strs.push(`(${inner.join(` ${key} `)})`);
+                res.strs.push(joinPredicates(inner, key as "AND" | "OR"));
```
angrykoala, line 96: To avoid casting the variable key, I think it is worth creating a new helper function :thinking: :
```ts
function isPredicateJoin(value: string): value is "AND" | "OR" {
    return ["AND", "OR"].includes(value)
}
```

And use this function instead of `if (key === "AND" || key === "OR") {`. This way key will be properly typed. Thoughts?

packages/graphql/src/translate/create-auth-and-params.ts
```diff
@@ -293,7 +294,7 @@ function createAuthAndParams({
                 predicateParams = { ...predicateParams, ...par };
             });
 
-            thisPredicates.push(predicates.join(` ${key} `));
+            thisPredicates.push(joinPredicates(predicates, key as "AND" | "OR"));
```
angrykoala, line 297: Like before, instead of just casting, I think it is worth it changing how we get the key to ensure it is properly typed.
In this case, instead of just looping through a string array
```
l.272  ["AND", "OR"].forEach((key) => {
```

We could explicitly set the types:
```ts
var keys:Array<"AND"|"OR"> = ["AND", "OR"]
keys.forEach((key)=>{
```

packages/graphql/src/utils/join-predicates.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (predicates.length === 0) {
+        return "";
+    }
+
+    if (predicates.length === 1) {
+        return predicates[0];
+    }
+
+    return `(${
+        joinType === "AND" ? predicates.join(" AND ") : predicates.map((predicate) => `(${predicate})`).join(" OR ")
```
angrykoala, line 30: I'm not sure I understand the reasoning behind treating AND and OR differently. We could wrap all predicated with parenthesis to improve consistency, making this function much simpler:

```ts
export default function joinPredicates(predicates: string[], joinType: "AND" | "OR"): string {
    if (predicates.length === 0) {
        return "";
    }

    return `(${
        predicates.map(predicate=>`(${predicate})`.join(` ${joinType} `);
     })`;
  }
```

This may lead to unnecessary parenthesis, but should not be a problem. Thoughts on this?.
This is not too important of a change so don't worry about it if it requires too many changes in the tests

packages/graphql/src/utils/join-predicates.test.ts
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import joinPredicates from "./join-predicates";
+import trimmer from "./trimmer";
+
+describe("join predicates", () => {
```
angrykoala, line 23: :clap: Great idea on creating this function and tests

angrykoala (changes requested):
Most of the changes look great :+1: , there are some minor improvements in typings and consistency that we could address before merging this.

rcbevans:
Great suggestions, I actually wanted to implement the typing you suggested but wasn't sure how to express it in typescript.  I'll update the code today

packages/graphql/src/translate/create-auth-and-params.ts
rcbevans: Yes! Much better!

packages/graphql/src/translate/create-auth-and-params.ts
rcbevans: Makes sense to me me

packages/graphql/src/utils/join-predicates.ts
rcbevans: This is how I originally planned to implement the change, I agree that it's better to be explicit with parentheses both for OR and for AND.

The code generates a lot of ANDs and not too many ORs, and the unit tests therefore largely contain tests which result in ANDs.

When I implemented the initial design to wrap all predicates in additional parentheses, several dozen unit tests were failing which would have taken a lot of time to go through and fix, and created an incredibly noisy PR.  In the majority of cases, where only ANDs were being generated, there was no issue as the order of evaluation wouldn't matter; it's only when operators of different precedence become mixed it's a problem.

As a consequence, to minimize noise (and hopefully get this fix accepted ASAP, I decided to target the issue around ORs specifically.

I think it would be a good idea to revisit later, but it's not a change I want to make right now.

rcbevans:
Thanks for the review @angrykoala.  I've implemented the changes you suggested as well as added a pretty exhaustive integration test for the issue described, to verify the issue is resolved.

angrykoala (approved):

PR #507, opened 2021-09-22
Remove Node casting in tests

angrykoala:
# Description
Using NodeBuilder in tests we can avoid typecasting and remove several mock properties


darrellwarde (approved):
LGTM!

PR #477, opened 2021-09-08
Fix: Project Interface Fields

dmoree:
# Description

Proposed fix for #476. Current behavior is to project fields by node. This add the fields of the implemented interfaces to projection fields.

# Issue

- #476 

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [ ] Documentation has been updated
- [x] TCK tests have been updated
- [x] Integration tests have been updated
- [ ] Example applications have been updated
- [x] New files have copyright header
- [x] CLA (https://neo4j.com/developer/cla/) has been signed

Issue #233, opened 2021-06-03
Allow relationships on interfaces - neo4j-graphql-js supported this


cramatt:
Using the latest GRANDStack starter which recently changed from using the old neo4j-graphql/neo4j-graphql-js library to this one, I have some regressions related to how we define relationships with interfaces. (I didn't open it as a regression since technically its a different library with different opinions :) ). 

Specifically, in neo4j-graphql/neo4j-graphql-js there were two PRs which addressed adding this functionality, referenced at the end of this issue: https://github.com/neo4j-graphql/neo4j-graphql-js/issues/201

In this neo4j/graphql, there was a decision to not support this functionality, as documented in this PR https://github.com/neo4j/graphql/pull/17 and this test: https://github.com/neo4j/graphql/blob/119a81ca4127f7d3036b408f124f2a2228f9d40a/packages/graphql/src/schema/make-augmented-schema.test.ts#L92-L105  

**Describe the solution you'd like**
Relationships on interfaces would be useful and are a common use case. Consider this: 

```graphql
interface Component {
  name: String!
}

type Battery implements Component {
  name: String!
  maxCurrent: Int!
  maxTemp: Int!
}

type Propeller implements Component {
  name: String!
  diameter: Int!
  numBlades: Int!
}

type System {
  name: String!
  description: String!
  components: [Component] @relationship(type: "USES_COMPONENT", direction: OUT)
}
```

The above use case seems valid. Additionally, although mentioned in #17 that Interfaces give "...users no real database support therefor no; query, update, delete, filter support." I believe this is incorrect. In neo4j-graphql/neo4j-graphql-js an interface would apply multiple labels to nodes, so a battery created with the above schema would have the labels `Component, Battery` which makes it very easy to write a cypher query for all `Components` in the graph. Also this was supported by the graphql query system.

**Describe alternatives you've considered**
None

**Additional context**
None

dmoree:
It was stated that this would not be looked into until after the implementation of relationship properties by @danstarns here https://github.com/neo4j/graphql/issues/146#issuecomment-831809546

That was a little over a month ago and perhaps there is some update to this.

danstarns:
@dmoree Still no update on this. We are working on 2.0 where interfaces are going to be used as relationship properties. Due to our opinionated choice of interfaces, it's fair to assume that we won't support relationships on interfaces any time soon.

dmoree:
[...]

PR #509, opened 2021-09-23
Fix: Project Interface Fields In Union Relationship

dmoree:
# Description

Similar to #477 that failed to address an issue raised in https://github.com/neo4j/graphql/issues/233#issuecomment-925371199. This PR projects fields of interfaces implemented by union types in a relationship.

# Checklist


rcbevans:
This is a great change! Most of my use of unions they also implement a common interface which is often what I need, this will make querying much more concise!

rcbevans:
In the example neo-push project/integration tests I think this could be tested using an 'authored' interface on comments and posts.  On a user you could query an 'authoredBy' field which returns a union of Post | Comment.

PR #515, opened 2021-09-27
Feature: Subscriptions

### LINKED ISSUE #191: Subscriptions
Looking for the ability to subscribe to events via the Neo4j GraphQL API. 

**Additional context**
Came from here https://github.com/neo4j/graphql/issues/179

KleptoKat:
This PR publishes events for every mutation created by Neo4j Graphql and allows GraphQL and OGM to subscribe to those events in an auth friendly way.

It allows the client to subscribe to updates for any node type with subscribeTo`Node`Updates. You can pass the subscription a `where` variable to filter nodes. Additionally, authentication can be taken into account as the system uses the existing `translateRead` function to execute the filter after the subscription has been alerted that a node has updated. This system takes into account recursive updates (ie, updateNode.update.relationship.update.node).

Work still required by Neo4j Team:
- [ ] Add authorization info to type defs for subscriptions. Right now anyone can subscribe to nodes (so long as they have query permissions on the resolved node)
- [ ] Add subscription exclusions
- [ ] Additional Documentation
- [ ] Create example application
- [ ] Additional tests

I'll be handing off this application for 2022 so the team can incorporate it into the next major release (:

See [Getting Started Documentation](https://neo4j-graphql-docs-515.surge.sh/graphql-manual/2.0/getting-started/#_advanced_setup_websockets_subscriptions_with_apollo) for setup instructions. There is also documentation on how to use an external pubsub engine, create subscriptions from graphql, and how to use the OGM `observe` function.

Usage:
 - Follow the setup procedure
 - In the playground, subscribe to a node type with `subscribeToTileUpdates`
 - In the playground (in a seperate tab or window) update that node type directly or in a nested update
 - You should see that update reflected in the subscribed playground window.

## WithProjector

To improve readability, I created a class called "WithProjector" that replaces withVars in each translation. It's essentially a stack of variables that allows you to pass the mutationMeta variable along with any other variables which you may be passing to a translation. It required a significant change in the code base to implement properly because it requires that you call each `nextWith` and `nextReturn` in order.

The following logic is used to declare to the WithProjector that there was a mutation which occured and, in the next WITH block, that we should add an event to the mutationMeta block.

```
withProjector.markMutationMeta({
    type: 'Deleted',
    idVar: `id(${ _varName })`,
    name: refNode.name,
});
```

>> Results in >>

```
MATCH (this:Movie)
WHERE this.name = $this_name
WITH this, [ metaVal IN [{type: 'Deleted', name: 'Movie', id: id(this)}] WHERE metaVal IS NOT NULL AND metaVal.id IS NOT NULL AND (metaVal.toID IS NOT NULL OR metaVal.toName IS NULL) ] as mutateMeta
DETACH DELETE this
RETURN mutateMeta
```

Note that overly complex mutations (such as a an update query which updates and creates multiple nodes ()-[]-> ()-[]-> ()) may see lower performance because of this.

## Use of internal node ID's for identification

Internal node ID's are used to resolve nodes after mutations have occurred. IE, if a user updates a node then the event will include that nodes internal ID. The subscription will resolve the node using the internal ID provided by that event.

## Possible future considerations:

- State change information. IE, if a user updates the `released` property on a movie, perhaps we can include the state that the movie was in previously?
- Add a toggle to enable/disable query subscriptions
- Add a toggle to enable/disable mutateMetas to improve all query performance << an "off-by-default" solution might work well here?
- Add a filter for "handles". A handle could be a unique ID which is implemented by the user, eg a UUID. One would also need to implement "handle" in the WithProjector/mutate meta.

# Issue

Closes #191 

# Checklist

This PR is still a proof of concept and is up for discussion.


KleptoKat:
@darrellwarde Ah okay. Should I ignore/exclude TCK tests for now then?

darrellwarde:
> @darrellwarde Ah okay. Should I ignore/exclude TCK tests for now then?

It quite possibly might be worthwhile - assuming that it's broken existing tests that you need to work on, this should be an automated update when we're done with this work.

packages/graphql/src/translate/translate-update.ts
```diff
@@ -108,52 +106,75 @@ function translateUpdate({ node, context }: { node: Node; context: Context }): [
             }
 
             if (relationField.interface) {
```
KleptoKat, line 108: These functions were re-ordered and placed in the `cypher` array because the order by which withProjector.nextWith is called matters. This also enhances readability.

packages/graphql/src/translate/create-update-and-params.ts
```diff
@@ -51,25 +51,35 @@ function createUpdateAndParams({
...
 
+    /**
+     * **NOTE**
+     * Update input is now being sent to neo4j twice. Instead of flattening each input parameter,
+     * perhaps we can utilize the object being sent to neo4j?
+     * `objUpdateParams` is necessary for subscriptions to sent an object containing
+     * the updated properties.
+     */
+    const objUpdateParams = {}; 
+    const objUpdateParamsKey = chainStr || `${parentVar}_update`;
```
KleptoKat, line 78: please advise:
```
     * Update input is now being sent to neo4j twice. Instead of flattening each input parameter,
     * perhaps we can utilize the object being sent to neo4j?
     * `objUpdateParams` is necessary for subscriptions to sent an object containing
     * the updated properties.
```

darrellwarde: Sorry @KleptoKat, missed this one. We can definitely pass to Neo4j as an object - feel free to clutter the Cypher parameter somewhat because this is an area we need to clean anyway.

darrellwarde:
@angrykoala and I spent a good little while getting this up and running this morning! We struggled a bit with the example code given in the description, but had great success with the following:

```ts
import neo4j from "neo4j-driver";
import { createServer } from "http";
import express from "express";
import { execute, subscribe } from "graphql";
import { ApolloServer, gql } from "apollo-server-express";
import { SubscriptionServer } from "subscriptions-transport-ws";
import { Neo4jGraphQL } from "@neo4j/graphql";

(async () => {
    const PORT = 4000;
    const app = express();
    const httpServer = createServer(app);

    const typeDefs = gql`
        type Movie {
            title: String!
        }

        type Actor {
            name: String!
        }
    `;

    const driver = neo4j.driver("bolt://127.0.0.1:7687", neo4j.auth.basic("admin", "password"));

        driver,
        typeDefs,
    });

    const schema = neoSchema.schema;

        schema,
    });
    await server.start();
    server.applyMiddleware({ app });

    SubscriptionServer.create(
        {
            schema,
            onConnect: (connectionParams, webSocket, context) => {
                console.log(`WebSocket connected.`);

                neoSchema.populateContext(context);

                return context;
            },
            onDisconnect: (webSocket, context) => {
                console.log(`WebSocket disconnected.`);
            },
            execute,
            subscribe,
        },
        { server: httpServer, path: server.graphqlPath }
    );

    httpServer.listen(PORT, () => {
        console.log(`🚀 Query endpoint ready at http://localhost:${PORT}${server.graphqlPath}`);
        console.log(`🚀 Subscription endpoint ready at ws://localhost:${PORT}${server.graphqlPath}`);
    });
})();
```

For anybody looking for a complete solution to get this up and running and have a play. 🙂

KleptoKat:
Hello @darrellwarde and @oskarhane ! I hope you have both been having a great Holiday (:

I have updated subscriptions to a point where I am happy with it. I have also updated my initial post and added some documentation related to subscriptions. I believe it is in a state where I can now hand it off to the Neo4j Team to review and implement any changes that they think is necessary to get it to the Neo4j standard. As it stands I won't have much time this year (2022) but if there are any changes required or if you have any questions about the code I am available to answer questions or do a code walkthrough! Please just reach out here or on Discord for that (: I've done my best to document the code in-line so your team has a better understanding of why I implemented these things the way I did.

The biggest changes to the code-base here is with the WithProjector. It is responsible for passing variables and mutateMeta information along with every WITH and RETURN clause. It likely needs the most scrutiny and consideration.

Following that, I have the subscribeToNode subscription queries and the resolvers, along with their respective "filters." Filters are a way to improve the performance of subscriptions by filtering out events _before_ they are resolved.

Next, OGM has a new "observe" method that allows the server to subscribe to events & resolve nodes using similar logic to the graphql subscriptions above. Additional internal ID logic needed to be implemented to solve mapping events to nodes.

The documentation should be sufficient enough for you to get started. I hope the integration tests are sufficient enough.

Have a safe and happy new year! Please let me know if you have any questions or would like to do an in-call code review!

Best,
KleptoKat

corysimmons:
@darrellwarde Please don't let this feature go stale. Subscriptions have been the only reason me and my friends don't switch to Neo4j.

inc16sec:
> 

I'm also watching this closely for months now. Subscriptions are crucial to us and they are what keep us from releasing our project till now.

helenesanche:
@corysimmons @inc16sec Hello, I'm Hélène the new Product Manager for Neo4j GraphQL. Thank you both so much for your contributions and interest in this feature. Please rest assured that we have kicked off the work on Subscriptions and it will be one of our main focuses for the next major release.

corysimmons:
Awesome, thank you so much for the update!

angrykoala:
Hi @inc16sec @corysimmons 
We have been closely reviewing this PR for a while now. It is a complex piece of work that, sadly, does not meet all our requirements for some scenarios and has grown a bit outdated after 3.0 release.
However, inspired by this solution (kudos to @KleptoKat ), we are currently working to bring a subscriptions solution (finally!).

You can see our current progress in https://github.com/neo4j/graphql/pull/1076 and https://github.com/neo4j/graphql/pull/1090

izakfilmalter:
In the same boat as @corysimmons. Excited to see this happen.

jdgamble555:
Note to the people writing this code and @angrykoala:

`subscriptions-transport-ws` is depreciated!  Please use [graphql-ws](https://github.com/enisdenjo/graphql-ws) whenever you implement this!

@darrellwarde - You might also want to update your code.

J

angrykoala:
Hi @jdgamble555 
Thanks for the advice. The library itself will not be using any WebSocket library (that's up to the server implementation) but our tests will be done against a server using `graphql-ws`

darrellwarde:
All, today we released 3.1.0 which contains the beginning of our journey into GraphQL Subscriptions. Despite the current implementation not being as comprehensive as what was originally built by @KleptoKat, implementing Subscriptions was just one of those pieces of work that we had to implement ourselves, to fully understand the problem and to learn a lot along the way. It is a feature that we are going to continue working on, building towards a stable, production ready release in the coming months. I think it is finally time to close this PR - please raise any issues with the implementation released today as new GitHub Issues. Thank you again for your work here @KleptoKat, we are very much stood on your shoulders!

jdgamble555:
Nice! Does this use polling or pubsub?

jdgamble555:
[Reference for Searching](https://github.com/neo4j-graphql/neo4j-graphql/issues/183)

Thanks @darrellwarde! 

**How do you just subscribe to ALL changes?** 

Every other GraphQL subscriptions can subscribe to ANY changes, not just the specific mutation.  Can you subscribe to all?

Thanks!

J

corysimmons:
@angrykoala @darrellwarde Sorry I wasn't timely with feedback, but I didn't have a chance to properly review this, but it did look odd that it wasn't generically monitoring a dataset, but rather only monitoring `CREATE` `UPDATE` and `DELETE` events. 🤔 

Now I'm trying to lean into Neo4j hard for a startup I'm working on and ran into needing Subscriptions (real time drawing and commenting and stuff).

@jdgamble555's comment immediately above is correct. There doesn't seem to be a way to just open a subscription to `READ` some dataset...

That is, I want to be able to just write something like this...

```gql
subscription Users {
  users {
    name
    email
  }
}
```

...and get data streamed in as it is created/deleted/updated.

Hasura does it like this. I think Dgraph does it like this. Pretty much every db platform that actually supports GraphQL subscriptions does it like this. They don't even offer stuff like `movieCreated`,  `movieUpdated`, and `movieDeleted`.

Is this an easy change to make?

angrykoala:
Hi @corysimmons 
Having a subscription for any change is technically doable. But there are a few consequences that made us decide towards having them as separate subscriptions. Mainly the fact that the data returned is slightly different in each case.

With that said, the presumption is also that something like what you need could be done with 3 separate subscriptions. One for delete, create, and update. Would there a problem or limitation with this approach?

jdgamble555:
So I don't think subscriptions are supposed to be separated that way.  You are supposed to subscribe to ALL changes.  

If I subscribe to a `Posts` query, any changes within that query should be returned.  The front end developer should not have to think about this.

That is why most database platforms actually use polling under the hood.  Dgraph, Supabase, and Hasura all use polling instead.  Firestore and Firebase don't need to use polling since it is a noSQL database.  You subscribe to the document itself.  

In a perfect world, you have something like [Quine](https://quine.io/) that is built for streaming. However, it could get complicated to write **Change Data Capture** in a graph database due to the complexities of graph queries.

I suggest you build a combination of the two.  If the create, delete, or update gets triggered, just refetch the query at that point.  Basically a combination of polling and changes.  If not, direct polling should be an option.

Just my thoughts.

J

corysimmons:
>Mainly the fact that the data returned is slightly different in each case.

Could you explain this a bit more with an example?

Couldn't we just let Apollo Client (or SWR, or any other modern fetcher) handle caching? For instance, I think a huge chunk of these two [very similar] queries would be cached:

```gql
subscription Users {
  users {
    name
    email
  }
}
```

```gql
subscription Authors {
  users(where: { type: "author" }) {
    name
    email
    booksWritten
  }
}
```

angrykoala:
@corysimmons Sorry, I think I didn't explain well enough. My point was that a [create event](https://neo4j.com/docs/graphql-manual/current/subscriptions/events/create/) has a different payload (movie created) than an [update event](https://neo4j.com/docs/graphql-manual/current/subscriptions/events/update/) that contains both the previous state and the next, having the same query for both would mean that we cannot make exact typings, as we will need to rely on nullable values.


@jdgamble555 There is not really any requirement to how subscriptions should work afaik. Separating them into the different possible events also avoid needing to check the event type every time is triggered, like the following example (taken from Firebase docs) which is quite common, as front-end behaviour is rarely the same for each kind of event:
```javascript
    if (change.type === "added") {
        console.log("New city: ", change.doc.data());
    }
    if (change.type === "modified") {
        console.log("Modified city: ", change.doc.data());
    }
    if (change.type === "removed") {
        console.log("Removed city: ", change.doc.data());
    }
```

Regarding Polling, which is a different topic, it was unfeasible, as it will lead to 1+ query per subscription, as well as a delay on the event triggering that would impact the real-time capabilities of the subscriptions.

I hope this explanation throws some light on the design decisions and tradeoffs of the current solution

jdgamble555:
So Firebase still gets triggered no matter what on the front end the same way without having to think about it, you can just separate the kinds of trigger manually if you need to.  Firebase Functions work the same way.

Polling definitely will be more pressure on the database, but that is how neo4j's competitors do it under the hood for the moment.  I don't think the delay is noticeable on any of them, although we are talking about GO, Euphoria, and Haskell languages instead of Node.js on their backends.

Perhaps there could be something like this:

```graphql
subscription {
  postsALL {
    ...
  }
}
```

Or something that would get triggered in all cases.  I think there is a bigger use case to get all changes, than to separate them.  That way people can have their choice.

My thoughts...

J

corysimmons:
I appreciate the explanation. Would the team mind keeping subscriptions in beta for a little longer? I would love to try these patterns in a real project with comments and drawing and give some feedback but it'll probably be like a month before I get there.

angrykoala:
Hi @jdgamble555 I still don't understand your case for polling. I'm not sure what limitations are you suffering that polling would solve.
Polling **does** put a big pressure on the database, which could be alleviated with caching and batching, but its effectiveness is completely dependant on the user queries, and may not be valid for everyone.
Polling also have a noticeable delay. You may think it is not noticeable. But most polling systems work by polling ~1 second. To which you need to add the extra round-trip to the database. Compared to a non-polling system, that only suffers the delay of server to client network. Keep in mind that polling delay is mostly Database and Networking. So usually it is negligible on what language you are building the server.

Finally, keep in mind that the discussion on whether to batch all subscriptions in a single subscribe or as separate subscriptions is an API design decision, completely irrelevant to the underlying mechanism used for getting those events, if there is a use case that cannot be solved with this API, please let us know, otherwise there isn't really a strong case for changing this.

jdgamble555:
What I was saying is that MOST of your competitors, if not all, use polling.  None of them separate the types of subscriptions out by the type of mutation.  I personally find this awkward and not very useful.

If I have a todo list, for example, I want it to be automatically updated no matter what type of change is incurred to the data (an item is added, deleted, modified).  It is awkward and not efficient to make 3 different subscriptions.  This is by far the most common use case.  Polling simply makes this easy.  Change Data Capture would be the preferred method by far over polling, but the code necessary to write this may be extremally complicated.  I don't know.

All I know is there is a reason other databases chose to use polling.  I can tell you the nano-second delays are not a problem in Dgraph, Supabase, Hasura, etc.

> Finally, keep in mind that the discussion on whether to batch all subscriptions in a single subscribe or as separate subscriptions > is an API design decision, completely irrelevant to the underlying mechanism used for getting those events, 

Respectfully, this is not true.  While having the core code being written for this makes it easier, it does not make it impossible on the GraphQL level.  Again, see your competitors which use polling on the GraphQL layer.  Perhaps that is why they use polling.

I am not advocating for polling per se, but I am arguing against it not being worth the extra database stress.  If anything, neo4j hosts could charge for this stress.

If you want to just look from a problem solving basis, the simplest example is subscribing to a TODO list which gets updated by all changes.  How can we achieve this basic example?

J

corysimmons:
I do like the idea of polling because it's easy but can't we already do that with just Apollo? https://www.apollographql.com/docs/react/data/queries/#polling

It also _does_ seem expensive to be polling all the time. Imagine 1000 people are looking at your website and it is making a db query every 500ms for each of them... I'm not sure how Hasura & Co. get away with that, but it feels bad.

I'm wondering if this is more a _"we need a good pattern and some examples for this"_ kind of thing...

@angrykoala Could you provide a simple example of how you'd watch a todos list using your pub/sub method? Curious how it looks vs the polling implementation which is just...

```gql
subscription Foo {
  todos {
    id
    text
  }
}
```

izakfilmalter:
I think the thing that is frustraighting about this as a Firebase user is that the burden is now on the FE engineer to merge all this together and do the right thing. This causes a lot of future bugs where someone doesn't do the local merge properly. I think everyone just wants to subscribe to a collection, and then receive changes to that array based on create, updated, and deleted that are merged together. Firebase in many ways pioneered the api that FE devs want to work with. If neo4j isn't gonna get close to that, I'll have to look at other graph options that are not as nice in other areas.

PR #519, opened 2021-09-29
Feat/where aggregations

danstarns:
# Where Aggregations

The second piece of aggregations after [Top Level Aggregations](https://github.com/neo4j/graphql/pull/470).

You can find the live documentation for this work [here](https://neo4j-graphql-docs-519.surge.sh/graphql-manual/2.0/filtering/#_aggregation_filtering).

## Technical Details

The work in this PR introduces more usage of `apoc.cypher.runFirstColumn` & this makes generating the Cypher easier plus allows for greater flexibility when adding more aggregation operators.

Given the GraphQL Query with a where aggregation:

```graphql
query {
    posts(where: { likesAggregate: { count_GT: 5 } }) {
        content
    }
}
```

The generated Cypher would look like:

```
MATCH (this:Post)
WHERE apoc.cypher.runFirstColumn("
    MATCH (this)<-[:LIKES]-(user:User)
    RETURN count(user) = 5 ",
    { this: this },
    false
)
RETURN this { .content } as this
```

The return value of evoking `runFirstColumn` is a boolean value that can be added to the normal `where` predicate such as:

```
MATCH (this:Post)
WHERE apoc.cypher.runFirstColumn("RETURN <condition>", { this: this }, false) AND <condition> OR <condition>
RETURN this { .content } as this
```

One may be asking why not do something like this:

```
MATCH (this:Post)<-[:LIKES]-(user:User)
WHERE count(user) = 5
RETURN this { .content } as this
```

Firstly, you would be fine in doing that if you were to write the cypher by hand, however because this is auto-generated and because of the way the codebase is designed its much easier and contained to do it via the `runFirstColumn` approach. This is due to the fact we do not need to change the way we do matching and simply adds code vs alerting or leaking into the existing.

Secondly, when it comes to more complex aggregations such as:

"Find Posts where the shortest like name is less than 10 characters":

```graphql
{
    posts(where: { likesAggregate: { node: { name_SHORTEST_LT: 10 } } }) {
        content
    }
}
```

The generated cypher has to include a `WITH` statement to collect the size of the `user.name`:

```
MATCH (this:Post)
WHERE apoc.cypher.runFirstColumn("
    MATCH (this)<-[:LIKES]-(user:User)
    WITH user, size(user.name) AS nameSize
    RETURN min(nameSize) < 10 ",
    { this: this },
    false
)
RETURN this { .content } as this
```

As nesting the `size` aggregation inside the `min` aggregation gives unexpected results:

⚠ Unexpected results, need to collect the `user.name` size first using `WITH`.

```
RETURN min(size(user.name)) < 10
```

Adding a `WITH` after the `MATCH`, containing all the aggregations, would add additional complexity plus introduce unexpected behaviour.

Finally, using `runFirstColumn` opens up the ability to do nested aggregation traversals, something not supported in this pr, with ease as we would have the full power of Cypher inside the `WHERE` clause.

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-519.surge.sh

packages/graphql/src/translate/create-aggregate-where-and-params.ts
```diff
@@ -0,0 +1,339 @@
...
+    inputValue: any;
+    nodeOrRelationship: Node | Relationship;
+    chainStr: string;
+    variable: string;
+}): { aggregations: string[]; params: any; withStrs: string[] } {
+    const aggregations: string[] = [];
+    let withStrs: string[] = [];
+    let params = {};
+
+    Object.entries(inputValue).forEach((e) => {
```
angrykoala, line 68: Using destructuring in e may give a slightly more readable variables and could be useful to provide some context on what e[0] and e[1] is
```suggestion
    Object.entries(inputValue).forEach(([e1, e2]) => {
```

docs/modules/ROOT/pages/filtering.adoc
```diff
@@ -89,3 +89,239 @@ query {
...
+This library offers, for each relationship, a aggregation key inside the where argument. You can use the aggregation key to satisfy questions such as: 
+
+* Find the posts where the number of likes are greater than 5
+* Find flights where the average age of passengers is greater than or equal to 18
+* Find movies where the shortest actor screen time is less than 10 minutes
+
+You can use this where aggregation on both the `node` and `edge` of a relationship. 
+
+
+=== Aggregation Filtering Usage Examples
```
oskarhane, line 104: I like these examples to show the usage 👍

packages/graphql/src/translate/create-where-and-params.ts
```diff
@@ -63,6 +64,37 @@ function createWhereAndParams({
             (x) => key.startsWith(x.fieldName) && x.typeMeta.name === "Duration"
         );
 
+        if (key.endsWith("Aggregate")) {
+            const [, reversedFieldName] = key
+                .split("")
+                .reverse()
+                .join("")
+                .split("Aggregate".split("").reverse().join(""));
+            const fieldName = reversedFieldName.split("").reverse().join("");
```
oskarhane, line 73: Am I missing something here, or is this a very complicated way to drop the `Aggregate` at the end?

What about `key.replace("Aggregate", "")` or even `key.slice(0, -1*"Aggregate".length)` if we really want to make sure it's only the last bit that gets stripped?

packages/graphql/tests/integration/aggregations/where/edge/bigint.int.test.ts
```diff
@@ -0,0 +1,378 @@
...
+            charset: "alphabetic",
+            readable: true,
+        });
+
+        const neoSchema = new Neo4jGraphQL({ typeDefs });
+
+        try {
+            await session.run(
+                `
+                    CREATE (:Post {testString: "${testString}"})<-[:LIKES {someBigInt: ${bigInt}}]-(:User {testString: "${testString}"})
```
oskarhane, line 67: `someBigInt` won't be an integer in the db here I believe.
You should have it as a cypher parameter and wrap it in `neo4j.int(bigInt)`.

oskarhane: Actually, you could probably use the Cypher function `toInteger()` on a number to convert it without using a param.

packages/graphql/tests/integration/issues/505.int.test.ts
```diff
@@ -24,7 +24,7 @@ import { generate } from "randomstring";
 import neo4j from "../neo4j";
 import { Neo4jGraphQL } from "../../../src/classes";
 
-describe("https://github.com/neo4j/graphql/issues/550", () => {
+describe.skip("https://github.com/neo4j/graphql/issues/550", () => {
```
oskarhane, line 27: Intentional? IF so, why?

oskarhane (commented):
Damn, those TCK tests are really good for documentation purposes but those must have taken a whole day to write.

Anyway, I love this feature and generally it looks really good! 
Thanks for explaining the use of `runForstColumn`, I think it makes sense for now and we don't need to rush our way from APOC.

I had a few more detailed comments as well.

packages/graphql/src/translate/create-aggregate-where-and-params.ts
oskarhane: Yes, a better description on what `e[0]` etc is would help when reading the code in the future.

packages/graphql/tests/integration/aggregations/where/edge/bigint.int.test.ts
danstarns: Good observation

packages/graphql/src/translate/create-where-and-params.ts
danstarns: Yes, I don't know why I made this soo complicated lol :O.

oskarhane (approved):
LGTM 🍰

PR #521, opened 2021-10-01
Field level aggregations

angrykoala:
# Description

This PR adds support for field aggregations, auth and where **not** supported in field aggregations


.eslintrc.js
```diff
@@ -64,6 +64,10 @@ const baseTypeScriptRules = {
     "@typescript-eslint/quotes": "off",
     "@typescript-eslint/lines-between-class-members": ["error", "always", { exceptAfterSingleLine: true }],
     "@typescript-eslint/no-unnecessary-type-assertion": ["off"],
+    "import/prefer-default-export": "off",
```
angrykoala, line 67: I think we are using eslint-disable a lot for this rule, so I'm thinking on disabling it. Thoughts?

.eslintrc.js
```diff
@@ -64,6 +64,10 @@ const baseTypeScriptRules = {
...
+    // disallow certain syntax forms
+    // http://eslint.org/docs/rules/no-restricted-syntax
+    "no-restricted-syntax": ["error", "ForInStatement", "LabeledStatement", "WithStatement"],
```
angrykoala, line 70: this is to avoid airbnb eslint disallowing `for ... of` clauses.

danstarns:
I noticed that the `@alias` directive is not working with this. Here is my typeDefs: 

```gql
type Movie {
    title: String
    actors: [Person] @relationship(type: "ACTED_IN", direction: IN)
}

type Person {
    myName: String! @alias(property: "name")
}
```

And my query: 

```gql
{
  movies {
    actorsAggregate {
      node {
        myName {
          shortest
        }
      }
    }
  }
}
```

And the cypher produced: 

```
MATCH (this:Movie)
RETURN this { actorsAggregate: { node: { myName: head(apoc.cypher.runFirstColumn(" MATCH (this)<-[r:ACTED_IN]-(n:Person)

        WITH n as n
        ORDER BY size(n.myName) DESC
        WITH collect(n.myName) as list
        RETURN {longest: head(list), shortest: last(list)} ", { this: this })) } } } as this
```

The expected cypher should be: 

```diff
MATCH (this:Movie)
RETURN this { actorsAggregate: { node: { myName: head(apoc.cypher.runFirstColumn(" MATCH (this)<-[r:ACTED_IN]-(n:Person)

        WITH n as n
-        ORDER BY size(n.myName) DESC
+       ORDER BY size(n.name) DESC
-       WITH collect(n.myName) as list
+      WITH collect(n.name) as list
        RETURN {longest: head(list), shortest: last(list)} ", { this: this })) } } } as this
```

PR #527, opened 2021-10-05
Fix transformation of arguments to @cypher directives

### LINKED ISSUE #526: Int Argument on Custom Query Converted to Float
**Describe the bug**
I'm unable to use Int's as arguments when using custom cypher queries with the @cypher directive.

Simply including it and passing the paramater to the LIMIT clause of a cypher query will trigger the following error:

```"Failed to invoke function `apoc.cypher.runFirstColumn`: Caused by: org.neo4j.exceptions.InvalidArgumentException: LIMIT: Invalid input. '25.0' is not a valid value. Must be a non-negative integer."```

image

**Type definitions**
```
type Movie {
  title: String
  tags: [Tag] @relationship(type: "HAS", direction: OUT)
}

type Tag {
  id: String!
  name: String!
  papers: [Movie] @relationship(type: "HAS", direction: IN)
}

type Query {
  movie_tags(tagName: String = "", limit: Int): [Movie]
    @cypher(
      statement: """
      MATCH (tag:Tag{type: 'task'})<-[:HAS]-(movie:Movie)
      WHERE tag.name = $tagName
      RETURN movie
      LIMIT $limit
      """
    )
}
```

**To Reproduce**

1. Run a server and simply feed those typeDefs to it
2. Execute the query:
```
{
  movie_tags(tagName: "test", limit: 25) {
    title
    tags {
      name
    }
  }
}
```
4. Error should arise

**Expected behavior**
The Int type is able to be used as an argument for custom queries that use the @cypher directive

-   OS: Ubuntu 18.04 LTS
-   Version: @neo4j/graphql 2.2.0
-   Node.js version: v14.17.5

**Extra Info**
Could be useful to know, I am using neo4j 4.3.2

### PR DESCRIPTION
**Pull request by oskarhane**
# Description

This PR fixes an issue where query arguments sent to a `@cypher` directive wasn't transformed properly before sent as Cypher params (that Neo4j driver Integer type).
See the issue for more information.

# Issue

Fixes #526 

# Checklist

### CODE REVIEW

File: packages/graphql/tests/integration/issues/526.int.test.ts [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+                )
+        }
+    `;
+
+    beforeAll(async () => {
+        driver = await neo4j();
+        const session = driver.session();
+
+        try {
+            await session.run(
```
angrykoala, line 58: Is there a reason for creating multiple sessions within the same suite?

If we create the session in beforeAll and close it in afterAll we can avoid all the try...catch and sessions setup, may lead to slightly simpler code :thinking: .
```ts
let session: Session

beforeAll(()=>{
     session=driver.session();
})

afterAll(()=>{
     session.close()
})
```

Thoughts?

If a new session is needed per test, we could do the same in before...after

angrykoala (approved):
LGTM, just a minor alternative to handling sessions in tests

packages/graphql/tests/integration/issues/526.int.test.ts
oskarhane: It's good practice to create a new session per query, so I'd prefer it that way. Sessions are very light weight.

danstarns (approved):
LGTM 👍

angrykoala (approved):

Issue #174, opened 2021-04-26
OGM Auto Generated TypeScript 


danstarns:
**Is your feature request related to a problem? Please describe.**
As a user of the OGM, I would like to see autogenerated typescript types. Currently, I have to use generics and cross-check with the GraphQL API. 

**Describe the solution you'd like**
A script I can call to produce typings for the OGM. 

**Additional context**
Please express interest in this issue so we can gauge demand.

TreeMan360:
I would definitely benefit from this feature. Thumbs up from me 👍

PR #528, opened 2021-10-07
Feat: Autogenerated Typescript for OGM

danstarns:
# Description

Please watch this video that shows how to use it and bonus points for anyone who records how many times I say the words "For Example" 😂

https://user-images.githubusercontent.com/35999252/141288522-a811f497-bdf5-4a04-90bc-5f267773d636.mp4

# Issue

#174


KleptoKat:
Now that the types are incorporated into the OGM, would it be possible to type the output based on the string input? An additional generic is redundant if the OGM has the type file.

This example demonstrates what I mean:

```typescript
interface GenericModel {}
interface Actor {}
interface Movie {}

interface ModelMap {
	Movie: Movie;
	Actor: Actor;
}

export type PropType<TObj, TProp extends keyof TObj> = TObj[TProp];

class OGM {

	model<T extends keyof ModelMap | string = string>(name: T) {

		type ModelType = T extends keyof ModelMap ? PropType<ModelMap, T> : GenericModel;
		const model: ModelType = {} as any;

		return model;
	}

}

const ogm = new OGM();

const actor = ogm.model('Actor');
const genericModel = ogm.model('Blah');
const movie = ogm.model('Movie');
```

Additionally, do we _need_ to pass the driver to the OGM when generating types? It seems unnecessary if you are doing type generation in something like gulp or webpack as a pre-build task.

redbar0n:
@danstarns @darrellwarde Wouldn't it be better to go the other way around: to do code-first auto-generation of the graphql schema typeDefs from TypeScript types? Like [Nexus](https://nexusjs.org/) does. Since it brings [some important benefits](https://bedrock.mxstbr.com/tools/nexus/). You'd get a single-source-of-truth and incredible TS autocomplete support (as opposed to having types in strings using string interpolations; e.g. [typeDefs](https://github.com/neo4j/graphql/tree/master/packages/graphql)).

>“Having a schema and a separate list of resolvers that I have to manually match is prone to bugs and impossible to do type-safely. With Nexus, I have a single source of truth for both.” - Max Stoiber

This would apply to Custom Resolvers, which currently has to be [manually kept in sync](https://neo4j.com/docs/graphql-manual/current/custom-resolvers/#_custom_object_type_field_resolver) in `neo4j/graphql-ogm`.

See also [The problems of Schema-first GraphQL development](https://www.prisma.io/blog/the-problems-of-schema-first-graphql-development-x1mn4cb0tyl3), which outlines why Prisma went the way of code-first instead. Prisma being an ORM, similar to `neo4j/graphql-ogm` being an OGM, it could be beneficial to take inspiration from their learnings.

I imagine the [directives](https://neo4j.com/docs/graphql-manual/current/type-definitions/relationships/) neo4j adds to graphql (like `@relationship(type: "IN_GENRE", direction: OUT)`) could also be auto-generated by the OGM, all from the TS code. _Virtually eliminating the developer having to deal with the GraphQL schema at all._ I also imagine code-first would be far easier than schema-first trying to target the TS types, since types in TS have greater variability and complexity than the GraphQL SDL. It seems to make sense to work in the most powerful language (TS), while targeting the least powerful language for static output (GraphQL SDL). As opposed to working in the GraphQL SDL (with neo4j directives etc.) and trying to target the more powerful type system of TS.

Please correct me if I'm wrong, or if its something I've misunderstood.

danstarns:
> @danstarns @darrellwarde Wouldn't it be better to go the other way around: to do code-first auto-generation of the graphql schema typeDefs from TypeScript types? Like [Nexus](https://nexusjs.org/) does. Since it brings [some important benefits](https://bedrock.mxstbr.com/tools/nexus/). (Nexus even works with plain JavaScript, so no TS required).
> 
> You'd get a single-source-of-truth and incredible TS autocomplete support (as opposed to having types in strings using string interpolations; e.g. [typeDefs](https://github.com/neo4j/graphql/tree/master/packages/graphql)):
> 
> > "Nexus auto-generates type-definitions as we develop, and infers them in our code, giving us IDE completion and type error catching out of the box" [ref](https://dev.to/novvum/typegraphql-and-graphql-nexus-a-look-at-code-first-apis-6k0)
> 
> It also makes Custom Resolvers easier, which currently has to be [manually kept in sync](https://neo4j.com/docs/graphql-manual/current/custom-resolvers/#_custom_object_type_field_resolver) in `neo4j/graphql-ogm`:
> 
> > “Having a schema and a separate list of resolvers that I have to manually match is prone to bugs and impossible to do type-safely. With Nexus, I have a single source of truth for both.” - Max Stoiber
> 
> See also [The problems of Schema-first GraphQL development](https://www.prisma.io/blog/the-problems-of-schema-first-graphql-development-x1mn4cb0tyl3), which outlines why Prisma went the way of code-first instead. Prisma being an ORM, similar to `neo4j/graphql-ogm` being an OGM, it could be beneficial to take inspiration from their learnings.
> 
> I imagine the [directives](https://neo4j.com/docs/graphql-manual/current/type-definitions/relationships/) neo4j adds to graphql (like `@relationship(type: "IN_GENRE", direction: OUT)`) could also be auto-generated by the OGM, all from the TS code. _Virtually eliminating the developer having to deal with the GraphQL schema at all._ I also imagine code-first would be far easier than schema-first trying to target the TS types, since types in TS have greater variability and complexity than the GraphQL SDL. It seems to make sense to work in the most powerful language (TS), while targeting the least powerful language for static output (GraphQL SDL). As opposed to working in the GraphQL SDL (with neo4j directives etc.) and trying to target the more powerful type system of TS.
> 
> Please correct me if I'm wrong, or if its something I've misunderstood.

Thank you for your input. We are all about ease of developer ramp up. Using the minimal GraphQL type definitions and outputting an extensive GraphQL schema, backed by Neo4j, appeals to the many Frontend/Fullstack developers plus the developers who don't know much Javascript. Having our input a single string means you can build a schema using various tools such as https://arrows.app/, Nexus or soon to be schema inference.

packages/ogm/src/classes/Model.ts
```diff
@@ -352,8 +352,7 @@ class Model {
         ];
 
         Object.entries(aggregate).forEach((entry) => {
-            // Must be count
-            if (!Object.keys(entry).length) {
+            if (((entry[0] as unknown) as string) === "count") {
```
angrykoala, line 355: :clap:

packages/ogm/src/classes/OGM.ts
```diff
@@ -23,27 +23,29 @@ import { filterDocument } from "../utils";
...
 
-        const neoSchema = new Neo4jGraphQL({
+        this.neoSchema = new Neo4jGraphQL({
             ...rest,
             typeDefs: filterDocument(typeDefs),
         });
 
         this.checkNeo4jCompat = function checkNeo4jCompat() {
-            return neoSchema.checkNeo4jCompat({
+            return this.neoSchema.checkNeo4jCompat({
```
angrykoala, line 42: Now that `neoSchema` is a class parameter, couldn't we make `checkNeo4jCompat` a simple public function?

```ts
checkNeo4jCompat(): Promise<void>{
    return this.neoSchema.checkNeo4jCompat({
        drive: this.rest.driver,  // rest should be a private parameter
        ...(rest.config?.driverConfig ? { driverConfig: this.rest.config.driverConfig } : {})
       })
}
```

I think is simpler and more readable than a curryfied function in the constructor

angrykoala (approved):
LGTM

packages/ogm/src/classes/OGM.ts
danstarns: Good observation, I'm going to merge this seeming the pipelines passed for this and knowing making a change will leave another PR unmerged until it's fixed.

PR #531, opened 2021-10-08
Add GraphElement and inherit Node and Relationship from it

angrykoala:
# Description

This PR adds a new abstract class `GraphElement` to be inherited by Node and Relationship.
The reasoning is to extract common properties and code to that parent class as well as using it in several methods that accept both elements. Conceptually, Node and Relationship share concepts as are graph elements that exists in the database.

Additionally, node properties `mutableFields` and `authableFields` are moved into getters to explicitly define their roles as computed properties and ensure consistency with other properties. 
If performance if a concern (now or in the future) due to these variables being created on each call, [memoization](https://www.npmjs.com/package/typescript-memoize) would fix that while keeping consistency.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -173,7 +173,9 @@ jobs:
         neo4j-version:
           - 4.1.5
           - 4.1
+          - 4.2.9
           - 4.2
+          - 4.3.2
           - 4.3
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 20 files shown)


angrykoala:
Regarding naming of this class:
:rocket: -> GraphElement
:eyes:  -> GraphEntity
:heart: No preference

Issue #556, opened 2021-10-30
GraphQLError: Input Object type ArticleCreateInput must define one or more fields.


KleptoKat:
**Describe the bug**
When you create a type which only has _only_ auto generated and/or relationship fields on it and then attempt to run the generated schema through the "validateSchema" function in graphql (used by Apollo and the like) you will get multiple errors: GraphQLError: Input Object type ArticleCreateInput must define one or more fields; GraphQLError: Input Object type ArticleUpdateInput must define one or more fields. There is no way to disable this functionality in the graphql validateSchema function.

**Type definitions**
```graphql
type Journalist {
    articles: [Article]! @relationship(type: "HAS_ARTICLE", direction: OUT, properties: "HasArticle" )
}

interface HasArticle @relationshipProperties {
    createdAt: DateTime! @timestamp
}

type Article {
    id: ID! @id
}
```

**To Reproduce**
Run an apollo server with the above types:
```typescript

    const apolloServer = new ApolloServer({
        schema: neoSchema.schema,
        context: ({ request }) => ({ request }),
    });
```

Alternatively, run the schema through validateSchema:

```typescript
import { validateSchema } from "graphql";
const errors = validateSchema(neoSchema.schema);
expect(errors).toEqual([]);
```

errors should be an empty array.

**Expected behavior**
Empty inputs should be either:
a) Removed from the schema. Would break create and update functionality.
b) Have a placeholder field placed on them. Introduces possibility of frontend regressions if a client uses the placeholder field
c) Replaced with a scalar. Scalars can be empty objects.

**Additional context**
I am running on the latest master branch.

This should also take interfaces into account.

angrykoala:
As discussed in #557, This should be properly documented with the workaround of adding a field so users can apply it themselves without adding unexpected fields in the final schema

KleptoKat:
Hey @angrykoala @danstarns apologies for the delay in response, been doing some travelling.

My issue with simply documenting this issue is twofold:

 1. There are many cases where users have nodes that have no manually provided properties on them, just relationships and/or autogenerated properties. The error that Apollo Graphql gives the user is not verbose enough and would require the user to google the error to find that specific piece of documentation if they haven't thoroughly read the documentation. Then, they have to implement a workaround which, in my opinion, doesn't seem very professional because...
[...]

PR #557, opened 2021-10-30
Fix/556: Input Object type ArticleCreateInput must define one or more fields

KleptoKat:
# Description

This merge adds an `_emptyInput` field to empty `NodeUpdateInput` and `NodeCreateInput`'s to ensure that they will compile. This seemed like the quickest way to get a fix out without changing to much about the code.

# Issue

#556 

# Checklist


packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -510,6 +507,21 @@ function makeAugmentedSchema(
         });
     });
 
+    function ensureNonEmptyInput(inputName: string | InputTypeComposer<any>) {
+        let input;
+        if (typeof inputName === 'string') {
+            input = composer.getITC(inputName);
+        } else {
+            input = inputName;
+        }
```
angrykoala, line 516: Here could we consider using a ternary operator? :thinking: it'll be less verbose and input will be properly typed and const:

```suggestion
        const input = isString(inputName) ? composer.getITC(inputName) : inputName;
```

Also, in the suggestion I'm using the method `isString` from `import { isString } from "../utils/utils";`

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -510,6 +507,21 @@ function makeAugmentedSchema(
         });
     });
 
+    function ensureNonEmptyInput(inputName: string | InputTypeComposer<any>) {
```
angrykoala, line 510: `inputOrName` may be a slightly more accurate name for the param

 ```suggestion
     function ensureNonEmptyInput(inputOrName: string | InputTypeComposer<any>) {
```

File: packages/graphql/tests/schema/issues/556.test.ts [hunk 3]
```diff
@@ -0,0 +1,81 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { gql } from "apollo-server";
+import { validateSchema } from "graphql";
+import { Neo4jGraphQL } from "../../../src/classes";
+
+
+describe("https://github.com/neo4j/graphql/issues/556", () => {
```
angrykoala, line 25: :clap:

packages/graphql/tests/schema/issues/556.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                featuredIn: [ Article! ] @relationship(type: "HAS_IMAGE", direction: IN )
+                url: String!
+            }
+        `;
+ 
+        const neoSchema = new Neo4jGraphQL({ typeDefs });
+        expect(neoSchema.schema).toBeDefined();
+
+        const errors = validateSchema(neoSchema.schema);
+        expect(errors).toEqual([]);
```
angrykoala, line 79: What would happen if we have a truly empty field after these changes?

e.g.:
```graphql
type MyField {

}
```

I reckon it will still fail due to parsing error (which is desirable), but it may be worth adding a new test case just to make sure the new `_emptyField` does not hide/remove expected errors

angrykoala (changes requested):
Great job, a few minor changes but overall it seems to be working fine and looks like a reasonable approach

KleptoKat:
Changes implemented, thank you :)

angrykoala:
Hi @KleptoKat, after some discussion, we decided that it is better to avoid adding extra fields as it may cause unexpected behavior.

Because this seems to be a relatively uncommon use case (nodes with no properties and relations),we decided to document this case so the users are aware of this and, if they want to, add an extra field (like the one you are proposing) themselves as a workaround, this way users will be aware of this extra field in their model.

angrykoala:
Hi @KleptoKat 
PR reopened. As I mentioned, everything looks good, the only thing missing would be adding a comment regarding the `_emptyInput` field in the docs (the FAQs section would be a good place) so if someone finds this field in their schema they can find out what it is about

Thanks again!

KleptoKat:
I added it to both the docs and the property description, thanks for re-opening @angrykoala !

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -512,8 +512,13 @@ function makeAugmentedSchema(
         const input = isString(nameOrInput) ? composer.getITC(nameOrInput) : nameOrInput;
 
         if (input.getFieldNames().length === 0) {
+            const faqURL = `https://neo4j.com/docs/graphql-manual/current/troubleshooting/faqs/`;
             input.addFields({
-                _emptyInput: 'Boolean',
+                _emptyInput: {
+                    type: 'Boolean',
+                    description: `Appears because this input type would be empty otherwise because this type is ` +
```
angrykoala, line 519: :clap:

angrykoala (approved):

danstarns (changes requested):
Great, just some conflicts that I think are due to auto alphabetical ordering of imports.

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-557.surge.sh

darrellwarde:
I've been thinking about this a lot since I saw it raised, it feels like a bit of a hacky workaround but I can of course see the value.

I just want to run through some actual usage use cases, so taking the following type definitions:

```graphql
type User {
  name: String!
  things: [Thing!]! @relationship(type: "HAS_THINGS", direction: OUT)
}

type Thing {
  id: ID! @id
}
```

How would the actual generated types be used in practice? Ideally, the `_emptyInput` field would not even need to be specified, so I would be able to execute the following to create a user and 3 "things":

```graphql
mutation {
  createUsers(
    input: {
      name: "Darrell"
      things: {
        create: [
          { }
          { }
        ]
      }
    }
  ) {
    users {
      name
    }
  }
}
```

Is this possible, and if so, could we a test case for this in action?

However, if we have to do the following:

```graphql
mutation {
  createUsers(
    input: {
      name: "Darrell"
      things: {
        create: [
          { _emptyInput: "Some random thing" }
          { _emptyInput: "Some random thing" }
        ]
      }
    }
  ) {
    users {
      name
    }
  }
}
```

This just feels a _little_ too hacky for me. If this is the case, I did have another slightly wildcard idea that the nested `create` for an "empty" node could take a number instead of an array of objects:

```graphql
mutation {
  createUsers(
    input: {
      name: "Darrell"
      things: {
        create: 2
      }
    }
  ) {
    users {
      name
    }
  }
}
```

Not something that has been discussed at all, just my wild musings! Sorry that it's taken a while for me to get to this, I look forward to hearing your thoughts @KleptoKat!

KleptoKat:
Hey Darrell,

Yeah it feels a little hacky; I wish you could disable the error on the graphql side of things instead of this workaround. The alternative of course is to rebuild _way_ too much of this repo to accommodate this issue though.

Your first case is correct as graphql _does_ allow empty interfaces for client side input values, where creating an object would look like this:

```
mutation {
  createUsers(
    input: {
      name: "Darrell"
      things: {
        create: [
          { }
          { }
        ]
      }
    }
  ) {
    users {
      name
    }
  }
}
```

but of course, you could _also_ do the latter but we should discourage users from doing that because _emptyInput will be removed if they add a user-defined property which may break their frontend code.

I like your last thought but it might cause some inconsistencies, IE you can't use `create: 2` on objects that require user-defined values unless you _also_ define those values for all of the objects which would have to be defined somewhere in that schema. If we only have it for values that don't have any user-defined values then it would break a users create code.

darrellwarde:
> I like your last thought but it might cause some inconsistencies, IE you can't use `create: 2` on objects that require user-defined values unless you _also_ define those values for all of the objects which would have to be defined somewhere in that schema. If we only have it for values that don't have any user-defined values then it would break a users create code.

I was purely exploring a slightly crazy idea in my head, I think you're right that it's definitely not a good idea for now! 

> ```graphql
> mutation {
>   createUsers(
>     input: {
>       name: "Darrell"
>       things: {
>         create: [
>           { }
>           { }
>         ]
>       }
>     }
>   ) {
>     users {
>       name
>     }
>   }
> }
> ```

If we could just see an integration test or two for a Mutation like the above, this will definitely be approved and merged. No pressure, but releasing early next week, so if you can find time before then, then you will see this fix in the next release. Cheers! 🙂

darrellwarde:
Hey @KleptoKat, I hope you don't mind, but I had the integration test I requested above - just super keen to get this merged in the New Year!

KleptoKat:
> Hey @KleptoKat, I hope you don't mind, but I had the integration test I requested above - just super keen to get this merged in the New Year!

Hey @darrellwarde happy holidays! Thanks for adding that - it looks good!

packages/graphql/tests/integration/issues/557.int.test.ts
```diff
@@ -0,0 +1,90 @@
...
+import { graphql } from "graphql";
+import { gql } from "apollo-server";
+import neo4j from "../neo4j";
+import { Neo4jGraphQL } from "../../../src/classes";
+
+describe("https://github.com/neo4j/graphql/issues/557 - Input Object type ArticleCreateInput must define one or more fields", () => {
+    let driver: Driver;
+    let bookmarks: string[];
+    const typeDefs = gql`
+        type User557 {
```
angrykoala, line 30: Not too important, but I think it is a better practice using unique, pseudorandom values for our types in integration tests, not only to solve name collision (in this case, it would be extremely rare for it to happen) but also for consistency (i.e. all integration tests using the same tooling that ensures no-collision, instead of relying on naming conventions)

packages/graphql/tests/integration/issues/557.int.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    beforeAll(async () => {
+        driver = await neo4j();
+    });
+
+    afterAll(async () => {
+        const session = driver.session();
+
+        try {
+            await session.run(`MATCH (u:User557) DETACH DELETE u`);
```
angrykoala, line 48: :+1:

packages/graphql/tests/integration/issues/557.int.test.ts
```diff
@@ -0,0 +1,90 @@
...
+            schema: neoSchema.schema,
+            source: query,
+            contextValue: { driver, driverConfig: { bookmarks } },
+        });
+
+        expect(result.errors).toBeFalsy();
+
+        expect(result.data?.createUser557s.user557s).toHaveLength(1);
+        expect(result.data?.createUser557s.user557s[0].name).toEqual("Darrell");
+        expect(result.data?.createUser557s.user557s[0].things).toHaveLength(2);
```
angrykoala, line 87: Should we check that these nodes contain the `_emptyInput` field?

packages/graphql/tests/integration/issues/557.int.test.ts
```diff
@@ -0,0 +1,90 @@
...
+ * limitations under the License.
+ */
+
+import { Driver } from "neo4j-driver";
+import { graphql } from "graphql";
+import { gql } from "apollo-server";
+import neo4j from "../neo4j";
+import { Neo4jGraphQL } from "../../../src/classes";
+
+describe("https://github.com/neo4j/graphql/issues/557 - Input Object type ArticleCreateInput must define one or more fields", () => {
```
angrykoala, line 26: :thinking: I'm not sure why using a separate file (`557`) instead of adding these tests to the (`556`) as we are testing the same piece of work.

I'm happy to leave it as is, just pointing out that seems a bit weird

angrykoala (approved):
Approved, a few comments on possible improvements for the integration tests @darrellwarde

packages/graphql/tests/integration/issues/557.int.test.ts
darrellwarde: They shouldn't do, the `_emptyInput` should never make it through to the database.

packages/graphql/tests/integration/issues/557.int.test.ts
darrellwarde: 556 doesn't exist yet (it's a schema test), but you're absolutely right, renaming to 556 now.

packages/graphql/tests/integration/issues/557.int.test.ts
darrellwarde: I've literally just tried to do this now, and until we fix our inconsistent pluralization, this is just so broken (the whole camel-casing thing). GIven I clean up in these tests, let's leave it given that it works for now?

PR #564, opened 2021-11-05
Added OpenID Json Web Key Sets for verification via the @auth directive

dvanmali:
# Description

According to OpenID configuration, verification of JWTs should be validated against the authentication server's JWKS to ensure that the server signed it. Instead of interpreting the JWT via a secret or PEM file (which if maliciously stolen, could rewrite JWTs), we verify via the signature represented by the token's kid at the endpoint `https://YOUR_DOMAIN/.well-known/jwks.json`. 

These changes are currently non-breaking changes.

# Issue

#252 

Passing a decoded token only partially solves this verification issue. We'd rather perform verification via the GraphQL @auth directive so checks occur at each request and are done **without secrets**.

# Checklist

### DIFF
--- a/packages/graphql/src/auth/get-jwt.ts
+++ b/packages/graphql/src/auth/get-jwt.ts
@@ -76,7 +76,7 @@ function getJWT(context: Context): any {
             // Create a JWK Client with a rate limit that
             // limits the number of calls to our JWK endpoint
             client = new JwksClient({
-                jwksUri: jwtConfig.jwkEndpoint || "",
+                jwksUri: jwtConfig.jwkEndpoint,
                 rateLimit: true,
                 jwksRequestsPerMinute: 10, // Default Value
                 cache: true, // Default Value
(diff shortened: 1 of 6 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-564.surge.sh

packages/graphql/src/auth/get-jwt.ts
```diff
@@ -68,8 +70,33 @@ function getJWT(context: Context): any {
...
             result = jsonwebtoken.decode(token);
-        } else {
-            debug("Verifying JWT");
+        } else if (jwtConfig.jwkEndpoint) {
+            debug("Verifying JWT using OpenID Public Key Endpoint");
+
+            // Create a JWK Client with a rate limit that
+            // limits the number of calls to our JWK endpoint
+            client = new JwksClient({
+                jwksUri: jwtConfig.jwkEndpoint || "",
```
oskarhane, line 79: Do we really need a default value here? At this point the `jwtConfig.jwkEndpoint` should always be set, right?

oskarhane (commented):
Thank you for this PR @dvanmali, really appreciate it.
Looks good in general, just a small comment.

Also, could you come up with a way to test this (that it takes the code path when the config value is set) so we don't introduce regressions in the future?

oskarhane (approved):
LGTM, thanks a lot @dvanmali

Issue #560, opened 2021-11-01
Cannot read property 'y' of null error on nullable Points


d-johnston-fugro:
**Describe the bug**
When the schema contains a nullable point, the field resolver blows up when it tries to resolve the longitude and latitude fields. I believe this is due to how the auto generated Cypher is created.

**Type definitions**
```graphql
type Log {
  location: Point
}
```

**To Reproduce**

Any schema with a nullable Point type should exhibit the issue.

**Expected behavior**
Field resolves as null without error

**System (please complete the following information):**

-   OS: Linux
-   Version: 2.3.0,
-   Node.js version: v14.17.0

**Additional context**
As I mentioned above, when a schema with a nullable point field is missing from the queried data, the field resolver from the Point object is called and then blows up on the nullable object. I believe this is because the auto generated cypher wraps the result in the `{ point: ... }` object so the result is `{ location: { point: null } }` instead of `{ location: null }`. This additional object indirection tricks the graphql resolver into calling the field resolver on the `{ point: null }` object because it doesn't realise it is null.

Marking the field as required makes the whole query fail, marking it as not required allows the data to return but the response is populated with the resolver error which by default causes the Apollo client to reject the "partial" data.

d-johnston-fugro:
Right now I'm able to work around this by patching it with a custom resolver.

```ts
  const neo4jGraphQL = new Neo4jGraphQL({
    typeDefs,
    driver,
    resolvers: {
      Log: {
        location: (parent: {location: { point: unknown }}) => {
          if (parent.location.point === null) return null;

          return parent.location;
        },
      },
    },
  });
  ```
  
  This intercepts the resolution of the field and sets it to null when the inner point is null, thus properly skipping the field resolvers on the Point type.

angrykoala:
Hi @d-johnston-fugro 
Thanks for sharing your workaround for anyone who may be experiencing the same issue while this is fixed

PR #565, opened 2021-11-05
Fix/null points

danstarns:
# Description

Uses `runFirstColumn` with a `CASE` statement to ensure we only project point elements that are not `NULL`. 

# Issue

#560


oskarhane (approved):
LGTM 👍 
Even if we don't want to use more APOC, I think in cases like this we don't have much of an option.

Issue #567, opened 2021-11-05
Error when only selecting info on update with disconnect


danstarns:
**Describe the bug**
When only selecting the `info` object from mutation error is thrown: 

```
Cannot destructure property 'fieldsByTypeName' of 'Object.values(...).find(...)' as it is undefined
```

**Type definitions**

```gql
type Movie {
    title: String
    imdbRating: Int
}
```

**Query** 

```gql
mutation {
  updateMovies(
    where: { title: "Forrest Gump" }
    update: { imdbRating: 123 }
  ) {
    info {
      nodesCreated
      nodesDeleted
    }
  }
}
```

**Error**

```json
{
  "errors": [
    {
      "message": "Cannot destructure property 'fieldsByTypeName' of 'Object.values(...).find(...)' as it is undefined.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "updateMovies"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: Cannot destructure property 'fieldsByTypeName' of 'Object.values(...).find(...)' as it is undefined.",
            "    at Object.translateUpdate (C:\\code\\neo4j\\graphql\\packages\\graphql\\src\\translate\\translate-update.ts:63:13)",
            "    at C:\\code\\neo4j\\graphql\\packages\\graphql\\src\\schema\\resolvers\\update.ts:29:34",
            "    at Generator.next (<anonymous>)",
            "    at C:\\code\\neo4j\\graphql\\packages\\graphql\\src\\schema\\resolvers\\update.ts:26:71",
            "    at new Promise (<anonymous>)",
            "    at __awaiter (C:\\code\\neo4j\\graphql\\packages\\graphql\\src\\schema\\resolvers\\update.ts:22:12)",
            "    at resolve (C:\\code\\neo4j\\graphql\\packages\\graphql\\src\\schema\\resolvers\\update.ts:35:16)",
            "    at C:\\code\\neo4j\\graphql\\packages\\graphql\\dist\\schema\\src\\addSchemaLevelResolver.js:35:24",
            "    at C:\\code\\neo4j\\graphql\\packages\\graphql\\node_modules\\value-or-promise\\src\\ValueOrPromise.ts:78:36",
[...]

PR #568, opened 2021-11-05
fix: error when only selecting info on update and create operations

danstarns:
# Issue

#567

### DIFF
--- a/packages/graphql/src/translate/translate-create.ts
+++ b/packages/graphql/src/translate/translate-create.ts
@@ -26,19 +26,18 @@ import createConnectionAndParams from "./connection/create-connection-and-params
 import createInterfaceProjectionAndParams from "./create-interface-projection-and-params";
 
 function translateCreate({ context, node }: { context: Context; node: Node }): [string, any] {
+    const { resolveTree } = context;
     const connectionStrs: string[] = [];
     const interfaceStrs: string[] = [];
     let connectionParams: any;
     let interfaceParams: any;
 
-    const { resolveTree } = context;
+    const mutationResponse =
+        resolveTree.fieldsByTypeName[`Create${node.getPlural({ camelCase: false })}MutationResponse`];
 
-    // Due to potential aliasing of returned object in response we look through fields of CreateMutationResponse
-    // and find field where field.name ~ node.name which exists by construction
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const { fieldsByTypeName } = Object.values(
-        resolveTree.fieldsByTypeName[`Create${node.getPlural({ camelCase: false })}MutationResponse`]
-    ).find((field) => field.name === node.getPlural({ camelCase: true }))!;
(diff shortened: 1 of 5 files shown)


packages/graphql/src/translate/translate-create.ts
```diff
@@ -26,19 +26,18 @@ import createConnectionAndParams from "./connection/create-connection-and-params
...
 
-    // Due to potential aliasing of returned object in response we look through fields of CreateMutationResponse
-    // and find field where field.name ~ node.name which exists by construction
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const { fieldsByTypeName } = Object.values(
-        resolveTree.fieldsByTypeName[`Create${node.getPlural({ camelCase: false })}MutationResponse`]
-    ).find((field) => field.name === node.getPlural({ camelCase: true }))!;
+    const nodeProjection = Object.values(mutationResponse).find(
+        (field) => field.name === node.getPlural({ camelCase: true })
+    );
```
angrykoala, line 40: :clap:

packages/graphql/src/translate/translate-create.ts
```diff
@@ -68,57 +67,78 @@ function translateCreate({ context, node }: { context: Context; node: Node }): [
...
-                field: connectionField,
-                context,
-                nodeVariable: "REPLACE_ME",
-            });
-            connectionStrs.push(connection[0]);
-            if (!connectionParams) connectionParams = {};
-            connectionParams = { ...connectionParams, ...connection[1] };
+    let replacedProjectionParams: Record<string, unknown> = {};
+    let projectionStr: string | undefined = undefined;
+    let authCalls: string | undefined = undefined;
```
angrykoala, line 72: I think there is no need to assign these as are `undefined` by default

```suggestion
    let projectionStr: string | undefined;
    let authCalls: string | undefined;
```

packages/graphql/tests/integration/issues/only-info.test.ts
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+import { Driver } from "neo4j-driver";
+import { graphql } from "graphql";
+import { gql } from "apollo-server";
+import { generate } from "randomstring";
+import neo4j from "../neo4j";
+import { Neo4jGraphQL } from "../../../src/classes";
+
+describe("only-info", () => {
```
angrykoala, line 27: We may want to have a few more details in the description for the future :sweat_smile:

angrykoala (dismissed):
Looks good to me!, just a couple of proposals

packages/graphql/tests/integration/issues/only-info.test.ts
danstarns: Sure

packages/graphql/src/translate/translate-create.ts
danstarns: Good call

Issue #549, opened 2021-10-25
Missing input validation on nested mutations for mandatory reference fields


AccsoSG:
**Describe the bug**
As in the documentation described (https://neo4j.com/docs/graphql-manual/current/type-definitions/relationships/) it shouldn't be possible to create a new node, when it needs a connected node. For Example: According to the GQL scheme (see below), a movie should have a director of type Person. The exclamation marks the director-field as mandatory. The documentation says "You can’t create a Movie without adding a director [...]", however it is possible.

**Type definitions**
```
type Person {
    name: String!
    born: Int!
    actedInMovies: [Movie!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: OUT)
    directedMovies: [Movie!]! @relationship(type: "DIRECTED", direction: OUT)
}

type Movie {
    title: String!
    released: Int!
    actors: [Person!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: IN)
    director: Person! @relationship(type: "DIRECTED", direction: IN)
}

interface ActedIn @relationshipProperties {
    roles: [String!]
}
```

**To Reproduce**

1. Run a server with the following code...
2. Execute the following Mutation...
**Mutation:**
```
mutation {
  createMovies(input: [{title: "Test", released: 2022}]) {
    movies {
      title
    }
  }
}
```
**Result:**
```
{
  "data": {
    "createMovies": {
      "movies": [
        {
          "title": "Test3"
        }
      ]
    }
  }
}
```

**3. Then run the following Query...**
```
query {
  movies {
    title
    director {
      name
    }
  }
}
```

**4. See error**
```
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Movie.director.",  [...]
...}]}
```
**Expected behavior**
It should't be possible to create a movie without director.

darrellwarde:
The attempted fix for this added in #572 introduced quite a nasty regression which we're now reverting. This will get looked at soon, with a more comprehensive fix.

darrellwarde:
If you were affected by the regression introduced in 2.5.0, please upgrade to 2.5.3 where this should now be resolved! 📢

PR #572, opened 2021-11-09
Fix/non null relationships

danstarns:
# Description

Enforces that Nodes have relationships, on update and create, if the field is annotated with an `!`. This PR uses `apoc.util.validate` to throw an error inside a transaction. 

# Issue

#549

### DIFF
--- a/packages/graphql/src/translate/create-create-and-params.ts
+++ b/packages/graphql/src/translate/create-create-and-params.ts
@@ -21,9 +21,10 @@ import { Node, Relationship } from "../classes";
 import { Context } from "../types";
 import createConnectAndParams from "./create-connect-and-params";
 import createAuthAndParams from "./create-auth-and-params";
-import { AUTH_FORBIDDEN_ERROR } from "../constants";
+import { AUTH_FORBIDDEN_ERROR, RELATIONSHIP_REQUIREMENT_PREFIX } from "../constants";
 import createSetRelationshipPropertiesAndParams from "./create-set-relationship-properties-and-params";
 import mapToDbProperty from "../utils/map-to-db-property";
+import createRelationshipValidationStr from "./create-relationship-validation-str";
 
 interface Res {
     creates: string[];
@@ -243,6 +244,12 @@ function createCreateAndParams({
     if (meta?.authStrs.length) {
         creates.push(`WITH ${withVars.join(", ")}`);
         creates.push(`CALL apoc.util.validate(NOT(${meta.authStrs.join(" AND ")}), ${forbiddenString}, [0])`);
+    }
+
+    let relationshipValidationStr = createRelationshipValidationStr({ node, context, varName });
+    if (relationshipValidationStr) {
+        creates.push(`WITH ${withVars.join(", ")}`);
+        creates.push(relationshipValidationStr);
     }
 
     return [creates.join("\n"), params];
--- a/packages/graphql/src/translate/create-create-and-params.ts
+++ b/packages/graphql/src/translate/create-create-and-params.ts
@@ -246,7 +246,7 @@ function createCreateAndParams({
         creates.push(`CALL apoc.util.validate(NOT(${meta.authStrs.join(" AND ")}), ${forbiddenString}, [0])`);
     }
 
-    let relationshipValidationStr = createRelationshipValidationStr({ node, context, varName });
+    const relationshipValidationStr = createRelationshipValidationStr({ node, context, varName });
     if (relationshipValidationStr) {
         creates.push(`WITH ${withVars.join(", ")}`);
         creates.push(relationshipValidationStr);
--- a/packages/graphql/src/translate/create-relationship-validation-str.ts
+++ b/packages/graphql/src/translate/create-relationship-validation-str.ts
@@ -1,0 +1,48 @@
+import { Node } from "../classes";
+import { RELATIONSHIP_REQUIREMENT_PREFIX } from "../constants";
+import { Context } from "../types";
+
+function createRelationshipValidationStr({
+    node,
+    context,
+    varName,
+}: {
+    node: Node;
+    context: Context;
+    varName: string;
+}): string {
+    let relationshipValidationStr = "";
+
+    const nonNullRelationFields = node.relationFields.filter((field) => {
+        const isRequired = field.typeMeta.required;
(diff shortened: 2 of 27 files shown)


packages/graphql/src/translate/create-create-and-params.ts
```diff
@@ -245,6 +246,12 @@ function createCreateAndParams({
         creates.push(`CALL apoc.util.validate(NOT(${meta.authStrs.join(" AND ")}), ${forbiddenString}, [0])`);
     }
 
+    let relationshipValidationStr = createRelationshipValidationStr({ node, context, varName });
```
angrykoala, line 249: I'm not sure why this is not const?

packages/graphql/src/translate/create-relationship-validation-str.ts
```diff
@@ -0,0 +1,48 @@
...
+        const isArrayTypeRequired = field.typeMeta.arrayTypeRequired;
+        const isArray = field.typeMeta.array;
+        const isNotUnionOrInterface = Boolean(field.union) || Boolean(field.interface);
+
+        if (isArray) {
+            return isArrayTypeRequired && !isNotUnionOrInterface;
+        }
+
+        return isRequired && !isNotUnionOrInterface;
+    });
```
angrykoala, line 27: Consider moving this to a separate function: 
```suggestion
    const nonNullRelationFields = filterNonNullFields(node.relationFields);
```
(This is a suggestion, I'm happy to keep it as is too)

packages/graphql/src/translate/create-relationship-validation-str.ts
```diff
@@ -0,0 +1,48 @@
...
+            const relTypeStr = `[:${field.type}]`;
+            const toNode = context.neoSchema.nodes.find((n) => n.name === field.typeMeta.name) as Node;
+            const exitsStr = `EXISTS((${varName})${inStr}${relTypeStr}${outStr}(${toNode.getLabelString(context)}))`;
+
+            return `apoc.util.validatePredicate(NOT(${exitsStr}), '${RELATIONSHIP_REQUIREMENT_PREFIX}${node.name}.${field.fieldName} required', [0])`;
+        });
+
+        relationshipValidationStr = `CALL apoc.util.validate(NOT(${nonNullPredicates.join(
+            " AND "
+        )}), '${RELATIONSHIP_REQUIREMENT_PREFIX}', [0])`;
```
angrykoala, line 42: Thoughts on just returning this?

```suggestion
       return `CALL apoc.util.validate(NOT(${nonNullPredicates.join(
            " AND "
        )}), '${RELATIONSHIP_REQUIREMENT_PREFIX}', [0])`;
```

This way we could get rid of the variable `relationshipValidationStr` and just return empty string at the end

packages/graphql/src/translate/create-relationship-validation-str.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (nonNullRelationFields.length) {
+        const nonNullPredicates = nonNullRelationFields.map((field) => {
+            const inStr = field.direction === "IN" ? "<-" : "-";
+            const outStr = field.direction === "OUT" ? "->" : "-";
+            const relTypeStr = `[:${field.type}]`;
+            const toNode = context.neoSchema.nodes.find((n) => n.name === field.typeMeta.name) as Node;
+            const exitsStr = `EXISTS((${varName})${inStr}${relTypeStr}${outStr}(${toNode.getLabelString(context)}))`;
+
+            return `apoc.util.validatePredicate(NOT(${exitsStr}), '${RELATIONSHIP_REQUIREMENT_PREFIX}${node.name}.${field.fieldName} required', [0])`;
+        });
```
angrykoala, line 38: Again, just an optional suggestion, but maybe moving this to a separate function could give a bit of extra clarity and context :thinking: :

```suggestion
        const nonNullPredicates = nonNullRelationFields.map((field) => {
            return createRelationFieldPredicate({field, context, node});
         });
```

WIth the separate function:

```ts
createRelationFieldPredicate({field, context, node}): string{
            const inStr = field.direction === "IN" ? "<-" : "-";
            const outStr = field.direction === "OUT" ? "->" : "-";
            const relTypeStr = `[:${field.type}]`;
            const toNode = context.neoSchema.nodes.find((n) => n.name === field.typeMeta.name) as Node;
            const exitsStr = `EXISTS((${varName})${inStr}${relTypeStr}${outStr}(${toNode.getLabelString(context)}))`;

            return `apoc.util.validatePredicate(NOT(${exitsStr}), '${RELATIONSHIP_REQUIREMENT_PREFIX}${node.name}.${field.fieldName} required', [0])`;

}

```

packages/graphql/tests/tck/tck-test-files/connections/projections/update.test.ts
```diff
@@ -79,6 +79,8 @@ describe("Cypher -> Connections -> Projections -> Update", () => {
             "MATCH (this:Movie)
             WHERE this.title = $this_title
             WITH this
+            WITH this
```
angrykoala, line 82: Having 2 consecutive `WITH this` is ok?
I feel we should try to avoid that, if its easy to do so

angrykoala (approved):
Approved! :+1:  just a few style suggestions.

packages/graphql/src/translate/create-create-and-params.ts
danstarns: Good call

packages/graphql/src/translate/create-relationship-validation-str.ts
danstarns: I'm going to keep this as it is. I feel the function is pretty small in size and breaking it down will just add unnecessary context switching for future readers.

packages/graphql/src/translate/create-relationship-validation-str.ts
danstarns: This is to ensure eslint does not complain about the consistent return.

packages/graphql/tests/tck/tck-test-files/connections/projections/update.test.ts
danstarns: This is an edge case that needs to be there, it wont effect the query in any way and is documented here https://github.com/neo4j/graphql/blob/master/packages/graphql/src/translate/translate-update.ts#L391

Issue #161, opened 2021-04-15
Autogenerated `MERGE` Mutations (`updateOrCreate`)


bobmhong:
Please consider adding autogenerated merge mutations similar to those in neo4j-graphql-js.

Merge mutations are very convenient for my use case of ingesting event data since it simplifies the code and allows idempotent replay of events if needed.

This feature would help transition from neo4j-graphql-js to neo4j/graphql

mobsean:
Why ist nobody talking about this? 

In my perspective, this is a major feature and while it´s not implemented we can not migrate to @neo4j/graphql.

@darrellwarde Can you tell us, when or if there are any plans implenting MERGE?

darrellwarde:
> Why ist nobody talking about this?
> 
> In my perspective, this is a major feature and while it´s not implemented we can not migrate to @neo4j/graphql.
> 
> @darrellwarde Can you tell us, when or if there are any plans implenting MERGE?

It'll definitely get done, but there is no timeline yet.

At the moment we're looking at aggregations and interface relationship fields - they were prioritised as more important based on community and customer feedback.

danstarns:
I would like to chime in and mention that we left MERGE out for a few reasons; firstly, you need to MERGE on two unique fields and so that implies that nodes must have an id field, or something unique - something we don't enforce and then secondly, we couldn't find a nice way to get it to work with nested mutations. Finally, we found the mechanics of MERGE hard to reason about especially for your average GraphQL developer with less experience with Neo4j. 

Of course, this doesn't negate the importance of MERGE, and in imparticular if you want to perform an atomic transaction. 

Your ideas and suggestions are welcome here.

rcbevans:
I'm working on a system for our datamodel which required idempotent insertion of data and had to hand write all the upsert mutations by hand.  It's not a huge deal, but would certainly be nice if there was merge/upsert mutations that could be generated.

bobmhong:
@darrellwarde Wondering if you're able to provide a rough timeline when generated MERGE mutations could be addressed? 

Does it help if the implementation uses the suggestion provided by @mobsean 

> Maybe this will work:
> a MERGE mutation will only be generated if atleast one @unique field ist defined in typeDefs.

tapnair:
This would be fantastic, but to your earlier comment it would be particularly valuable if it works relatively seamlessly with nested mutations/queries.

AnsonHwang86:
waiting for this inplement

helenesanche:
[...]

PR #575, opened 2021-11-10
Feature/connectOrCreate

angrykoala:
# Description
This PR implements the parameter `connectOrCreate` in create and update mutations for relations with nodes having `@unique` directives.

#161 

## Example

```cypher
type Actor {
    name: String!
    movies: [Movie] @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
}

type Movie {
    title: String
    id: String! @unique
    actors: [Actor]! @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
}
```

```cypher
mutation Mutation {
  createActors(input: {
    name: "Tom Hanks",
    movies: {
      connectOrCreate: {
        where: {node: {id: "1234"}}
        onCreate: {node: {title: "Forrest Gump", id: "1234"}}
      }
    }
  }) {
    info {
      nodesCreated
    }
  }
}
```

This mutation creates a new actor and connects to a movie with id `1234`. If the movie with that id does not exists, it creates it using the given title and id.

## Other changes
Additionally, this PR have the following changes:

* connectOrCreate in update mutations
* cypher-builder helpers to construct general, reusable statements (**merge**, **node** and **relation**)
* "Statement" types and utilities. A statement being defined as a pair [string, params] that can be concatenated to other params. 
* General utils and helpers


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-575.surge.sh

docs/modules/ROOT/pages/type-definitions/constraints.adoc
```diff
@@ -59,6 +59,54 @@ type Colour @node(label: "Color", additionalLabels: ["Hue"]) {
 }
 ----
 
+=== `connectOrCreate` mutations
```
danstarns, line 62: I don't think this is the correct file for this.

danstarns (changes requested):
This is cool, however, after playing around I noticed a hole with `@auth`. 

**TypeDefs:** 

```gql
type Movie {
    title: String
    genres: [Genre] @relationship(type: "IN_GENRE", direction: OUT)
}

type Genre @auth(rules: [{ operations: [CONNECT, CREATE], roles: ["admin"] }]) {
    name: String @unique
}
```

**Query:** 

```gql
mutation {
  createMovies(
    input: [
      {
        title: "Cool Movie"
        genres: {
          connectOrCreate: [
            {
              where: { node: { name: "Horror" } }
              onCreate: {
                node: { name: "Horror" }
              }
            }
          ]
        }
      }
    ]
  ) {
    movies {
      title
    }
  }
}
```

**Generated Cypher:** 

```
CALL {
CREATE (this0:Movie)
SET this0.title = $this0_title
MERGE (this0_genres_connectOrCreate0:Genre { name: $this0_genres_connectOrCreate0_node_name })
ON CREATE SET
this0_genres_connectOrCreate0.name = $this0_genres_connectOrCreate0_on_create_name
MERGE (this0)-[this0_relationship_this0_genres_connectOrCreate0:IN_GENRE]->(this0_genres_connectOrCreate0)
RETURN this0
}
RETURN
this0 { .title } AS this0
```

Notice how the auth roles check is not inserted in the cypher ?

danstarns:
This requires some changes to the OGM, that is because we are adding the `createOrConnect` key to the 'Top-Level' of the update mutation. I would like to be able to call the following: 

```gql
mutation {
  updateMovies(
    where: { title: "My Cool Movie" }
    connectOrCreate: {
      genres: {
        where: { node: { name: "Horror" } }
        onCreate: { node: { name: "Horror" } }
      }
    }
  ) {
    movies {
      title
    }
  }
}
```

With the ogm like so: 

```js
await ogm.model("Movie").update({
    where: { title: "My Cool Movie" },
    connectOrCreate: {
        genres: {
            where: { node: { name: "Horror" } },
            onCreate: { node: { name: "Horror" } },
        },
    },
});
```

This will also require changes to the type generation that was merged #528 .

docs/modules/ROOT/pages/type-definitions/constraints.adoc
angrykoala: Maybe not, the reasoning is because it is part of unique fields only, it would make sense to have it as part of that, I guess mutations/create and /update are better places

packages/graphql/src/translate/field-aggregations/apoc-run-utils.ts
```diff
@@ -1,5 +1,6 @@
 import { escapeQuery } from "./utils";
 import { AggregationAuth } from "./field-aggregations-auth";
+import { serializeObject } from "../utils";
 
 /** Wraps a query inside an apoc call, escaping strings and serializing params */
 export function wrapApocRun(query: string, params: Record<string, string> = {}): string {
```
darrellwarde, line 6: Slightly more detailed naming?

```suggestion
export function wrapWithApocRunFirstColumn(query: string, params: Record<string, string> = {}): string {
```

packages/graphql/src/translate/cypher-builder/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CypherStatement, CypherParams } from "../types";
+import { serializeObject } from "../utils";
+
+/** Generates a string to be used as parameter key */
+export function generateParameterKey(preffix: string, key: string): string {
```
darrellwarde, line 24: `prefix`

```suggestion
export function generateParameterKey(prefix: string, key: string): string {
```

packages/graphql/src/utils/test/builders/relation-field-builder.ts
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { RelationField } from "../../../types";
+import { Builder } from "./builder";
+
+export class RelationFieldBuilder extends Builder<RelationField, RelationField> {
```
darrellwarde, line 23: My usual question - does it need to be a builder? Are we likely to ever add or change anything following construction?

packages/graphql/src/schema/resolvers/update.ts
```diff
@@ -48,21 +49,25 @@ export default function updateResolver({ node }: { node: Node }) {
...
+              connect: `${node.name}ConnectInput`,
+              disconnect: `${node.name}DisconnectInput`,
+              create: `${node.name}RelationInput`,
+              delete: `${node.name}DeleteInput`,
+          }
+        : {};
 
+    if (schemaComposer.has(`${node.name}ConnectOrCreateInput`)) {
+        relationFields.connectOrCreate = `${node.name}ConnectOrCreateInput`;
+    }
```
darrellwarde, line 63: In what scenario would just checking for `node.relationFields.length` not be sufficient here?

packages/graphql/src/translate/cypher-builder/build-merge-statement.ts
```diff
@@ -0,0 +1,120 @@
...
+    } else {
+        leftStatement = buildNodeStatement({
+            ...leftNode,
+            context,
+        });
+    }
+
+    const mergeNodeStatement = joinStatements(["MERGE ", leftStatement], "");
+
+    const onCreateSetQuery = onCreateStatements.length > 0 ? "ON CREATE SET" : "";
```
darrellwarde, line 107:
```suggestion
    const onCreateSetQuery = onCreateStatements.length ? "ON CREATE SET" : "";
```

darrellwarde (commented):
Just had an initial read through of code - still need to check the documentation and give all of this a go! But left some general comments.

Leaving comments for multiple lines broke at one point during review, so sorry if it's not super obvious what a comment is for.

packages/graphql/src/utils/utils.ts
```diff
@@ -28,3 +28,30 @@ export function isString(value: unknown): value is string {
...
     return isInt(value);
 }
+
+/** Joins all strings with given separator, ignoring empty or undefined statements */
+export function joinStrings(statements: string | Array<string | undefined>, separator = "\n"): string {
+    return filterTruthy(arrayfy(statements)).join(separator);
+}
+
+/** Makes sure input is an array, if not it turns into an array (empty array if input is null or undefined) */
+export function arrayfy<T>(raw: T | Array<T> | undefined | null): Array<T> {
```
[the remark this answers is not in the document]
angrykoala: Mmm, agree that this may not be the best name, but assert may be confusing, as an element is converted into an array if it is not, instead of failing.

what about `asArray` or `toArray`?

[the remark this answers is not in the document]
darrellwarde: Either of those sound good!

packages/graphql/src/translate/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+
+import { joinStrings, isString, arrayfy, filterTruthy } from "../utils/utils";
+import { CypherStatement, CypherParams } from "./types";
+
+/** Wraps a string in a CALL statement */
+export function wrapInCall(statement: string, withVarName: string, returnStatement = "RETURN COUNT(*)"): string {
+    const withStatement = `WITH ${withVarName}`;
+    return joinStrings([withStatement, "CALL {", withStatement, statement, returnStatement, "}"]);
```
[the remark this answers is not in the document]
angrykoala: I'm happy to add this, although keep in mind that this will not make the output completely prettified as `statement` may be a multi-line statement, only the first line will be indented.

IMO We should consider separating the query composition and its visual aspect at some point (e.g prettify at the end)

packages/graphql/src/translate/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { joinStrings, isString, arrayfy, filterTruthy } from "../utils/utils";
+import { CypherStatement, CypherParams } from "./types";
+
+/** Wraps a string in a CALL statement */
+export function wrapInCall(statement: string, withVarName: string, returnStatement = "RETURN COUNT(*)"): string {
```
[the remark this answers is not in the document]
angrykoala: sure? prepositions are not my strong suit but I always though this was wrap _[something]_ **in**

packages/graphql/src/translate/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+        return getStatementString(statement);
+    });
+    const statementsParams = statementsArray.reduce((acc, statement) => {
+        return { ...acc, ...getStatementParams(statement) };
+    }, {} as CypherParams);
+    return [joinStrings(statementsStrings, separator), statementsParams];
+}
+
+/** Serializes object into a string for Cypher objects */
+export function serializeObject(fields: Record<string, string | undefined | null>): string {
```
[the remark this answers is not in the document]
angrykoala: And you didn't like **arrayfy**? :cry: 

I could go both ways here

packages/graphql/src/translate/utils.ts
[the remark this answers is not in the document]
darrellwarde: Honestly, you could be right: https://english.stackexchange.com/questions/152759/what-preposition-should-be-used-with-the-wrap-verb

packages/graphql/src/translate/utils.ts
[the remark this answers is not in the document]
darrellwarde: Stringify is an industry accepted term, you made up "arrayfy" in this branch 😉

packages/graphql/src/translate/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+    return `{ ${Object.entries(fields)
+        .map(([key, value]): string | undefined => {
+            if (value === undefined || value === null || value === "") return undefined;
+            return `${key}: ${value}`;
+        })
+        .filter(Boolean)
+        .join(", ")} }`;
```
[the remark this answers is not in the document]
angrykoala: Careful here. `JSON.stringify()` is intended to convert an object into its JSON representation, here we are trying to create a cypher friendly representation of the object:

```ts
JSON.stringify({name: 'Arthur'}) // {"name":"Arthur"}   <- MATCH (n: Node {"name": "Arthur"}) will fail
stringifyObject({name: '"Arthur"'}) // {name: "Arthur"}
```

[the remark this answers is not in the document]
danstarns: What about if one of the params is one of the values: `""` or "null" ? 

```
movies(where: { title: "" })
```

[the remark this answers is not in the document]
darrellwarde: I see, didn't catch that nuance!

packages/graphql/src/translate/field-aggregations/apoc-run-utils.ts
```diff
@@ -8,16 +9,6 @@ export function wrapApocRun(query: string, params: Record<string, string> = {}):
     return `head(apoc.cypher.runFirstColumn(" ${escapedQuery} ", ${serializedParams}))`;
```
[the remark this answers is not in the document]
angrykoala: Ideally no, `escapeQuery` handles escaping of double quotes in the inside query, regardless of its contents. If multiple .runFirstColumn are used nested, this method should be called n times, correctly escaping every quote n times, so:

```ts
const query1=wrapApocRun("'mything'") -> head(apoc.cypher.runFirstColumn(" \\'mything\\' ")
const query2=wrapApocRun(query1) ->head(apoc.cypher.runFirstColumn("  head(apoc.cypher.runFirstColumn(\\" \\\\'mything\\\\' \\") ")
```
In fact this was the problem that this piece of code was intended to solve originally in field-level aggregations. If you check #587 you'll see that could be solved by using this method, as `connection[0].replace(/"/g, '\\"')` is how `escapeQuery` works

packages/graphql/src/translate/cypher-builder/utils.ts
angrykoala: good catch

packages/graphql/src/utils/test/builders/relation-field-builder.ts
angrykoala: Consistency and flexibility.

* If we have a few builders already, I would keep using them for the rest of dummy data, just to make all work the same
* A builder allows us slightly more flexibility for adding/modifying dummy data or changing its structure (quite common in tests) and I found them really useful in the past. It is true that we may not need that in the moment, and we could go with functions, in this case, I would advocate for moving all of these to functions (again, consistency) rather than mixing different types of dummy data generators.

I'm happy to do a review on the builders that exists at the moment to decide whether to go for functions or builders (or other option) but I would really keep them consistent

packages/graphql/src/translate/utils.ts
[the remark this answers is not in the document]
angrykoala: It is called innovation :laughing:

packages/graphql/src/schema/resolvers/update.ts
angrykoala: connectOrCreate only exists for relations with `@unique` fields.

On a related note, I think it is a bit more safe to check for the existence of a type rather than duplicating the checks used for creating the type, if in a couple of months, `connectOrCreate` require other conditions, we won't need to change that in 2 separate places, as the type creation holds the single source of truth of whether the type exists or not

packages/graphql/src/schema/create-relationship-fields/index.ts
```diff
@@ -427,6 +449,31 @@ function createRelationshipFields({
                         [n.name]: disconnect,
                     });
                 }
+
+                if (n.uniqueFields.length > 0) {
```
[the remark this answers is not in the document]
angrykoala: Is this really a better alternative in terms of readability?

There is also the nuance of what happens if length is not a number, but that is a bit farfetched

docs/modules/ROOT/pages/type-definitions/constraints.adoc
```diff
@@ -0,0 +1,80 @@
+[[type-definitions-constraints]]
```
darrellwarde, line 1: I'm confused as to why this is showing as a new file - this already exists in `master`?

docs/modules/ROOT/pages/mutations/create.adoc
```diff
@@ -96,3 +96,48 @@ mutation {
...
 
 This will create a User with name "John Doe", an introductory Post, both of which will be returned with their autogenerated IDs.
+
+== `connectOrCreate` relations
+If a related node has a `@unique` directive defined, `connectOrCreate` can be used in a
+nested create to perform a `MERGE` operation on the related node, creating a new
+relation and the related node if it doesn't exists.
+
+Consider the following type definition:
+[source, graphql, indent=0]
```
darrellwarde, line 106:
```suggestion
Consider the following type definitions:

[source, graphql, indent=0]
```

docs/modules/ROOT/pages/mutations/create.adoc
```diff
@@ -96,3 +96,48 @@ mutation {
 ----
 
 This will create a User with name "John Doe", an introductory Post, both of which will be returned with their autogenerated IDs.
+
+== `connectOrCreate` relations
+If a related node has a `@unique` directive defined, `connectOrCreate` can be used in a
+nested create to perform a `MERGE` operation on the related node, creating a new
+relation and the related node if it doesn't exists.
```
darrellwarde, line 103: Pointing out redundant new lines, and also the fact that the `@id` directive should also work to enable `connectOrCreate` operations.

```suggestion
== `connectOrCreate` relations

If a related node has at least one `@unique` or `@id` (without `unique` argument set to `false`) directive defined, `connectOrCreate` can be used in a nested create to perform a `MERGE` operation on the related node, creating a new relation and the related node if it doesn't exists.
```

docs/modules/ROOT/pages/mutations/create.adoc
```diff
@@ -96,3 +96,48 @@ mutation {
...
+[source, graphql, indent=0]
+----
+type Actor {
+    name: String!
+    movies: [Movie] @relationship(type: "ACTED_IN", direction: OUT)
+}
+
+type Movie {
+    title: String
+    id: String! @unique
```
darrellwarde, line 115: I think documenting this example like this encourages "bad" behaviour, as the `@id` directive also leads to a unique node property constraint being added, with the added bonus of ID values being generated automatically. Perhaps give an example that uses both the `@id` and `@unique` directives, and then point out that you can use either of them to match the node to connect to.

```suggestion
    id: String! @id
```

darrellwarde (changes requested):
Just a review of documentation this time around, a few broad comments:

- I'm unsure if it's you or your tooling, but I'm pretty sure all of these new lines in the middle of sentences are going to lead to funky formatting when published. I've pointed out a couple, but they're all over the place. New lines should only be in when actually intended, i.e. new paragraphs.
- I don't think the "`connectOrCreate` relations" headers sit well in the "Create" and "Update" pages, especially as the two pages are essentially duplicates. This isn't on you, I just don't think there's presently a good home for them. I think we would benefit from a page within "Mutations" called "Nested operations". Within that page, there should be a page for each nested operation, at the top of which it is documented which Mutation type they are available under. And then also, from each Mutation page, add a matrix of all of the nested operations it supports with links. I think this will be really tidy, and give these sections a better home. I'd be happy to look into this, I love myself some documentation changes!

docs/modules/ROOT/pages/type-definitions/constraints.adoc
angrykoala: mmm, I think there has been a merge issue, as this file exists with a different name. I'll remove this version as it should not have any relevant additions

docs/modules/ROOT/pages/mutations/create.adoc
angrykoala: Regarding new lines:

These were not produced by my tooling, I wrote it like that just to improve readability as plain text (happy to remove them for consistency). I think there is no need to worry regarding formatting, I already checked and adoc works like markdown, in that newlines are not taken into account when formatting (with the idea of making the text legible in plain text without automatic word wrapping) only multiple new lines are considered as separate paragraphs

angrykoala:
@darrellwarde Regarding your comment on duplicate docs I'm slightly torn.
While it makes sense to move it to a separate doc, having them in the same page makes it easier (in my opinion) to follow for people learning.

I think it is time to make some changes to the docs structure, as some information is hard to find in some cases. We need a bit of further discussion on this.

docs/modules/ROOT/pages/mutations/create.adoc
darrellwarde: We need to choose one approach, and that approach should be enforced by automatic linting and formatting. At the moment we have half the team doing one thing and half the other, so let's agree on an approach and find the right tool to do it for us.

oskarhane: I was taught that you should format md and asciidoc by having a max line width. One of the reasons should be easier diffing if I remember correctly. And the other one that we don't have to rely on the editors wrapping lines to make it easy to read.

Single newlines has no effect in the output.  
If you actually want a newline in the output you should add two newlines or insert a double space at the end of a line.

Setting up automatic linting and formatting is a great idea, but I would vote on inserting newlines rather than have the editors wrap lines.

darrellwarde: Happy to run with that, know this was the case for markdown but was unsure on the effects of newlines with AsciiDoc.

It's a draft document, but an interesting recommendation from AsciiDoc here: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

packages/graphql/src/schema/create-relationship-fields/index.ts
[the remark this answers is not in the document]
darrellwarde: More just consistency throughout the codebase, I think this is the first time this has been done like this.

Your second sentence is extremely farfetched. 😉

packages/graphql/src/translate/connect-or-create/create-connect-or-create-and-params.ts
```diff
@@ -0,0 +1,186 @@
...
+                refNode,
+                context,
+            });
+        }
+    );
+
+    return joinStatements(statements);
+}
+
+function createConnectOrCreateSubquery({
```
darrellwarde, line 70: `createConnectOrCreatePartialStatement`? Let's be careful not to use "subquery" where a function is not actually producing a subquery.

packages/graphql/src/translate/cypher-builder/build-merge-statement.ts
```diff
@@ -0,0 +1,123 @@
...
```
darrellwarde, line 110: Let's add a new line between `ON CREATE` and `SET`,  makes it more consistent with the style established in the official docs: https://neo4j.com/docs/cypher-manual/current/clauses/merge/#merge-merge-with-on-match-setting-multiple-properties

darrellwarde (changes requested):
I feel like this is really close now, just down to super fussy things.

Let's not worry too much about documentation, and do a bit of a general documentation refactor before release.

packages/graphql/tests/tck/tck-test-files/connections/connect-or-create/connect-or-create-unions.test.ts
```diff
@@ -0,0 +1,243 @@
...
+            CREATE (this0:Actor)
+            SET this0.name = $this0_name
+            MERGE (this0_actedIn_Movie_connectOrCreate0:Movie { isan: $this0_actedIn_Movie_connectOrCreate0_node_isan })
+            ON CREATE SET
+            this0_actedIn_Movie_connectOrCreate0.title = $this0_actedIn_Movie_connectOrCreate0_on_create_title,
+            this0_actedIn_Movie_connectOrCreate0.isan = $this0_actedIn_Movie_connectOrCreate0_on_create_isan
+            MERGE (this0)-[this0_relationship_this0_actedIn_Movie_connectOrCreate0:ACTED_IN]->(this0_actedIn_Movie_connectOrCreate0)
+            ON CREATE SET
+            this0_relationship_this0_actedIn_Movie_connectOrCreate0.screentime = $this0_relationship_this0_actedIn_Movie_connectOrCreate0_on_create_screentime
+            MERGE (this0_actedIn_Series_connectOrCreate0:Series { isan: $this0_actedIn_Series_connectOrCreate0_node_isan })
```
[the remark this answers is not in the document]
angrykoala: If a larger refactor is needed we may want to discuss this a bit further and create a separate task for this?

darrellwarde (approved):
I think this is about as reviewed as it can be, let's merge it! 🥳

Issue #559, opened 2021-10-31
Error: Cannot read property 'args' of undefined in subscriptions with args


Dajust:
**Describe the bug**
When you publish a subscription that accepts arguments, the above error is thrown.

**Type definitions**
```
type Query {
  greet: String
}
type Subscription {
  newMessage(to: String!): String
}
```

**Resolver function**

```
{
   Subscription: {
    newMessage: {
      subscribe: () => pubsub.asyncIterator('newMessage'),
    },
}
```
**To Reproduce**

1. Run a server with the above Subscription type definition
2. Subscribe for `newMessage` event ( on Apollo Studio/Playground )
3. Make the server Publish a `newMessage` event.
5. Remove the arguments from the Subscription type definition
6. Repeat step 3
7. See there is NO error

**Expected behavior**
Neo4jGraphQL should be able to publish a Subscription that accepts arguments on its type definition. But instead of doing so, it currently throws the `Error: Cannot read property 'args' of undefined`

**System (please complete the following information):**

-   OS: macOS
-   Version: "@graphql-tools/schema": "^8.3.1",
    "@neo4j/graphql": "^2.3.0",
    "subscriptions-transport-ws": "^0.9.19",
    "apollo-server-core": "^3.4.0",
    "apollo-server-express": "^3.4.0",
-   Node.js version: "16.6.0"

**Additional context**
It might also help to know that everything works well when you replace Neo4jGraphQL schema with `makeExecutableSchema`.

```js
const neoSchema = new Neo4jGraphQL({
  driver,
  typeDefs,
  resolvers,
});
const schema = neoSchema.schema;
// Everything works if we swtich back to  `makeExecutableSchema`, but that's old school 🤧🥴
// const schema = makeExecutableSchema({ typeDefs, resolvers })
```

KleptoKat:
Have you got a sample of your newMessage resolver?

Dajust:
Thanks, @KleptoKat, updated with the resolver function now.

danstarns:
I wrote an int test for this in `/packages/graphql/tests/integration/issues/` and it works before and after, am I missing something or does this not fix the associated issue? 

```ts
import { Driver } from "neo4j-driver";
import { createSourceEventStream, parse } from "graphql";
import { gql } from "apollo-server";
import { generate } from "randomstring";
import neo4j from "../neo4j";
import { Neo4jGraphQL } from "../../../src/classes";
import { PubSub } from "graphql-subscriptions";

[...]

PR #576, opened 2021-11-10
fix/559: add subscription type fields to resolve tree field resolver

KleptoKat:
# Description

Adds subscriptions to the resolve tree field resolver. This fixes the args is undefined issue.

I didn't see any tests for getNeo4jResolveTree so I just left it as is, I hope that's OK. Note that this issue was first resolved in https://github.com/neo4j/graphql/pull/515 I've just added the solution to a separate PR to fix the relevant issue.

# Issue

https://github.com/neo4j/graphql/issues/559

# Checklist


danstarns:
Hey, could you please provide a test case that shows that the linked issues are no longer happening?

KleptoKat:
This was a little tricky to test but I _think_ this should be sufficient. It doesn't have total coverage of that file but it's better than nothing.

danstarns:
I wrote an int test for this in `/packages/graphql/tests/integration/issues/` and it works before and after your changes, am I missing something? Is the error happening somewhere else? Is this test testing the right thing? 

```ts
import { Driver } from "neo4j-driver";
import { createSourceEventStream, parse } from "graphql";
import { gql } from "apollo-server";
import { generate } from "randomstring";
import neo4j from "../neo4j";
import { PubSub } from "graphql-subscriptions";

describe("https://github.com/neo4j/graphql/issues/559", () => {
    let driver: Driver;

    beforeAll(async () => {
        driver = await neo4j();
    });

    afterAll(async () => {
        await driver.close();
    });

    test("should not throw when subscription is subscribed to", async () => {
        const pubsub = new PubSub();

        const to = generate({
            charset: "alphabetic",
        });

        const typeDefs = gql`
            type Query {
                greet: String
            }

            type Subscription {
                newMessage(to: String!): String
            }
        `;

            typeDefs,
            resolvers: {
                Subscription: {
                    newMessage: {
                        subscribe: () => pubsub.asyncIterator("newMessage"),
                    },
                },
            },
        });

        const query = `
            subscription {
                newMessage(to: "${to}")
            }
        `;

        const gqlResult = await createSourceEventStream(neoSchema.schema, parse(query), null, { driver });

        setTimeout(() => {
            pubsub.publish("newMessage", to);
        }, 1000);

        // @ts-ignore
        const next = await gqlResult.next();
        expect(next.value).toEqual(to);
    });
});
```

KleptoKat:
@danstarns I tried debugging the test that you wrote but `createSourceEventStream` & parse never seems to touch the inside of the function within `createWrappedSchema` and therefore `getNeo4jResolveTree` is never called. I am not really sure why nor do I know of an alternative that would call the wrapped resolver.

When I create a separate application and link the build result to this one, it _does_ hit `getNeo4jResolveTree` which is where the error occurs. Is there some other way that you know of to test this that would call the wrapped resolver?

darrellwarde:
The fix in question here looks like it _should_ be there, and @angrykoala has expressed in the past that it is difficult to test subscription functionality, so I think based on that, we should be looking to merge.

KleptoKat:
> The fix in question here looks like it _should_ be there, and @angrykoala has expressed in the past that it is difficult to test subscription functionality, so I think based on that, we should be looking to merge.

Awesome. Might be worthwhile to spin up an _actual_ apollo subscriptions server in an integration test and test that. This PR blocks subscriptions with args in it though so it should probably be in a different PR. I realize that may make for some fragile tests but it _is_ something we should test for.

Also, what's up with the CI/CD? it seems to fail every once in moon for connection reasons.

darrellwarde:
> > The fix in question here looks like it _should_ be there, and @angrykoala has expressed in the past that it is difficult to test subscription functionality, so I think based on that, we should be looking to merge.
> 
> 
> 
> Awesome. Might be worthwhile to spin up an _actual_ apollo subscriptions server in an integration test and test that. This PR blocks subscriptions with args in it though so it should probably be in a different PR. I realize that may make for some fragile tests but it _is_ something we should test for.
> 
> 
> 
> Also, what's up with the CI/CD? it seems to fail every once in moon for connection reasons.

Apologies about the CI/CD, some issues with APOC and the 4.3.7 release mean they've been a bit flakey - we should be back to normal soon.

darrellwarde:
> Adds subscriptions to the resolve tree field resolver. This fixes the args is undefined issue.

@KleptoKat, where you have mentioned this in the main PR body, how did you validate that it fixes it? Did you try with Apollo Server?

KleptoKat:
> > Adds subscriptions to the resolve tree field resolver. This fixes the args is undefined issue.
> 
> @KleptoKat, where you have mentioned this in the main PR body, how did you validate that it fixes it? Did you try with Apollo Server?

Yeah, I spun up a separate Apollo server and linked the local build to it. This could be tested if we spun up another apollo instance and used websockets with it, perhaps? Or if we could somehow pass a blank context to it without the resolve tree... I haven't spent a ton of time trying to figure out how to test this properly, honestly, but it does work with a manual test. I have an application which I am working on separately and using the existing subscriptions PR + trying out the master branch every once in awhile. I noticed this issue initially when I was working on the subscription features and included the fix in the subs PR.

My setup included Apollo, Fastify (although express could be used), and "SubscriptionServer" from subscriptions-transport-ws. Perhaps we could reverse engineer `server.ts` in `subscriptions-transport-ws` to create a better testing method for subscriptions that doesn't rely on websockets? Alas, this is why I didn't dig too deep into this test...

darrellwarde:
> > > Adds subscriptions to the resolve tree field resolver. This fixes the args is undefined issue.
> > 
> > 
> > @KleptoKat, where you have mentioned this in the main PR body, how did you validate that it fixes it? Did you try with Apollo Server?
> 
> Yeah, I spun up a separate Apollo server and linked the local build to it. This could be tested if we spun up another apollo instance and used websockets with it, perhaps? Or if we could somehow pass a blank context to it without the resolve tree... I haven't spent a ton of time trying to figure out how to test this properly, honestly, but it does work with a manual test. I have an application which I am working on separately and using the existing subscriptions PR + trying out the master branch every once in awhile. I noticed this issue initially when I was working on the subscription features and included the fix in the subs PR.
> 
> My setup included Apollo, Fastify (although express could be used), and "SubscriptionServer" from subscriptions-transport-ws. Perhaps we could reverse engineer `server.ts` in `subscriptions-transport-ws` to create a better testing method for subscriptions that doesn't rely on websockets? Alas, this is why I didn't dig too deep into this test...

Manual testing is good enough for now given how difficult it is to set up comprehensive testing for subscriptions - that is a task in itself. Just wanted a bit of an overview of the manual testing you have performed, sounds good to me! 🙂

darrellwarde (approved):

PR #587, opened 2021-11-16
Querying DateTime in the edge of a connection throws an error

mhlz:
The error that gets thrown:

```
Neo4jError: Invalid input 'iso_zoned_date_time': expected
  "!="
  "%"
  ")"
  "*"
  "+"
  ","
  "-"
  "."
  "/"
  ":"
  "<"
  "<="
  "<>"
  "="
  "=~"
  ">"
  ">="
  "AND"
  "CONTAINS"
  "ENDS"
  "IN"
  "IS"
  "OR"
  "STARTS"
  "XOR"
  "["
  "^" (line 6, column 97 (offset: 375))
"WITH collect({ node: { birthday: apoc.date.convertFormat(toString(this_movies_actor.birthday), "iso_zoned_date_time", "iso_offset_date_time") } }) AS edges"
```

Generated Cypher:

```
MATCH (this:Genre)
WHERE this.id = $this_id
RETURN this { movies: [ (this)-[:MOVIE]->(this_movies:Movie)   | this_movies { actorsConnection: apoc.cypher.runFirstColumn("CALL {
WITH this_movies
MATCH (this_movies)-[this_movies_actor_relationship:ACTOR]->(this_movies_actor:Actor)
WITH collect({ node: { birthday: apoc.date.convertFormat(toString(this_movies_actor.birthday), "iso_zoned_date_time", "iso_offset_date_time") } }) AS edges
RETURN { edges: edges, totalCount: size(edges) } AS actorsConnection
} RETURN actorsConnection", { this_movies: this_movies }, false) } ] } as this
```

# Description

Because runFirstColumn gets called with a string there are numerous places in the current code where the library will generate invalid cypher, because of unescaped quotation marks. I wrote a quick test to demonstrate one of those places, but I'm fairly certain there are more. The fix I also implemented would fix at least this case, however, I feel like it might not be the best solution. I am unsure however, what a better solution would look like. There might have to be some sort of way to track if the cypher that is generated is in quotation marks or not (and maybe even how many levels deep if that is an issue?).

I'm open for suggestions as I would love to get this bug fixed.

# Checklist

### DIFF
--- a/packages/graphql/src/translate/create-projection-and-params.ts
+++ b/packages/graphql/src/translate/create-projection-and-params.ts
@@ -541,7 +541,7 @@ function createProjectionAndParams({
                 : `{ ${chainStr}: ${chainStr} }`;
 
             res.projection.push(
-                `${field.name}: apoc.cypher.runFirstColumn("${connection[0].replace(/"/g, '\\"')} RETURN ${
+                `${field.name}: apoc.cypher.runFirstColumn("${connection[0].replace(/("|')/g, "\\$1")} RETURN ${
                     field.name
                 }", ${runFirstColumnParams}, false)`
             );
--- a/packages/graphql/src/translate/create-projection-and-params.ts
+++ b/packages/graphql/src/translate/create-projection-and-params.ts
@@ -541,7 +541,9 @@ function createProjectionAndParams({
                 : `{ ${chainStr}: ${chainStr} }`;
 
             res.projection.push(
-                `${field.name}: apoc.cypher.runFirstColumn("${connection[0]} RETURN ${field.name}", ${runFirstColumnParams}, false)`
+                `${field.name}: apoc.cypher.runFirstColumn("${connection[0].replace(/"/g, '\\"')} RETURN ${
+                    field.name
+                }", ${runFirstColumnParams}, false)`
             );
             res.params = { ...res.params, ...connection[1] };
             return res;
--- a/packages/graphql/tests/integration/issues/date-in-edge.test.ts
+++ b/packages/graphql/tests/integration/issues/date-in-edge.test.ts
@@ -24,7 +24,7 @@ import { generate } from "randomstring";
 import neo4j from "../neo4j";
 import { Neo4jGraphQL } from "../../../src/classes";
 
-describe("issue", () => {
+describe("587: Dates in edges can cause wrongly generated cypher", () => {
     let driver: Driver;
 
     beforeAll(async () => {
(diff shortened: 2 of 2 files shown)


packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -541,7 +541,9 @@ function createProjectionAndParams({
                 : `{ ${chainStr}: ${chainStr} }`;
 
             res.projection.push(
-                `${field.name}: apoc.cypher.runFirstColumn("${connection[0]} RETURN ${field.name}", ${runFirstColumnParams}, false)`
+                `${field.name}: apoc.cypher.runFirstColumn("${connection[0].replace(/"/g, '\\"')} RETURN ${
```
angrykoala, line 544: Please, note that single quotes `'` needs to be escaped as well, replace should cover both:
```suggestion
                `${field.name}: apoc.cypher.runFirstColumn("${connection[0].replace(/("|')/g, "\\$1")} RETURN ${
```

packages/graphql/tests/integration/issues/date-in-edge.test.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+import { Driver } from "neo4j-driver";
+import { graphql } from "graphql";
+import { gql } from "apollo-server";
+import { generate } from "randomstring";
+import neo4j from "../neo4j";
+import { Neo4jGraphQL } from "../../../src/classes";
+
+describe("issue", () => {
```
angrykoala, line 27: Please, add a more descriptive name to this test suite

packages/graphql/tests/integration/issues/date-in-edge.test.ts
```diff
@@ -0,0 +1,113 @@
...
+            `);
+
+            const result = await graphql({
+                schema: neoSchema.schema,
+                source: query,
+                contextValue: { driver },
+            });
+
+            if (result.errors) {
+                console.log(JSON.stringify(result.errors, null, 2));
```
angrykoala, line 105: This log is unnecessary, the expectation should already provide enough information

packages/graphql/src/translate/create-projection-and-params.ts
angrykoala: FYI, We already have a couple of utilities (`escapeQuery` and `wrapApocRun`) to take care of this issue in other places, however, these are not generic enough at the moment to be used here directly.

angrykoala (changes requested):
Good catch here, as you pointed out, there are several uses of apoc.cypher.runFirstColumn (or similar) that may be susceptible of this same issue. We are already tracking this issue and try to fix these for other cases

angrykoala (approved):

PR #588, opened 2021-11-16
RFC: Global Node Resolution

darrellwarde:
# Description

Intended to supersede #285 for final discussions on the solution for Global Node Resolution before implementation.

A change in circumstances (mainly, the introduction of unique node property constraint management) means that we can likely take more efficient approaches now than we could have done at the time this was initially discussed.


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-588.surge.sh

docs/rfcs/01-global-object-identification/global-object-identification.md
```diff
@@ -0,0 +1,105 @@
...
+#### Implementation of the `Node` interface
+
+In <https://github.com/neo4j/graphql/pull/285>, there was talk of using the `@node` directive for this purpose. However, it now has other purposes, which means that using this would mean the directive plus an argument within, at which point it is less boilerplate to explicity add `implements Node` and the necessary `id` field to the type definition. So a `Movie` type with global object identification should be defined using:
+
+```graphql
+type Movie implements Node {
+    id: ID!
+    title: String!
+}
+```
```
darrellwarde, line 48: The main point of contention for me in regards to this proposal is `implements Node` versus an argument in the `@node` directive.

An example of what the latter might look like:

```graphql
type Movie @node(relay: true) {
    title: String!
}
```

Advantages of `implements Node` are mainly in its explicitness - what the user sees is what they get. This is why I opted for it in the current proposal.

Whilst an argument in the `@node` directive is maybe a bit less boilerplate, there is some "magic" in the addition of the `id` field, and it's a less familiar syntax for those already familiar with Relay.

docs/rfcs/01-global-object-identification/global-object-identification.md
```diff
@@ -0,0 +1,105 @@
+# Relay Object Identification
```
darrellwarde, line 1:
```suggestion
# Global Object Identification
```

docs/rfcs/01-global-object-identification/global-object-identification.md
litewarp: I'm completely on the fence about this. On the one hand, the naming collision is messy, especially if you don't live and breath schemas. The user needs to know that "Node" means two different things in different contexts -- if "implements" is used, it's a global relay id in the graphql layer context; if it's the directive "@node" then it's a mapping to the neo4j db. 

In an ideal world, you could rename the interface type to something `RelayNode` but given the strict rules of relay, that's not possible. I also don't imagine you want to rename the `@node` directive.

I like your point that `implements Node` explicitly tells you what you're getting.  My concern is that if you don't live and breath schemas, the naming collision will be very confusing unless explicitly spelled out. This could perhaps be an argument for adding a parameter to the `@node` directive -- it's the single source of truth for describing how to resolve the entity in both the db and the graphql layer. 

In terms of syntax, I think the second option is probably cleaner. Just my two cents. 

```graphql
type Documentary implements Node @node(
  label: "Movie", 
  additionalLabels: "Documentary"
) {
  //
}

type Documentary @node(
  label: "Movie",
  additionalLabels: "Documentary",
  global: true
) {
  //
} 
```

darrellwarde: Thanks for your thoughts here too @litewarp, it's great to still see you around and making such valuable contributions to the project! I think definitely either the second option which you have proposed, or what is currently proposed with explicit `implements Node` are our two choices. Just need to have some discussions and settle on one!

Appreciate it has taken a while to get around to this. Rough plan of action is to close #285 (probably in an unrecoverable state of merge conflicts now I imagine), but use the work done there by you as the basis for the solution we agree on here.

litewarp:
Hey @darrellwarde! Just wanted to check in and see if there is a way I can help push this forward. I was thinking of putting together a small wrapper to use as a plugin with graphql-tools to try and implement this feature, but it'd probably be less work to push the PR forward. Has there been any further discussion in which approach you all would like to take w/r/t the directive?

I feel like regardless of the API design, it'd be nice to get this functionality merged soon. Or maybe I just speak for the relay weirdos who would love to see a fully-compatible package (there are dozens of us! dozens!).

Cheers!

docs/rfcs/01-global-object-identification/global-object-identification.md
darrellwarde: @angrykoala and @danstarns, can we get some thoughts here on the following options?

@litewarp is very right in that we need to make some progress on this RFC so that it can be implemented, either by ourselves or the community.

## Option 1

Explicit implementation of the `Node` interface. `id` field must be defined by users to fulfil the interface contract.

**Pros:**
- Very clear for users who are familiar with Relay (which really is our target audience for this feature)
- No confusion about any generated `id` field, where it came from, why it is there or what it returns
- Clear separation of concerns for `Node` and `@node` (see related con)

**Cons:**
- More boilerplate
- Potential confusion of `Node` vs `@node` (see related pro)

```gql
type Movie implements Node {
    id: ID!
    title: String!
}
```

## Option 2

Adding a `global` argument to the existing `@node` directive. @litewarp, would your vision for this approach be that `id` is automatically added, or manually specified by users?

Pros:
- Minimum amount of boilerplate for implementation

Cons:
- `@node` is no longer a single purpose directive, sets a bad precedent

```gql
type Movie @node(global: true) {
    id: ID! # Would this need to be defined by user or automatically added?
    title: String!
}
```

darrellwarde: Personally, I think I prefer option 1. I think the only users who are going to want to turn this on are hardcore Relay users, who will be familiar with the syntax and the underlying behaviours.

darrellwarde:
> Hey @darrellwarde! Just wanted to check in and see if there is a way I can help push this forward. I was thinking of putting together a small wrapper to use as a plugin with graphql-tools to try and implement this feature, but it'd probably be less work to push the PR forward. Has there been any further discussion in which approach you all would like to take w/r/t the directive?
> 
> I feel like regardless of the API design, it'd be nice to get this functionality merged soon. Or maybe I just speak for the relay weirdos who would love to see a fully-compatible package (there are dozens of us! dozens!).
> 
> Cheers!

Hey @litewarp, you are absolutely right, we need to get a move-on with this piece of work! 

It's something I would like to see done too - you're not alone! Let's aim for an outcome on what this should look like this week, and then if you fancy the task of implementing the final solution, then we would welcome that with open arms! I wouldn't like to see you have to implement this with an external tool, this should be part of the library.

docs/rfcs/01-global-object-identification/global-object-identification.md
angrykoala: I'm also more inclined with option 1, although I'm slightly concerned about the `Node` vs `@node` naming collision

darrellwarde:
@litewarp, we're going to run with an argument in the `@node` directive - see the changes to the RFC to see how this should function. This _should_ be enough to run with for now, the only thing we're slightly unsure about is the naming of the `global` argument which we might want to tweak down the line.

Please have a review of the PR and note down anything we may have missed, but all going well, we can merge this PR and work can begin whenever you or we are ready! 🙂

litewarp:
Thanks @darrellwarde. One quick question about the RFC. In the implementation section, can you clarify what the line below means?

`- There are no field with either `@id` or `@unique` in the type`

Does this mean that only one unique field is permitted by node? So I couldn't, for example, have a type like the following?

```graphql
type Person @node(global: true) {
  email: String! @unique
}
```

If not, is this a db issue, or a design choice?

dmoree:
I read that as if the flag `global` is set to `true` there must exist at least one field that is either an `@id` or `@unique` field. By going with an argument in `@directive` it is not guaranteed that there exists such a field.

litewarp:
That makes more sense! Thanks @dmoree

litewarp:
Another question -- what kind of actions should be available using the `id` field? 

Sort and aggregations don't make sense. 

A where filter argument does, but it is unclear whether it should be anything more than an equality or presence check. 

What do you all think?

```graphql

input MovieWhere {
  ...
  // include
  id: ID
  id_NOT: ID
  id_IN: [ID]
  id_NOT_IN: [ID]

  // proposal to exclude
  id_CONTAINS: ID
  id_ENDS_WITH: ID 
  id_NOT_CONTAINS: ID
  id_NOT_ENDS_WITH: ID
  id_STARTS_WITH: ID
  id_NOT_STARTS_WITH: ID
  ...
}

```

darrellwarde:
Precisely @dmoree. @litewarp, essentially the solution needs to:

If the `global` argument is set to `true` in the `@node` directive, ensure that there is at least one field with `@id` or `@unique` in the type. If there is zero, throw an error. If there is more than one, choose one arbitrarily (perhaps take preference to `@id` over `@unique`?).

Issue #583, opened 2021-11-13
Cannot resolve properties from common interface with new interface


KleptoKat:
**Describe the bug**

When you add a separate interface to a type it is resolved with inline fragments but its properties are not projected. See below:

**Type definitions**
If applicable, add your type definitions to help us reproduce the bug.

```graphql
interface Show {
  title: String
}

interface Awardable {
  awardsGiven: Int!
}

type Actor implements Awardable {
  name: String
  awardsGiven: Int!
  actedIn: [ Show! ] @relationship(type: "ACTED_IN", direction: OUT)
}

type Movie implements Show & Awardable {
  title: String
  awardsGiven: Int!
}

type Series implements Show & Awardable {
  title: String
  awardsGiven: Int!
}

type ShortFilm implements Show {
  title: String
}
```

**To Reproduce**

1. Run a server with the above type definitions
2. Run the following neo4j query:

```
CREATE (a:Actor { name:  "aaa", awardsGiven: 0 })
CREATE (a)-[:ACTED_IN]->(:Series { title: 'stranger who', awardsGiven: 2 })
CREATE (a)-[:ACTED_IN]->(:Movie { title: 'doctor things', awardsGiven: 42 })
CREATE (a)-[:ACTED_IN]->(:ShortFilm { title: 'all too well' })
```

3. Run the following graphql query:
```graphql

query shows {
  actors {
    name
    __typename
 	  actedIn {
      title
      ...on Awardable {
        awardsGiven
      }
    }  
  }
}


```

4. See error: `Cannot return null for non-nullable field Movie.awardsGiven.`

**Expected behavior**

Because awardsGiven is set for both the series and the movie, we would expect it to return but it doesn't so we get an error. When we debug the cypher query, it looks something like this:

```
MATCH (this:Actor)
WITH this
CALL {
WITH this
MATCH (this)-[:ACTED_IN]->(this_Movie:Movie)
RETURN  { __resolveType: "Movie", title: this_Movie.title } AS actedIn
UNION
WITH this
MATCH (this)-[:ACTED_IN]->(this_Series:Series)
RETURN  { __resolveType: "Series", title: this_Series.title } AS actedIn
UNION
WITH this
MATCH (this)-[:ACTED_IN]->(this_ShortFilm:ShortFilm)
RETURN  { __resolveType: "ShortFilm", title: this_ShortFilm.title } AS actedIn
}
RETURN this { .name, actedIn: collect(actedIn) } AS this
```

[...]

PR #590, opened 2021-11-16
Fix: Project All Interface Fields For Types Implementing `@relationship` Interface

dmoree:
# Description

Proposed fix for #583. Removes unnecessary redefinition of `fieldsByTypeName` that was omitting additional implemented interfaces on types used in `@relationship`. The fields for any node that implements `@relationship` interface are already accounted for in `createProjectionAndParams`

# Issue

- #583 

# Checklist


dmoree:
@darrellwarde I noticed that the `fieldsByTypeName` is being redefined to include only the fields of the node and interface of the `@relationship`. Perhaps I'm missing somthing, but was there any intent to this beyond being sure to include the fields of the interface? If not, then this is actually already being taken care of in `createProjectionAndParams` itself and the entire `resolveTree.fieldsByTypeName` can be passed. This will account for not only the interface used in the `@relationship` directive, but any additional interfaces implemented by the underlying types.

darrellwarde (approved):
To be very honest, I don't think there were any test cases using multiple interfaces as you have probably seen. However, this fix seems to work, and doesn't appear to break anything, so let's get it merged!

Issue #593, opened 2021-11-17
Sum Field On Aggregations


danstarns:
**Is your feature request related to a problem? Please describe.**
When aggregating numerical properties I would like the ability to sum those numerical values. 

**Describe the solution you'd like**
Each numerical aggregation type has a sum property. Given the following Type Definitions:

```gql
type Movie {
    title: String!
    released: BigInt!
    actors: [Person] @relationship(type: "ACTED_IN", direction: IN)
}

type Person {
    name: String!
    born: BigInt
}
```
 

You would be able to query sum top level like so: 

```gql
{
  moviesAggregate {
    released {
      sum
    }
  }
}
```

Pluss you should be able to query sum on the field aggregations like so: 

```gql
{
  movies {
    title
    actorsAggregate {
      node {
        born {
          sum
        }
      }
    }
  }
}
```

PR #597, opened 2021-11-18
Feat/Aggregate Sum

danstarns:
# Description

Adds `SUM` aggregation to all three: 

1. Top Level Aggregations
2. Field Level Aggregations
3. Where Aggregations

### Top Level Aggregations

_Find the length of time to watch all the matrix movies_

**Type Definitions:** 

```gql
type Movie {
    title: String!
    length: Int!
}
```

**Query:** 

```gql
query {
    moviesAggregate(where: { title_CONTAINS: ["Matrix"] }) {
        length {
            sum
        }
    }
}
```

### Field Level Aggregations

_Find the total gas used in a block._ 

**Type Definitions:** 

```gql
type Block {
    hash: String!
    transactions: [Transaction] @relationship(type: "HAS_TRANSACTION", direction: OUT)
}

type Transaction {
    hash: String!
    gasUsed: Int!
}
```

**Query:** 

```gql
query {
    blocks(where: { hash: "0xc2702baffd4228dbafab12952ec4d33b81724d19e2e444390f9a4837256cfffd" }) {
        transactionsAggregate {
            node {
                gasUsed {
                    sum
                }
            }
        }
    }
}
```

### Where Aggregations

_Find students where the sum of all test scores are greater than 100._

**Type Definitions:** 

```gql
type Student {
    name: String!
    tests: [Test] @relationship(type: "HAS_TEST", direction: OUT)
}

type Test {
    name: String!
    score: Int!
}
```

**Query:** 

```gql
query {
    students(where: { testsAggregate: { node: { score_SUM_GT: 100 } } }) {
        name
    }
}
```

# Issue

#593


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-597.surge.sh

packages/graphql/tests/integration/aggregations/where/edge/int.int.test.ts
```diff
@@ -751,4 +751,79 @@ describe("aggregations-where-edge-int", () => {
...
+
+            const testString = generate({
+                charset: "alphabetic",
+                readable: true,
+            });
+
+            const someInt1 = Math.floor(Math.random() * Math.random());
+            const someInt2 = Math.floor(Math.random() * Math.random());
+            const someInt3 = Math.floor(Math.random() * Math.random());
+
```
angrykoala, line 782: I'm not sure what is the advantage of using random data here :confused:, I think random data adds unnecessary complexity to a test.

In fact, I think this setup is already problematic, as `Math.random()` gives a value between 0 and 0.99999..., multiplied with another `Math.random` we will always get a value between 0 and 0.999...,  using `.floor` means that `someInt1`, `someInt2`, `someInt3` are **always** 0

IMO, random data in tests make repeatability and readability harder (+ prone to bugs such as this one), without really giving any wider range of testing (with exceptions, of course), at best you get the same test as hard-coded data, at worst you get a test that randomly fails.

packages/graphql/tests/integration/aggregations/where/node/int.int.test.ts
```diff
@@ -729,4 +729,77 @@ describe("aggregations-where-node-int", () => {
...
+            `;
+
+            const testString = generate({
+                charset: "alphabetic",
+                readable: true,
+            });
+
+            const someInt1 = Math.floor(Math.random() * Math.random());
+            const someInt2 = Math.floor(Math.random() * Math.random());
+            const someInt3 = Math.floor(Math.random() * Math.random());
```
angrykoala, line 756: Same as before

angrykoala (changes requested):
:+1:  looking good, just some discussion/issues in a couple of tests

packages/graphql/tests/integration/aggregations/where/edge/int.int.test.ts
darrellwarde: Given Dan's absence this week, the fact that the tests pass and the fact that this is littered _everywhere_, let's leave this as it is for now and refactor all instances of this in a different branch? I've created a card in Trello for this. I also think we could be making better use of holistic test data generators such as `faker`, instead of using `generate` as we currently do.

darrellwarde (approved):
Nothing too controversial here, looks good!

Issue #586, opened 2021-11-16
Type with name "BigInt" does not exists after upgrading from 2.1.0


Kubera2017:
**Describe the bug**
I tried to upgrade an existing project from 2.1.0 and got this error: "Type with name "BigInt" does not exists". I tried 2.2.0, 2.3.0 and 2.4.0

Kubera2017:
I found the issue, basically it doesn't work now with this syntax:
```
input ExampleInput {
  value: BigInt = 0
}
```
If I remove default value it works
```
input ExampleInput {
  value: BigInt
}
```

darrellwarde:
I don't seem to be able to reproduce this @Kubera2017! I'm assuming you're seeing this error on server startup?

Kubera2017:
Yes, here is it with blank setup
![Code_DNCMd3F32T](https://user-images.githubusercontent.com/29781060/142000722-ec2f980d-e190-4bb4-bee1-d3792fd5de4d.png)

```
const { Neo4jGraphQL } = require("@neo4j/graphql");
const { ApolloServer, gql } = require("apollo-server");
const neo4j = require("neo4j-driver");

const typeDefs = gql`
    input TestInput {
        id: BigInt = "0"
    }

    type Movie {
        id: BigInt!
    }
`;

const driver = neo4j.driver(
    "bolt://localhost:7687",
    neo4j.auth.basic("neo4j", "123")
);

const neoSchema = new Neo4jGraphQL({ typeDefs, driver });

const server = new ApolloServer({
    schema: neoSchema.schema,
});

server.listen().then(({ url }) => {
    console.log(`🚀 Server ready at ${url}`);
});
```

darrellwarde:
Weird, I definitely can't produce on my end.

Out of interest, what version of `graphql` do you have installed?

Kubera2017:
My steps:
```
npm init
npm i -save @neo4j/graphql
npm i -save apollo-server
npm i -save neo4j-driver
npm i -save graphql
```
For graphql it installs graphql@16.0.1

And run the code above with `node index.js`

darrellwarde:
So we don't support `graphql@16.x.x` yet, sadly the version specification in our `peerDependencies` doesn't seem to be working as expected.

Can you please instead install `graphql` using `npm i --save graphql@^15.0.0` and try again?

Kubera2017:
I tried now, the same thing
[...]

PR #599, opened 2021-11-18
Generate scalar typedefs before extra definitions

angrykoala:
# Description

Updating graphql-compose from 9.0.3 to 9.0.4 causes `make-augmented-schema` to break when using the following typedefs:
```
    input TestInput {
        id: BigInt = "0"
    }
```
This is due to type BigInt not being defined before `extraDefinitions` are added

A test was added to ensure this behavior does not happen again, but keep in mind that this is not reproducible with graphql-compose < 9.04

# Issue

#586

### DIFF
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -67,7 +67,7 @@
         "debug": "^4.3.2",
         "deep-equal": "^2.0.5",
         "dot-prop": "^6.0.1",
-        "graphql-compose": "^9.0.2",
+        "graphql-compose": "^9.0.4",
         "graphql-parse-resolve-info": "^4.12.0",
         "graphql-relay": "^0.8.0",
         "jsonwebtoken": "^8.5.1",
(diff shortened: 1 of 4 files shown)


darrellwarde:
Looks good! Worthwhile performing an upgrade of `graphql-compose` so that our `yarn.lock` and `package.json` entries are at 9.0.4 and we can have higher confidence that this test case will actually be triggered? 🙂

angrykoala:
@darrellwarde I was planning on doing that, but wanted to check before what version the CI is using first

darrellwarde:
> @darrellwarde I was planning on doing that, but wanted to check before what version the CI is using first

Because we do our CI installs using `yarn --immutable`, it will install whatever is in the `yarn.lock` file, so have a check of the entry in there. 🙂

darrellwarde:
Looks good! Do you also want to update the GitHub Actions workflows as we discussed earlier in this branch?

angrykoala:
@darrellwarde Happy to do it, but I think it is better in a separate PR to keep these as small as possible

darrellwarde (approved):

PR #602, opened 2021-11-22
Fix on some lint errors

angrykoala:
This PR contains a batch of simple eslint errors fixes, to bring down the number of eslint problems from 114 -> 75

## Considerations
* The rule `prefer-destructuring` has been disabled, as this rule enforces destructuring in cases where it doesn't make sense or even directly contradicts `prefer-const`. We may want to discuss if this is ok


darrellwarde (dismissed):
LGTM! We need to add a linting step to our GitHub Actions workflows to stop violations sneaking in. 👀

angrykoala:
@darrellwarde Agree on the step, but there are still a big chunk of errors to fix, and some of them require some discussion to see what approach do we want to take

PR #606, opened 2021-11-24
Fix changelog generator config

darrellwarde:
# Description

If a change had been made to both the GraphQL Library and OGM, these changes would not be in the changelog because the config for the GraphQL changelog ignored "ogm" labels, and vice versa.

This fix instead focuses on combinations of labels, so labels "graphql" and "feature" would add the PR to the changelog for GraphQL Library as a feature. It doesn't matter if "ogm" is another label, compared to before where the entire PR would be ignored.


angrykoala (approved):

oskarhane (approved):
Nice 👾

PR #607, opened 2021-11-24
Eslint fix

angrykoala:
# Description

Another run on fixing some lint errors and warnings.

## Notable changes
* Allows `__typename` with dangling underscore


packages/graphql/tests/integration/sort.int.test.ts
```diff
@@ -243,108 +243,136 @@ describe("sort", () => {
...
-                            actors(
-                                where: { name_IN: $actorNames }
-                                options: { sort: [{ totalScreenTime: $direction }] }
-                            ) {
-                                name
-                                totalScreenTime
-                            }
-                        }
-                    `;
+        test("should sort DESC on top level", async () => {
```
darrellwarde, line 246: Any reason these tests have been split? 🧐

darrellwarde (approved):
Looks good once the test failure has been addressed!

packages/graphql/tests/integration/sort.int.test.ts
angrykoala: Oh, good catch, forgot to mention in the description.

Generally, dynamic flows in tests are considered a bad practice (with some exceptions) as that increases the complexity of the test, and thus, the chances of it having bugs (as you can see, the disabled rule `jest/no-conditional-expect` exists because of this).

darrellwarde (approved):
LGTM!

PR #614, opened 2021-11-30
Introspector fixes

oskarhane:
# Description

Two fixes for the introspector:

- Create correct non-nullable arrays for relationships.
- Don't generate invalid type/interface names with regards to node labels and rel-types that starts with invalid characters (i.e. numbers).

For the last point, the integration test assertion with Neo4j GraphQL is commented out due to lack of support in the library for labels starting with numbers.

fyi: I deliberately named this in a way so it doesn't show up in changelogs.

### DIFF
--- a/packages/introspector/src/transforms/neo4j-graphql/graphql.ts
+++ b/packages/introspector/src/transforms/neo4j-graphql/graphql.ts
@@ -26,6 +26,7 @@ import generateRelationshipPropsName from "./utils/generate-relationship-props-n
 import { RelationshipPropertiesDirective } from "./directives/RelationshipProperties";
 import createRelationshipFields from "./utils/create-relationship-fields";
 import { ExcludeDirective } from "./directives/Exclude";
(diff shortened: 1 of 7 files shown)


angrykoala (approved):

darrellwarde:
Is it sensible to produce type definitions that we know won't work with the library in the case of labels beginning with numbers? 

It might throw users off as they would expect a tool written by us to work with our library.

oskarhane:
> Is it sensible to produce type definitions that we know won't work with the library in the case of labels beginning with numbers?
> 
> It might throw users off as they would expect a tool written by us to work with our library.

@darrellwarde I do see your point, but not supporting labels starting with numbers in Neo4j GraphQL is a bug and I don't think we should avoid doing the right thing in one place because of a bug in another place.
Also, if we don't do this, the produced GraphQL typedefs will be invalid according to the GraphQL spec and I think that would be even worse.

This is also an edge case that no one has bumped into yet (in the Neo4j GraphQL lib).

PR #616, opened 2021-11-30
Set all aggregation fields as nullable

angrykoala:
# Description

This PR reverts a fix done in #605 to avoid breaking changes, instead of the correct types, all aggregation types will be nullable with the typename unchanged.

Comments on the breaking fix left for v 3.0.0

### DIFF
--- a/packages/graphql/src/translate/translate-aggregate.ts
+++ b/packages/graphql/src/translate/translate-aggregate.ts
@@ -92,9 +92,10 @@ function translateAggregate({ node, context }: { node: Node; context: Context })
 
         if (field) {
             const thisProjections: string[] = [];
-            const aggregateFields =
-                selection[1].fieldsByTypeName[`${field.typeMeta.name}AggregateSelectionNullable`] ||
-                selection[1].fieldsByTypeName[`${field.typeMeta.name}AggregateSelectionNonNullable`];
+            const aggregateFields = selection[1].fieldsByTypeName[`${field.typeMeta.name}AggregateSelection`];
+            // const aggregateFields =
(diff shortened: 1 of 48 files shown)


packages/graphql/src/translate/translate-aggregate.ts
```diff
@@ -92,9 +92,10 @@ function translateAggregate({ node, context }: { node: Node; context: Context })
 
         if (field) {
             const thisProjections: string[] = [];
-            const aggregateFields =
-                selection[1].fieldsByTypeName[`${field.typeMeta.name}AggregateSelectionNullable`] ||
-                selection[1].fieldsByTypeName[`${field.typeMeta.name}AggregateSelectionNonNullable`];
+            const aggregateFields = selection[1].fieldsByTypeName[`${field.typeMeta.name}AggregateSelection`];
+            // const aggregateFields =
```
danstarns, line 96: I understand the reason for leaving this code comented out but I don't know if it's a good habit to get into. Could it make it acceptable to do it elsewhere and then a codebase filled with commented out code?

angrykoala: I agree it is not a good habit, but this is something that we should be able to fix pretty soon after this release, and I think it would be a bit of unnecessary overhead to remove all these references and then having to repeat all the work of adding the fix

darrellwarde: I'd normally agree, but I think given we want to reintroduce this in the next major release then it's sensible to leave it in.

100% though, this is not the setting of a precedent.

darrellwarde (approved):

PR #617, opened 2021-12-01
Fix: Escape all @node directive labels

angrykoala:
# Description

Escape all labels set in the @node directive label to ensure these are valid cypher labels


darrellwarde:
Only 13 changed files? Seems like... Not enough. 🧐

darrellwarde (approved):
Looks good, smart move only doing this for the `@node` directive!

PR #637, opened 2021-12-13
Fix: Removes escaped labels from filter IN labels statements

angrykoala:
# Description
This PR removes label escaping from queries using the pattern `"mylabel" IN labels(param)`


darrellwarde (approved):
Looks good to me! Nice work getting this in so quickly. 🏅

Issue #620, opened 2021-12-02
Cannot convert undefined or null to object in 2.5.0


danielfein:
Hi I upgraded to 2.5.0 and now a custom cypher query is failing. I downgraded to 2.4.0 to rerun the exact same query and it's totally fine, so it's isolated to 2.5.0.

It's a standard query:

```
query explore{
    explore_topics{
      id
      name
      description
      image
    }
    topic_count
  }
```

Built with a fairly simple custom definition:

```
explore_topics: [Topic] @cypher(statement: """
        MATCH (t:Topic)-[:TAGGED]-(p:Post)
        WITH t, COLLECT(p) as posts
        RETURN t
        ORDER BY SIZE(posts) DESC
        LIMIT 10
    """),
```

And if it's helpful here's the full error:

```
{
  "errors": [
    {
      "message": "Cannot convert undefined or null to object",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "explore_topics"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: Cannot convert undefined or null to object",
            "    at Function.entries (<anonymous>)",
            "    at Object.createProjectionAndParams [as default] (/PATH_TO_APP/node_modules/@neo4j/graphql/dist/translate/create-projection-and-params.js:441:49)",
            "    at /PATH_TO_APP/node_modules/@neo4j/graphql/dist/schema/resolvers/cypher.js:66:71",
            "    at Generator.next (<anonymous>)",
            "    at /PATH_TO_APP/node_modules/@neo4j/graphql/dist/schema/resolvers/cypher.js:26:71",
            "    at new Promise (<anonymous>)",
            "    at __awaiter (/PATH_TO_APP/node_modules/@neo4j/graphql/dist/schema/resolvers/cypher.js:22:12)",
            "    at resolve (/PATH_TO_APP/node_modules/@neo4j/graphql/dist/schema/resolvers/cypher.js:44:16)",
            "    at /PATH_TO_APP/node_modules/@graphql-tools/schema/index.cjs.js:40:24",
            "    at processTicksAndRejections (internal/process/task_queues.js:93:5)"
          ]
        }
      }
    }
  ],
  "data": {
    "explore_topics": null
  }
}
```

angrykoala:
Hi @danielfein 
[...]

PR #639, opened 2021-12-14
Fix context being overriden in for multiple queries #620

angrykoala:
# Description
This PR fixes an error caused by global context being shared by multiple revolvers in the same query and being overridden. The patch consist on moving the resolve tree generation (from global resolver) to each resolver after the asynchronous causing the race condition has finished 

This PR only covers a quick patch the underlying problem of shared context is not addressed here

# Issue

#620


packages/graphql/src/schema/resolvers/count.ts
```diff
@@ -16,15 +16,22 @@
...
 import { Context } from "../../types";
 import { isNeoInt } from "../../utils/utils";
+import getNeo4jResolveTree from "../../utils/get-neo4j-resolve-tree";
 
 export default function countResolver({ node }: { node: Node }) {
-    async function resolve(_root: any, _args: any, _context: unknown) {
+    async function resolve(_root: any, _args: any, _context: unknown, info?: GraphQLResolveInfo) {
         const context = _context as Context;
+        if (info) {
+            // FIXME: info is always available, this is only to support unit tests at count.test.ts
```
darrellwarde, line 32:
```suggestion
            // TODO: info is always available, this is only to support unit tests at count.test.ts
```

darrellwarde (approved):
LGTM! 🙂  Let's get this released!

PR #641, opened 2021-12-14
RFC: Redirect Relationships

darrellwarde:
RFC for redirecting relationships.


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-641.surge.sh

darrellwarde:
> @darrellwarde Seeing as this is stale, and no relevant information is actually in the RFC maybe we could close this?

Agreed, we have a card for this on the backlog, so we can revisit writing an RFC for this when we have time!

Issue #354, opened 2021-07-23
Mandatory relationships when creating nodes


Kubera2017:
**Is your feature request related to a problem? Please describe.**
It would be great to check the absence of related nodes and throw errors when creating nodes with "connect".
I see now it generates OPTIONAL MATCH for that.

Example: when you add comment to the post that doesn't exist. Now if we provide missing id:

mutation {
  createComments (input: [
   comment_id: 1
   connect: {where: {post_id: "missing node"}}
]) {id}
}

This is old unresolved issue https://github.com/neo4j-graphql/neo4j-graphql-js/issues/534

darrellwarde:
This one is tricky, easier to enforce in some parts and very difficult in others.

The easy part is enforcing 1-1 and 1-m relationships, so if the `Comment` type is:

```graphql
type Comment {
  id: ID!
  post: Post! @relationship(type: "HAS_COMMENT", direction: IN)
}
```

We could and should enforce the non-nullability of that relationship field, which we currently don't. 🙁 

Much more difficult is when the type definition is:

```graphql
type Comment {
  id: ID!
  post: [Post!]! @relationship(type: "HAS_COMMENT", direction: IN)
}
```

Because technically, according to the data model, there is absolutely nothing wrong with a comment not having any posts! So we would need a different way of enforcement here (an argument on the directive perhaps, `minimumConnections`?!).

As such, I imagine we will address the former first, but when we do get around to it, it will be a major change where we start treating those cardinalities properly, such as not allowing 'disconnect' operations.

Sorry for no firm "yes" or "no", just wanted to fill you in on my thoughts on the matter! 😄

Kubera2017:
Yes, at least the first 1:1 to be implemented would be great, because it's frequent case in many models, for example - Post always have User (author), Comment always have User (author) etc.

danstarns:
Something like this would need validation inside the Cypher query using something like `apoc.util.validate`, similar to how we enforce auth, which could be used to ensure that the pattern definitely exists after connecting.

Kubera2017:
I think current workaround is neosemantics plugin and SHACL transaction validation, you could also take a look on this for possible cases (https://neo4j.com/labs/neosemantics/4.0/validation/#_validating_transactions)

Kubera2017:
@danstarns just checked this with such schema:
```
type Post {
        id: Int!
[...]

PR #643, opened 2021-12-15
Reverts #572 non nullable relation checks

angrykoala:
# Description

This PR reverts changes made in #572 that caused an unexpected behaviour on common queries

# Issue

#354
#549


darrellwarde (approved):
LGTM, let's get it out there!

PR #646, opened 2021-12-16
Feature: Sort On Interface Relationship

dmoree:
# Description

This is an implementation of sorting on relationship fields that use interfaces. Current behavior is to treat interfaces in much the same way as unions. However, since the field is returning an interface it can be guaranteed that the fields of the interface will be present. This proposes a creation of `{Interface}Sort` (and `{Interface}Options`) that holds the sortable fields of the interface to be more in line with the treatment of the underlying nodes.


rcbevans:
Awesome stuff!

rcbevans (dismissed):

packages/graphql/src/translate/connection/create-connection-and-params.ts
```diff
@@ -52,7 +52,18 @@ function createConnectionAndParams({
...
+    // Fields of {edge, node} to sort on. A simple resolve tree will be added if not in selection set
+    // Since nodes of abstract types and edges are constructed sort will not work if field is not in selection set
+    const sortFields: Record<keyof ConnectionSortArg, string[]> = Object.assign(
+        {},
+        ...(["edge", "node"] as Array<keyof ConnectionSortArg>).map((element) => ({
+            [element]: ([] as string[]).concat(
+                ...sortInput.map(({ [element]: sortField = {} }) => Object.keys(sortField))
+            ),
+        }))
+    );
```
darrellwarde, line 65: I'm honestly having a hard time understanding what this is actually doing so can't make a suggestion as to how it could be a bit more readable... But the `assign` -> `map` -> `concat` -> `map` makes this pretty hard to comprehend. Any chance it could be simplified a touch - even if it's not quite as concise!!!

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -622,25 +629,36 @@ function createProjectionAndParams({
...
+        ...Object.entries(sort).map(([fieldName, direction]) => {
+            if (direction === "DESC") {
+                return `'${fieldName}'`;
+            }
+
+            return `'^${fieldName}'`;
+        }),
+    ];
+}
+
```
darrellwarde, line 644: 👏

darrellwarde (commented):
There's literally one part of this that I would like to see simplified a touch just to make it a little easier to comprehend! But otherwise, great work!

packages/graphql/src/translate/connection/create-connection-and-params.ts
dmoree: I think this one got away from me.
Refactored in https://github.com/neo4j/graphql/pull/646/commits/fcfd3091089cec432f34064eea2e61ccaadc33bd

dmoree:
Converted to draft in order to add tests for aliased sort fields.

darrellwarde (dismissed):
This looks really good! Just needs changes merging from `dev` and then good to go. 🙂

dmoree:
This can be factored to include the utility functions. I can push those changes in a couple of minutes.

darrellwarde (approved):

Issue #591, opened 2021-11-17
Cypher fields projected before options limiting causing unusably slow performance


rcbevans:
Given a GraphQL type with an expensive cypherField

```
type MyType {
    id: String!
    expensiveField: [String!]! @cypher(statement: "<some expensive sub-query>")
}
```

called with a query

```
query MyExpensiveQuery {
    myTypes(options: {offset: 0, limit: 10, sort: [{id: "ASC"}]}) {
        id
        expensiveField
    }
}
```

The generated Cypher looks like

```
MATCH (this:MyType)
RETURN this { .id, expensiveField: apoc.cypher.runFirstColumn("<some expensive sub-query>") } AS this
ORDER BY this.id ASC
SKIP 0
LIMIT 10
```

Running this query in neo4j desktop with `PROFILE` I can see that the cypher planner is selecting all `MyType` nodes, generating the projection for them *all*, then throwing all but the first 10 away.

Further, the cypher planner does not seem to be caching the result of the custom cypher, meaning it is re-evaluated projecting *all* nodes *every* call, even as you offset/limit paginate through the results.

Even worst, the cypher query does not seem to be parallelized, causing the expensive sub query to run for every node in a single threaded manner.

All of these issues combine to yield a GraphQL query that is unbearably slow, and execution speed slows linearly with the number of nodes of type `MyQuery`.

To avoid the unbearably slow execution time the GraphQL it's necessary to ensure that the expensive projection happens *after* the initial filtering of nodes has occurred.

If the generated cypher is rearranged to something like:

```
CALL {
    MATCH (this:MyType)
    RETURN this, this { .id } AS this_selection
    ORDER BY this_selection.id ASC
    SKIP 0
    LIMIT 10
}
RETURN this { .id, expensiveField: apoc.cypher.runFirstColumn("<some expensive sub-query>") } AS this
```

The cypher planner will then evaluate the final projection for only the 10 `MyType` nodes returned by the initial `CALL`.

[...]

PR #649, opened 2021-12-17
LIMIT results before projecting custom Cypher

darrellwarde:
# Description

Implement the Cypher optimisations as suggested by @rcbevans in #591.

For a massive increase in performance, limits node numbers _before_ executing custom Cypher on all of the nodes. If there are 100 nodes for a type with a custom Cypher field and you query with a limit 10, this PR will ensure the custom Cypher is only executed for those 10 nodes and not all 100. However, if you have opted to sort by the custom Cypher field, the limit will happen after the custom Cypher projection.

Use cases:
- Existing behaviour for:
  - No custom Cypher fields
  - No limit option provided
  - Custom Cypher fields where sort is performed on Cypher field
- `LIMIT` in subquery before custom Cypher projection for:
  - Queries on type with custom Cypher with no sort
  - Queries on type with custom Cypher where sort is on non-Cypher field

A note on the various `package.json` and `tsconfig.json` changes in this branch. The use of `Object.fromEntries` by @rcbevans took me down the path of looking out our TypeScript configuration to get this working. Turns out we weren't really using the correct configuration given our minimum support of Node.js 12. So our base has now been set to extend from https://github.com/tsconfig/bases/blob/main/bases/node12.json. Then it was a matter of getting `neo-push` client working - this now extends https://github.com/tsconfig/bases/blob/main/bases/create-react-app.json but I had to upgrade TypeScript a bit to 4.2.4 to get this working properly (some might say this was long overdue anyway).

# Issue

#591

# Checklist


darrellwarde:
Hey @rcbevans, I don't seem to able to add you as a reviewer, but feel free to take a look at this one, implementing your changes from the branch linked to from your comment https://github.com/neo4j/graphql/issues/591#issuecomment-989582175

Probably needs a little polishing, just wanted to quickly get something together to start with. Includes newer features from `master` including the ability to be able to sort by Cypher fields, which complicates things somewhat!

darrellwarde:
Some frustrating TypeScript issues causing test failures it would seem... 😠

rcbevans:
@darrellwarde the code in my branch has the ability to sort on a cypher field.

```
                const orderByProjection = createProjectionAndParams({
                     node,
                     context,
                     fieldsByTypeName: {
                         [node.name]: Object.fromEntries(
                             Object.entries(fieldsByTypeName[node.name]).filter(([key]) => sortFields.includes(key))
                         ),
                     },
                     varName,
                 });
```

In the initial call block I build a partial projection of *only* the fields being sorted on, and this projection is done within the call block.  This includes cypher fields.

One thing that I didn't yet figure out is how to allow sorting on fields which aren't part of the query fieldset.

packages/graphql/src/translate/translate-read-new.ts
```diff
@@ -0,0 +1,201 @@
...
+            whereStrs.push(where[0]);
+            cypherParams = { ...cypherParams, ...where[1] };
+        }
+    }
+
+    const whereAuth = createAuthAndParams({
+        operation: "READ",
+        entity: node,
+        context,
+        where: { varName, node },
```
rcbevans, line 102: This might be an addition from another PR, or perhaps it was a property removed since I forked?

packages/graphql/src/translate/translate-read-new.ts
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        if (hasLimit) {
+            limitStr = `LIMIT $${varName}_limit`;
+            cypherParams[`${varName}_limit`] = optionsInput.limit;
+        }
+
+        if (optionsInput.sort && optionsInput.sort.length) {
+            if (hasLimit) {
+                const sortFields = optionsInput.sort.map((sort) => Object.keys(sort)[0]);
```
rcbevans, line 140: Collect the fields the caller is sorting on

packages/graphql/src/translate/translate-read-new.ts
```diff
@@ -0,0 +1,201 @@
...
+                const orderByProjection = createProjectionAndParams({
```
rcbevans, line 141: Build a sub projection of *only* the minimal set of fields and properties required for sorting

packages/graphql/src/translate/translate-read-new.ts
```diff
@@ -0,0 +1,201 @@
...
+                    node,
+                    context,
+                    fieldsByTypeName: {
+                        [node.name]: Object.fromEntries(
```
rcbevans, line 145: `Object.fromEntries` required me to bump the ts target (from es2015?) I think.  There may be polyfills available for older targets.

packages/graphql/src/translate/translate-read-new.ts
```diff
@@ -0,0 +1,201 @@
...
+            offsetStr,
+        ];
+    } else {
+        cypher = [
+            "CALL {",
+            matchStr,
+            whereStr,
+            authStr,
+            ...(sortStr ? [`WITH ${varName}, ${varName} ${limitOrderProjStr} AS ${limitOrderVarName}`, sortStr] : []),
+            ...(projAuth ? [`WITH ${varName}`, projAuth] : []),
```
rcbevans, line 188: I'm not 100% sure what projAuth is;  but this will be evaluated for every match pre-limit, though it doesn't seem to be participating in the sorting/limiting process itself, so perhaps it could be moved out of the call black to be evaluated only for the nodes that are going to be returned?

packages/graphql/src/translate/translate-read-new.ts
rcbevans: The polyfill for this is simple to implement if it's beneficial to maintain older target language support: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#polyfill

packages/graphql/src/translate/translate-read.ts
```diff
@@ -124,30 +126,77 @@ function translateRead({ node, context }: { context: Context; node: Node }): [st
...
+                ? [`WITH ${varName}, ${varName} ${limitOrderProjStr} AS ${limitOrderVarName}`, sortStr]
+                : []),
+            ...(projAuth ? [`WITH ${varName}`, projAuth] : []),
+            `RETURN ${varName}`,
+            ...(!cypherSort ? [offsetStr, limitStr] : []),
+            "}",
+            ...connectionStrs,
+            ...interfaceStrs,
+            `RETURN ${varName} ${projStr} as ${varName}`,
+            ...(sortStr && cypherSort
```
rcbevans, line 190: I don't think this can go *after* the final return?

I was performing a `WITH` sort,

```
WITH this, this { ... } AS this_sort
ORDER BY this_sort.foo
```

Which cause the result set to be sorted in-line.

packages/graphql/src/translate/translate-read.ts
```diff
@@ -124,30 +126,77 @@ function translateRead({ node, context }: { context: Context; node: Node }): [st
...
+            `RETURN ${varName} ${projStr} as ${varName}`,
+            ...(sortStr && cypherSort ? [sortStr] : []),
+            offsetStr,
+        ];
+    } else {
+        cypher = [
+            "CALL {",
+            matchAndWhereStr,
+            authStr,
+            ...(sortStr && !cypherSort
```
rcbevans, line 180: The `limitOrderProjStr` handles a projection including cypher fields, and so can be sorted here.

The downside of this is that the cypher projection would potentially be evaluated twice, once in the CALL block, and once outside.

It may be possible to optimize this by projecting fields participating in the sort inside the `CALL` block, and those which don't outside, then merging the two projections into one final projection which is returned.  I don't know if Cypher supports something like spreading.

`RETURN this { **this_sort, foo: ... } AS this`

packages/graphql/tests/tck/tck-test-files/directives/cypher.test.ts
```diff
@@ -181,6 +181,88 @@ describe("Cypher directive", () => {
...
+                }
+            }
+        `;
+
+        const req = createJwtRequest("secret", {});
+        const result = await translateQuery(neoSchema, query, {
+            req,
+        });
+
+        expect(formatCypher(result.cypher)).toMatchInlineSnapshot(`
```
rcbevans, line 198: So glad to see jest snapshots now!

packages/graphql/tests/tck/tck-test-files/directives/cypher.test.ts
```diff
@@ -181,6 +181,88 @@ describe("Cypher directive", () => {
...
+            "CALL {
+            MATCH (this:Actor)
```
rcbevans, line 240: I would expect to see:

```
CALL {
    MATCH (this:Actor)
    WITH this, this { randomNumber: apoc.cypher.runFirstColumn(\\"RETURN rand()\\", {this: this, auth: $auth}, false) } AS this_sort
    ORDER BY this_sort.randomNumber ASC
    RETURN this
    LIMIT $this_limit
}

RETURN this { randomNumber: apoc.cypher.runFirstColumn(\\"RETURN rand()\\", {this: this, auth: $auth}, false) } as this
```

rcbevans: In terms of performance, this version is worse, because it will perform the apoc projection once for every node, then once again for the `LIMITed` nodes being returned, however, if there were additional expensive cypher projections (or even properties) being returned *they* wouldn't be evaluated until the end for the subset of nodes being returned.

rcbevans (changes requested):

packages/graphql/src/translate/translate-read-new.ts
darrellwarde: This `translate-read-new.ts` file wasn't even meant to be committed, I was just using it as a scratch pad whilst trying to fit your changes into code, apologies! 🙈

packages/graphql/src/translate/translate-read.ts
darrellwarde: Damn, good spot, I forgot that `sortStr` was prepended with `WITH`.

darrellwarde:
Sorry @rcbevans, a plethora of errors in this first hash, I clearly wasn't concentrating enough, and misunderstood how much is in the `sortStr` variable. Will revisit, likely on Monday morning.

rcbevans:
No problem!

darrellwarde:
Hey @rcbevans, I've had a revisit. It might be an oversimplification, but I've changed this up so that the only time when a subquery is used to encompass limiting/offsetting/sorting is when there are custom Cypher fields on the type and sorting has not been applied to any custom Cypher fields. In any other situation, the Cypher will be as it was before this branch. Furthermore, the projection is only done once after the subquery, which should perform better than performing it twice whilst achieving the same results.

Knowing my pre-Christmas brain fog, I've probably missed something vital, so do tell me if there's something I've missed! 🙈

packages/graphql/src/translate/translate-read.ts
```diff
@@ -32,14 +32,17 @@ function translateRead({ node, context }: { context: Context; node: Node }): [st
     const optionsInput = resolveTree.args.options as GraphQLOptionsArg;
     const varName = "this";
 
+    const hasOffset = Boolean(optionsInput?.offset) || optionsInput?.offset === 0;
+    const hasLimit = Boolean(optionsInput?.limit) || optionsInput?.limit === 0;
```
angrykoala, line 36: What does it mean a limit or offset of 0?
I would expect that be the same as no offset or no limit. :thinking:

packages/graphql/src/translate/translate-read.ts
```diff
@@ -137,17 +140,36 @@ function translateRead({ node, context }: { context: Context; node: Node }): [st
...
-        ...connectionStrs,
-        ...interfaceStrs,
-        `RETURN ${varName} ${projStr} as ${varName}`,
-        ...(sortStr ? [sortStr] : []),
-        offsetStr,
-        limitStr,
-    ];
+    let cypher: string[] = [];
+
+    if (node.cypherFields.length && hasLimit && !cypherSort) {
```
angrykoala, line 145: Maybe a variable could give a bit more context on what this logic implies?

packages/graphql/src/translate/translate-read.ts
```diff
@@ -137,17 +140,36 @@ function translateRead({ node, context }: { context: Context; node: Node }): [st
...
+            `RETURN ${varName}`,
+            ...(sortStr ? [sortStr] : []),
+            ...(offsetStr ? [offsetStr] : []),
+            ...(limitStr ? [limitStr] : []),
+            "}",
+            ...connectionStrs,
+            ...interfaceStrs,
+            `RETURN ${varName} ${projStr} as ${varName}`,
+        ];
+    } else {
```
angrykoala, line 160: Consider moving this to separate functions to split a bit this large codebase

angrykoala (approved):

packages/graphql/src/translate/translate-read.ts
darrellwarde: I think this makes sense - if I set the limit to zero then really I would expect zero results back. Without this, setting offset to zero would return all results, which is misleading.

PR #658, opened 2021-12-21
Pin graphql-compose to 9.0.4 until we have upgraded to GraphQL 16.x

darrellwarde:
# Description

It appears that the 9.0.5 release of `graphql-compose` expects GraphQL 16.x, as can be seen in the recent failures of our package tests. This PR pins it to 9.0.4 for now - this should be undone when we upgrade to GraphQL 16.x.

### DIFF
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -73,7 +73,7 @@
         "debug": "^4.3.2",
         "deep-equal": "^2.0.5",
         "dot-prop": "^6.0.1",
-        "graphql-compose": "^9.0.4",
+        "graphql-compose": "9.0.4",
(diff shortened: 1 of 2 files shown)


wheresrhys:
Any idea when we can expect to see you upgrade to graphql 16?

darrellwarde:
Hey @wheresrhys, for sure, this is planned for after Christmas as well as updating `@graphql-tools`. Expect a major release with breaking changes once this is done (with some good features, not just the dependency upgrades!).

darrellwarde:
This needs to be released as anyone currently installing the library will also get the incompatible `graphql-compose@9.0.5` - merging without approval.

Issue #489, opened 2021-09-16
Feature Request: Allow multiple relationship types


dmoree:
**Is your feature request related to a problem? Please describe.**
It can be helpful to have a field on a GraphQL type that corresponds to multiple relationship types in the underlying graph. As a familiar example, take for instance the `typeDefs`:
```graphql
type Director {
    id: ID!
    name: String!

    directed: [Movie!]! @relationship(type: "DIRECTED", direction: OUT)
}

type Actor {
    id: ID!
    name: String!

    movies: [Movie!]! @relationship(type: "ACTED_IN", direction: OUT)
}

type Movie {
    id: ID!
    title: String!

    director: Director! @relationship(type: "DIRECTED", direction: IN)
    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
}
```
This is a one-to-one representation of the nodes in the database where `(:Director)` nodes are related to a `(:Movie)` node by a single relationship `[:DIRECTED]` and `(:Actor)` nodes are related to `(:Movie)` nodes by a single relationship `[:ACTED_IN]`.

A typical question of `Movie` would be: who are the associated people? In order to pull these nodes from the database a cypher query of the form
```cypher
MATCH (person)-[:DIRECTED|ACTED_IN]->(this:Movie)
RETURN person
```
will return what is being sought after i.e. `(:Director)` and `(:Actor)` nodes. Fortunately,  GraphQL has an abstract type `Union` that can be used to describe the associated types. This would lead to extending the `typeDefs` with:
```graphql
union Person = Director | Actor

extend type Movie {
  people: [Person!]!
}
```
The obvious type for the `@relationship` directive would be `"DIRECTED|ACTED_IN"` as it is the only way not to bring the abstraction of `Person` into the database (one wouldn't want to add superfluous relationships like `[:SEARCH]` or `[:PERSON]` in order to facilitate a query that is as simple as the one above). Therefore, a proposal would be to allow multiple types to the `@relationship` directive so that a natural extension would be:
```graphql
union Person = Director | Actor

extend type Movie {
  people: [Person!]! @relationship(type: "DIRECTED|ACTED_IN", direction: IN)
}
```
[...]

PR #676, opened 2022-01-02
Feature: Allow Multiple Relationship Types

dmoree:
# Description

This PR is an updated implementation of the Feature Request raised in #489 which goes into further detail.

# Checklist

# TODO
- [ ] Update Documentation


darrellwarde:
Just as a heads up, this is quite a big change so quite likely that this one will be sitting in limbo for a while as we finish up the next release. Apologies.

packages/graphql/src/constants.ts
```diff
@@ -56,10 +56,16 @@ export const RESERVED_TYPE_NAMES = [
     },
 ];
 
+export const RELATIONSHIP_TYPE_FIELD = "_type";
```
dmoree, line 55: A decision on what the field name should be for the `edge` type can be made on this line. The implementation makes use of this variable and the tests can be updated with a simple find and replace.

dmoree:
@darrellwarde If you'll allow me to make the case that this is not a big change. 

I first want to point out that this is not a breaking change. Secondly, the only change really that this PR seeks to implement is to allow an `edge` to be more than just the properties. This adds a field  (currently `_type` but can be changed of course) that can be used in a `where` clause and will be added to the edge of a connection to query.

Issue #670, opened 2021-12-23
The @writeonly directive does not work when the @relationship directive is used at the same time.


AccsoSG:
**Describe the bug**
The @writeonly directive does not work when the @relationship directive is used at the same time. You can still query the field.

**Type definitions**
````graphql
    type IdentifierType{
        termId: String! @readonly @unique
        termName: String! @readonly @unique
        equivalents: [IdentifierType] @writeonly @relationship(type: "IS_EQUIVALENT_TO", direction: OUT)
        equivalenceClass: [IdentifierType] @cypher(statement: "MATCH (this)-[:IS_EQUIVALENT_TO*]-(c:IdentifierType) WHERE this <> c RETURN DISTINCT c")
    }
````

**Expected behavior**
It should be possible to create the relationship using a mutation, but it should not be possible to read it using a query.
![image](https://user-images.githubusercontent.com/49721200/147275680-9c9329c7-8c61-443b-8447-1deb9b7c22bf.png)

**Additional context**
This should be a workaround for my issue https://github.com/neo4j/graphql/issues/596

neo4j-team-graphql:
Many thanks for raising this bug report @AccsoSG. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @AccsoSG! :pray: We will now prioritise the bug and address it appropriately.

PR #678, opened 2022-01-03
Fix: `@writeonly` on `@relationship` Fields

dmoree:
# Description

Proposed fix for #670. This PR only adds a `{field}` `{field}Aggregate` and `{field}Connection` `@relationship` fields if it is not write-only as designated by the `@writeonly` directive.

# Issue

- #670


darrellwarde (approved):
LGTM, thanks @dmoree!

PR #713, opened 2022-01-05
Bump to GraphQL 16.x

darrellwarde:
# Description

Bumps GraphQL to version 16.x and makes the necessary changes to get this working.

# Checklist


packages/graphql/src/schema/scalars/Date.ts
```diff
@@ -17,16 +17,28 @@
  * limitations under the License.
  */
 
-import { GraphQLScalarType } from "graphql";
-import neo4j from "neo4j-driver";
+import { GraphQLError, GraphQLScalarType } from "graphql";
+import neo4j, { Date as Neo4jDate, isDate } from "neo4j-driver";
```
darrellwarde, line 21: One of two points of contention here as unsure if this will work with older driver versions. If not, we need to either update our minimum, or find an alternative route.

packages/graphql/src/schema/scalars/DateTime.ts
```diff
@@ -17,16 +17,28 @@
  * limitations under the License.
  */
 
-import { GraphQLScalarType } from "graphql";
-import neo4j from "neo4j-driver";
+import { GraphQLError, GraphQLScalarType } from "graphql";
+import neo4j, { DateTime, isDateTime } from "neo4j-driver";
```
darrellwarde, line 21: One of two points of contention here as unsure if this will work with older driver versions. If not, we need to either update our minimum, or find an alternative route.

darrellwarde (commented):
I need to do some testing with older driver versions to make sure all of this works.

packages/graphql/src/schema/scalars/DateTime.ts
danstarns: I'm curious about why it would not work?

darrellwarde: I believe these exports were only added in more recent release of the driver, hence why we did the weird `neo4j.types.DateTime` thing.

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-713.surge.sh

darrellwarde:
Pretty happy that this can be reviewed now @danstarns and @angrykoala. Actually pretty gutted that we can't maintain backwards compatibility with `^15.0.0` given the limited amount of changes, so I'm thinking about that now.

darrellwarde:
So it looks like the scalar types didn't need as much of a dramatic refactor as originally thought - tests seem to pass for both GraphQL 15 and 16 now. The only thing that can't run over both is the schema tests as each version seems to format slightly differently (new lines between `"""` comments).

packages/graphql/src/schema/scalars/BigInt.ts
```diff
@@ -17,32 +17,44 @@
...
+import { GraphQLError, GraphQLScalarType, Kind, ValueNode } from "graphql";
+import { int, isInt } from "neo4j-driver";
 
 export default new GraphQLScalarType({
     name: "BigInt",
     description:
         "A BigInt value up to 64 bits in size, which can be a number or a string if used inline, or a string only if used as a variable. Always returned as a string.",
-    serialize(value: Integer) {
-        return value.toString(10);
+    serialize(outputValue: any) {
```
angrykoala, line 27: We may want to consider start using `unknown` here and in `parseValue`

```suggestion
    serialize(outputValue: unknown) {
```

packages/graphql/tests/integration/interface-relationships/read.int.test.ts
```diff
@@ -125,7 +125,7 @@ describe("interface relationships", () => {
 
             expect(gqlResult.errors).toBeFalsy();
 
-            expect(gqlResult.data?.actors[0].actedIn).toHaveLength(2);
+            expect((gqlResult.data as any)?.actors[0].actedIn).toHaveLength(2);
```
angrykoala, line 128: I see a lot of these changes in the tests, anyway we can improve this so tests are not filled with these?.

maybe a test utility function that takes care of that when running graphql? something akin to this:
```ts
function runGraphql(options) {
      return graphql(options) as Promise<{errors?: Error, data: any}>
}

```

angrykoala (approved):
A few comments on easy possibilities to improve typings, but overall all looking good :+1:

packages/graphql/tests/integration/nested-unions.int.test.ts
```diff
@@ -466,25 +466,9 @@ describe("Nested unions", () => {
                 contextValue: { driver, driverConfig: { bookmarks: session.lastBookmark() } },
             });
             expect(gqlResult.errors).toBeFalsy();
-            // expect(gqlResult.data?.createMovies.movies).toEqual([
```
danstarns, line 469: Nice!

packages/graphql/tests/integration/interface-relationships/read.int.test.ts
danstarns: I agree that some utility like this would be useful though, It may not be necessary to add it to this PR.

danstarns (approved):
Cool! I thought this work was going to be much larger, and so nice work on proving otherwise.

angrykoala (approved):

darrellwarde:
@danstarns @angrykoala - final approach - `graphql` peer dependency at `^16.0.0` for now, but `graphql@^15.0.0` being tested as follows:

- Unit tests
- TCK tests
- Integration tests
- Next.js test
- Package test

That way, if we get a call to add official support for `^15.0.0` back in, we are likely be able to do so. Note that it is only a peer dependency, so users can technically use 15 anyway, we just don't endorse it.

Issue #595, opened 2021-11-17
Missing optional arguments on cypherFields are not passed to cypher query causing runtime error


rcbevans:
Given the following simple schema:

```gql
type Account {
    id: ID!
    name: String!
}

input ListAccountOptions {
    offset: Int
    limit: Int
}

type Query {
    listAccounts(options: ListAccountOptions): [Account!]!
        @cypher(
            statement: "MATCH (accounts:Account) RETURN accounts SKIP coalesce($options.offset, 0) LIMIT coalesce($options.limit, 30)"
        )
}
```

Executing the GraphQL query

```gql
query MyListAccounts {
    listAccounts {
        id
        name
    }
}
```

The library does not pass an `options` parameter as part of the query execution, causing the execution to throw a runtime error: 

>Failed to invoke function `apoc.cypher.runFirstColumn`: Caused by: org.neo4j.exceptions.ParameterNotFoundException: Expected parameter(s): options

Unfortunately, it doesn't seem to be possible to coalesce non-existent parameters to a value either, so starting the statement with

```
WITH coalesce($options, NULL) AS options
```

still throws the same error.

As a consequence, it doesn't seem possible to have cypher fields with optional arguments.

maieulChevalier:
As a workaround, I think you can provide `null` to the optional arguments. It seems to work for simple arguments at least.

For example, given the following type defs:

``` graphql
type Post {
  title: String!
  text: String
}

type Mutation {
  createPost(title:String!, text:String): Post!
        @cypher(
            statement: "CREATE (p:Post{title:$title, text:$text) RETURN p"
        )
}
```
Executing following mutation

```graphql
mutation {
  createPost(title:"neo4j/graphql is cool") {
    title
  }
}
```
throws a **similar** error (It's not `apoc.cypher.runFirstColumn` but `apoc.cypher.doIt`)

> Failed to invoke procedure `apoc.cypher.doIt`: Caused by: org.neo4j.exceptions.ParameterNotFoundException: Expected parameter(s): text

But if you add `text:null`

```graphql
mutation {
  createPost(title:"neo4j/graphql is cool", text:null) {
    title
  }
}
```

The mutation works. Though having to add `null` to all optional arguments doesn't result in a good developer experience. 

Right now I rely on multiple almost identical queries and mutations with the `@cypher directive` to achieve unsupported features, where I could have only a few queries and mutations with the `@cypher directive` if we could use optional arguments/parameters.

dmoree:
[...]

Issue #444, opened 2021-08-20
Null default values with cypher directive


danstarns:
**Describe the bug**
When the default value is `null` it's not passed through to cypher. If you want to use the null variable in your statement error is thrown. 

**Type definitions**
```gql
type Destination {
   ...
   preposition(posId: Int!, caseName: String = null): String @cypher(statement: """
      MATCH... # if you want to use $caseName error will be thrown
   """)
}
```

**To Reproduce**

Run the following Query...
```gql
query {
    destinations {
        preposition(posId: "some-id")
    }
}
```

See error

```
{
  "errors": [
    {
      "message": "Failed to invoke function `apoc.cypher.runFirstColumn`: Caused by: org.neo4j.exceptions.ParameterNotFoundException: Expected parameter(s): caseName",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "destinations"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "code": "Neo.ClientError.Procedure.ProcedureCallFailed",
          "name": "Neo4jError",
          "stacktrace": [
            "Neo4jError: Failed to invoke function `apoc.cypher.runFirstColumn`: Caused by: org.neo4j.exceptions.ParameterNotFoundException: Expected parameter(s): caseName",
            ": ",
            "    at captureStacktrace (C:\\code\\neo4j\\graphql\\node_modules\\neo4j-driver-core\\lib\\result.js:239:17)",
            "    at new Result (C:\\code\\neo4j\\graphql\\node_modules\\neo4j-driver-core\\lib\\result.js:59:23)",
            "    at newCompletedResult (C:\\code\\neo4j\\graphql\\node_modules\\neo4j-driver-core\\lib\\transaction.js:372:12)",
            "    at Object.run (C:\\code\\neo4j\\graphql\\node_modules\\neo4j-driver-core\\lib\\transaction.js:226:20)",
            "    at Transaction.run (C:\\code\\neo4j\\graphql\\node_modules\\neo4j-driver-core\\lib\\transaction.js:98:34)",
            "    at C:\\code\\neo4j\\graphql\\packages\\graphql\\src\\utils\\execute.ts:88:16",
            "    at TransactionExecutor._safeExecuteTransactionWork (C:\\code\\neo4j\\graphql\\node_modules\\neo4j-driver-core\\lib\\internal\\transaction-executor.js:92:26)",
[...]

PR #769, opened 2022-01-10
Fix: Missing or Default `null` `@cypher` Arguments 

dmoree:
# Description

Proposed fix for #444 and #595 wherein nullable arguments of the cypher directive that are either missing or have a default value of `null` are not passed as parameters into the `apoc` procedure thereby causing a runtime error to occur.

This PR adds an arbitrary key-value pair of `{ [argument]: null }` to the arguments of the resolve tree in order to generate the correct `apocParam.strs` used in the `apoc` procedure generated by either a top-level query or projection.

# Issue

- #444 
- #595 

# Checklist


darrellwarde:
Please ignore the ESLint violations being reported by Reviewdog - we're having a play with the ESLint rules for this as we speak.

packages/graphql/src/schema/resolvers/cypher.ts
```diff
@@ -153,13 +153,25 @@ export default function cypherResolver({
...
+        const nullArgumentValues = field.arguments.reduce(
+            (res, argument) => ({
+                ...res,
+                ...{ [argument.name.value]: null },
+            }),
+            {}
+        );
+
+        // Overwrite null arguments with those defined in resolve tree
+        const argumentsWithNullValues = { ...nullArgumentValues, ...resolveTree.args };
```
darrellwarde, line 168: I don't know why, but I find this variable name very confusing, it sounds like "all of the arguments in this object have a null value". Maybe just `arguments` if that's not already declared in this scope?

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -265,7 +265,21 @@ function createProjectionAndParams({
...
+            const nullArgumentValues = cypherField.arguments.reduce(
+                (r, argument) => ({
+                    ...r,
+                    ...{ [argument.name.value]: null },
+                }),
+                {}
+            );
+
+            // Overwrite null arguments with those defined in resolve tree
+            const argumentsWithNullValues = { ...nullArgumentValues, ...field.args };
```
darrellwarde, line 280: Same comment on this variable naming as in the other file.

darrellwarde (changes requested):
Brilliant, this will make a lot of people very happy. Just a small comment about the naming of a single variable, but generally looks great!

It would be even better if we could see a little integration test of this working so that we can have 100% confidence that this works end-to-end if you don't mind?

packages/graphql/src/schema/resolvers/cypher.ts
dmoree: Sorry, it does look confusing. Unfortunately, `arguments` is a reserved word in javascript and `args` is taken. Perhaps `apocArgs`? I do agree that it should be changed to be clearer.

It is only ever used in generating the `apocParams`. We could skip the entire problem and put it directly in `Object.entries` and let the comment above `nullArgumentValues` do the work.

darrellwarde: Of course it is, easy to forget all of the reserved words!

I think just pop it directly in `Object.entries` given that it's only spreading a couple of objects and not a huge multiline function. 🙂

dmoree: Fixed in https://github.com/neo4j/graphql/pull/769/commits/b27b37efc39a595019b8668b235737d14b7d6b65

dmoree:
Do you mean TCK tests? Integration tests were added in https://github.com/neo4j/graphql/pull/769/commits/2a4c0be4452873f20f65747ff06a88df606369e7

darrellwarde:
> Do you mean TCK tests? Integration tests were added in [2a4c0be](https://github.com/neo4j/graphql/commit/2a4c0be4452873f20f65747ff06a88df606369e7)

It's quite possible that I'm either blind or was just very confused when reviewing multiple PRs yesterday - let me take another look following the removal of that confusing variable. 🙂

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -265,7 +265,21 @@ function createProjectionAndParams({
...
+
+            const apocParams = Object.entries(argumentsWithNullValues).reduce(
```
darrellwarde, line 282: Looks to me that this one still needs moving down into `Object.entries`?

```suggestion
            const apocParams = Object.entries({ ...nullArgumentValues, ...field.args }).reduce(
```

darrellwarde (changes requested):
Just the other `argumentsWithNullValues` needs removing and then I think this is good to go - I added it as a suggestion so should just be able to commit from the UI. 🙂  I don't know why I missed the integration test - I think I'm losing it!

darrellwarde (approved):
Great, this looks good to me! 😄

PR #776, opened 2022-01-11
Upperfirst util

angrykoala:
# Description
This PR creates an `upperFirst` util to reduce dependencies to `graphql-compose`

Additionally, `upperFirst` is no longer exported in the `@neo4j/graphql`. This is a breaking change and requires minor changes to the `OGM` library. This poses a slight duplication of the same util existing on both libraries, this can be further improved by making a "common" utilities folder


darrellwarde (changes requested):

darrellwarde (approved):
Marvellous, thanks for adding those tests!

PR #777, opened 2022-01-11
TypeScript configuration work

darrellwarde:
# Description

There have been some recent issues surrounding TypeScript configuration, particularly with ESLint.

This PR somewhat simplifies and unifies the TypeScript configuration across the packages:

- Upgrades TypeScript to 4.2.4
- Uses official base configurations to reduce boilerplate and improve default configuration
- Replaces `tsconfig.json` in each projects root, `src` and `tests` directories with one at the root for development purposes
- Adds a `tsconfig.production.json` to the root of each package for production builds
  - Note that this differs slightly for the main library which needs access to the `package.json` file - this has `tsconfig.production.json` in `src` for the code, and `tsconfig.package.json` at the root for the `package.json` file

The configuration for ESLint and Jest is now more foolproof because all `src` and `tests` files are included in the root `tsconfig.json` file for each package.


angrykoala (approved):

darrellwarde:
Some slight updates here @angrykoala, can I get a re-review? Changes mainly reflect the need for a production `tsconfig.json` in `packages/graphql/src` due to the fact that that project needs access to the `package.json` file.

angrykoala (approved):

Issue #780, opened 2022-01-12
Bug: Sort fails if field is not selected


dmoree:
#### Describe the bug
When sorting on a node if the field is not in the selection set then the sort fails. See additional context for debugger statements, but the error is coming from the fact that the sort is ultimately on a map that is generated from the projection in the case of top level queries or in an `apoc.coll.sortMulti` in the case of nested relationships.

The solution seems to be looking at the `options` argument and manually adding the fields to the fields in the selection set.

#### Type definitions
Based on `neo-push` example
#### To Reproduce

1. Clone repository
```
git@github.com:dmoree/neo4j-graphql-issue-sort-on-missing-fields.git
```
2. Seed and start server
```
cd neo4j-graphql-issue-sort-on-missing-fields && yarn install && yarn seed && yarn start
```
3. Run the queries in GraphiQL at `http://localhost:4001/graphql` (or as defined in `.env`)

###### Query

```graphql
query SortOnTopLevelWithFieldSelected {
  blogs(options: {sort: {name: DESC}}) {
    id
    name
  }
}
```

###### Response
```
{
  "data": {
    "blogs": [
      {
        "id": "2",
        "name": "b"
      },
      {
        "id": "1",
        "name": "a"
      }
    ]
  }
}
```

###### Query

```graphql
query SortOnTopLevelWithFieldNotSelected {
  blogs(options: {sort: {name: DESC}}) {
    id
  }
}
```

###### Response
```
{
  "data": {
    "blogs": [
      {
        "id": "1"
      },
      {
        "id": "2"
      }
    ]
  }
}
```

###### Query

```graphql
query SortOnNestedLevelWithFieldSelected {
  blogs(where: {id:1}) {
    id
    posts(options: {sort: {title: ASC}}) {
      id
      title
    }
  }
}
```

###### Response
```
{
  "data": {
    "blogs": [
      {
        "id": "1",
        "posts": [
          {
            "id": "3",
            "title": "d"
          },
          {
            "id": "4",
            "title": "e"
          }
        ]
      }
    ]
  }
}
```

###### Query

```graphql
query SortOnNestedLevelWithFieldNotSelected {
  blogs(where: {id:1}) {
    id
    posts(options: {sort: {title: ASC}}) {
      id
    }
  }
}
```

###### Response
```
{
  "data": {
    "blogs": [
      {
        "id": "1",
        "posts": [
          {
            "id": "4"
          },
          {
            "id": "3"
          }
        ]
      }
    ]
  }
[...]

Issue #160, opened 2021-04-14
Field must be included in selection set when sorting in a nested selection


johnymontana:
**Describe the bug**

When sorting a nested selection the ordering is not applied if the field used for ordering is not included in the selection set. Sorting works as expected when the field is included in the selection set.

**Type definitions**

This [Codesandbox](https://codesandbox.io/s/priceless-pasteur-d92b3) can be used to demonstrate the issue.

**To Reproduce**

This query (note that `year` is not included in the `acted_in` selection set:

```GraphQL
{
  movies(where: { title_CONTAINS: "Matrix Reloaded, The" }) {
    title
    actors(options: { limit: 1 }) {
      name
      acted_in(options: { limit: 3, sort: { year: DESC } }) {
        title
      }
    }
  }
}
```

produces this result:

```json
{
  "data": {
    "movies": [
      {
        "title": "Matrix Reloaded, The",
        "actors": [
          {
            "name": "Keanu Reeves",
            "acted_in": [
              {
                "title": "Matrix Reloaded, The"
              },
              {
                "title": "Side by Side"
              },
              {
                "title": "Matrix Revolutions, The"
              }
            ]
          }
        ]
      }
    ]
  }
}
```

however, when `year` is included in the selection set:

```GraphQL
{
  movies(where: { title_CONTAINS: "Matrix Reloaded, The" }) {
    title
    actors(options: { limit: 1 }) {
      name
      acted_in(options: { limit: 3, sort: { year: DESC } }) {
        title
        year
      }
    }
  }
}
```

then ordering of the `acted_in` array is properly sorted in the result:

```json
{
  "data": {
    "movies": [
      {
        "title": "Matrix Reloaded, The",
        "actors": [
          {
            "name": "Keanu Reeves",
            "acted_in": [
              {
                "title": "The Neon Demon",
                "year": 2016
              },
              {
                "title": "John Wick",
                "year": 2014
              },
              {
                "title": "47 Ronin",
                "year": 2013
              }
            ]
          }
        ]
      }
    ]
  }
}
```

**Expected behavior**

I expect ordering to be applied to the nested field even if the field used to order is not included in the selection set.

**System (please complete the following information):**

-   Version:  @neo4j/graphql@1.0.0-beta.2

danstarns:
@johnymontana Thanks for the clarity and steps to reproduce I can produce this locally also.

danstarns:
Digging deeper - If we take the query that's failing: 

```graphql
{
[...]

PR #781, opened 2022-01-12
Fix: Sort On Missing Fields

dmoree:
# Description

Proposed fix for #780 wherein sort fails if the field being sorted on is not in the selection set. This PR looks through the `options` argument and adds a basic resolve tree in order to make sure it is projected. This allows sorting on the resultant map.

# Issue

- #780
- #160

# Checklist


packages/graphql/src/translate/translate-read.ts
```diff
@@ -29,17 +29,33 @@ import translateTopLevelMatch from "./translate-top-level-match";
...
+    );
+
+    sortFields.forEach((sortField) => {
+        if (!Object.values(fieldsByTypeName[node.name]).find((r) => r.name === sortField)) {
+            fieldsByTypeName[node.name] = {
+                ...fieldsByTypeName[node.name],
+                [sortField]: { alias: sortField, args: {}, fieldsByTypeName: {}, name: sortField },
+            };
+        }
+    });
```
darrellwarde, line 57: This is a beautiful hack if I've ever seen one! 😅  Whilst I will admit to modifying the resolve tree in the past, it always gets me a bit on edge. But it's clearly an effective use of it here. I'm assuming there was no way of calculating the missing fields whilst generating the projection? That would likely be the preferred approach.

darrellwarde (commented):
This clearly works very well, just modifying the resolve tree gets me slightly on edge! Take a look at my comment and let me know what you think?

packages/graphql/src/translate/translate-read.ts
dmoree: You're right of course. In hindsight that seems careless and easy to avoid. Fixed in https://github.com/neo4j/graphql/pull/781/commits/e401d5ca0cc424c9204a2e731146bda8032018af

Should `createProjectionAndParams` be refactored to accept a `resolveTree` argument instead of the `fieldsByTypeName` as that would already be in `resolveTree`. It would make it easier to reason about in my opinion and one would have access to `args` as well which would greatly simplify this PR.

dmoree: A refactor is quite trivial as it turns out. Here are the necessary changes:
https://github.com/dmoree/graphql/compare/fix/sort-on-missing-fields...dmoree:refactor/create-projection-and-params

This seems clearer to me for some reason as it is evident that what you are really passing into `createProjectionAndParams` is _a_ resolve tree and when it needs to be modified from above it is quite apparent.  It did in fact simplify this PR as the modification of the fields now resides within `createProjectionAndParams` and not above it in `translateRead` or below in the `relationField` of the reducer. This is because `sort` is simply on the `args` of a resolve tree to be projected on a node.

darrellwarde: Interesting! A `ResolveTree` object certainly feels more sensible to pass around than `FieldsByTypeName`, which feels more like an "internal" type to the `parse-graphql-resolve-info` library. This actually looks like a really sound change, my only interest is https://github.com/dmoree/graphql/compare/fix/sort-on-missing-fields...dmoree:refactor/create-projection-and-params#diff-a9bb71892975416036eb25742d83ed7e1bbf074bf21ff16db77d5f67e98008f2R125 - what makes this call different that `node` needs to be spread into the `resolveTree`?

I would say merge that branch into this so that we can review in full. 🙂

dmoree: That link is broken on my end, but I think you are talking about in `createConnectionAndParams`? As this was simply a refactor of `createProjectionAndParams`, no variable names were changed in this file.

`node` here was defined as a `ResolveTree` coming from the edge of the connection. I would probably rename it, but I just left everything as is for the time being.

darrellwarde (commented):
I'm more than happy with the changed function signature of `createProjectionAndParams` - any strong feelings about this @danstarns and @angrykoala?

Broadly looks very good to me, just a question about one of the calls of `createProjectionAndParams`. 🙂

packages/graphql/src/translate/connection/create-connection-and-params.ts
```diff
@@ -122,7 +122,7 @@ function createConnectionAndParams({
                     };
 
                     const nodeProjectionAndParams = createProjectionAndParams({
-                        fieldsByTypeName: nodeFieldsByTypeName,
+                        resolveTree: { ...node, fieldsByTypeName: nodeFieldsByTypeName },
```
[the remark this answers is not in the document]
darrellwarde: Ignore this, as I have just read your explanation above!

packages/graphql/src/translate/translate-read.ts
darrellwarde: Brilliant, thanks for the explanation - I like this refactor!

darrellwarde (approved):
I think this looks great, the changes to `createProjectionAndParams` are really nice! Thanks @dmoree!

PR #784, opened 2022-01-12
Remove top level count

angrykoala:
# Description

Remove all references to top-level count, as it is now deprecated in favour of aggregate -> count. 
Instead of
```graphql
{
    moviesCount
}
```

The same operation can be achieved with:
```graphql
{
    moviesAggregate {
        count
    }
}
```

* Some top-level count tests have been ported to ensure that aggregate->count support those cases
* Documentation will be updated in a separate PR
* OGM updated accordingly


darrellwarde (approved):
LGTM!

PR #785, opened 2022-01-12
remove references to top-level count in docs

angrykoala:
# Description

Removes references to top-level count and ogm .count in docs. These are the docs updated to changes in #784


darrellwarde (approved):
LGTM!

darrellwarde:
Interesting!! So we disabled the test runs for purely documentation PRs, but GitHub is expecting tests to run, hence this current state. Safe to merge, but we might want to take a look at that.

PR #792, opened 2022-01-13
Eslint fix

angrykoala:
# Description

* Sets `--cache` option in `yarn lint`
* Fix some eslint errors


packages/graphql/src/translate/create-connect-and-params.ts
```diff
@@ -375,7 +375,7 @@ function createConnectAndParams({
         return { subquery: subquery.join("\n"), params };
     }
 
-    function reducer(res: Res, connect: any, index): Res {
+    function reducer(res: Res, connect: any, index:number): Res {
```
darrellwarde, line 378:
```suggestion
    function reducer(res: Res, connect: any, index: number): Res {
```

packages/graphql/src/auth/get-jwt.ts
```diff
@@ -100,28 +100,27 @@ async function getJWT(context: Context): Promise<any> {
...
             },
             (err, decoded) => {
-                return err ? reject(err) : resolve(decoded);
+                if (err) reject(err);
+                else resolve(decoded);
             }
-        )
-    );
+        );
+    });
```
darrellwarde, line 125: Is this right? Having an arrow function with curly braces but no return value doesn't quite seem correct to me.

angrykoala: This is actually the fix to an eslint rule of a promise returning. within the promise callback, the return value is irrelevant, as the key is the resolve/reject callbacks

darrellwarde: So that ESLint rule would have been referring to the return on line 121 (`return resolve(decoded)`) for example. I'm talking about line 113/114 - shouldn't it be `return jsonwebtoken.verify(`?

angrykoala: No, the callback inside a Promise (`new Promise...`) does not return anything, the key part is the resolve/reject that are called within the `.verify` callback. Eslint complains about that one, keep in mind that the callback inside .verify is not really the promise callback, is a plain one

darrellwarde: Interesting! Thanks for the background. 🙂

Issue #144, opened 2021-03-05
Proposal: add support for relation filters _some, _none, _single and _every


Andy2003:
In the [java version](https://github.com/neo4j-graphql/neo4j-graphql-java/blob/master/core/src/main/kotlin/org/neo4j/graphql/Predicates.kt#L198-L209) we currently support the relation filters `_some`, `_none`, `_single` and `_every`

As stated by @danstarns in slack:

> Good point - `_single` and `_some` are missing. We removed `_none` because its the same as:
> ```graphql
> {
>   User(filter: { posts_not: { id: 1 } }) {
>     id
>   }
> }
> ```
> And `_every` is the same as:
> ```graphql
> {
>   User(filter: { posts: { id: 1 } }) {
>     id
>   }
> }
> ```

---

I would suggest to distinguish between `n..1` and `n..m` relations:

### Filter for `n..m`-relations

I realized that users may use the `_every` but expecting it to act like the `_some` filter:
```graphql
{
  User(filter: { posts: { id: 1 } }) {
    id
  }
}
```

That's why I would vote for make all 4 filters explicit:

- `_every`: Filters only those `${type.name}` for which all `${field.name}`-relationships matches this filter
- `_some`: Filters only those `${type.name}` for which at least one `${field.name}`-relationship matches this filter
- `_single`: Filters only those `${type.name}` for which exactly one `${field.name}`-relationship matches this filter
- `_none`: Filters only those `${type.name}` for which none of the `${field.name}`-relationships matches this filter

and remove `<field>_not` and `<field>` for `n..m` relations!
**It's much clearer to the user what he can expect as result, especially with the given description on each field.**

### Filter for `n..1`-relations

The filter above does not make sense for `n..1` relations, so I would keep the current API with `<field>_not` and `<field>` for these types of relations.

wheresrhys:
@darrellwarde is there any decision on whether these will be added, and if so what the timeline is likely to be.

I _think_ it's the only blocker we have to migrating from neo4j-graphql-js

darrellwarde:
No decision yet (there hasn't really been a _massive_ call for it), but it will certainly be compelling if this is one of the last remaining blockers to migration.

Another one of the reasons we haven't gone ahead with this is because we believe that despite some of these absent filters, there are alternative ways of achieving the same outcomes using the already implemented filters.

[...]

PR #736, opened 2022-01-06
Fix: Schema validation for the correct definition of n..m relationships - must be non-nullable lists and entries

danstarns:
# Description

Enforces the validation stated in this RFC : [https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-003-relationship-cardinality.md#invalid-relationship-definition](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-003-relationship-cardinality.md#invalid-relationship-definition)

PR #796, opened 2022-01-15
Feature: Additional Relationship Filters

dmoree:
# Description

This PR implements the additional relationship filters `{EVERY,NONE,SINGLE,SOME}` taken from the proposal in #144. The filters simply map to the cypher list predicates:
| Relationship Filter | Cypher List Predicate    |
|--------------|-----------|
| `EVERY` | `ALL`      |
| `NONE` | `NONE`      |
| `SINGLE` | `SINGLE`      |
| `SOME` | `ANY`      |

It can be difficult or even impossible with a given data model to get the correct data without having the full set of list predicates available in neo4j. @wheresrhys has recently provided two use cases for such filters: https://github.com/neo4j/graphql/issues/144#issuecomment-1012991415 https://github.com/neo4j/graphql/issues/144#issuecomment-1013000782. It also affects updates to nodes where one may want to update all nodes where `ALL` relationships satisfy some filter.

Currently a filter on a relationship field generates the cypher(in a simplified form):
```
EXISTS(pattern)
AND ANY(param IN [ pattern | param ])
```
The only current filter `_NOT` simply changes the list predicate to `NONE`. This leaves no ability to use the remaining two predicates `ALL` or `SINGLE`. This PR is the completion of the list predicates available in neo4j.
# Issue

- #144 

# Checklist


dmoree:
In the original proposal @andy3000 made this comment:
> and remove `<field>_not` and `<field>` for n..m relations!
It's much clearer to the user what he can expect as result, especially with the given description on each field.

This was not done for backward compatibility, but all four filters were added in order to be specific with filters.

darrellwarde:
Cool! This feels much needed and has been an open issue for a long time, many thanks for the useful contribution as always. A few comments below.

## Naming

A bit on naming - I think the naming is good for the most part, but personally, I would like to see `_EVERY` renamed to `_ALL` as I think this reads more naturally.

It would be nice if the chosen naming mapped into the same sentence structure, "where <predicate> of the <plural-type>", for example:

- "where ALL of the Actors"
- "where SOME of the Actors"
- "where NONE of the Actors"
- "where ONE of the Actors"

However, I think `SINGLE` is most appropriate choice for the last one because a field `actors_ONE` I think wouldn't make much sense.

## Field descriptions

With the above sentences in mind, I would then like to see a field description in the schema for each of these new fields, because I think relationship filtering might become a bit intimidating without some help. It would be super cool if these could be specific to each type, for example:

```gql
type MovieWhere {
    """Return Movies where all of the related Actors match this filter"""
    actors_ALL: ActorWhere
    """Return Movies where some of the related Actors match this filter"""
    actors_SOME: ActorWhere
    """Return Movies where none of the related Actors match this filter"""
    actors_NONE: ActorWhere
    """Return Movies where one of the related Actors matches this filter"""
    actors_SINGLE: ActorWhere
}
```

## Deprecations

With these added filters, I don't think we can maintain the existing filters for `n..m` relationships forever - 6 filters for each relationship is far too many. However, we shouldn't remove them right away, this would be a far too breaking change for pretty much all users.

So can you please mark the existing relationship fields for `n..m` with the following, keeping in mind that they will need to remain for `1..n` relationships:

```gql
type MovieWhere {
    actors: ActorWhere @deprecated(reason: "Use `actors_SOME` instead.")
    actors_NOT: ActorWhere @deprecated(reason: "Use `actors_NONE` instead.")
}
```

These will then be removed at a yet-to-be-determined point in the future.

@danstarns has some work on detecting relationship cardinality in #736 as well as more in the pipeline, so it might be worthwhile holding off until that is merged and can be utilised to distinguish between different relationship cardinalities. 


Thanks again for the hard work @dmoree, it's always appreciated!

dmoree:
I'm not sure how to test the `@deprecated` directive because it seems as if it is being stripped out during testing.

PS. It accidentally got lumped into the description commit, but it is there.

darrellwarde:
I've checked this branch out and generated a simple schema, but the `@deprecated` directive doesn't appear to be present on those fields. I've had a bit of a play around but it's seemingly very difficult to get this to work... Unsure where the problem lies here, but this is definitely something we're going to want to have!

dmoree:
Most likely I am doing it wrong. I had a look at the [documentation for `setFieldDirectiveByName`](https://graphql-compose.github.io/docs/api/InputTypeComposer.html#setfielddirectivebyname) and I thought this would be the correct one to use.

darrellwarde:
> Most likely I am doing it wrong. I had a look at the [documentation for `setFieldDirectiveByName`](https://graphql-compose.github.io/docs/api/InputTypeComposer.html#setfielddirectivebyname) and I thought this would be the correct one to use.

It doesn't look wrong to me! I also tried explicitly adding the `@deprecated` directive definition to the schema and it still didn't work. A confusing one!

dmoree:
I simply did `console.log` on the `parsedDoc` from `make-augmented-schema` and it shows that `@deprecated` has been added to the correct fields. Were you also trying the inline snapshots? I wonder if it is being stripped out of `printedSchema`.

dmoree:
I believe this means that `printSchemaWithDirectives` skips specified directives like `@deprecated`
https://github.com/ardatan/graphql-tools/blob/master/packages/utils/src/print-schema-with-directives.ts#L70-L77

darrellwarde:
> I simply did `console.log` on the `parsedDoc` from `make-augmented-schema` and it shows that `@deprecated` has been added to the correct fields. Were you also trying the inline snapshots? I wonder if it is being stripped out of `printedSchema`.

The mystery thickens!

I was actually creating a server using Apollo Server and checking for the directives in Playground/Sandbox.

Digging into this, I print the type definitions before `makeExecutableSchema` and the directives are there, however printing the schema output from `makeExecutableSchema` and they are not!

Clearly we need to do something in `makeExecutableSchema` to preserve these - this is really bizarre given that they are default directives!

dmoree:
I believe I may have solved the mystery. [According to the latest spec](https://spec.graphql.org/October2021/#sec--deprecated) the definition of `@deprecated` is:

```graphql
directive @deprecated(
  reason: String = "No longer supported"
) on FIELD_DEFINITION | ENUM_VALUE
```

which means that `INPUT_FIELD_DEFINITION` are not a valid location. Therefore, the documentation will have to make clear that these will be deprecated.

darrellwarde:
> I believe I may have solved the mystery. [According to the latest spec](https://spec.graphql.org/October2021/#sec--deprecated) the definition of `@deprecated` is:
> 
> ```graphql
> directive @deprecated(
>   reason: String = "No longer supported"
> ) on FIELD_DEFINITION | ENUM_VALUE
> ```
> 
> which means that `INPUT_FIELD_DEFINITION` is not a valid location. Therefore, the documentation will have to make clear that these will be deprecated.

Interestingly, I did try this with GraphQL 16 installed and it did seem to work!!! I have just merged the GraphQL 16 branch into `dev` - can you merge those changes, do a `yarn install` and try this again? I have a feeling it could possibly work!

dmoree:
You're right. The spec must not have been updated because here is the actual implementation:
https://github.com/graphql/graphql-js/blob/main/src/type/directives.ts#L168-L194

where `INPUT_FIELD_DEFINITION` is valid. I can revert those changes and add it to the tests.

On tests, the schema test `packages/graphql/tests/schema/pluralize-consistency.test.ts` is failing after this change (it takes a few seconds to load):
https://github.com/neo4j/graphql/commit/12fedf42dba98cd81724426ed868b1d54ac75fe2#diff-e21cce676d2a592e4ac9e5d82e1de2de7a650c26a3b212f4578b895d0cc73ce8

This seems to comport with the other changes to `Sort` descriptions. I'm going to revert this change for now to pass the checks.

darrellwarde:
> You're right. The spec will be updated: [graphql/graphql-spec@0e1f9b8](https://github.com/graphql/graphql-spec/commit/0e1f9b8f812de2ebee3cd95395f3c76fd86f3266). I can revert those changes and add it to the tests.

That sounds like a plan to me!

> No idea what happened here: https://github.com/neo4j/graphql/runs/4910828870?check_suite_focus=true

On this one, just update the snapshot. On upgrading to GraphQL 16, `"""` comments now have a newline character around the comment. 🙂

darrellwarde:
Just as a heads up, this PR is apparently missing the license header in the following files:

- packages/graphql/src/schema/create-connection-fields.ts
- packages/graphql/src/translate/create-interface-projection-and-params.ts
- packages/graphql/src/schema/create-relationship-fields/index.ts
- packages/graphql/src/utils/map-to-db-property.ts
- packages/graphql/src/translate/where/utils.ts

This report comes from an automated check we added recently. 🙂

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-796.surge.sh

dmoree:
I just want to flag that there appears to have been a dependency cycle introduced between `createConnectionWhere` and `createNodeWhere` (now `createElementWhere`) in some prior commit.

darrellwarde:
I was in the middle of reviewing this and something just really started to bug me, the API is just getting messier and messier with changes like this.

So I thought about an alternative way in which we could achieve this, and actually reduce the number of relationship filtering fields from two to one. What if to get all movies where the actor was born after 1970 you could do:

```gql
query {
  movies(where: { actors(predicate: ALL): { born_GT: 1970 } }) {
    title
  }
}
```

The thing that bugged us the most is that the nice default `actors` becomes `actors_SOME` as it is now, which just isn't nice to use. With this, `SOME` could be the default value for `predicate`, and changed as appropriate.

I realise that this is a big change to propose at this stage - what do you think @dmoree? Would also like to get your views @Andy2003.

What we'll do with `actors_NOT` as it is now is deprecate it with the `@deprecated` directive saying to use `actors(predicate: NONE)` instead.

The only thing we're not happy with is the argument name `predicate` - open to suggestions here.

Apologies for these proposals at the ninth hour.

Andy2003:
@darrellwarde I like this idea, but this would be inconsistent with all the other (scalar) filters and the user will no longer be able to filter by multiple predicates in one query, like:

```graphql
query {
  movies(where: { actors_some: { born_GT: 1970 } , actors_all: { name: 'Max' } }) {
    title
  }
}
```

Ok this is a really constructed example ;-)

darrellwarde:
> @darrellwarde I like this idea, but this would be inconsistent with all the other (scalar) filters and the user will no longer be able to filter by multiple predicates in one query, like:
> 
> ```graphql
> query {
>   movies(where: { actors_some: { born_GT: 1970 } , actors_all: { name: 'Max' } }) {
>     title
>   }
> }
> ```
> 
> Ok this is a really constructed example ;-)

They wouldn't really be inconsistent as scalars (and their list form) have their own filters which are specific to them.

The example which you gave also wouldn't be possible right now, but would be possible with:

```graphql
query {
  movies(where: { AND: [ { actors_some: { born_GT: 1970 } }, { actors_all: { name: 'Max' } } ] }) {
    title
  }
}
```

So no major problems there.

Andy2003:
> The example which you gave also wouldn't be possible right now, but would be possible with:
> ```graphql
> query {
>   movies(where: { AND: [ { actors: { born_GT: 1970 } }, { actors(predicate: ALL): { name: 'Max' } } ] }) {
>     title
>   }
> }
> ```

you are right, nice one!

> The only thing we're not happy with is the argument name `predicate` - open to suggestions here.

what about `matches` or `requires` ?

dmoree:
I think what @Andy2003 was saying is that all other filters are using a suffix `_FILTER` and this would treat relationships differently by placing a filter within an argument as opposed to appending it at the end. I'm not sure why, but I think I prefer this (the current) approach



> The thing that bugged us the most is that the nice default `actors` becomes `actors_SOME` as it is now, which just isn't nice to use. With this, `SOME` could be the default value for `predicate`, and changed as appropriate.

I can see going the other way as to not lose the nice `actors` clause; I just don't know if `ANY` should be the default. It seems as if it should be `ALL`. I think this may have tripped up people in the past.



> So I thought about an alternative way in which we could achieve this, and actually reduce the number of relationship filtering fields from two to one. What if to get all movies where the actor was born after 1970 you could do:
> 
> ```graphql
> query {
>   movies(where: { actors(predicate: ALL): { born_GT: 1970 } }) {
>     title
>   }
> }
> ```

Maybe `with`?

```graphql
query {
  movies(where: { actors(with: ALL): { born_GT: 1970 } }) {
    title
  }
}
 ```


```graphql
query {
  movies(where: { AND: [ { actors(with: SOME): { born_GT: 1970 } }, { actors(with: ALL): { name: 'Max' } } ] }) {
    title
  }
}
```

dmoree:
> what about `matches` or `requires` ?

I think `matches` or perhaps even `matching` works

> ```graphql
> query {
>   movies(where: { actors(matching: ALL): { born_GT: 1970 } }) {
>     title
>   }
> }
> ```

Andy2003:
> I just don't know if ANY should be the default. It seems as if it should be ALL. I think this may have tripped up people in the past.

In my opinion, there should not be any default, the user should select what he want, as even we are having different expectations of what the default should be.

darrellwarde:
> > what about `matches` or `requires` ?
> 
> I think `matches` or perhaps even `matching` works
> 
> > ```graphql
> > query {
> >   movies(where: { actors(matching: ALL): { born_GT: 1970 } }) {
> >     title
> >   }
> > }
> > ```

Oh, `matching` sings out to me! Thoughts @angrykoala @danstarns @tbwiss?

> In my opinion, there should not be any default, the user should select what he want, as even we are having different expectations of what the default should be.

There should be a default and it should be the same as it is now in terms of behaviour, and this is purely for reasons of easy migration. All of the changes for 3.0.0 so far are server side only. This would introduce a client side change, and from our experience of 2.0.0, that made it to be a very unpleasant release.

dmoree:
> In my opinion, there should not be any default, the user should select what he want, as even we are having different expectations of what the default should be.

I think this is correct because all of our examples are using `query` and it should be remembered that these `where` clauses are also used in `mutations`.

darrellwarde:
> I think what @Andy2003 was saying is that all other filters are using a suffix `_FILTER` and this would treat relationships differently by placing a filter within an argument as opposed to appending it at the end. I'm not sure why, but I think I prefer this (the current) approach

And regarding this, it's actually a really good point. This is inconsistent, and would actually lean us towards having some like `number(operator: LT): 6` for `Int` values, which I think nobody would like!

dmoree:
In a mutation, if you were to `update` all the movies where `actors: { born_GT: 1970 }` the current implementation will update all the movies where `ANY` actors are born after 1970. That doesn't seem right.

By not having a default, the user will have to specify that they want `ANY` or `ALL`.

darrellwarde:
> In a mutation, if you were to `update` all the movies where `actors: { born_GT: 1970 }` the current implementation will update all the movies where `ANY` actors are born after 1970. That doesn't seem right.
> 
> By not having a default, the user will have to specify that they want `ANY` or `ALL`.

Please do see above about breaking _client_ changes for this next release - our users are used to this current behaviour, and giving no default value means that _every single_ client with a Neo4j GraphQL Library server will need to make code changes.

darrellwarde:
Server breaking changes are very easy!! Client side breaking changes provide a lot of pain for everybody involved.

dmoree:
> Please do see above about breaking _client_ changes for this next release - our users are used to this current behaviour, and giving no default value means that _every single_ client with a Neo4j GraphQL Library server will need to make code changes.

This is understandable which is why I think the correct move was to `deprecate` the current fields and give clients the opportunity to move over to more specific filters.

Andy2003:
> Please do see above about breaking client changes for this next release - our users are used to this current behaviour, and giving no default value means that every single client with a Neo4j GraphQL Library server will need to make code changes.

the fields (`_some`, `_any`, ...) currently do not exist, so there is no braking change.
if we go with the `matching`-argument, we cannot make it non-nullable, so we need a default if we don't want to break client code. But we should add a big comment saying, that this argument will become non-nullable in future and the user should explicitly start using a concrete value.

darrellwarde:
> In a mutation, if you were to `update` all the movies where `actors: { born_GT: 1970 }` the current implementation will update all the movies where `ANY` actors are born after 1970. That doesn't seem right.
> 
> By not having a default, the user will have to specify that they want `ANY` or `ALL`.

To quickly give an alternative approach on getting the best of best worlds with no client side breaking changes, we could deal with this with an additional argument to the `@relationship` directive:

```gql
type Movie {
  actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN, defaultFilter: ANY)
}
```

This would result in the default `matching` argument being `ANY`. This means that server developers could change the behaviour of queries and mutations from the server, with no changes needed in the client. We've taken a similar approach with `queryDirection`.

darrellwarde:
Personally, I think the `matching` argument should _always_ have a default, but we let the type definitions decide what that is. There's just something super user friendly about being able to just query an `actors` field!

Andy2003:
> This would result in the default matching argument being ANY. This means that server developers could change the behaviour of queries and mutations from the server, with no changes needed in the client. We've taken a similar approach with queryDirection.

+1

dmoree:
> There's just something super user friendly about being able to just query an `actors` field!

While I do agree, I think we're missing mutations in this discussion which share the same filters. Deprecations are used to warn the client about changing behaviors and being specific with filters will help any new users avoid the pitfalls of the interpretation of `actors`.

While in the schema one can set a `defaultFilter`, a consumer of this GraphQL API may not know that.

darrellwarde:
This has been a really interesting conversation. I think we might have to press pause on this for a short while as this is quite a big decision to make and we want to get it right, rather than rushing it into the 3.0.0 release.

> While in the schema one can set a `defaultFilter`, a consumer of this GraphQL API may not know that. This would have to be documented somewhere and it seems easier to force the client to specify what they are looking to do. Be it querying a field or disconnected nodes, etc.

My thinking here is that really, it's the _server_ owner who cares about the data in the database the most, and by allowing them to set the default behaviour, all clients will abide by that behaviour unless they specifically do otherwise (and we might even want to specify a way to stop this). But can definitely see both sides of the discussion!

dmoree:
> This has been a really interesting conversation. I think we might have to press pause on this for a short while as this is quite a big decision to make and we want to get it right, rather than rushing it into the 3.0.0 release.

I am in favor of working through the semantical issues here in order to get this in the next release. There currently is not way of filtering for `ALL` (or `SINGLE`). Which affects both queries and mutations. Going back to the previous toy example
```graphql
  mutation {
    updateMovies(input: { where: { actors: { born_GT: 1970 } } }) {
      movies {
        actors {
          born
        }
      }
    }
  }
```
This will update all movies where `ANY` actor is born after 1970. In the current version of this library there is no way to only update movies where `ALL` actors are born after 1970 unless I am missing something.

packages/graphql/src/translate/where/utils.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export const whereRegEx =
+    /(?<fieldName>[_A-Za-z]\w*?)(?<isAggregate>Aggregate)?(?:_(?<operator>NOT|NOT_IN|IN|NOT_INCLUDES|INCLUDES|MATCHES|NOT_CONTAINS|CONTAINS|NOT_STARTS_WITH|STARTS_WITH|NOT_ENDS_WITH|ENDS_WITH|LT|LTE|GT|GTE|DISTANCE|ALL|NONE|SINGLE|SOME))?$/;
+export type WhereRegexGroups = {
+    fieldName: string;
+    isAggregate?: string;
+    operator?: WhereOperator;
+};
+
+export const createWhereClause = ({
```
darrellwarde, line 73: **suggestion (non-blocking):** This seems to be an extension of `createFilter` which previously existed, and also appears to be quite a unit testable function, but doesn't appear to have any tests? Even though we have the TCK tests for this, it would be super cool to see some unit tests validating the output of this function!

darrellwarde (changes requested):
I think the conversation on Friday spiralled a little and you're right in that we should get this feature merged for the next release.

There's a few things I would like to see here, but it broadly looks great!

Just as a general note, there was quite a few refactors in this PR which made it quite hard to review _just_ the feature being introduced. We of course welcome these refactors as they improve our code quality, but if at all possible we would like them in their own PR in which we can review the merits of the refactor. Then feature PRs can be used to purely review new features which means we should be able to review and merge them a lot more quickly.

packages/graphql/src/translate/where/utils.ts
dmoree: Added unit tests in https://github.com/neo4j/graphql/pull/796/commits/74a18321c7a5c5d4b84a178cf6ffd15467a4985d

dmoree:
I apologize for the difficult review. I try to front load all refactors to make it easier, but there was quite a bit with this one. I actually forked off of a refactor locally, which probably should have been a cue to initiate a separate PR.

packages/graphql/src/translate/where/create-where-and-params.ts
```diff
@@ -0,0 +1,264 @@
...
+                const labels = refNode.getLabelString(context);
+                const collectedMap = `${thisParam}_map`;
+
+                if (value === null) {
+                    res.clauses.push(
+                        `${isNot ? "" : "NOT "}EXISTS((${varName})${inStr}[:${
+                            connectionField.relationship.type
+                        }]${outStr}(${labels}))`
+                    );
+                    return res;
```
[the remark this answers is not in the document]
dmoree: Nice. Good catch. The other ESLint issue is the dependency cycle introduced in a previous commit.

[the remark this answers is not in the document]
darrellwarde: For sure - we're not 100% enforcing ESLint just yet because as you say, there are a lot of legacy violations getting in the way at the moment. Thanks!

packages/graphql/src/translate/where/create-where-clause.ts
```diff
@@ -0,0 +1,101 @@
...
+    NOT_ENDS_WITH: "ENDS WITH",
+    ENDS_WITH: "ENDS WITH",
+    // Regex
+    MATCHES: "=~",
+    // Array
+    NOT_IN: "IN",
+    IN: "IN",
+    NOT_INCLUDES: "IN",
+    INCLUDES: "IN",
+};
```
darrellwarde, line 46: This is now declared twice - here and also in `packages/graphql/src/translate/where/utils.ts`. Can you choose the one which is most appropriate?

darrellwarde (changes requested):
This is basically approved now - just this one oversight needs looking at before merging. 🙂

packages/graphql/src/translate/where/create-where-clause.ts
dmoree: I'm really falling down here. Sorry. Fix in https://github.com/neo4j/graphql/pull/796/commits/f0118d1cf2e1bc0fc6c2fb2e99c3027e7c38fdee

darrellwarde (approved):

packages/graphql/src/translate/where/create-where-clause.ts
darrellwarde: Not at all, this has been open for some time, it's easily done!

Let's get this merged. 🙂

PR #813, opened 2022-01-18
Feat/Validate Relationships In Cypher

danstarns:
# Description

This is the work to add Cypher checks for relationship cardinality. The work in this PR is based on [RFC 003](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-003-relationship-cardinality.md). 

### DIFF
--- a/packages/graphql/src/translate/create-relationship-validation-string.ts
+++ b/packages/graphql/src/translate/create-relationship-validation-string.ts
@@ -1,0 +1,57 @@
+import { Node } from "../classes";
+import { RELATIONSHIP_REQUIREMENT_PREFIX } from "../constants";
+import { Context } from "../types";
+
+function createRelationshipValidationString({
+    node,
+    context,
+    varName,
+}: {
+    node: Node;
+    context: Context;
+    varName: string;
+}): string {
+    let relationshipValidationStr = "";
+
+    const nonNullRelationFields = node.relationFields.filter((field) => {
+        if (field.typeMeta.array) {
+            return false;
+        }
+
+        const isRequired = field.typeMeta.required;
+        const isNotUnionOrInterface = Boolean(field.union) || Boolean(field.interface);
+
+        return isRequired && !isNotUnionOrInterface;
+    });
+
+    if (!nonNullRelationFields.length) {
+        return relationshipValidationStr;
+    }
+
+    const nonNullPredicates = nonNullRelationFields.map((field) => {
+        const toNode = context.neoSchema.nodes.find((n) => n.name === field.typeMeta.name) as Node;
+
+        const inStr = field.direction === "IN" ? "<-" : "-";
+        const outStr = field.direction === "OUT" ? "->" : "-";
+        const relTypeStr = `[:${field.type}]`;
+
+        const pathMatchStr = `MATCH p=(${varName})${inStr}${relTypeStr}${outStr}(${toNode.getLabelString(context)})`;
+        const countNodesStr = `count(nodes(p))`;
+        const relationshipLength = 1;
+        const comparisonOperator = `=`;
+        const comparison = `${countNodesStr} ${comparisonOperator} ${relationshipLength}`;
+        const apocRunFirstCol = `
+            apoc.cypher.runFirstColumn('${pathMatchStr}\nRETURN ${comparison}', { ${varName}: ${varName} }, false)
+        `;
+
+        return `apoc.util.validatePredicate(NOT(${apocRunFirstCol}), '${RELATIONSHIP_REQUIREMENT_PREFIX}${node.name}.${field.fieldName} required', [0])`;
+    });
+
+    relationshipValidationStr = `CALL apoc.util.validate(NOT(${nonNullPredicates.join(
+        " AND "
+    )}), '${RELATIONSHIP_REQUIREMENT_PREFIX}', [0])`;
+
+    return relationshipValidationStr;
+}
+
+export default createRelationshipValidationString;
--- a/packages/graphql/tests/integration/update.int.test.ts
+++ b/packages/graphql/tests/integration/update.int.test.ts
@@ -384,13 +384,16 @@ describe("update", () => {
             charset: "alphabetic",
         });
 
-        const name = generate({
+        const actorName1 = generate({
+            charset: "alphabetic",
+        });
+        const actorName2 = generate({
             charset: "alphabetic",
         });
 
         const mutation = `
-            mutation($id: ID, $name: String) {
-                updateMovies(where: { id: $id }, delete: { actors: { where: { node: { name: $name } } } }) {
+            mutation($id: ID, $actorName1: String) {
+                updateMovies(where: { id: $id }, delete: { actors: { where: { node: { name: $actorName1 } } } }) {
                     movies {
                         id
                         actors {
@@ -405,26 +408,29 @@ describe("update", () => {
             await session.run(
                 `
                 CREATE (m:Movie {id: $id})
-                CREATE (a:Actor {name: $name})
-                MERGE (a)-[:ACTED_IN]->(m)
+                CREATE (a1:Actor {name: $actorName1})
+                CREATE (a2:Actor {name: $actorName2})
+                MERGE (a1)-[:ACTED_IN]->(m)
+                MERGE (a2)-[:ACTED_IN]->(m)
             `,
                 {
                     id,
-                    name,
+                    actorName1,
+                    actorName2,
                 }
             );
 
             const gqlResult = await graphql({
                 schema: neoSchema.schema,
                 source: mutation,
-                variableValues: { id, name },
+                variableValues: { id, actorName1 },
                 contextValue: { driver, driverConfig: { bookmarks: session.lastBookmark() } },
             });
 
             expect(gqlResult.errors).toBeFalsy();
 
             expect(gqlResult?.data?.updateMovies).toEqual({
-                movies: [{ id, actors: [] }],
+                movies: [{ id, actors: [{ name: actorName2 }] }],
             });
         } finally {
             await session.close();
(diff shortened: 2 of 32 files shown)


danstarns:
https://github.com/neo4j/graphql/pull/814

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-813.surge.sh

docs/modules/ROOT/pages/troubleshooting/faqs.adoc
```diff
@@ -34,3 +34,57 @@ type Cookie {
     # f: String # If you don't want _emptyInput, uncomment this line.
 }
 ```
+
+== Relationship nullability isn't being enforced in my graph
+Currently, Neo4j GraphQL enforces on creating and updating that a movie has a director:
```
angrykoala, line 39: An example has not been set here, so `enforce ... that a movie has a director` may not be easy to understand, Maybe use a more generic description and add this referencing the example below?

packages/graphql/tests/integration/update.int.test.ts
```diff
@@ -897,7 +920,7 @@ describe("update", () => {
         }
     });
 
-    test("!", async () => {
```
angrykoala, line 900: :+1:

docs/modules/ROOT/pages/troubleshooting/faqs.adoc
danstarns: Sure!

darrellwarde:
> What I have done is prepend a subquery after each `create` and `update` operation, this includes nested too. The subquery will match a path based on the required relationship and then count the nodes in that path. If the length of nodes is not `>=` 1 then an error is thrown.

Just to confirm, this has been updated to include `delete` Mutations as well now?

darrellwarde:
Disregard hidden comment, I remember we discussed that this would be impossible (or very difficult).

darrellwarde (changes requested):
Broadly looks good to me, I would just like to see a few more tests:

- The following throw if count would be over 0:
  - create -> create
  - create -> connect
  - update -> create
  - update -> connect
  - create -> connectOrCreate
  - update -> connectOrCreate
- The following throw if count would be zero:
  - update -> delete
  - update -> disconnect

Seems like a lot writing it out, but we've definitely been tripped up for not having full coverage in the past.

Additionally, the tests added specifically for this functionality are all for negative cases (error would be thrown). I would hope that all positive cases will be caught by tests elsewhere in our tests, but worth thinking about anything that might have been missed.

packages/graphql/src/translate/create-relationship-validation-string.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    const nonNullPredicates = nonNullRelationFields.map((field) => {
+        const toNode = context.neoSchema.nodes.find((n) => n.name === field.typeMeta.name) as Node;
+        const inStr = field.direction === "IN" ? "<-" : "-";
+        const outStr = field.direction === "OUT" ? "->" : "-";
+        const relVarname = `${varName}_${field.fieldName}_${toNode.name}_unique`;
+
+        const subQuery = [
+            `CALL {`,
```
darrellwarde, line 55: **question:** Could there be issues by not prepending `WITH` before each subquery here, rather than prepending all relationship validation with a single `WITH`?

For instance, taking these type definitions:

```gql
type Director {
    id: ID!
}

type Genre {
    name: String!
}

type Movie {
    id: ID!
    director: Director! @relationship(type: "DIRECTED", direction: IN)
    genre: Genre! @relationship(type: "HAS_GENRE", direction: OUT)
}
```

And executing:

```gql
mutation {
  updateMovies(
    where: {id: "${movieId}"}
    disconnect: {director: {where: {node: {id: "${directorId1}"}}}}
    connect: {genre: {where: {node: {name: "${directorId2}"}}}}
  ) {
    movies {
      id
      director {
        id
      }
    }
  }
}
```

You get the following Cypher:

```
MATCH (this:Movie)
WHERE this.id = $this_id
WITH this
CALL {
        WITH this
        OPTIONAL MATCH (this_connect_genre0_node:Genre)
        WHERE this_connect_genre0_node.name = $this_connect_genre0_node_name
        FOREACH(_ IN CASE this WHEN NULL THEN [] ELSE [1] END | 
                FOREACH(_ IN CASE this_connect_genre0_node WHEN NULL THEN [] ELSE [1] END | 
                        MERGE (this)-[:HAS_GENRE]->(this_connect_genre0_node)
                )
        )
        RETURN count(*)
}
WITH this
CALL {
WITH this
OPTIONAL MATCH (this)<-[this_disconnect_director0_rel:DIRECTED]-(this_disconnect_director0:Director)
WHERE this_disconnect_director0.id = $updateMovies.args.disconnect.director.where.node.id
FOREACH(_ IN CASE this_disconnect_director0 WHEN NULL THEN [] ELSE [1] END | 
DELETE this_disconnect_director0_rel
)
RETURN count(*)
}
WITH this
CALL {
        WITH this
        MATCH (this)<-[this_director_Director_unique:DIRECTED]-(:Director)
        WITH count(this_director_Director_unique) as c
        CALL apoc.util.validate(NOT(c = 1), '@neo4j/graphql/RELATIONSHIP-REQUIREDMovie.director required', [0])
        RETURN c AS this_director_Director_unique_ignored
}
CALL {
        WITH this
        MATCH (this)-[this_genre_Genre_unique:HAS_GENRE]->(:Genre)
        WITH count(this_genre_Genre_unique) as c
        CALL apoc.util.validate(NOT(c = 1), '@neo4j/graphql/RELATIONSHIP-REQUIREDMovie.genre required', [0])
        RETURN c AS this_genre_Genre_unique_ignored
}
RETURN this { .id, director: head([ (this)<-[:DIRECTED]-(this_director:Director)   | this_director { .id } ]) } AS this
```

Note the following:

```
WITH this
CALL {
        WITH this
        MATCH (this)<-[this_director_Director_unique:DIRECTED]-(:Director)
        WITH count(this_director_Director_unique) as c
        CALL apoc.util.validate(NOT(c = 1), '@neo4j/graphql/RELATIONSHIP-REQUIREDMovie.director required', [0])
        RETURN c AS this_director_Director_unique_ignored
}
CALL {
        WITH this
        MATCH (this)-[this_genre_Genre_unique:HAS_GENRE]->(:Genre)
        WITH count(this_genre_Genre_unique) as c
        CALL apoc.util.validate(NOT(c = 1), '@neo4j/graphql/RELATIONSHIP-REQUIREDMovie.genre required', [0])
        RETURN c AS this_genre_Genre_unique_ignored
}
```

This has caused us issues in the past (I'm actually surprised this runs) because subqueries generally always need to be prepended with `WITH`. Potentially instead return a `WITH` here?

```suggestion
        const subQuery = [
            `WITH ${varName}`,
            `CALL {`,
```

darrellwarde (approved):
Broadly looks good to me, nice work! I do just have one question which I would like to be looked at before merging. 🙂

darrellwarde:
I've just been thinking about this, and had a thought - are we validating that _nullable_ relationships count cannot be _more than_ 1? For instance:

```gql
type Movie {
  title: String!
  director: Director @relationship(type: "DIRECTED", direction: IN)
}
```

I'm just thinking about this now because I only remember seeing validation of count _equals_ 1 for non-nullable relationships, and not checking that the count isn't _over_ 1 for nullable relationships.

Feel free to correct me if I'm wrong!

darrellwarde:
> > I've just been thinking about this, and had a thought - are we validating that _nullable_ relationships count cannot be _more than_ 1? For instance:
> > ```graphql
> > type Movie {
> >   title: String!
> >   director: Director @relationship(type: "DIRECTED", direction: IN)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm just thinking about this now because I only remember seeing validation of count _equals_ 1 for non-nullable relationships, and not checking that the count isn't _over_ 1 for nullable relationships.
> > Feel free to correct me if I'm wrong!
> 
> Currently, no validation is applied to the director relationship if it does not have an `!`

Okey doke! I think we definitely get this merged anyway, let's chat about this tomorrow!

danstarns:
> > > I've just been thinking about this, and had a thought - are we validating that _nullable_ relationships count cannot be _more than_ 1? For instance:
> > > ```graphql
> > > type Movie {
> > >   title: String!
> > >   director: Director @relationship(type: "DIRECTED", direction: IN)
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I'm just thinking about this now because I only remember seeing validation of count _equals_ 1 for non-nullable relationships, and not checking that the count isn't _over_ 1 for nullable relationships.
> > > Feel free to correct me if I'm wrong!
> > 
> > 
> > Currently, no validation is applied to the director relationship if it does not have an `!`
> 
> Okey doke! I think we definitely get this merged anyway, let's chat about this tomorrow!

I have just re-merged this and if it passes I will merge into itdev. I agree with your point about validating greater than 1 and let's do this in another PR!

packages/graphql/src/translate/create-relationship-validation-string.ts
danstarns: This will be fine if there is nothing in between the CALL statements!

packages/graphql/src/translate/create-disconnect-and-params.ts
```diff
@@ -152,13 +152,15 @@ function createDisconnectAndParams({
...
-           Replace with subclauses https://neo4j.com/developer/kb/conditional-cypher-execution/
-           https://neo4j.slack.com/archives/C02PUHA7C/p1603458561099100
+        Replace with subclauses https://neo4j.com/developer/kb/conditional-cypher-execution/
+        https://neo4j.slack.com/archives/C02PUHA7C/p1603458561099100
         */
         subquery.push(`FOREACH(_ IN CASE ${_varName} WHEN NULL THEN [] ELSE [1] END | `);
         subquery.push(`DELETE ${_varName}_rel`);
         subquery.push(`)`); // close FOREACH
 
+        // TODO - relationship validation - Blocking, if this were to be enforced it would stop someone from 'reconnecting'
```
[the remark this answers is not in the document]
danstarns: It is still valid because this is a potential gap in the rel validation!

PR #814, opened 2022-01-18
RFC/Relationship Count Validation

danstarns:


docs/rfcs/relationship-length-validation.md
```diff
@@ -0,0 +1,43 @@
...
+
+Expose new `validation` parameter to the `@relationship` directive, where the following can be specified:
+
+1. `min`
+2. `max`
+3. `*_GT`
+4. `*_GTE`
+5. `*_LT`
+6. `*_LTE`
+7. `EQUAL`
```
darrellwarde, line 17: We don't need all of these arguments, as `max` is actually equivalent to `LT` - you could actually just have `min` and `max` - a few examples:

- Must have count of 3 - `min: 3, max: 3`
- Maximum of 3 - `max: 3`
- Greater than 1, less than or equal to 5 - `min: 1, max: 5`

docs/rfcs/relationship-length-validation.md
```diff
@@ -0,0 +1,43 @@
...
+
+For example:
+
+"Movies must have at least one Actor":
+
+```gql
+type Movie {
+    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: OUT, validation: { min_GTE: 1 })
```
darrellwarde, line 25: `count` instead of `validation`?

darrellwarde (changes requested):

darrellwarde (approved):

Issue #764, opened 2022-01-10
Augmented schema is missing description for relationship fields 


Roei-Levi:
**Describe the bug**
Every field in a GraphQL schema can have a description associated with it. However, the augmented schema doesn't seem to include the description on fields with the "relationship" directive specified to them. 

**Type definitions**
```graphql
type Application {
	"Application ID"
	id: ID

	**"Related Object Description"
	object: [Service] @relationship(type: "has_object", direction: OUT)**
}
type Service{
	"Service ID"
	id: ID
}
```

**To Reproduce**
Run the server with a GraphQL schema similar to the one above. The description property is missing from the "object" field. 

**Expected behavior**
The description "Related Object Description" should be shown in the graphQL documentation of the "object" field. 

**Additional context**
I presume that the description is dropped when the function createRelationshipFields() is executed to create relationship fields.

neo4j-team-graphql:
Many thanks for raising this bug report @Roei-Levi. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @Roei-Levi! :pray: We will now prioritise the bug and address it appropriately.

PR #820, opened 2022-01-19
Fix: Missing Relationship Description

dmoree:
# Description

This PR addresses #764 by adding the description field.

# Issue

- #764 

# Checklist

### DIFF
--- a/packages/graphql/src/schema/create-relationship-fields/index.ts
+++ b/packages/graphql/src/schema/create-relationship-fields/index.ts
@@ -81,6 +81,7 @@ function createRelationshipFields({
                             options: "QueryOptions",
                             where: `${rel.typeMeta.name}Where`,
                         },
+                        description: rel.description,
                     },
                 });
             }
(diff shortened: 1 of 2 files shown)


dmoree:
@darrellwarde There should probably be a decision on whether to add a description to the generated `{field}Connection` or even the `{field}Aggregate` field and what that would look like. I would be happy to include that in this PR if desired.

darrellwarde:
> @darrellwarde There should probably be a decision on whether to add a description to the generated `{field}Connection` or even the `{field}Aggregate` field and what that would look like. I would be happy to include that in this PR if desired.

Tricky one that! What would it look like? If description is not `null`, output (using your test) "Actors in Movie (Relay Connection field)" and "Actors in Movie (Aggregate values)"?

I think we should just address the bug at hand - the purpose of the other two fields can be inferred from the description on the "main" relationship field in my opinion.

Thanks for this work.

darrellwarde (approved):
LGTM!

Issue #619, opened 2021-12-02
input.map is not a function (with connectOrCreate)


danielfein:
Hi thanks for releasing 2.5.0, really exciting and trying to get going with it. I have a fairly complicated query, and when I run it it's currently giving me this error:

```json
 {
      "message": "input.map is not a function",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "updateWidgets"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: input.map is not a function",
            "    at Object.createConnectOrCreateAndParams 
....
```

And here's my query. 

It's a little complicated but here's what the deal is:

The node types are: `Fact`, `Topic`, `User`, and `Vote`.

They connect like this:

```
(t:Topic)-[:VOTE_HAS_THING]-(v:Vote)-[:VOTE_HAS_FACT]-(d:Detail)
(u:User)-[:VOTED]-(v:Vote)
```
So below I'm trying to say: take this Detail that I have, and if I have a Vote ID then find that and add a new user_votes. If I don't have one, onCreate then it should add the `type`, and connect `Topic`, `Fact`, and to `User`. I'm not positive it will add the user_votes relationship if there is a connect, I would have imagined there would have been an `onConnect` function but I don't see that one exists. Any idea what this error message is from?

```gql
mutation {
updateFacts(
  where: { id: "5d081f2e-7663-4ac1-9c18-f087c35b7042"},
  connectOrCreate: {
    votes: {
      where: { node: { id: "lmnop"}},
      onCreate: { 
        node: { 
          type: "detailVote",
          topic: { connect: { where: { node: { id: "21732715-2aa2-4c4b-9c45-ea4985f2113c" }}}},
          fact: { connect: { where: { node: { id: "5d081f2e-7663-4ac1-9c18-f087c35b7042" }}}},
          user_votes: { connect: {where: {node: { id: "YSCbEHFxm0gnXOPEKmRb8WUVO7j2"}}}}
        }
      }
    }
  }
)
  {
    facts	{
      id
      dataValue
      votes	{
        id
         user_votes{
          id
          name
        }
      }
    }
  }

  
}```

mobsean:
I can confirm that bug. Also getting `input.map is not a function`

`"TypeError: input.map is not a function",
[...]

PR #824, opened 2022-01-19
Fix/619 input.map is not a function error in connectOrCreate

angrykoala:
# Description

This PR fixes an error when `connectOrCreate` is used in 1-to-1 relationships

# Issue

#619


darrellwarde (approved):
Looks great!

Issue #582, opened 2021-11-12
Different results for relationship query


tonxxd:
**Describe the bug**
I have a query that tries to filter nodes based on a `type` field of children nodes recursively
e.g 
[Cat] --> [Dog] <-- [Bird]

Now, I want to get all Cats where there exists at least one child Dog that has a Bird parent
My where clause looks like
```json
{
  "where": {
   "type": "Cat",
      "children": {
          "type": "Dog",
          "parents": {
             "type":"Bird"
          } 
       }
    }
}
```
It works fine.

Though, If I try to use the generated `childrenConnection` query filter (so that I can also use edge properties in the query) I get an empty result

```json
{
    "where": {
        "type": "Cat",
        "childrenConnection": {
            "node": {
                "type": "Dog",
                "parentsConnection": {
                    "node": {
                        "type": "Bird"
                    }
                }
            }
        }
    }
}
```

**Type definitions**
```graphql
type Entity {
    children: [Entity]
      @relationship(type: "EDGE", properties: "Edge", direction: OUT)
    parents: [Entity]
      @relationship(type: "EDGE", properties: "Edge", direction: IN)
    type: String!
}
interface Edge @relationshipProperties {
  type: String!
}
```

**Expected behavior**
They should return the same result(?)

Any idea why the output is different? Are they not equivalent?

tonxxd:
After logging the generated cypher it seems the query is different and that in the second case, instead of generating a nested statement for the `parentsConnection`, the field is treated as non nested

CASE1
```lua
MATCH (this:Entity)
WHERE this.type = $this_type AND EXISTS((this)-[:EDGE]->(:Entity)) 
AND ANY(
    this_children IN [(this)-[:EDGE]->(this_children:Entity) | this_children] 
    WHERE this_children.type = $this_children_type 
    AND EXISTS((this_children)<-[:EDGE]-(:Entity)) 
    AND ANY(
        this_children_parents IN [(this_children)<-[:EDGE]-(this_children_parents:Entity) | this_children_parents] 
        WHERE this_children_parents.type = $this_children_parents_type
    )
)
RETURN this { .type } as this
```

CASE 2
```lua
MATCH (this:Entity)
WHERE this.type = $this_type 
AND EXISTS((this)-[:EDGE]->(:Entity)) 
AND ANY(
[...]

PR #826, opened 2022-01-19
Fix: #582, cannot filter nested connection fields

darrellwarde:
# Description

The logic to perform filtering on nested connection fields was just completely absent. Added in this branch with tests.

# Issue

#582

# Checklist


angrykoala (approved):

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-826.surge.sh

packages/graphql/src/translate/where/create-node-where-and-params.ts
```diff
@@ -60,11 +62,12 @@ function createNodeWhereAndParams({
...
         };
 
-        const re = /(?<field>[_A-Za-z][_0-9A-Za-z]*?)(?:_(?<not>NOT))?(?:_(?<operator>INCLUDES|IN|MATCHES|CONTAINS|STARTS_WITH|ENDS_WITH|LT|GT|GTE|LTE|DISTANCE))?$/gm;
+        const re =
+            /(?<field>[_A-Za-z][_0-9A-Za-z]*?)(?:_(?<not>NOT))?(?:_(?<operator>INCLUDES|IN|MATCHES|CONTAINS|STARTS_WITH|ENDS_WITH|LT|GT|GTE|LTE|DISTANCE))?$/gm;
 
         const match = re.exec(key);
 
-        const fieldName = match?.groups?.field;
+        const fieldName = match?.groups?.field as string;
```
angrykoala, line 70: **suggestion (blocking-ish)**: mmm... I think is great we are adding typings to existing code, but I'm worried, the typings here are lying, as `fieldName` is actually `string | undefined`. If we are **sure** match and groups exist, I would rather use `match!.groups!.field` instead of casting to string (which may silently fail).

If, however, we are trying to handle the case of `match?.groups?` being `undefined`, I think it is better to explicitly define what should happen if `fieldName` is undefined:
```suggestion
        const fieldName = match?.groups?.field;
        if(fieldName===undefined){
            throw new Error(...)
        }
```

packages/graphql/src/translate/where/create-node-where-and-params.ts
```diff
@@ -144,7 +148,76 @@ function createNodeWhereAndParams({
...
+            }
+
+            if (connectionField) {
+                let nodeEntries: Record<string, any> = value;
+
+                if (!connectionField?.relationship.union) {
+                    nodeEntries = { [connectionField.relationship.typeMeta.name]: value };
+                }
+
+                Object.entries(nodeEntries).forEach((entry) => {
```
angrykoala, line 162: **suggestion (non-blocking):** This callback is a prime candidate to be extracted into a (fully) separate function to reduce the complexity of the `reduce` function (we are easily surpassing a cyclomatic complexity of 5 here), size, and scope of its variables.

packages/graphql/src/translate/where/create-node-where-and-params.ts
```diff
@@ -144,7 +148,76 @@ function createNodeWhereAndParams({
...
+                        `RETURN ${existsStr}`,
+                        `AND ${
+                            not ? "NONE" : "ANY"
+                        }(${collectedMap} IN [(${safeNodeVariable})${inStr}[${relationshipVariable}:${
+                            connectionField.relationship.type
+                        }]${outStr}(${relatedNodeVariable}${labels}) | { node: ${relatedNodeVariable}, relationship: ${relationshipVariable} } ] INNER_WHERE `,
+                    ];
+
+                    const connectionWhere = createConnectionWhereAndParams({
+                        whereInput: entry[1] as any,
```
angrykoala, line 198: **question**:  Is there a reason we are not using `GraphQLWhereArg` here? :eyes:
It feels like the recasting of `nodeEntries` to `Record<string, any>` may not be necessary?

angrykoala (changes requested):
Tests and code itself look good :+1: 
I'm a bit concerned about the typings though, I feel we could try to improve them a bit

packages/graphql/src/translate/where/create-node-where-and-params.ts
darrellwarde: This is an interesting one! We have 100% confidence in the match, because we own and control the input.

However, `match!.groups!.field` is currently an ESLint violation (null assertion) - is this a rule we want to disable perhaps?

angrykoala: I'm not sure tbh, if we want to keep the rule (which somehow makes sense) I think we could go with some `assertString` helper, that returns the proper type and fails if it is not a string.

My point is that this should fail if it is not a string imo

darrellwarde: I guess the thing for me in this instance is that it will _always_ be a string - what do we do in that case? Kind of renders a helper redundant no?

angrykoala: well, assertions are for that, ensuring that something that should not happen does not happen
I think is reasonable when typings system comes a bit short

angrykoala (approved):

Issue #635, opened 2021-12-10
OGM generating duplicate names for interfaces/types that include AggregateInput in type name.


Tanner-Scadden:
**Describe the bug**
OGM generate creating duplicate names

**Type definitions**
```gql
enum Role {
    SUPER_ADMIN
    ADMIN
    OWNER
    MANAGER
    EMPLOYEE
  }

  enum FeatureName {
    NONE
  }

  enum BodyStyle {
    MICRO
    SEDAN
    CUV
    SUV
    HATCHBACK
    MINIVAN
    CONVERTABLE
    COUPE
    SUPERCAR
    ROADSTER
    PICKUP
    VAN
    LIMOUSINE
    CAMPERVAN
    TRUCK
  }

  enum TimeClockType {
    IN
    OUT
    BREAK
    LUNCH
  }

  enum IntegrationNames {
    STRIPE
  }

  enum PriceTier {
    FREE
    PREMIUM
    PARTNER
  }
  ## END ENUMS ##

  ## START NODES ##
  type User {
    id: ID! @id
    email: String! @unique
    firstName: String!
    lastName: String!
    fullName: String! @ignore
    role: Role!
    hashedPassword: String! @private
    createdAt: DateTime! @timestamp(operations: [CREATE])
    updatedAt: DateTime! @timestamp
    company: Company! @relationship(type: "WORKS_AT", direction: OUT)
    workHistory: [WorkOrderItem]
      @relationship(type: "WORKED_ON", direction: OUT)
    timeHistory: [TimeClock] @relationship(type: "CLOCK_FOR", direction: IN)
  }

  type TimeClock {
    id: ID! @id
    type: TimeClockType
    createdAt: DateTime! @timestamp(operations: [CREATE])
    user: User! @relationship(type: "CLOCK_FOR", direction: OUT)
  }

  type Company {
    id: ID! @id
    name: String! @unique
    stripeCustomerId: String @unique
    pricingTier: PriceTier
    createdAt: DateTime! @timestamp(operations: [CREATE])
    updatedAt: DateTime! @timestamp
    employees: [User!]! @relationship(type: "WORKS_AT", direction: IN)
    customers: [Customer] @relationship(type: "CUSTOMER_AT", direction: IN)
    workOrders: [WorkOrder] @relationship(type: "WORK_OWNED_BY", direction: IN)
    features: [CompanyFeatures]
      @relationship(type: "FEATURE_TOGGLED", direction: OUT)
[...]

PR #830, opened 2022-01-20
Fix/Duplicate Ogm Type Names

danstarns:
# Description

Fixes issue #635

# Issue

https://github.com/neo4j/graphql/issues/635

# Checklist


angrykoala (dismissed):
Technically, these would be breaking changes right? However, if we consider the original types to be broken, we should be able to release this as a patch

darrellwarde:
Maybe this one should be merged into `dev` and released in 3.0.0 in the end, just in case?

angrykoala (approved):

PR #831, opened 2022-01-20
GitHub Actions: pipeline on `master` PRs, high priority label/comment

darrellwarde:
# Description

Pull request jobs will now run off pull requests to `master`.

If a bug is given high priority, add a label and create a comment about branching from `master`. Didn't see the point of doing this for low priority bugs as this is "business as usual".

Adds new feature requests to project just to keep things tidy whilst we decide what to do here.

Any other tweaks we need to make?

### DIFF
--- a/.github/workflows/project-card-moved.yml
+++ b/.github/workflows/project-card-moved.yml
@@ -1,4 +1,4 @@
-name: Bug confirmed label
+name: Project card moved
 
 on:
   project_card:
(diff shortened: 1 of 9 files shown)


.github/workflows/project-card-moved.yml
```diff
@@ -17,3 +17,14 @@ jobs:
         with:
           add-labels: "confirmed"
           repo-token: ${{ secrets.NEO4J_TEAM_GRAPHQL_PERSONAL_ACCESS_TOKEN }}
+  add_high_priority_label:
+    if: github.event.project_card.column_id == '17286751'
```
angrykoala, line 21: This is going to be fun someday

angrykoala (approved):
Lovely changes, :clap:

.github/workflows/project-card-moved.yml
darrellwarde: Hahaha, it's the only way to do it, crazy right?!

PR #834, opened 2022-01-20
removes isNeoInt util

angrykoala:
# Description

The reasoning behind `isNeoInt` util was to improve typings, this is no longer needed, as the base `isInt` now returns the correct typings


angrykoala:
@darrellwarde That's a good point, it should not be a problem for users as these types are not exposed, but indeed tests fail when run with an older driver.

Given this, it is probably not worth the trouble of removing this just now

darrellwarde:
> @darrellwarde That's a good point, it should not be a problem for users as these types are not exposed, but indeed tests fail when run with an older driver.
> 
> Given this, it is probably not worth the trouble of removing this just now

Also given this, I will add back in automated tests against every minor version of the driver which we support as I think this is an important quality metric which we're currently missing. Will do this when the GraphQL 16 branch is merged as this contains similar functionality.

PR #836, opened 2022-01-20
RFC Cypher Based Hooks

danstarns:


maieulChevalier:
Coincidence @darrellwarde ! Following our talks about the cypher method with the OGM, I think that what you were reasoning about with @cypher fields to implement custom logic via the OGM wouldn't even be necessary if this Hooks feature was to be implemented.

maieulChevalier:
@danstarns Seems to me like a great approach to add custom logic to generated Resolvers like `createUsers`, `updateUsers` or `deleteUsers` since it allows running only one transaction.

Just two small remarks from me:

1. Adding a CALL (subquery or apoc) can cause some cypher queries to not be able to use the PIPELINED runtime and fallback to SLOTTED, leading to performance issues in big data sets. So it might be best to add custom CALLS in a separate transaction by using graphql-middleware for this usecase. I guess the PIPELINED runtime will be optimized by the cypher-team in the future but in the meantime it would be nice to have it documented if this Hooks feature is to be implemented.

2. Being able to toggle the added custom logic would be nice. For example, users might have two `updateUsers` mutations in their app for two different purposes, where only one of the `updateUsers` mutation is to be subscribed. There could be a `subscription:Boolean` argument (or `pre:Boolean` and `post:Boolean`)  like the `directed` argument in [RFC-004].(https://github.com/neo4j/graphql/blob/60d6d04551ba870e2f66b7fff4969298ec11b21c/docs/rfcs/rfc-004-undirected-relationship-queries.md)

docs/rfcs/cypher-based-hooks.md
```diff
@@ -0,0 +1,254 @@
...
+            jwt: $auth.jwt.sub
+        }
+        RETURN count(*)
+    }
+    RETURN u
+}
+RETURN u { .name, .email } AS u
+```
+
+### Subscriptions
```
darrellwarde, line 126: Let's just completely remove this example from this RFC for brevity - we discussed this a long time ago and this just wouldn't work in a production environment.

docs/rfcs/cypher-based-hooks.md
```diff
@@ -0,0 +1,254 @@
...
+}
+RETURN u { .name, .email } AS u
+```
+
+Triggering a HTTP call to our server and feeding the subscribers.
+
+### Extending Auth
+
+Finally, I would like to point out that our `@auth` directive cannot cover all complexities related to auth and that Cypher Hooks would enable users to enforce sophisticated auth patterns before or after a Cypher operation.
+
```
darrellwarde, line 247:
```suggestion

## Risks

- We'll face similar issues as we did with relationship cardinality - if a node is disconnected as a side effect, we just wouldn't know to execute the hook

```

Issue #204, opened 2021-05-12
Return Primitive Fields to GraphQL layer


litewarp:
**Is your feature request related to a problem? Please describe.**
Currently, if you want to add a computed field on the Node through a custom resolver you need the client to fetch all the dependent fields in order to compute the result. For example, to combine and first and last name as a fullName (as shown in the docs), the client needs to request firstname and lastname in the query in order for the server to have access to the first and last name. This is inefficient from a client / graphql standpoint as we should only be fetching the data we need from the server. It would be better if we had access to all, or certain, of the primitive fields on the server such that there is a parent reference or object to extend. 

**Describe the solution you'd like**
Ideally one of the following: (i) return all primitive fields on the node from the db to the GraphQL layer, (ii) a directive to include fields in the server request, and/or (iii) at least return the primary key or indicated @id for the node.

**Describe alternatives you've considered**
N/A. Without a reference to the parent node, one can't even query for the additional fields. 
**Additional context**
This may also help you reduce the demand for the relationship properties as users like myself could build whatever connections we wanted rather than having to rely on the enhanced schema. This is akin to more traditional GraphQL endpoints where a query on a Model or document returns the row or document reference.

PR #842, opened 2022-01-21
Feature: Add Required Fields For Custom Resolver

dmoree:
# Description

Proposed solution for #204 that adds a `require` argument to `@ignore` thereby making the definition:

```graphql
directive @ignore(
  require: [String!]
) on FIELD_DEFINITION
```
If any field that is marked with `@ignore(require: ["field1", ...])` is found in the selection set of the parent node, then the fields in the `require` argument will be added to the fields in the `resolveTree` that ultimately get projected and only if those fields are not already in the selection set. This ensures that regardless if all fields are selected, they will be available on the parent object in the resolver.

### Current Implementation of `@ignore`
Taking the example from the [documentation](https://neo4j.com/docs/graphql-manual/current/custom-resolvers/):

```js
const typeDefs = `
    type User {
        firstName: String!
        lastName: String!
        fullName: String! @ignore
    }
`;

const resolvers = {
    User: {
        fullName(source) {
            return `${source.firstName} ${source.lastName}`;
        },
    },
};

    typeDefs,
    resolvers,
});
```

It is accompanied by a note:

> [Due to the nature of the Cypher generation in this library, you must query any fields used in a custom resolver. For example, in the first example below calculating fullName, firstName and lastName must be included in the selection set when querying fullName. Without this being the case, firstName and lastName will be undefined in the custom resolver.](https://neo4j.com/docs/graphql-manual/current/custom-resolvers/)

As @litewarp pointed out in the request, this is inefficient from a GraphQL perspective.

### Proposed Implementation of `@ignore`
If the `typeDefs` were then defined as:

```graphql
type User {
    firstName: String!
    lastName: String!
    fullName: String! @ignore(require: ["firstName", "lastName"])
}
```
The resolvers will then have access to the fields on the `source` object
```js
const resolvers = {
    User: {
        fullName(source) {
            // source.firstName and source.lastName are always defined here
            return `${source.firstName} ${source.lastName}`;
        },
    },
};
```

Which means even if the query were for example:
```graphql
query UserWithoutFirstAndLastName {
    users {
        fullName
    }
}
    
```
The field `fullName` would resolve correctly. 

# Issue

- #204 

# Checklist


darrellwarde:
This is a really innovative solution @dmoree, excited to take a look at this during the week!

darrellwarde:
> Due to the nature of the Cypher generation in this library, you must query any fields used in a custom resolver. For example, in the first example below calculating fullName, firstName and lastName must be included in the selection set when querying fullName. Without this being the case, firstName and lastName will be undefined in the custom resolver.

Before I even jump into code review, it looks like this sentence hasn't been removed from the docs in this PR. Can you find and remove it? 🙂

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-842.surge.sh

mhlz:
I was working on something similar for a while, but had to stop because other things came up.

One thing I had in my implementation was to also be able to require nested fields:

```graphql

type Offer {
  price: Money! // @relationship etc
  vat: Money! // @relationship etc
  
  gross: Int! @ignore(requires: [ "price.amount", "vat.amount" ])
}

type Money {
  amount: Int!
  currency: String!
}
```

This is just an example of course, but it would make a few use cases we have possible.

My code is [here](https://github.com/opencreek/graphql/tree/require-fields), but it's really not finished or cleaned up (and it might contain some other things I was working on).

dmoree:
Thanks for the suggestion @mhlz. I really like the `buildResolveTree` that you're using which I'll come back to.

I considered something similar, but was using a selection set instead of the array in the PR. It would have looked something like:
```graphql
type User {
    firstName: String!
    lastName: String!
    fullName: String!
        @ignore(
            require: """
            {
                firstName
                lastName
            }
            """
        )
}
```
This generalization would allow for similar use cases that you describe. Such that, if the schema is:
```graphql
type User {
    id: ID!
    username: String!
}

type Post {
    id: ID!
    title: String!
    content: String!
    author: User! @relationship(type: "HAS_POST", direction: IN)
    likes: [User!]! @relationship(type: "LIKED_POST", direction: IN)
}
```
one could add a field with a custom resolver that can traverse the graph:
```graphql
extend type User {
    customResolver: String!
        @ignore(
            require: """
            {
                id
                posts {
                    id
                    title
                    likes {
                        id
                    }
                }
            }
            """
        )
}
```
I'm not sure why one would want this particular selection, but in any case the resolver:
```js
const resolvers = {
  User: {
    customResolver: (source) => {
      // get unique user ids of those that have liked any post
      const userIds = Array.from(
        new Set(
          source.posts.map((post) => post.likes.map((like) => like.id)).flat()
        )
      )
      // do something with userIds
    },
  },
}
```

There were two immediate issues (probably more): validation and arguments. Both, I think, are solvable.
### Validation
My attempt at validation was to use the `validate` function from `graphql` against a constructed query in the constructor of `Neo4jGraphQL`.
```js
const ignoreValidationErrors = nodes
    .filter((node) => node.ignoredFields.length)
    .map((node) =>
        node.ignoredFields
            .filter((ignoredField) => ignoredField.require)
            .map((ignoredField) => validate(schema, gql`query { ${node.plural} ${ignoredField.require} }`))
            .flat()
    )
    .flat();

if (ignoreValidationErrors.length) {
    throw new Error(ignoreValidationErrors[0].message);
}
```
The drawback was that this is relying on a root `users` query that may not exist in the schema since one could always `@exclude(operations: [READ])` on the `User` type. This would prevent the generation of `users` query thereby always throwing a validation error. One could exclude those fields _after_ `makeAugmentedSchema`, but _before_ `makeExecutableSchema` to ensure this works.

This was the approach because I couldn't find a way to validate a selection set against a particular type. It seems hacky. Ideally, it would be something like `validateAgainstType('User', selectionSet)` I tried searching through `@graphql-tools` but came up empty. It may already exist; I don't know. I'm sure at the end of the day it could be done.

If validation is accounted for, then `buildResolveTree` can go through the parsed `selectionSet` and construct the resolve tree. The library uses `graphql-parse-resolve-info` to get the initial `resolveTree`, but I don't see a way to achieve the same thing from a selection set. This should be doable, but it would have to be done to get the projection right. The only problem I see here is that what would happen if `posts` returned an interface? I'm not sure if fields of interfaces are kept track of like `nodes` are. And even then scalars would be fine, it would be an issue for related fields.

### Arguments
If one were to query `users` with:
```graphql
query ($postTitle: String!) {
  users {
    id
    username
    posts(where: { title: $postTitle }) {
      id
      title
      content
      likes {
        id
      }
    }
    customResolver
  }
}
```
This should throw an error because `posts` would have differing arguments since in `require` no arguments are present. A simple solution to this would be to require any non-scalar fields to be aliased within the selection set, but this also feels a little off, but it would solve the problem of merging fields in the selection set by having a completely different tree to project.

I apologize for the long post, but as you can see it ended up being a nontrivial problem. I don't want to trivialize this PR, but it is relatively straightforward and satisfies the initial request in #204. Perhaps in the future this could be revisited.

mhlz:
Using SelectionSets is also a nice idea!

And I also agree that this doesn't have to be part of this PR, just wanted to put the idea somewhere :)

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -645,3 +638,41 @@ function createProjectionAndParams({
...
+        new Set(
+            node.ignoredFields
+                // Filter any that have required fields
+                .filter((f) => f.requiredFields?.length)
+                // Filter those ignored fields that exist in selection set
+                .filter((ignoredField) => Object.values(nodeFields).find((f) => f.name === ignoredField.fieldName))
+                // Extract required fields
+                .map((f) => f.requiredFields)
+                .flat()
+        )
```
angrykoala, line 662: **suggestion (non-blocking** I think we may be able to avoid some comments and nested code with a couple of extra variables and/or functions:

```suggestion
    const fieldsWithRequiredFields = node.ignoredFields.filter((f) => f.requiredFields?.length);
    const fieldsInSelectionSet = filterFieldsInSelectionSet(nodeFields, fieldsWithRequiredFields);
    const requiredFields = fieldsInSelectionSet.map((f) => f.requiredFields).flat();

    return Array.from(
        // Create set to remove duplicate fields required across different `@ignore` fields
        new Set(
            requiredFields
        )
```


```typescript
function filterFieldsInSelectionSet<T extends BaseField>(nodeFields: Record<string, ResolveTree>, fields: T[]): T[] {
    const fieldsValues = Object.values(nodeFields);
    return fields.filter((field) => fieldsValues.find((f) => f.name === field.fieldName));
}
```

This is just a suggestion on how we may do it, maybe there are better solutions

packages/graphql/tests/tck/tck-test-files/directives/ignore.test.ts
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+
+import { gql } from "apollo-server";
+import { DocumentNode } from "graphql";
+import { Neo4jGraphQL } from "../../../../src";
+import { createJwtRequest } from "../../../utils/create-jwt-request";
+import { formatCypher, translateQuery, formatParams } from "../../utils/tck-test-utils";
+
+describe("Cypher ignore directive", () => {
```
angrykoala, line 26: Tests looking really good :+1:

angrykoala (commented):
Great job. It is almost good to go IMO.
I've made a few comments regarding some areas where code complexity could be reduced, would love to get your thoughts on these before merging @dmoree

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -645,3 +638,41 @@ function createProjectionAndParams({
...
+}: {
+    node: Node;
+    resolveTree: ResolveTree;
+}): Record<string, ResolveTree> => {
+    const nodeFields = resolveTree.fieldsByTypeName[node.name];
+
+    return Array.from(
+        // Create set to remove duplicate fields required across different `@ignore` fields
+        new Set(
+            node.ignoredFields
```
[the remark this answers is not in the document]
dmoree: Agreed: https://github.com/neo4j/graphql/pull/842/commits/3c0057632e83a8ace7c6c041995af753c031c972

packages/graphql/src/schema/get-obj-field-meta.ts
```diff
@@ -357,11 +362,19 @@ function getObjFieldMeta({
...
+                const requiredFields = Array.from(
+                    new Set(
+                        (
+                            ignoreDirective?.arguments?.find((arg) => arg.name.value === "require")
+                                ?.value as ListValueNode
+                        )?.values.map((v) => (v as StringValueNode).value) ?? []
+                    )
+                );
+
+                res.ignoredFields.push({ ...baseField, requiredFields });
```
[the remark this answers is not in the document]
dmoree: I opted for created a `getIgnoreMeta` function similar to what is done for `@cypher` and `@relationship`: https://github.com/neo4j/graphql/pull/842/commits/ce1a9d705a642253692bc2a21394cf4b488fadf5

packages/graphql/src/translate/create-projection-and-params.ts
dmoree: There are other PRs that are open that can make use of utility functions on the resolve tree: https://github.com/neo4j/graphql/pull/842/commits/a9f37e2f4f964e1e3dd85016497b3e8306d86624#diff-55c04c06457213093b423b2d53edd073daf6dd676e3aca304af9f3234bdbff23

packages/graphql/src/schema/get-obj-field-meta.ts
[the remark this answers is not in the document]
angrykoala: That sounds like a great solution :+1:  thanks!

packages/graphql/src/translate/utils/resolveTree.ts
```diff
@@ -0,0 +1,68 @@
+/*
```
angrykoala, line 1: Great idea on adding a set of utilities for handling the resolveTree :clap:

packages/graphql/src/translate/utils/resolveTree.ts
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+
+import { ResolveTree } from "graphql-parse-resolve-info";
+import { BaseField } from "../../types";
+import { removeDuplicates } from "../../utils/utils";
+
+/* Finds a field of selection based on field name */
+export const fieldNameExistsInSelection = (selection: Record<string, ResolveTree>) => (fieldName: string) =>
+    Object.values(selection).find((resolveTree) => resolveTree.name === fieldName);
```
angrykoala, line 26: **question:** Are we getting all the value from curry functions here?

I feel we could be achieving the same with a much simpler function:

```suggestion
export function fieldNameExistsInSelection(selection: Record<string, ResolveTree>, fieldName: string): ResolveTree {
    return Object.values(selection).find((resolveTree) => resolveTree.name === fieldName);
}
```

packages/graphql/src/translate/utils/resolveTree.ts
```diff
@@ -0,0 +1,68 @@
...
+    const fieldNameExists = fieldNameExistsInSelection(selection);
+    const fieldNameAliased = fieldNameAliasedInSelection(selection);
+
+    return removeDuplicates(fieldNames).reduce((acc, fieldName) => {
+        const exists = fieldNameExists(fieldName);
+        const aliased = fieldNameAliased(fieldName);
+        if (!exists || aliased) {
+            return { ...acc, ...generateProjectionField({ name: fieldName }) };
+        }
+        return acc;
```
angrykoala, line 66: :100: 
Thanks for the effort in making this so readable

packages/graphql/src/translate/utils/resolveTree.ts
```diff
@@ -0,0 +1,68 @@
...
+export const fieldNameAliasedInSelection = (selection: Record<string, ResolveTree>) => (fieldName: string) =>
+    Object.values(selection).find((resolveTree) => resolveTree.name === fieldName && resolveTree.alias !== fieldName);
+
+export const filterFieldsInSelection =
+    (selection: Record<string, ResolveTree>) =>
+    <T extends BaseField>(fields: T[]) =>
+        fields.filter((field) => Object.values(selection).find((f) => f.name === field.fieldName));
+
+/* Generates a field to be used in creating projections */
+export const generateProjectionField = ({ name }: { name: string }): Record<string, ResolveTree> => {
```
angrykoala, line 38: **suggestion (if-minor):** Now that we have these utils, it could be useful to unit-test them.
I realize that most of these deal with a `ResolveTree` object, which may be hard to mock. Thoughts?

angrykoala (commented):
Thanks for making the effort on all of those changes @dmoree 
It is looking great, I just added a couple of comments/questions regarding `resolveTree.ts`

Everything else looks good and ready to merge after we discuss those comments

packages/graphql/src/translate/utils/resolveTree.ts
dmoree: I think some things can be unnecessarily complicated since most things are with respect to a "`field`" which has two meanings here: a field of the node and `Record<string, ResolveTree>`. Simply manipulating a set of `ResolveTree`s will make this easier.

In a recent refactor (https://github.com/neo4j/graphql/commit/bd391a851360fa7bcda758abc02cd90ab172a18c) `createProjectionAndParams` now takes a `ResolveTree`. It would make things easier to reason about if the reducer in that function were to also be refactored to take `ResolveTree[]` as opposed to a "field" of `Record<string, ResolveTree>`. This would be even more trivial by only requiring a change from `key` to `field.alias` which is what `key` is anyway by construction.

Thoughts on this?

packages/graphql/src/translate/utils/resolveTree.ts
dmoree: Some of these are really simple and straightforward (which I understand should never be a response to whether unit tests are necessary), but otherwise doable.

packages/graphql/src/translate/utils/resolveTree.ts
dmoree: In one iteration of the refactor I was reusing this function, but at the end it makes more sense to have an uncurried function here.

packages/graphql/src/translate/utils/resolveTree.ts
dmoree: Refactor: https://github.com/dmoree/graphql/compare/feature/ignore-directive-required-fields...dmoree:refactor/field-to-resolve-tree

Note: I also changed the names of the utility functions and added unit tests.

angrykoala: Hi @dmoree Your changes in that refactor make a lot of sense, I think overall it is an improvement :+1: 
My only minor, (non-blocking) comment on those would be to use a variable for `field.alias` in the reducer function just to avoid accessing to `.alias` all the time.
Feel free to add that commit to this PR so I can make a last review on those changes and hopefully merge this soon

packages/graphql/src/translate/utils/resolveTree.ts
dmoree: Tests in https://github.com/neo4j/graphql/pull/842/commits/3b465a2ce89074e05baf82bc955d7e92ed27c547

packages/graphql/src/translate/utils/resolveTree.ts
dmoree: Refactored in https://github.com/neo4j/graphql/pull/842/commits/59e8d9b2848938364adfaf1402f3494d3071182a

packages/graphql/src/translate/utils/resolveTree.ts
dmoree: Refactored in https://github.com/neo4j/graphql/pull/842/commits/339cbc2928aee6e30ccd6b5794355d0c389c4b31

angrykoala (approved):
LGTM

oskarhane:
I'm not sure I like the naming of `@ignore(require: [""])`. 
Not sure what it is, but those words combined makes no sense. :)

Is it only me?

Personally I think `@ignore(dependsOn: [""])` or `@ignore(dependencies: [""])` reads better and is easier to understand what it means.

dmoree:
It's not just you. I went back and forth on `required` `requires` and `require`. But really I have no strong opinions either way; I'm willing to change it to wherever the consensus lands.

oskarhane:
I think the real issue for me here is `@ignore`. I think `@computed` would be a better name for that directive.

```
type Person {
  firstName: String!
  lastName: String!
  fullname: String! @computed(requires: ["firstName", "lastName"])
}
```

is much more clear imo.

Sorry for the curve ball here @angrykoala and @dmoree. 
I just happened to see this in on of my feeds and figure I'd mention it before v.3 goes out (while you still can break API:s).

dmoree:
If the directive were to change, it should be more along the lines of `@custom` because `require` isn't required and
```graphql
type Person {
  firstName: String!
  lastName: String!
  customField: String! @computed
}
```
seems like the library should be doing something as opposed to the user taking responsibility for the field.

(I always thought `@computed` should have been used for `@cypher` fields returning a Scalar type.)

oskarhane:
Yeah, you're right on the `@computed`, it might be too narrow in this case.  
`@custom` makes sense to me as well.

Not sure what the team thinks though.

angrykoala:
@dmoree @oskarhane 
I agree on `require` name change (and now it's a great time to rename stuff!).
I'm not sure about `ignore` to `computed` though, so maybe we could go with `@ignore(dependsOn)`? Thoughts @darrellwarde @danstarns @tbwiss ?

darrellwarde:
This is tricky! Whilst we do have 3.0.0 coming up, we should still keep breaking changes to a minimum if possible, and renaming `@ignore` to `@custom` will result in a breaking change for everybody that uses this directive, as opposed to the non-breaking manner of this feature as it is.

I always look at our directives as "instructions" to our library - so always saw it that the `@ignore` directive was instructing the library to ignore the field. Personally, I would go with `@ignore(dependsOn: [...])` but I can understand all of the arguments here.

dmoree:
>I always look at our directives as "instructions" to our library - so always saw it that the @ignore directive was instructing the library to ignore the field.

I think this may be slightly confusing here because it is true that `@ignore` is telling the library to ignore the field, but I'm not so sure one would want to _totally_ ignore the field. The library should ignore the field for the purposes of schema generation, but *not* ignore the field when it comes to projection. This, I think, is the basis for issues like #204 and must come with a note in the documentation due to the ignoring of the field in projections.

I like the concept of directives as "instructions", I'm not so sure if `@ignore` is it. Personally, I think a different directive should be used that captures the distinction in generating the schema and generating the projection.

angrykoala:
Hi @dmoree , @oskarhane @darrellwarde
The conversation has creeped completely out of scope for this PR. We should keep this PR only for the newly added feature. I opened issue #937 so we can move any discussion regarding improvements to the `@ignore` directive there, as it doesn't make sense to block this PR for this kind of discussion.

As of now, it seems that `dependsOn` is the preferred name for this new argument, so if there are no objections to that, I'm happy to merge this PR once that rename is made.

dmoree:
@angrykoala @darrellwarde @oskarhane 
Refactored `require` to `dependsOn` in https://github.com/neo4j/graphql/pull/842/commits/2b4091a9ddd5da450a07a11e31318ab72577a30e

packages/graphql/src/schema/get-ignore-meta.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    if (
+        directiveDependsOn?.value.kind !== Kind.LIST ||
+        directiveDependsOn?.value.values.some((value) => value.kind !== Kind.STRING)
+    ) {
+        throw new Error(ERROR_MESSAGE);
+    }
+
+    // `@ignore(require: [String!])`
```
angrykoala, line 52: **nitpick non-blocking**
```suggestion
    // `@ignore(dependsOn: [String!])`
```

angrykoala (approved):

angrykoala (approved):

Issue #692, opened 2022-01-04
feature request: custom logic on top of generated queries and mutations in resolvers


maieulChevalier:
**Is your feature request related to a problem? Please describe.**

I'd like to keep using generated queries and mutations and add custom logic on top of them (e.g. request to another api/database) in a resolver with an identical name, but now it overwrites those generated queries and mutations.

Given the following type definitions [from the docs](https://neo4j.com/docs/graphql-manual/current/type-definitions/cypher/):

```graphql
type Actor {
    actorId: ID!
    name: String
}

type Mutation {
    createActor(name: String!): Actor
        @cypher(
            statement: """
            CREATE (a:Actor {name: $name})
            RETURN a
            """
        )
}
```
Let's say I'd like to be able to add custom logic on top of the `createActor` and `deleteActors` generated mutations;

But right now, the following code

```graphql
const resolvers = {
  Mutation: {
    createActor: async (_source, args, context, info) => {
      console.log("I can't create an Actor anymore!")
    }
    deleteActors: async (_source, args, context, info) => {
      console.log("I can't delete Actors anymore!")
    }
  },
};
```
completely overwrites the generated `createActor` and `deleteActors` mutations. No cypher query gets executed.

As of now, if I want to add custom logic to queries and mutations, I have to do it in a separate resolver, leading to 2 network requests from client to server instead of 1. And I have to rewrite the cypher code manually, which not too hard when I can use the OGM but not easy when using the @cypher directive.

**Describe the solution you'd like**

This was actually possible in neo4j-graphql-js, simply by returning the `neo4jgraphql` function in each resolvers:

```graphql
const resolvers = {
  Mutation: {
    createActor: async (_source, args, context, info) => {
      console.log("I can add custom logic on top of the generated mutation thanks to the neo4jgraphql function below!")
      return neo4jgraphql(_source, args, context, info);
    }
    deleteActors: async (_source, args, context, info) => {
      console.log("I can add custom logic on top of the generated mutation thanks to the neo4jgraphql function below!")
      return neo4jgraphql(_source, args, context, info);
    }
  },
};
```

**Additional context**

[...]

PR #846, opened 2022-01-21
RFC: Pre/Post resolver hooks

angrykoala:
# Description
This RFC proposes a solution to the issue raised at #692 by defining pre and post hooks on the generated resolvers


darrellwarde:
This is probably redundant given the "workaround" using `graphql-middleware` - let's document and test that a bit more and then make a decision.

PR #857, opened 2022-01-24
Fix: Sort Fails for Aliased Field

dmoree:
# Description

This PR addresses an issue that leads to sorting failing if the field being sorted on is aliased.

# Checklist


packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -639,3 +621,31 @@ function createProjectionAndParams({
...
+            // If fieldname is not found in fields of selection set
+            ...(!Object.values(nodeFields).find((field) => field.name === sortFieldName) ||
+            // or does exist but is aliased
+            Object.values(nodeFields).find((field) => field.name === sortFieldName && field.alias !== sortFieldName)
+                ? // generate a basic resolve tree
+                  generateProjectionField({ name: sortFieldName })
+                : {}),
+        }),
+        {}
+    );
```
angrykoala, line 650: **suggestion:** This reduce statement is starting to rack up some complexity. I wonder if we could improve readability a bit by splitting it into multiple statements inside the function, something akin to:

```suggestion
    return Array.from(new Set(sortFieldNames)).reduce((acc, sortFieldName) => {
        const fieldMissing = !Object.values(nodeFields).find((field) => field.name === sortFieldName);
        const fieldIsAliased = Boolean(
            Object.values(nodeFields).find((field) => field.name === sortFieldName && field.alias !== sortFieldName)
        );

        if (fieldMissing || fieldIsAliased) {
            return {
                ...acc,
                ...generateProjectionField({ name: sortFieldName }),
            };
        }
        return acc;
    }, {} as Record<string, ResolveTree>);
```

Thoughts @dmoree ?

angrykoala (commented):
This is a nasty bug, thanks for fixing it @dmoree 
The solution looks good, and the tests are thorough :clap: 
Just a comment to discuss readability on a complex piece of code, to see what you think.

packages/graphql/src/translate/create-projection-and-params.ts
dmoree: Thanks for the suggestion. I do agree that this reads better.

Refactored in https://github.com/neo4j/graphql/pull/857/commits/96040d368a877e7096e9f7a0f45ad52f897d3466

angrykoala (approved):
LGTM

Issue #851, opened 2022-01-21
Top Level Connections


litewarp:
I would like the ability to add top-level connections, or to fetch relationships using connections even if no relationship properties are defined.

Currently, if I want to do pagination client-side, I have to handle two different types of pagination requests -- those using the connections model, and those without. Pagination is not a trivial task, and there is a high potential for confusion (as evidenced by all the discord and issues) when users have to switch between offsets and cursors.

For example, I have a relay-nextjs app I am putting together to where the user can select multiple companies. Since those companies are not yet associated with a node, I have to query them using the default top-level findMany resolver, i.e.
 
```graphql

## currently

type Query {
  companies(where: CompanyWhere, options: CompanyOptions): [Company!]!
}

## proposed

type Query {
  companiesConnection(first: Int, after: String, where: CompanyWhere, sort: [CompanySort]): CompanyConnection!
}
```

Since this general findMany query is not relay-compliant, I can't take advantage of all the built-in goodness that `react-relay` offers, e.g., [usePaginationFragment](https://relay.dev/docs/guided-tour/list-data/rendering-connections/). So if I am querying a top-level "node", in order to paginate through the list, I now have to keep track of the offset, calculate the next page and send the request. I also have to manually include the aggregation query in the top-level selection set and then divide the total number by the page size to determine if there is a next page. This information should come from the server, which is why connections are often preferable to simple paginated lists. 

I know there has been some debate over whether relationship fields without properties should have connections generated for them. I don't know the specifics of that debate, but allowing users to query all types of relationships through connections would certainly make it easier to do pagination seamlessly. For other packages, it seems like the criteria is whether a list of nodes is being fetched rather than when properties are present. For example, [Postgraphile implements connections for any "large list"](https://www.graphile.org/postgraphile/connections/#gatsby-focus-wrapper), and [Hasura has top-level connections as well](https://hasura.io/docs/latest/graphql/core/databases/postgres/schema/relay-schema.html) (see the "users_connection" in the screenshot).

Currently, I use a [custom graphql-tools plugin I built](https://github.com/litewarp/graphql-utils/blob/main/packages/neo4j-helpers/src/plugins/top-level-connections.ts) which repurposes the findMany function into a connection field, but it is not the easiest to maintain and comes with other drawbacks since it hijacks the aggregation count field as well and there may be conflicting arguments sent. 

[...]

PR #864, opened 2022-01-24
Root connection fields

litewarp:
Implements #851 

Adding root connection fields to the schema. The resolver hooks into translate read and prepares the connection using the existing tools, so there should be no extra configuration needed.

Last thing I need to resolve is to add an alias ability to the root connection field.

Could use some more tests as well. Let me know what you'd like to see.


packages/graphql/src/schema/resolvers/root-connection.ts
```diff
@@ -0,0 +1,102 @@
...
+import { isNeoInt } from "../../utils/utils";
+import { createConnectionWithEdgeProperties } from "../pagination";
+
+export default function rootConnectionResolver({ node, composer }: { node: Node; composer: SchemaComposer }) {
+    async function resolve(_root: any, _args: any, _context: unknown, info: GraphQLResolveInfo) {
+        const context = _context as Context;
+        const resolveTree = getNeo4jResolveTree(info);
+
+        const edgeTree = resolveTree.fieldsByTypeName[`${node.name}RootConnection`].edges;
+        const nodeTree = edgeTree.fieldsByTypeName[`${node.name}RootEdge`].node;
```
darrellwarde, line 36: Out of interest, did `${node.name}Connection` and `${node.name}Edge` (without `Root`) cause conflicts? I think they're much cleaner type names.

darrellwarde (commented):
Just a couple of initial comments as it's getting quite late here in the UK. 

The functionality looks great, but we are likely to want to see a few more integration tests. We'll have a chat about it in the team and discuss, sorry if there's a bit of radio silence for a while!

packages/graphql/src/schema/resolvers/root-connection.ts
```diff
@@ -0,0 +1,102 @@
...
+    } catch (e) {}
+
+    return {
+        type: rootConnection.NonNull,
+        resolve,
+        args: {
+            first: "Int",
+            last: "Int",
+            after: "String",
+            before: "String",
```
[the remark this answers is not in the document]
litewarp: Sounds good. I think I recall that decision when I pushed the pagination PR so many months ago. Will fix.

packages/graphql/src/schema/resolvers/root-connection.ts
litewarp: I would rather use the cleaner names if possible. My only thought on adding "Root" was to fit the convention on relationships (e.g., "ActorMoviesConnection") but the more I think about it, if following that convention, the name should have been "RootMoviesConnection" rather than "MovieRootConnection"

I'll go with your suggestion and just use `${node.name}Connection` and `${node.name}Edge`

darrellwarde:
Other than what I flagged, you do have several test failures. Taking a look, it looks like possibly your branching or merging from `dev` has not gone as planned, as a lot are to do with recent snapshot updates. Both schema and TCK tests are now managed by Jest Snapshots, so those can be easy to update.

Other than that, take a look at the checks in GitHub Actions, you should be able to see the full detail! Let me know if you need a hand with fixing any of the tests.

litewarp:
Thanks. I fixed the schema and integration tests and made the two changes you asked for above.

I agree more tests are needed. I'll try and think of more TCK tests tonight.

litewarp:
Ok, having a test issue -- On my local machine, all the tests pass:

image

So not sure why the ogm and schema checks are still failing

packages/graphql/tests/schema/aggregations.test.ts
```diff
@@ -48,9 +48,7 @@ describe("Aggregations", () => {
               mutation: Mutation
             }
 
-            \\"\\"\\"
-            A BigInt value up to 64 bits in size, which can be a number or a string if used inline, or a string only if used as a variable. Always returned as a string.
-            \\"\\"\\"
+            \\"\\"\\"A BigInt value up to 64 bits in size, which can be a number or a string if used inline, or a string only if used as a variable. Always returned as a string.\\"\\"\\"
```
dmoree, line 51: It looks like this PR is reverting the change in #713 here:
https://github.com/neo4j/graphql/pull/713/files#diff-c0a37dcd426c9fb8df65d78d4f03fc92f83a8698885a5fd4f5b1b84cebbdfbad

packages/ogm/src/generate.test.ts
```diff
@@ -51,16 +51,13 @@ describe("generate", () => {
         expect(generated).toMatchInlineSnapshot(`
             "import { SelectionSetNode, DocumentNode } from \\"graphql\\";
             export type Maybe<T> = T | null;
-            export type InputMaybe<T> = Maybe<T>;
```
dmoree, line 54: It looks like this is reverting a change in #713 here:
https://github.com/neo4j/graphql/pull/713/files#diff-7f1bf5c3db524a84130ef78d588a180f48e311278a2243315bb523df59f80e56

dmoree:
Something doesn't look right in https://github.com/neo4j/graphql/pull/864/commits/083cd1122cc96daa3c379418e8cadbef70c2907e. I believe that you should be using `InputMaybeWhere`.
Also in https://github.com/neo4j/graphql/pull/864/commits/5b4eb2df7d65da892cacb606e873167fee91586e, comments should be on their own line.

This seems to be reverting changes in #713.

If they are passing on your machines, did you happen to update the snapshots? Also, did you do a `yarn install` after merging to install `graphql@16.2.0`

litewarp:
> Something doesn't look right in [083cd11](https://github.com/neo4j/graphql/commit/083cd1122cc96daa3c379418e8cadbef70c2907e). I believe that you should be using `InputMaybeWhere`. Also in [5b4eb2d](https://github.com/neo4j/graphql/commit/5b4eb2df7d65da892cacb606e873167fee91586e), comments should be on their own line.
> 
> This seems to be reverting changes in #713.
> 
> Did you do a `yarn install` after merging to install `graphql@16.2.0` locally?
> 
> If they are passing locally, did you happen to update the snapshots? They may not after a `yarn install`.

You were right again, @dmoree. Hopefully it's all good now.

dmoree:
I was able to see failing tests when I cloned your repo and merged `dev`. Have you done `git fetch upstream` (assuming the upstream is set to `neo4j/graphql`) along with a `git pull upstream dev` to pull in the most recent changes? After doing that you should see tests failing in `packages/ogm/src/generate.test.ts`. It is precisely the failure seen in the failing check above.

packages/graphql/src/schema/resolvers/root-connection.ts
```diff
@@ -0,0 +1,124 @@
...
+    const rootEdge = composer.createObjectTC({
+        name: `${node.name}Edge`,
+        fields: {
+            cursor: "String!",
+            node: `${node.name}!`,
+        },
+    });
+
+    const rootConnection = composer.createObjectTC({
+        name: `${node.name}Connection`,
```
darrellwarde, line 98: Pluralize the connection name perhaps? Given that one connection contains multiple edges and thus multiple nodes.

This is also then consistent with relationships, for example `ActorMoviesConnection`.

```suggestion
        name: `${node.plural}Connection`,
```

packages/graphql/src/schema/resolvers/root-connection.ts
```diff
@@ -0,0 +1,124 @@
...
+export default function rootConnectionResolver({ node, composer }: { node: Node; composer: SchemaComposer }) {
+    async function resolve(_root: any, _args: any, _context: unknown, info: GraphQLResolveInfo) {
+        const context = _context as Context;
+        const resolveTree = getNeo4jResolveTree(info);
+
+        const edgeTree = resolveTree.fieldsByTypeName[`${node.name}Connection`].edges;
+        const nodeTree = edgeTree.fieldsByTypeName[`${node.name}Edge`].node;
+
+        context.resolveTree = { ...nodeTree, args: resolveTree.args };
+        context.isRootConnectionField = true;
```
darrellwarde, line 40: I realise we go against this ourselves by setting `context.resolveTree` this way, but we should avoid modifying the context during execution. This feels like it might be better placed as an argument on the `translateRead` function?

packages/graphql/src/schema/resolvers/root-connection.ts
```diff
@@ -0,0 +1,124 @@
...
+            defaultAccessMode: "READ",
+            context,
+        });
+
+        // TODO: Ask neo4j team how to resolve this issue
+        // TCK Tests Run the Resolver But Expect a Return Value
+        // So in order to get the tests to pass we have to return
+        // an empty connection. But it may be more preferable to
+        // throw an error if there is no record, but that might
+        // require modification to the "translateQuery" function
```
darrellwarde, line 56: I don't think an error should be thrown on no records found - errors should be reserved for actually unexpected behaviour, when no records is perfectly normal behaviour for this library. So this solution feels good! I believe this is also inline with the Relay specification.

packages/graphql/src/schema/resolvers/root-connection.ts
```diff
@@ -0,0 +1,124 @@
...
+            edges: rootEdge.NonNull.List.NonNull,
+        },
+    });
+
+    // since sort is not created when there is nothing to sort, we check for its existence
+    let sortArg: InputTypeComposer<any> | undefined;
+    try {
+        sortArg = composer.getITC(`${node.name}Sort`);
+        /* eslint-disable-next-line no-empty */
+    } catch (e) {}
```
darrellwarde, line 111: A couple of better options here:

```ts
if (composer.has(`${node.name}Sort`))
```

```ts
if (composer.hasInstance(`${node.name}Sort`, GraphQLInputObjectType))
```

```ts
if (composer.isInputObjectType(`${node.name}Sort`))
```

I believe these are all correct, take your pick!!

packages/graphql/tests/integration/root-connections.int.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      type Pilot {
+        name: String
+        aircraft: [Aircraft!]! @relationship(type: "FLIES_IN", direction: IN)
+      }
+
+      type Aircraft {
+        id: ID!
+        name: String!
+        pilots: [Pilot!]! @relationship(type: "FLIES_IN", direction: OUT)
+      }
```
darrellwarde, line 50: Can we use the `generateUniqueType` functionality here please? 🙂

darrellwarde (commented):
This broadly looks really good!!! Much simpler than I thought it would be.

I'm super confident that everything will work because it is just using existing functionality, but is it worth checking for some more advanced filtering and selection set use cases? Please tell me if you don't think this is worthwhile!!

darrellwarde (commented):
Slight misfire on the type naming - that first character just need upper casing!!

Looks like your integration tests are failing, I can't immediately tell what's causing this.

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -699,6 +700,10 @@ function makeAugmentedSchema(
             composer.Query.addFields({
                 [rootTypeFieldNames.aggregate]: aggregateResolver({ node }),
             });
+
+            composer.Query.addFields({
+                [`${node.plural}Connection`]: rootConnectionResolver({ node, composer }),
```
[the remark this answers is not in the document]
darrellwarde: > Perhaps you meant that I should uppercase the TypeName (makes more sense) in `root-connection.ts`?

That is exactly what I meant, I commented in the completely wrong file!! 🤦

darrellwarde (dismissed):
This is looking good to me!! I'll leave it open shortly in case any one else in the team wants to take a look. Thanks again for this contribution, and apologies again that we've been so slow on the PRs lately!!

darrellwarde (approved):

PR #866, opened 2022-01-25
Feature: Global node resolution

litewarp:
# Global Node Resolution

## Completed
- Add global argument to `@node` directive
- Add validations for `@id` and `@unique` on type
- Add validation for use of non-aliased `id` field
- Add the node interface to applicable types
- Add `id` field for applicable types
- Add toGlobalId function on Node class
- Add idField property to NodeDirectiveClass
- Auto project the required field for id computation
- Create tests for id field resolver
- Add translation for the id field when used in a `where` argument
- Add top-level resolver for the `node` query

## Todo
- Add top-level resolver for the `nodes [PLURAL]` query (see note below)
- Whatever else @darrellwarde tells me to do!

# Checklist


litewarp:
@darrellwarde I'm rounding the corner on getting the node resolver done and have begun thinking about the `nodes` query. 

So for starters, I took the path of least resistance and tied node resolution to the main label. If we want to add support for additional labels, I figure we can add that on top of the work here. 

The question I have about resolving nodes plural is what's the best way to fetch the data using a single cypher query?

I believe that the order of the return value must match the order of the argument array. So it should have the same number of nodes (or null if not found) returned. 

I'm having trouble mapping this concept to a cypher query. 

Three questions --

1. If the id doesn't translate to a label or field for some reason, there will be nothing to project in the query as the translation won't know what label or field to query on. But we have to keep the nodes in the order of the array of ids submitted or return null if no node is present. 

Is it appropriate or even possible to have a construct a clause that returns null? So if the 10th item is null, how do we project a cypher clause that will just return null and thus keep the list sequence in order?

2. An alternate option is to have the resolver keep track of the order of the ids, and then create a mapping of the null items and the expected nodes. 

The benefit of this approach is you could do batching of queries if you made sure that the id field is in the return projection. So if someone sent 8 node ids, but 3 of them are "Movies" and 5 of them are "TVSeries" you could construct two clauses rather than eight individual ones. And then you reconstruct the array based on the pre-request mapping. 

Not sure if that is faster in any way from a cypher perspective, but I'm curious what you think the best approach might be.

litewarp:
To follow up on the note above 👆🏽, I have come to the conclusion that the `nodes` query should not be part of this PR.

It's not in the spec. There are [plural identifying root fields](https://relay.dev/graphql/objectidentification.htm#sec-Plural-identifying-root-fields) but these are different than a `nodes` query since they operate on primitive fields and don't require translation of the ids. 

Some thought should be put into how to optimize a nodes query. It could be problematically expensive if there is no cap on the number of nodes, or if there is no batching. This would be in the "nice to have" category, imho.

darrellwarde:
Hey @litewarp, sorry for the slow reply here!

The `nodes` query was always very much an optional extra (and not an essential part of the Relay spec), this this can always be added later if requested. It also means a smaller PR here. 

The second option could work as a nice way for now though! There will surely be a way to do this natively in Cypher but I don't have time right now to delve into the editor and experiment sadly. 😞  @danstarns always loves a Cypher challenge, he might have some thoughts on this!

litewarp:
Thanks for the note, @darrellwarde. Could we push a `nodes` query to a later PR? I'd love to get some feedback on this initial implementation so I could help get it merged. I am in the midst of a project that would benefit greatly from this, as this unlocks two key things in Relay: 
(i) ability to use [refetchable fragments](https://relay.dev/docs/guided-tour/refetching/refetching-fragments-with-different-data/) and; 
(ii) advanced client-side caching (cache once, render anywhere).

darrellwarde:
> Thanks for the note, @darrellwarde. Could we push a `nodes` query to a later PR? I'd love to get some feedback on this initial implementation so I could help get it merged. I am in the midst of a project that would benefit greatly from this, as this unlocks two key things in Relay: (i) ability to use [refetchable fragments](https://relay.dev/docs/guided-tour/refetching/refetching-fragments-with-different-data/) and; (ii) advanced client-side caching (cache once, render anywhere).

Of course we can push that back!!

I'm going to try to spend some time on PR reviews today, and will try to include this. 🙂

litewarp:
Hey @darrellwarde any chance we can get this into the 3.0 release? Realize you're swamped with a ton else so don't want to be too pushy if it's not practical.

darrellwarde:
> Hey @darrellwarde any chance we can get this into the 3.0 release? Realize you're swamped with a ton else so don't want to be too pushy if it's not practical.

Hey @litewarp! Sorry for the radio silence here - as you say, we have been pretty swamped recently. If I'm being very honest with you, unlikely that this will make the 3.0.0 release. However, we're not at all constrained by a strict release schedule, so 3.1.0 could come with this included at any point after 3.0.0 - we won't keep you waiting forever. We're just in proper crunch mode for 3.0.0 at the moment which does have a strict deadline, and we have to stop the scope of that release at some point.

litewarp:
No worries. I'll check in again in a week or two.

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -598,6 +598,23 @@ function createProjectionAndParams({
...
+        if (!Object.values(existingProjection).find((field) => field.name === globalIdField)) {
+            existingProjection[globalIdField] = {
+                alias: node.getGlobalIdField(),
+                args: {},
+                fieldsByTypeName: {},
+                name: node.getGlobalIdField(),
+            };
+        }
+        delete existingProjection.id;
+    }
```
darrellwarde, line 616: I could do with an explanation of this logic surrounding the projection of the `id` field!

darrellwarde (commented):
My apologies against that this has taken so long to review, we've really fallen off the wagon when it comes to PR reviews recently. 

I realise it's a lot of points, but figured I would give this a good thorough review with a few questions to kick off!!

packages/graphql/src/schema/parse-node-directive.ts
```diff
@@ -17,15 +17,53 @@
...
+                    "Type `Movie` already has a field `id`. Either remove it, or if you need access to this property, consider using the `@alias` directive to access it via another field"
+                );
+            }
+        }
+
+        const candidates = fields.filter(
+            (x) =>
+                x.type.kind === "NonNullType" &&
+                x.type.type.kind === "NamedType" &&
+                (x.type.type.name.value === "ID" || x.type.type.name.value === "String")
```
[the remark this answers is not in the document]
litewarp: I suppose any unique type could work. Will change

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -787,7 +812,7 @@ function makeAugmentedSchema(
 
         composer.createInputTC({
             name: `${node.name}Where`,
-            fields: queryFields,
+            fields: node.isGlobalNode() ? { id: "ID", ...queryFields } : queryFields,
```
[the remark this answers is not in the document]
litewarp: If the user opts into using the global node, an ID field will be generated for it and returned. Without adding the "id" field to where, the only way the user can query a specific node by identity would be to call the top-level `node(id: ID!): Node!` query. If we add the id field to the where argument, users can query a Movie by "id"

Hypothetically, the user could manage the retrieval and querying of unique ids by themselves -- e.g., returning the underlying unique field and then querying for that. But one of the benefits of relay is having and using a single unique id, rather than having to request two different identifiers and decide which is appropriate "where" (pun intended).

packages/graphql/src/translate/create-projection-and-params.ts
litewarp: If a user includes the field "id" in the selectionSet for the node, then we need to make sure we include the underlying unique field in the projection for the `toGlobalId()` function.  E.g., with a type "Movie" that uses "_id" as it's unique field, if I query just for the "id" and "title", the field resolver for the ID will receive an object { "title": "something" }. That object needs to include the "_id" value.

litewarp: There is some interplay with [this](https://github.com/neo4j/graphql/pull/866#discussion_r832088478) issue above -- if you are right (and I think you are) and that "id" in the db can be the underlying field, then we'll need a conditional check before deleting the existingProject.id field. 

In general, if the underlying field is **not** "id" but rather "_id", "dbId", etc., then we have to remove the "id" field otherwise the db will err when it tries to return the null value.

packages/graphql/src/schema/parse-node-directive.test.ts
```diff
@@ -81,9 +84,86 @@ describe("parseNodeDirective", () => {
...
+                type Movie @node(global: true) {
+                    id: ID! @id @alias(property: "dbId")
+                    title: String!
+                }
+            `;
+
+            const definition = parse(typeDefs).definitions[0] as ObjectTypeDefinitionNode;
+            const directive = definition?.directives?.length ? (definition.directives[0] as DirectiveNode) : undefined;
+            expect(() => parseNodeDirective(directive, definition)).not.toThrow();
+        });
```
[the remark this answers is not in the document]
litewarp: Yeah, I might have gotten this one backward. Will fix!

packages/graphql/src/schema/make-augmented-schema.ts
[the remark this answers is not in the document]
darrellwarde: Sure, I'm convinced!!

packages/graphql/src/utils/global-ids.ts
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { GraphQLID } from "graphql";
+import { base64, unbase64 } from "graphql-relay/utils/base64";
+
+export function toGlobalId(label: string, field: string, id: string | number): string {
+    return base64([label, field, GraphQLID.serialize(id)].join(":"));
+}
```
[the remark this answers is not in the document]
litewarp: Sure. It adds another step, but is otherwise innocuous. Plus if you ever want to abstract labels and types, I suppose it would be easier.

packages/graphql/src/schema/resolvers/global-node.ts
```diff
@@ -0,0 +1,90 @@
...
+        if (!node) return null;
+
+        // modify the resolve tree and append the fragment selectionSet
+        const parseInfo = parseResolveInfo(info) ?? { fieldsByTypeName: [] };
+
+        const fieldsByTypeName = Object.entries(parseInfo.fieldsByTypeName).reduce((res, [key, value]) => {
+            if (key === "Node") return res;
+
+            if (key === label) {
+                const fields = { ...value, id: { name: "id", alias: "id", args: {}, fieldsByTypeName: {} } };
```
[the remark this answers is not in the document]
litewarp: I can't recall exactly, but it's necessary to resolve some nested or underlying fields. It's getting late here, but if you comment it out, there will be a test that fails which should give you an idea of why the projection is needed.

packages/graphql/src/schema/resolvers/global-node.ts
```diff
@@ -0,0 +1,90 @@
...
+            cypher,
+            params,
+            defaultAccessMode: "READ",
+            context,
+        });
+
+        // TODO: Determine workaround for TCK tests
+        // we should return null if no record is found, but if you
+        // return null here it throws off the tck tests. Same issue as in the
+        // root-connections PR
```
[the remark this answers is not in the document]
litewarp: I suppose (1) decide on what should happen -- should we return null if no node is found, make the return Node type non-null, and then (2) either modify the tck test runner to bail before it errs, or come up with some other workaround.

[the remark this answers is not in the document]
darrellwarde: Definitely return `null` if no node is found - so the query field should look like `node(id: ID!): Node`. With that being the case, no changes to the TCK tests should be needed?

packages/graphql/src/schema/resolvers/global-node.ts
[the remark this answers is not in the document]
darrellwarde: I think I see, fairly sure we do similar things elsewhere for interfaces!

packages/graphql/src/translate/create-projection-and-params.ts
darrellwarde: Yes, I think that conditional is needed - perfectly plausible that `id` will be the name of the property in the database.

packages/graphql/src/utils/global-ids.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { GraphQLID } from "graphql";
+import { base64, unbase64 } from "graphql-relay/utils/base64";
+
+export function toGlobalId(typeName: string, field: string, id: string | number): string {
```
darrellwarde, line 22: **suggestion (non-blocking):** wondering whether this is a touch nicer with named arguments in an object?

```suggestion
export function toGlobalId({ typename, field, id }: { typename: string, field: string, id: string | number }): string {
```

packages/graphql/tests/tck/tck-test-files/global-node.test.ts
```diff
@@ -0,0 +1,110 @@
...
+            variableValues: { id: toGlobalId("Movie", "title", "A River Runs Through It") },
+        });
+
+        expect(formatCypher(result.cypher)).toMatchInlineSnapshot(`
+            "MATCH (this:\`Movie\`)
+            WHERE this.title = $this_title
+            RETURN this { .title } as this"
+        `);
+    });
+    test("it should project the correct node and fields when id is the idField", async () => {
```
darrellwarde, line 70: 👏

packages/graphql/src/translate/where/create-where-and-params.ts
```diff
@@ -50,7 +50,7 @@ function createWhereAndParams({
         return ["", {}];
     }
 
-    function reducer(res: Res, [key, value]: [string, GraphQLWhereArg]): Res {
+    function reducer(res: Res, [key, value]: [string, any | GraphQLWhereArg | GraphQLWhereArg[]]): Res {
```
darrellwarde, line 53: **question:** what's up with loosening the type safety here with `any`?

packages/graphql/src/translate/where/create-where-and-params.test.ts
```diff
@@ -43,4 +43,47 @@ describe("createWhereAndParams", () => {
...
+        })
+            .withNodeDirective({ global: true, idField: "title" })
+            .instance();
+
+        const whereInput = {
+            id: node.toGlobalId("some title"),
+        };
+
+        // @ts-ignore
+        const context: Context = { neoSchema: { nodes: [] } };
```
darrellwarde, line 82: We have a nice `ContextBuilder` class which can be used for type safe test context objects here. 🙂

packages/graphql/src/schema/parse-node-directive.ts
```diff
@@ -37,4 +67,8 @@ function getArgumentValue<T>(directive: DirectiveNode, name: string): T | undefi
     return argument ? (valueFromASTUntyped(argument.value) as T) : undefined;
 }
 
+function alphabeticallySortFields(fields: FieldDefinitionNode[]): FieldDefinitionNode[] {
+    return fields.sort((a, b) => (a.name.value > b.name.value ? 1 : -1));
+}
```
darrellwarde, line 72: **question:** just a quick thought on how most people will grow their schema... It would be cool if between server restarts, and adding fields to types, the global IDs try their best to remain consistent. By sorting alphabetically, if a user adds a new unique field beginning the letter "a", it will automatically become the global ID. As users are quite likely to add new fields to the bottom of a type, are we better taking the first unique or ID field in the type as it is written? I would say that's more likely to remain at the top for longer, and remain a more consistent global ID. Thoughts?

packages/graphql/src/translate/create-where-and-params.ts
```diff
@@ -0,0 +1,723 @@
+/*
```
darrellwarde, line 1: Why is this showing as an entirely new file? 🤔

darrellwarde (commented):
Mostly questions now really, this is looking good!

packages/graphql/src/schema/parse-node-directive.ts
litewarp: The cleanest way to do this would be to have the user define the field in the directive. Then we could do a check to make sure that the field is decorated with the @unique or @id directives. 

There are directives where field names are used. This seems to be a good use case -- no need to infer something so important when the user can just define it.

packages/graphql/src/translate/where/create-where-and-params.ts
litewarp: Not sure if it's loosening -- it's just making the definition explicit for the top level 

This is the def in the types file 
```typescript

export interface GraphQLWhereArg {
    [k: string]: any | GraphQLWhereArg | GraphQLWhereArg[];
    AND?: GraphQLWhereArg[];
    OR?: GraphQLWhereArg[];
}

```

Without making it explicit I was getting ts errors.

packages/graphql/src/utils/global-ids.ts
litewarp: Up to you!

packages/graphql/src/translate/create-where-and-params.ts
litewarp: Befuddled by this one myself. I don't recall making any changes to it for the purposes of this PR.

litewarp:
Thanks, @darrellwarde 

Let me know if there are any immediate changes to make. I'll try and get to it when I can but a bit busy this week. 

- One random item I thought of while trying to fall asleep is that if we're using values other than strings or numbers for potential unique fields, there may be issues with either comparison (e.g., date === date) or serialization. No specific errors in mind yet, but it is foreseeable.

packages/graphql/src/utils/global-ids.ts
darrellwarde: Let's do it!! That way you can jump between `fromGlobalId` and `toGlobalId` without spreading the resulting object.

packages/graphql/src/translate/create-where-and-params.ts
darrellwarde: I see... So on `dev` this has now been moved to https://github.com/neo4j/graphql/blob/dev/packages/graphql/src/translate/where/create-where-and-params.ts and for some reason your merge of `dev` has not moved it to the right location. This will need moving there for the diff to be represented correctly. It will also be worthwhile confirming that filtering on the `id` field also works for connection fields, as the filtering is done in a different file for that.

packages/graphql/src/schema/parse-node-directive.ts
darrellwarde: Potentially... How do you propose we do this?

```gql
type Book @node(globalId: "isbn") {
  isbn: String! @unique
}
```

I think we discussed this before, and we didn't like it because it wasn't 100% apparent that this was a toggle for behaviour.

```gql
type Book @node(global: { enabled: true, id: "isbn" }) {
  isbn: String! @unique
}
```

Perhaps? Appreciate your thoughts!

darrellwarde:
> Thanks, @darrellwarde 
> 
> Let me know if there are any immediate changes to make. I'll try and get to it when I can but a bit busy this week. 
> 
> - One random item I thought of while trying to fall asleep is that if we're using values other than strings or numbers for potential unique fields, there may be issues with either comparison (e.g., date === date) or serialization. No specific errors in mind yet, but it is foreseeable. 

Can you expand on this? If the property has a unique node property constraint in the database, what issues do you foresee?

I also just had another thought whilst reading the security chapter of a GraphQL book - does the `node` query field enforce auth rules? I'm on my phone right now so can't easily check the code, is this tested already?

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -594,10 +594,48 @@ function createProjectionAndParams({
...
+            existingProjection[globalIdField] = {
+                alias: globalIdField,
+                args: {},
+                fieldsByTypeName: {},
+                name: globalIdField,
+            };
+        }
+        if (globalIdField !== "id" && !!existingProjection.id) {
+            delete existingProjection.id;
+        }
```
angrykoala, line 613: **question**: I don't completely understand the logic behind this bit, but why do we need to remove the id field in this case?

angrykoala (changes requested):
Great work here. Some style/organization suggestions.

packages/graphql/src/schema/parse-node-directive.ts
litewarp: Hmm, I'm probably in favor of the first -- I don't see a world in which someone wants to toggle global nodes but not set the idField. In fact, it will err if they don't, so I just see it as adding an unnecessary variable. 

But I get your hesitation, so I have no objection to 2, especially if the alternative is to try and infer it based on the order of the schema. But let me know what you think. This would be a breaking change if implemented later on, so we should probably come to a decision before adding the feature.

packages/graphql/src/translate/create-projection-and-params.ts
litewarp: Let's take two examples:

```graphql
type ExampleOne @node(global: true) {
  title: String! @unique
}

type ExampleTwo @node(global: true) {
  dbId: ID! @id @alias(property: "id")
}
```

For ExampleOne, if you query for the `id` -- i.e. the globalId -- then we need to add `title` to the selectionSet. But we *also* have to remove `id` from the selectionSet otherwise the cypher will attempt to project a field that doesn't exist and we get an error.

For ExampleTwo, if you query for the `id`, we don't need to do anything except let the `id` selection through since it maps to the underlying db field.

So this second conditional checks to see if the globalIdField is "id." If it isn't and the `id` selection is still in the projection, we need to delete it. 

I'm going to refactor this so it's clearer, but that's the explanation!

packages/graphql/src/classes/Node.test.ts
```diff
@@ -787,5 +787,69 @@ describe("Node", () => {
...
+
+            const relayId = node.toGlobalId(value);
+
+            expect(node.fromGlobalId(relayId)).toEqual({
+                typeName: "Film",
+                field: "title",
+                id: value,
+            });
+        });
+        test("should convert a relay id to an object of { typeName: string, field: string, id: any }", () => {
```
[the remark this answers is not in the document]
litewarp: Yup. Must have started out as something else, but now it's virtually identical.

litewarp:
@darrellwarde Alright I think I've gotten everything. It could probably use some more tests, but not sure how much more time I can spend on it in coming weeks. 

There's a weird error in the "app setup" tests where it says a compressed file is corrupted, but that seems unrelated to this PR. In any case I would not know how to fix.

packages/graphql/src/translate/create-projection-and-params.ts
angrykoala: mmm, I see. forgive me if this is a stupid question, but what would happen in a situation like this?:
```graphql
type ExampleOne @node(global: true) {
  title: String! @unique
  dbId: ID! @id @alias(property: "id")
}
```

We would add the global Id (I assume it would be title in this case), but will remove id from the projection

litewarp: This is an interesting question because it gets to whether we should have the user define the globalIdField or we should infer it.

The current inference strategy looks for an `@id` directive first and `@unique` directives second, so in this case, the globalIdField would be "id" and the id would not get deleted.

litewarp:
Hi @darrellwarde & @angrykoala --

I think there's one issue left to resolve in order to get this in final shape and that's how do we want to handle definition of the idField in the directive API. 

Specifically, do we want to infer what the globalIdField (i.e., the field use to calculate the globalId) is based on the `@id` and `@unique` directives, or do we want to allow the user to define it?

Currently, the user needs to know that if there are both `@id` and `@unique` decorated fields, the `@id` field will take precedence. The user also needs to know how to structure the schema if there are more than one `@unique` fields on a type, but no `@id` field. [There are two proposals for this](https://github.com/neo4j/graphql/pull/866#discussion_r836721416) -- we use the first alphabetical field with a `@unique` directive, or we use the first defined field with a `@unique` directive. 

Needless to say, this is complicated and not-so-straightforward to implement. There are also potential side effects -- see, e.g., [the question](https://github.com/neo4j/graphql/pull/866#discussion_r840708993) of how and when to project a field with name "id" when multiple decorated fields are queried.

A [much cleaner proposal](https://github.com/neo4j/graphql/pull/866#discussion_r837424957) would remove the complexity on the user's end as well as reducing complexity in implementation. It would also possibly permit use cases not possible with an inference strategy. For example, a user could utilize a field with a `@unique` directive even when an `@id` directive is also present; with the inference strategy, the `@id` field always trumps.

If this is the way to go (and I think it is), I'll just need your input / decision on what the API should look like. Open to other options as well.

Two Options from @darrellwarde's comment above:

```graphql
type Book @node(globalId: "isbn") {
  isbn: String! @unique
}
```

```graphql
type Book @node(global: { enabled: true, id: "isbn" }) {
  isbn: String! @unique
}
```

darrellwarde:
> Hi @darrellwarde & @angrykoala --
> 
> I think there's one issue left to resolve in order to get this in final shape and that's how do we want to handle definition of the idField in the directive API.
> 
> Specifically, do we want to infer what the globalIdField (i.e., the field use to calculate the globalId) is based on the `@id` and `@unique` directives, or do we want to allow the user to define it?
> 
> Currently, the user needs to know that if there are both `@id` and `@unique` decorated fields, the `@id` field will take precedence. The user also needs to know how to structure the schema if there are more than one `@unique` fields on a type, but no `@id` field. [There are two proposals for this](https://github.com/neo4j/graphql/pull/866#discussion_r836721416) -- we use the first alphabetical field with a `@unique` directive, or we use the first defined field with a `@unique` directive.
> 
> Needless to say, this is complicated and not-so-straightforward to implement. There are also potential side effects -- see, e.g., [the question](https://github.com/neo4j/graphql/pull/866#discussion_r840708993) of how and when to project a field with name "id" when multiple decorated fields are queried.
> 
> A [much cleaner proposal](https://github.com/neo4j/graphql/pull/866#discussion_r837424957) would remove the complexity on the user's end as well as reducing complexity in implementation. It would also possibly permit use cases not possible with an inference strategy. For example, a user could utilize a field with a `@unique` directive even when an `@id` directive is also present; with the inference strategy, the `@id` field always trumps.
> 
> If this is the way to go (and I think it is), I'll just need your input / decision on what the API should look like. Open to other options as well.
> 
> Two Options from @darrellwarde's comment above:
> 
> ```graphql
> type Book @node(globalId: "isbn") {
>   isbn: String! @unique
> }
> ```
> 
> ```graphql
> type Book @node(global: { enabled: true, id: "isbn" }) {
>   isbn: String! @unique
> }
> ```

Sorry for the delay here. Let's definitely discount the second option, no need for the redundant `enabled` argument, that's implied.

So we're left with the following:

```graphql
type Book @node(globalId: "isbn") {
  isbn: String! @unique
}
```

I'll also throw the following option into the ring:

```graphql
type Book {
  isbn: String! @unique @globalId
}
```

And also a new directive:

```graphql
type Book @relay(id: "isbn") {
  isbn: String! @unique
}
```

I'm for some reason really indecisive about this, I think it's just because I don't _love_ any of the proposals. I'll sleep on it tonight and be back with a decision tomorrow!

litewarp:
Hah. No worries! Let's get it right. 

I'll take a crack of implementing this weekend. 

Just to clarify -- if there is an `@id` directive with the `global` argument, it both (i) opts the Type into "implementing Node", and (ii) assigns the decorated field as the underlying field for global id generation?

I think my only stylistic comment is to rename the argument from `global: true` to `relayId: true`

Can't wait to get this merged!

darrellwarde:
> Hah. No worries! Let's get it right.
> 
> I'll take a crack of implementing this weekend.
> 
> Just to clarify -- if there is an `@id` directive with the `global` argument, it both (i) opts the Type into "implementing Node", and (ii) assigns the decorated field as the underlying field for global id generation?
> 
> I think my only stylistic comment is to rename the argument from `global: true` to `relayId: true`
> 
> Can't wait to get this merged!

Gaaahhhh, of course, I forgot all about implementing the `Node` interface, hence why we had the directive and argument on the type like we did!

However, I still feel like this is the tidiest approach - it's opt-in, and as long as we document it appropriately then I don't think there are any nasty surprises.

Let's go with `relay: true` - I feel like the fact that it's an ID is already implied through the fact that it's the `@id` directive?

litewarp:
> Let's go with `relay: true` - I feel like the fact that it's an ID is already implied through the fact that it's the `@id` directive?

So I realized that I need this to get merged soon because it's blocking a project of mine, so I ended up implementing the changes tonight, which should be available for review now. 

I opted to keep the argument `global: true` after some reconsideration.

packages/graphql/src/schema/get-nodes.ts
```diff
@@ -186,6 +186,33 @@ function getNodes(definitionNodes: DefinitionNodes, options: { callbacks?: Neo4j
             cartesianPointInTypeDefs = nodeFields.pointFields.some((field) => field.typeMeta.name === "CartesianPoint");
         }
 
+        const globalIdFields = nodeFields.primitiveFields.filter((field) => field.isGlobalIdField);
+
+        if (globalIdFields.length > 1) {
+            throw new Error(
+                "Only one field may be decorated with an '@id' directive where the global argument is set to true"
```
angrykoala, line 193: I'm not 100% sure, but maybe this is slightly clearer?
```suggestion
                "Only one field may be decorated with an '@id' directive when the global argument is set to true"
```

angrykoala (commented):
Looks good to me. Thanks for this awesome contribution @litewarp

packages/graphql/src/schema/get-nodes.ts
darrellwarde:
```suggestion
                "Only one field may be decorated with an '@id' directive with the global argument set to `true`"
```

packages/graphql/src/schema/make-augmented-schema.test.ts
```diff
@@ -429,4 +429,47 @@ describe("makeAugmentedSchema", () => {
...
+    describe("global nodes", () => {
+        test("should throw error if more than one @id directive field has the global argument set to true", () => {
+            const typeDefs = gql`
+                type User {
+                    email: ID! @id(global: true)
+                    name: ID! @id(global: true)
+                }
+            `;
+            expect(() => makeAugmentedSchema(typeDefs)).toThrow(
+                "Only one field may be decorated with an '@id' directive where the global argument is set to true"
```
darrellwarde, line 441:
```suggestion
                "Only one field may be decorated with an '@id' directive with the global argument set to `true`"
```

darrellwarde (dismissed):
This is looking great, once again a slow review, I am sorry!! I will just quickly tidy up that extra unneeded argument, and then merge!!

PR #867, opened 2022-01-25
Undirected aggregation and connection

angrykoala:
# Description
This PR adds support for _undirected relationships_ in _connection_ and _aggregate_ queries as defined in [this rfc](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-004-undirected-relationship-queries.md#connection-and-aggregations)

For instance, the following graphql:
```graphql
query Users {
  users {
    friendsAggregate(directed: false) {
        count
    }
  }
}
```

yields the following Cypher:
```cypher
MATCH (this:User)
RETURN this { friendsAggregate: { count: head(apoc.cypher.runFirstColumn(\\"
    MATCH (this)-[r:FRIENDS_WITH]-(n:User)      
    RETURN COUNT(n)\\", { this: this })) 
} } as this
```


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-867.surge.sh

docs/modules/ROOT/pages/queries.adoc
```diff
@@ -254,3 +270,19 @@ query {
     }
 }
 ----
+
+=== Aggregate undirected relationships
+When performing an aggregation on related nodes, the relationship
+can be defined as an xref::queries.adoc#_undirected_queries[undirected relationship] using the argument `directed: false`:
```
darrellwarde, line 276: **suggestion:** We should be careful about phrasing in relation to this - this makes it sound like the relationship has no direction when in fact we are only querying it disregarding its direction. Perhaps the following?

```suggestion
=== Undirected aggregation queries

When performing an aggregation on related nodes, the query against the relationship
can be defined as an xref::queries.adoc#_undirected_queries[undirected] using the argument `directed: false`:
```

darrellwarde (dismissed):
Looks good to me, would just like to see some of the phrasing in the documentation made more clear.

Issue #809, opened 2022-01-18
Make items within arrays required


Andy2003:
All array items should be non-nullable.

This is to track issues with array fields currently allowing `null` values:

- [x] `AND` and `OR` in unions where [ref](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/connect-or-create-unions.test.ts#L110-L111)
- [x] Sorting fields [ref](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/aggregations.test.ts#L176)
- [x] All `_IN_ clauses in the where type [ref](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/aggregations.test.ts#L214)

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @Andy2003! :pray: We will now prioritise the bug and address it appropriately.

neo4j-team-graphql:
This bug report has been assigned high priority to fix. If you wish to contribute a fix, please branch from `master` and submit your PR with the base set to `master`. Thanks!

darrellwarde:
First two fixed in #874, last one intentional because it allows you to check "value is either this value or unset on the node". Perhaps this could be changed for non-nullable fields.

Andy2003:
@darrellwarde I found another occurence of nullabel array items: https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/comments.test.ts#L939

PR #874, opened 2022-01-25
Fix/nullable array entries

darrellwarde:
# Description

Fixes the first two issues listed in #809.

# Issue

Closes:
- #809


packages/graphql/src/schema/get-where-fields.ts
```diff
@@ -60,8 +60,8 @@ function getWhereFields({
                     return res;
                 }
 
-                res[`${f.fieldName}_IN`] = `[${f.typeMeta.input.where.pretty}]`;
-                res[`${f.fieldName}_NOT_IN`] = `[${f.typeMeta.input.where.pretty}]`;
+                res[`${f.fieldName}_IN`] = `[${f.typeMeta.input.where.pretty}${f.typeMeta.required ? "!" : ""}]`;
```
angrykoala, line 63: **suggestion (non-blocking)**:
A helper function `makeRequired(str: string, required: boolean): string` to make a string required could be helpful to improve readability and make this common use case a bit more consistent in our codebase, something a bit like:

```suggestion
                res[`${f.fieldName}_IN`] = `[${makeRequired(f.typeMeta.input.where.pretty, f.typeMeta.required) }]`;
```

angrykoala (approved):
:clap:

packages/graphql/src/schema/get-where-fields.ts
darrellwarde: This use case is actually pretty uncommon - normally I would have just been able to use `f.typeMeta.input.where.pretty` but this new nullability logic for `_IN` means it needs to be slightly different here. So I think a helper would only be used in this one place anyway.

PR #875, opened 2022-01-25
Crude check if any empty interfaces or objects exist in the schema

### LINKED ISSUE #870: If all field of an interface are private an invalid schema is generated
Given the following schema ([taken from the tests](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/directives/private.test.ts#L28-L36)):

```graphql
interface UserInterface {
  private: String @private
}

type User implements UserInterface {
  id: ID
  password: String @private
  private: String
}
```

[An empty interface is generated](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/directives/private.test.ts#L113):

```graphql
interface UserInterface
```

Which [is not allowed by the GrapQL spec](https://spec.graphql.org/June2018/#sec-Interfaces):

> An Interface type must define one or more fields.

darrellwarde:
# Description

If after full augmentation, any interfaces or objects with no fields exist, throw an error, because this is a violation of the GraphQL spec.

# Issue

Closes #870


angrykoala (approved):

darrellwarde:
Let's get this merged but be aware of the potential impact on performance - we should probably amalgamate any and all "post" checks into one iteration of the output document.

Issue #841, opened 2022-01-20
`assertIndexesAndConstraints` with `create` option throws an error if fulltext index already exists


darrellwarde:
**Describe the bug**

We should check whether fulltext indexes already exist before trying to create them.

This is a side effect of having to use `db.index.fulltext.createNodeIndex` over `CREATE FULLTEXT INDEX` (to support DB 4.1).

Customer reported.

neo4j-team-graphql:
Many thanks for raising this bug report @darrellwarde. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @darrellwarde! :pray: We will now prioritise the bug and address it appropriately.

darrellwarde:
This can be reproduced by taking any type definitions with a `@fulltext` definition and then calling `assertIndexesAndConstraints` twice. The second time will throw an error. Partial example:

```ts
const typeDefs = gql`
    type Product @fulltext(indexes: [{ name: "ProductName", fields: ["name"] }]) {
        name: String!
    }
`;

const neoSchema = new Neo4jGraphQL({
    driver,
    typeDefs,
});

await neoSchema.assertIndexesAndConstraints({ options: { create: true } });
await neoSchema.assertIndexesAndConstraints({ options: { create: true } }); // throws
```

Assertions should be idempotent.

neo4j-team-graphql:
This bug report has been assigned high priority to fix. If you wish to contribute a fix, please branch from `master` and submit your PR with the base set to `master`. Thanks!

mobsean:
Hi,
I´m also running into this problem.

I can see this merged to **neo4j:dev** branch,  when will it be merged to the master branch?

PR #876, opened 2022-01-25
Check whether fulltext indexes exist before creating

darrellwarde:
# Description

Fixes the fact that `assertIndexesAndConstraints` throws an error if a fulltext index already exists.

Still throws if the index exists, but the fields don't match.

# Issue

Closes: #841


packages/graphql/tests/integration/indexes-and-constraints/fulltext.int.test.ts
```diff
@@ -530,4 +530,132 @@ describe("assertIndexesAndConstraints/fulltext", () => {
             `@fulltext index '${indexName}' on Node '${type.name}' is missing field 'description' aliased to field '${alias}'`
         );
     });
+
+    test("should create index if it doesn't exist and not throw if it does exist, and then query using the index", async () => {
+        // Skip if multi-db not supported
+        if (!MULTIDB_SUPPORT) {
```
danstarns, line 536: **thought (non-blocking)** I see this is repeated maybe we can make a function to abstract this.

danstarns (dismissed):

PR #645, opened 2021-12-15
RFC: Query Limits

darrellwarde:
RFC for query limits

PR #883, opened 2022-01-27
Feat/Query Limits

danstarns:
# Description

Adds the `@queryOptions` directive to nodes and allows users to specify the `defaultLimit` argument. 

I decided not to use a default value stored in the schema due to the fact that it would mean we would have to generate yet another type for each node. 

# RFC
#645


packages/graphql/src/translate/translate-read.ts
```diff
@@ -44,6 +44,10 @@ function translateRead({ node, context }: { context: Context; node: Node }): [st
     const connectionStrs: string[] = [];
     const interfaceStrs: string[] = [];
 
+    if (node.queryOptions?.defaultLimit && !optionsInput.limit) {
+        optionsInput.limit = node.queryOptions.defaultLimit;
+    }
```
angrykoala, line 49: **suggestion**: Two minor things that worry me about this approach
* this is duplicate from `create-projection-and-params`, I think some common functions to handle `limit` could help avoid duplication.
* Modifying an existing object (`field.args.options`)

Maybe it would be worth some function such as 

```typescript
function getLimit(node: Node, field: ResolveTree): { limit?: Integer} {
// .....
}
```

This way, we can obtain the real limit options (including default), without modifying our pseudo global object `field.args.options`

packages/graphql/tests/integration/rfcs/query-limits.int.test.ts
```diff
@@ -0,0 +1,212 @@
...
+                });
+
+                if (gqlResult.errors) {
+                    console.log(JSON.stringify(gqlResult.errors, null, 2));
+                }
+
+                expect(gqlResult.errors).toBeUndefined();
+                expect((gqlResult.data as any)[randomType.plural]).toHaveLength(2);
+            } finally {
+                await session.close();
```
angrykoala, line 82: **question/suggestion (non-blocking)**: Is there a reason for not using the `beforeEach` and `afterEach` hooks for this?

```typescript
let session: Session

beforeEach(()=>{
  session = driver.session();
});

afterEach(async()=>{
  await session.close();
});
```

The `afterEach` block should be executed regardless of the test failing or not, this way we can avoid ùsing `try...finally` block in our test case

angrykoala (changes requested):
Looking really good :clap: :clap:  , just a few small comments on some suggestions/concerns

angrykoala:
As discussed, the api may be changed to `limit` is actually an object with `max` and `default` values. I'll cover that in a PR to the rfc @danstarns

angrykoala:
According to the [rfc](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-005-query-limits.md) `QueryOptions` limit should have a default limit, I think this has not been implemented in this PR

i.e.
```graphql
input MovieOptions {
  offset: Int
  limit: Int! = 10
  sort: [MovieSort]
}
```

darrellwarde:
@angrykoala, love the thorough review, but unless you've agreed otherwise with @danstarns, can you please take the suggestions forward, either in this branch and PR, or by merging this branch and making the necessary changes in a branch of your own? Happy to discuss tomorrow!

angrykoala:
@danstarns I made the changes we discussed as well as the interface from `defaultLimit` to `limit: {default}` 
If you are happy with these changes this is ready merge

docs/modules/ROOT/pages/directives.adoc
```diff
@@ -108,3 +108,9 @@ Reference: xref::type-definitions/indexes-and-constraints.adoc#type-definitions-
 The `@writeonly` directive marks fields as write-only.
 
 Reference: xref::type-definitions/access-control.adoc#type-definitions-access-control-writeonly[`@writeonly`]
+
+== `@queryOptions`
+
+The `@queryOptions` is to be used on nodes, where applied will inject default values into a query such as the `defaultLimit`.
```
danstarns, line 114:
```suggestion
The `@queryOptions` is to be used on nodes, where applied will inject default values into a query such as a default limit. 
```

packages/graphql/src/schema/validation/directives.ts
```diff
@@ -253,3 +253,23 @@ export const fulltextDirective = new GraphQLDirective({
     },
     locations: [DirectiveLocation.OBJECT],
 });
+
+export const queryOptions = new GraphQLDirective({
+    name: "queryOptions",
+    description: "Instructs @neo4j/graphql to inject default values into a query such as the `defaultLimit`.",
```
danstarns, line 259:
```suggestion
    description: "Instructs @neo4j/graphql to inject default values into a query such as a default limit.",
```

docs/modules/ROOT/pages/type-definitions/default-values.adoc
```diff
@@ -38,3 +38,19 @@ directive @coalesce(
...
+== `@queryOptions`
+
+The `@queryOptions` is to be used on nodes, where applied will inject default values into a query such as the `defaultLimit`.
+
+=== Definition
+
+[source, graphql, indent=0]
+----
+"""The `@queryOptions` is to be used on nodes, where applied will inject default values into a query such as the `defaultLimit`."""
+directive @queryOptions(
```
danstarns, line 52: **(blocking)** This will need to change to reflect the new definition.

docs/modules/ROOT/pages/type-definitions/default-values.adoc
```diff
@@ -38,3 +38,19 @@ directive @coalesce(
     value: Scalar!,
 ) on FIELD_DEFINITION
 ----
+
+[[type-definitions-default-values-queryoptions]]
+== `@queryOptions`
+
+The `@queryOptions` is to be used on nodes, where applied will inject default values into a query such as the `defaultLimit`.
```
danstarns, line 45:
```suggestion
The `@queryOptions` is to be used on nodes, where applied will inject default values into a query such as a default limit.
```

danstarns (commented):
@angrykoala thank you! Just some doc changes I suggest and it's good to go!

PR #888, opened 2022-01-28
Query direction argument

angrykoala:
# Description

Adds queryDirection parameter to relationships as defined in [rfc-004](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-004-undirected-relationship-queries.md#querydirection-option) 

Documentation updates for this PR are in #889


packages/graphql/src/schema/validation/directives.ts
```diff
@@ -163,6 +168,9 @@ export const relationshipDirective = new GraphQLDirective({
         type: {
             type: new GraphQLNonNull(GraphQLString),
         },
+        queryDirection: {
+            type: RelationshipQueryDirectionEnum,
+        },
```
darrellwarde, line 173: **suggestion (blocking):** This should have a default value of `DEFAULT_DIRECTED` should it not? I know we don't use these directive definitions explicitly, but they should mirror our expectations of the directive.

packages/graphql/src/translate/cypher-builder/get-relationship-direction.ts
```diff
@@ -0,0 +1,61 @@
...
+            return undirectedArgs;
+        case RelationshipQueryDirectionOption.DIRECTED_ONLY:
+            if (fieldArgs.directed === false) throw new Error("Invalid direction in 'DIRECTED_ONLY' relationship");
+            return directedArgs;
+        case RelationshipQueryDirectionOption.UNDIRECTED_ONLY:
+            if (fieldArgs.directed === true) throw new Error("Invalid direction in 'UNDIRECTED_ONLY' relationship");
+            return undirectedArgs;
+        default:
+            throw new Neo4jGraphQLError(`Invalid queryDirection argument ${relationField.queryDirection}`);
+    }
```
darrellwarde, line 60: **suggestion (blocking):** I know this feels like a non-blocking suggestion as it's stylistic, but we've always had a hard line on `if` statements without curly braces. I'm amazed https://eslint.org/docs/rules/curly isn't enabled, I think that's something we should have enabled.

packages/graphql/src/schema/directed-argument.ts
```diff
@@ -0,0 +1,56 @@
...
+        defaultValue,
+    };
+}
+
+export function addDirectedArgument<T extends Record<string, any>>(
+    args: T,
+    relationField: RelationField
+): T & { directed?: DirectedArgument } {
+    const directedArg = getDirectedArgument(relationField);
+    if (directedArg) return { ...args, directed: directedArg };
```
darrellwarde, line 54: **suggestion (blocking):** Same comment here on `if` without braces, really affects readability.

darrellwarde (changes requested):
Really nice clean work! Sorry about the comments regarding code formatting, but I think for this case it is worth it.

darrellwarde (approved):
Looks great, thanks for making those changes!

darrellwarde (approved):

PR #892, opened 2022-01-28
Fix inconsistent ConnectionWhere type names for unions

### LINKED ISSUE #871: Harmonize names for connection-where of unions
Given the following schema ([taken from the tests](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/unions.test.ts#L28-L38)):

```graphql
union Search = Movie | Genre

type Genre {
  id: ID
}

type Movie {
  id: ID
  search: [Search!]! @relationship(type: "SEARCH", direction: OUT)
  searchNoDirective: Search
}
```

The following type is generated among others:

```graphql
input MovieSearchConnectionGenreWhere {
  AND: [MovieSearchConnectionGenreWhere!]
  OR: [MovieSearchConnectionGenreWhere!]
  node: GenreWhere
  node_NOT: GenreWhere
}
```

All other connection-where-types are named `${prefix}${node.name}ConnectionWhere` but for unions the name is `${prefix}Connection${node.name}Where`, we should harmonize them!

darrellwarde:
# Description

Fix inconsistent ConnectionWhere type names for unions

# Issue

Closes #871


packages/graphql/src/schema/create-connection-fields.ts
```diff
@@ -125,7 +125,13 @@ function createConnectionFields({
             const relatedNodes = nodes.filter((n) => connectionField.relationship.union?.nodes?.includes(n.name));
 
             relatedNodes.forEach((n) => {
-                const unionWhereName = `${connectionField.typeMeta.name}${n.name}Where`;
+                const connectionName = connectionField.typeMeta.name;
+
+                // Append union member name before "ConnectionWhere"
+                const unionWhereName = `${connectionName.substring(0, connectionName.length - "Connection".length)}${
```
angrykoala, line 131: **question**: I assume you already checked this and the answer is now, but there is no way of composing this name in a way that does not rely on the use of substring?
It feels a bit flaky

angrykoala (approved):

packages/graphql/src/schema/create-connection-fields.ts
darrellwarde: I would like to revisit this in the future - pinky promise that the `substring` won't live forever? 🥺

PR #896, opened 2022-01-31
Asynchronous schema generation and OGM initialization

darrellwarde:
# Description

This PR moves the heavy lifting of the `Neo4jGraphQL` and `OGM` constructors into asynchronous functions, `getSchema()` and `init()`, respectively.

This means that construction will be quick and these functions will have to be awaited following construction. Whilst this may seem like a nuisance right now, in the future this will open doors to more performant schema generation through asynchronous processing.

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-896.surge.sh

darrellwarde:
> Great that this is happening. I guess this means you'll be going for a new major release reasonably soon. Any idea of roughly when the release is scheduled?

Hey @wheresrhys! We wouldn't normally give timelines, but this release has a fairly strict schedule, so you can expect this next week! 🙂 It won't be as much of a jarring release as 3.0.0, but certainly some nice changes to come.

angrykoala (changes requested):
Overall, the solution looks solid, I just commented on a few style/legibility suggestions and improvements

angrykoala (approved):

packages/graphql/tests/integration/custom-resolvers.int.test.ts
```diff
@@ -213,6 +213,65 @@ describe("Custom Resolvers", () => {
         expect(next.value.id).toEqual(id);
     });
 
+    test("should accept an array of custom resolvers", async () => {
```
angrykoala, line 216: :clap:

packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -128,6 +128,50 @@ class Neo4jGraphQL {
         await assertIndexesAndConstraints({ driver, driverConfig, nodes: this.nodes, options: input.options });
     }
 
+    private wrapResolvers(resolvers: IResolvers, { schema }: { schema: GraphQLSchema }) {
```
angrykoala, line 132: **nitpick (non-blocking)**: Maybe we could move these functions to the bottom?
Usually, it is slightly easier to read functions in a class from more abstract to less abstract (kind of like articles are composed). So, first, we read `generateSchema` at the top, and then, if we need to, we keep going to the functions that are used by that one

packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -128,6 +128,50 @@ class Neo4jGraphQL {
...
+            relationships: this.relationships,
+            schema,
+        };
+
+        const resolversComposition = {
+            "Query.*": [wrapResolver(wrapResolverArgs)],
+            "Mutation.*": [wrapResolver(wrapResolverArgs)],
+        };
+
+        // this.schemaDefinition.resolvers can either be undefined, IResolvers or IResolvers[]
```
angrykoala: **nitpick (non-blocking)**: I think this comment is not too relevant, types should already make that clear

packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -128,6 +128,50 @@ class Neo4jGraphQL {
...
+        let allResolvers = [resolvers];
+        if (this.schemaDefinition.resolvers) {
+            if (Array.isArray(this.schemaDefinition.resolvers)) {
+                allResolvers = [...allResolvers, ...this.schemaDefinition.resolvers];
+            } else {
+                allResolvers.push(this.schemaDefinition.resolvers);
+            }
+        }
```
angrykoala: **suggestion:** I just though we could use `asArray` here:

```suggestion
        const allResolvers=[resolvers, ...asArray(this.schemaDefinition.resolvers)];
```

packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -128,6 +128,50 @@ class Neo4jGraphQL {
...
+    private addWrappedResolversToSchema(resolverlessSchema: GraphQLSchema, resolvers: IResolvers): GraphQLSchema {
+        const schema = addResolversToSchema(resolverlessSchema, resolvers);
+
+        // Assign a default field resolver to account for aliasing of fields
+        forEachField(schema, (field) => {
+            if (!field.resolve) {
+                // eslint-disable-next-line no-param-reassign
+                field.resolve = defaultFieldResolver;
+            }
+        });
```
angrykoala, line 161: **suggestion (non-blocking)**: I know it may be pushing a bit, but it could be sooo nice
```suggestion
    return addDefaultResolvers(schema)
```

angrykoala (approved):

PR #902, opened 2022-02-01
Adds default and max options to limit

angrykoala:
# Description
This PR updates the rfc to cover both options `default` and `max`.

The proposed solution, involves the following arguments for `queryOptions`:

```
@queryOptions(limit: {default: 5, max: 10}) {
```

Alternatively, 2 separate arguments could be used:

```
@queryOptions(defaultLimit: 5, maxLimit: 10) {
```

### DIFF
--- a/docs/rfcs/rfc-005-query-limits.md
+++ b/docs/rfcs/rfc-005-query-limits.md
@@ -6,7 +6,7 @@ Users want to be able to hard limit the maximum number of results that can be re
 
 ## Proposed Solution
 
-A new directive, `@queryOptions` to be added, which has an argument `limit` for specifying the maximum number of nodes for this type which should be returned from a query.
+A new directive, `@queryOptions` to be added, which has an argument `limit` for specifying the default **and** maximum number of nodes for this type which should be returned from a query.
 
 The minimum value for `limit` should be 1.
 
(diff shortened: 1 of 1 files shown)


docs/rfcs/rfc-005-query-limits.md
```diff
@@ -25,7 +25,7 @@ Would result in the following type definitions being generated:
 ```graphql
 input MovieOptions {
   offset: Int
-  limit: Int! = 10
+  limit: Int! = 5
```
darrellwarde, line 28: From my discussion with him earlier, @danstarns did not end up outputting this in the schema - so shall we remove this or are you going to add it in @angrykoala?

docs/rfcs/rfc-005-query-limits.md
```diff
@@ -91,7 +91,7 @@ LIMIT 10
 #### Relationship field
 
 ```graphql
-type Movie @queryOptions(limit: 10) {
+type Movie @queryOptions(limit: {default=10, max: 10}) {
```
darrellwarde, line 94:
```suggestion
type Movie @queryOptions(limit: {default: 10, max: 10}) {
```

darrellwarde (commented):
Broadly looks good - one typo and one question!

danstarns (dismissed):

docs/rfcs/rfc-005-query-limits.md
angrykoala: Agree, this is nice, but adding this means adding a lot of extra types
```suggestion
  limit: Int!
```

PR #914, opened 2022-02-03
Merge root-level args from resolvers when translating queries and mutations

nelsonpecora:
# Description

This PR makes the `graphql-middleware` pattern discussed in #692 more useful for mutations, as it allows the autogenerated Neo4J/GraphQL resolvers to make use of `args` which may have been modified by middleware resolvers.

In this code, root-level `args` from the resolver are merged with the `args` from the `resolveTree`, favoring the former. This allows a middleware resolver to modify those args before they're passed into the translation functions like `translateCreate()` or `translateUpdate()`.

For example, if I have a middleware resolver for a `createNodes` mutation that adds additional properties to the `input` array when saving, I can modify the `input` arg before sending it onwards:

```js
async function createNodes (resolve, root, args, ctx, info) {
  const nodesToCreate = args.input;
  const nodesWithExtraProps = nodesToCreate.map((node) => ({
    ...node,
    additionalProp: doSomeBusinessLogicTo(node.someProp)
  }))

  // Pass our new args to the autogenerated createNodes() resolver.
  return resolve(root, { input: nodesWIthExtraProps }, ctx, info);
}
``` 

# Issue

We originally discussed this scenario [on discord](https://discord.com/channels/787399249741479977/818578492723036210/938782849375551508), but I [added a comment](https://github.com/neo4j/graphql/issues/692#issuecomment-1029082690) to #692 summarizing it.

# Checklist


nelsonpecora:
The logic for pulling in the resolver `args` was fairly straightforward, but I'm struggling a bit to assign the right types in `translate-create.js` and `translate-update.js` 😓

packages/graphql/src/translate/translate-update.ts
```diff
@@ -33,14 +33,15 @@ import translateTopLevelMatch from "./translate-top-level-match";
...
-    const createInput = resolveTree.args.create;
-    const deleteInput = resolveTree.args.delete;
-    const connectOrCreateInput = resolveTree.args.connectOrCreate;
+    const rootArgs = { ...resolveTree.args, ...args };
+    const updateInput = rootArgs.update;
+    const connectInput = rootArgs.connect;
+    const disconnectInput = rootArgs.disconnect;
+    const createInput = rootArgs.create;
+    const deleteInput = rootArgs.delete;
+    const connectOrCreateInput = rootArgs.connectOrCreate;
```
nelsonpecora, line 44: These variables are the main reason tck tests seem to be failing. They get the correct types on `dev`, but I can't seem to figure out what those should be.

packages/graphql/src/schema/resolvers/create.ts
```diff
@@ -29,7 +29,7 @@ export default function createResolver({ node }: { node: Node }) {
     async function resolve(_root: any, _args: any, _context: unknown, info: GraphQLResolveInfo) {
```
darrellwarde, line 29: I think the underscore could be removed here, which means you can just do `translateCreate({ context, node, args });` below. It was just this way because it was unused.

```suggestion
    async function resolve(_root: any, args: any, _context: unknown, info: GraphQLResolveInfo) {
```

packages/graphql/src/schema/resolvers/create.ts
```diff
@@ -29,7 +29,7 @@ export default function createResolver({ node }: { node: Node }) {
...
         const context = _context as Context;
         context.resolveTree = getNeo4jResolveTree(info);
```
darrellwarde, line 31: As opposed to the solution that you've presented here, you could actually add an `args` argument to `getNeo4jResolveTree` and then augment the args in the resolve tree with the ones from `args`. Then you only need to do this in one place as opposed to doing it in each translate function. 

Happy to walk through this if you're unsure. 🙂

darrellwarde (changes requested):
This looks like a good solution, but I think there could be a more efficient way of achieving it! We will also need some integration tests to prove that this works in practice.

Great first contribution! But could you please sign the CLA if you haven't done so already? https://neo4j.com/developer/cla/#sign-cla

packages/graphql/src/schema/resolvers/create.ts
nelsonpecora: That makes sense to me! It was actually my initial approach, but I was hesitant to modify something so deep in the logic. I'll take a stab at that, thanks 😄

packages/graphql/src/schema/resolvers/create.ts
nelsonpecora: Oh nice!

nelsonpecora:
Pushed up a change that consolidates the code! That also seems to fix most of the type issues, but tests with `limit` and `offset` are failing (among others). It seems like something converts certain numerical args into `{ high, low }`, and those conversions are being clobbered when the resolver `args` overwrite them. I'm still digging into the code and trying to find out where that logic is happening, though.

darrellwarde:
This looks great - the solution is nice and tidy. However, we'll need to see at least one integration test, ideally using `graphql-middleware`, to get this merged in. 🙂

darrellwarde:
@nelsonpecora, can you please ensure that you give maintainers permission to update PRs when you create them please? Makes this very difficult to merge when we can't update the branch with changes from `dev`.

nelsonpecora:
Ah, sorry about that! I didn't see the option on these PRs, and I suspect it's because I forked the repo into an org instead of my personal account. I'll make sure to use my personal account for future PRs 😅

PR #921, opened 2022-02-03
Resolve fields on interfaces returned by custom cypher

AccsoSG:
# Description

As an extension of #495 and #487, I tried to resolve interface fields that use a custom cypher query. The implementation is similar to that for unions (#495). It generally works, but I don't know enough about the code to be sure. I'm sure there is still work to be done, but I still wanted to share my insights.

I found the following disadvantages/problems with my implementation (These actually also apply to the implementation of unions #495):
- If you use a sorting, things get mixed up again, because in a further step sorting is done according to the types of the interface.
- The comparison on Labels and AdditionalLabels is very inefficient if the interface has been implemented by many types and/or multiple additionallabels have been created. This is used to properly generate the __resolveType property.

# Issue

#487

Closes:

- 

# Checklist

### DIFF
--- a/.github/workflows/bug-confirmed-comment.yml
+++ b/.github/workflows/bug-confirmed-comment.yml
@@ -1,11 +1,11 @@
-name: Bug confirmed label
+name: Bug confirmed comment
 
 on:
   issues:
     types: [labeled]
 
 jobs:
-  remove_labels:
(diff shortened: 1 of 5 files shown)


AccsoSG:
I took the most recent version and repeated my changes. So I closed this PR and recreated #926 .

Issue #660, opened 2021-12-21
Allow use of @coalesce on enum fields


wheresrhys:
**Is your feature request related to a problem? Please describe.**
I was one of the original requesters of `@coalesce` but I've just discovered it doesn't meet my use case as it is restricted to primitive types. The primary use case I have is that records to filter records that _do not_ have a particular value of an enum stored in them, including those that do not have a value defined.

**Describe the solution you'd like**
Allow `@coalesce` to be used on enum fields. The value passed in to `@coalesce` should be allowed to be any string (not restricted to the values of the enum) as its purpose in my use case is to avoid the sometimes unintuitive behaviour caused by querying against a field that is sometimes `null`

**Additional context**
https://github.com/neo4j/graphql/issues/157

darrellwarde:
This sounds like a good idea!

> The value passed in to `@coalesce` should be allowed to be any string (not restricted to the values of the enum)

However, fairly sure this won't work because the GraphQL runtime would complain that the values aren't present in the enum.

wheresrhys:
> However, fairly sure this won't work because the GraphQL runtime would complain that the values aren't present in the enum.

Although with coalesce, it's only use in the neo4j querying and no value is returned by the API IIUC... so would this be an issue? Although I can see that it would be impossible to query explicitly for records with a field that is unset as GraphQL would not allow the coalesced value to be passed in to the query. While I am not a fan of reimplementing `null`, perhaps a way forward for us is to add `NULL` or `UNDEFINED` as an option in all our enums, and coalesce to this

A follow up use case I've also just stumbled across is temporal types (and I think the same could be said for spatial types too). e.g. to facilitate a query such as 'Return all systems where the last release timestamp is before this year or has never been set' it would be useful to avoid having to write more complex GraphQL queries to deal with `null`. Coalescing to the UNIX epoch feels like it might be a good approach, but then also we basically lose the ability to intuitively query for records where the date is not set.

I guess there is a trade off between ease of querying "X or null" vs "null" which we'll have to work out the right approach for us

darrellwarde:
> > However, fairly sure this won't work because the GraphQL runtime would complain that the values aren't present in the enum.
> 
[...]

PR #945, opened 2022-02-07
Add @default logic for enum fields

### LINKED ISSUE #944: @default support for Enum values
I love how flexible the `@default` directive is, and how it handles String, Int, etc primitive values. It would be really useful if it was also able to handle things like enums, since they're somewhat similar to primitive String values.

Something like this seems ideal to me:

```graphql
type SomeObject {
  field: SomeEnum! @default(value: A_VALUE)
}

enum SomeEnum {
  A_VALUE
  ANOTHER_VALUE
}
```

The idea being that it would be able to handle enums directly, and the I wouldn't be able to specify a default value that is not in the referenced enum. I believe a lot of this is already handled in GraphQL directly, e.g.

```
input SomeInput {
  field: SomeEnum = A_VALUE
}
```

An alternative is adding imperative code to _create_ mutations to set default values, though that is a less elegant solution than extending the existing `@default` logic.

**Additional context**
I took a stab at doing this in a PR, and it looks like we _do_ get a lot of type safety directly from GraphQL. I feel like `@default` could be extended to also include custom scalars, though that seems more dangerous (maybe it should only cover things like `Duration`?)

nelsonpecora:
# Description

This PR allows the use of the `@default` directive on enum fields (in addition to primitive types).

```graphql
enum SomeEnum {
  FIRST_VAL
  SECOND_VAL
}

type SomeObj {
  aField: SomeEnum! @default(value: FIRST_VAL)
}
```

When the directive is parsed, the generated input types have the correct values. I tested them in a basic Apollo Server setup and it turns out enum default values are perfectly valid GraphQL 😄 

```graphql
input SomeObjCreateInput {
  aField: SomeEnum! = FIRST_VAL
}

input SomeObjUpdateInput {
  aField: SomeEnum
}
```

# Issue

Fixes #944, allowing this on enum fields. But this PR does not extend the functionality to any other types for the reasons listed in that issue. Additionally, I wonder if the discussion in #660 is relevant to this as well?

Also, two questions:

* Should this have an example in the docs? (I feel like the answer is yes)
* Should the error message (when using `@default` on disallowed types) be changed from "only primitives" to "only primitives and enums"? I hesitated on making that change because it would touch a lot of places in the code, and I'm not sure how valuable that additional context would be in those error messages.

Closes:

- #944

# Checklist


packages/graphql/src/types.ts
```diff
@@ -166,6 +166,7 @@ export type CustomScalarField = BaseField;
 export interface CustomEnumField extends BaseField {
     // TODO Must be "Enum" - really needs refactoring into classes
     kind: string;
+    defaultValue?: any;
```
angrykoala, line 169: **nitpick (blocking)** I may be wrong, but for an enum field it looks like this is always going to be of type string, so maybe we can use a stricter type here?

```suggestion
    defaultValue?: string;
```

packages/graphql/src/schema/get-obj-field-meta.ts
```diff
@@ -318,6 +315,17 @@ function getObjFieldMeta({
...
+
+                if (defaultDirective) {
+                    const value = defaultDirective.arguments?.find((a) => a.name.value === "value")?.value;
+
+                    if (value?.kind !== Kind.ENUM) {
+                        throw new Error("@default value on enum fields must be an enum value");
+                    }
+
+                    enumField.defaultValue = (value as EnumValueNode).value;
+                }
```
angrykoala, line 327: **suggestion (blocking)**: We could improve these typings a bit, to avoid casting. By using a helper function to check the value kind:
```typescript
function isEnumValue(value: ValueNode): value is EnumValueNode {
     return value.kind === Kind.ENUM;
}
```
This function, for the time being, could be in the same file at the bottom.

This way,  this check can be safely typed:

```suggestion
                if (defaultDirective) {
                    const defaultValue = defaultDirective.arguments?.find((a) => a.name.value === "value")?.value;

                    if (!defaultValue || !isEnumValue(defaultValue)) {
                        throw new Error("@default value on enum fields must be an enum value");
                    }

                    enumField.defaultValue = defaultValue.value;
                }
```

angrykoala (changes requested):
Looking good, this looks like a great and intuitive feature, and it seems to be working as expected. I would like to see at least one happy-path integration test, to ensure that the default value is added to the database when a new item is created.

As you mentioned, it would be great to have an example of an enum [here](https://neo4j.com/docs/graphql-manual/current/type-definitions/default-values/#type-definitions-default-values-default)

darrellwarde:
I notice that https://github.com/newsela/graphql/blob/default-enums/packages/graphql/src/schema/validation/directives.ts#L79 has not been updated - and as such the schema validation will also not have been updated. The directive definition will need to have been updated to accept both scalar values and enums (this might be difficult) before we can merge this.

darrellwarde:
> I notice that https://github.com/newsela/graphql/blob/default-enums/packages/graphql/src/schema/validation/directives.ts#L79 has not been updated - and as such the schema validation will also not have been updated. The directive definition will need to have been updated to accept both scalar values and enums (this might be difficult) before we can merge this.

Okay, we've discussed this and this will actually be really awkward. So following the request changes, we'll get this merged, then get this also working with `@coalesce`, and then we can update the `ScalarType` to be `ScalarOrEnumType`. However, this is still a bit of a hack, but the directive is a touch hacky as it is, so this isn't such a huge problem.

packages/graphql/tests/integration/enums.int.test.ts
```diff
@@ -87,4 +87,58 @@ describe("enums", () => {
...
+        const session = driver.session();
+
+        const statusResolver = {
+            ACTIVE: "active",
+        };
+
+        const typeDefs = `
+            enum Status {
+                ACTIVE
+            }
```
angrykoala, line 101: Just to make this more realistic, I would add one or two extra options to this status
```suggestion
            enum Status {
                ACTIVE
                INACTIVE
                EATING
            }
```

packages/graphql/tests/integration/enums.int.test.ts
```diff
@@ -87,4 +87,58 @@ describe("enums", () => {
             await session.close();
         }
     });
+
+    test("should create a movie (with a default enum)", async () => {
+        const session = driver.session();
+
+        const statusResolver = {
+            ACTIVE: "active",
+        };
```
angrykoala, line 96: Interesting, I reckon we should add another test case **not** using the statusResolver (and `ACTIVE` being the expected result.

I would also consider adding this usage to the docs, as it is quite handy, and may not be obvious

angrykoala (changes requested):
:+1: I think this is almost ready to go, just a couple of things that would be nice to add before merging related to the integration tests

packages/graphql/tests/integration/enums.int.test.ts
nelsonpecora: That makes sense!

nelsonpecora: Ah, looking closer at these tests, I think they should cover all four cases: with/without default value, and with/without custom resolver. Added!

angrykoala: mmm, I think that was not needed, as those cases were already covered in the tck (and integration covers the actual database translation) but having them won't hurt. Thanks!

angrykoala (approved):

Issue #936, opened 2022-02-06
Support transaction management in @neo4j/graphql-ogm


ashlog:
**Is your feature request related to a problem? Please describe.**
Yes, currently, I am doing quite a bit of transactional mutations and would like to keep things atomic. Unfortunately I cannot use ogm since it doesn't have support for transactions, and therefore I am forced to write queries manually with no type hinting.

**Describe the solution you'd like**
Add support for transaction management in `@neo4j/graphql-ogm`.

Hosch250:
Nice to see a PR already open. Would like this to be resolved ASAP, as I'm hitting the exact same issue :).

dennisjlee:
FYI I think this was released in @neo4j/graphql 3.2.0, but #953 was not mentioned in the release notes.

After upgrading @neo4j/graphql and @neo4j/graphql-ogm to the latest (3.2.3), I was able to pass in an open transaction to an OGM operation like this:

```
model.create({
    /* other args here */
    context: {
      executionContext: tx,
    },
  });
```

darrellwarde:
Apologies @dennisjlee, I've now added this to the release notes.

Indeed, this is now implemented, and as such I will close this.

PR #946, opened 2022-02-08
feat: support existing transactions via context

ashlog:
Resolves https://github.com/neo4j/graphql/issues/936

### DIFF
--- a/packages/graphql/src/types.ts
+++ b/packages/graphql/src/types.ts
@@ -17,12 +17,12 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


marius56782:
I see now that we have implemented the same feature basically: https://github.com/neo4j/graphql/pull/953

ashlog:
> I see now that we have implemented the same feature basically: https://github.com/neo4j/graphql/pull/953

Ahh I didn't see there was already an issue and PR existing for this. Perfect, thanks! Let's go with your MR since you have most of the checklist complete already.

marius56782:
No worries! The important thing is that it gets fixed :)

Issue #464, opened 2021-08-26
Share request Neo4j transaction in context


Kubera2017:
**Is your feature request related to a problem? Please describe.**
I propose you to share active transaction for custom resolvers, it will be very useful.

darrellwarde:
The problem here is that we open and close a database session immediately before and after executing the Cypher, respectively. If we did this, we wouldn't be able to close the session, which would put the onus on the user, which is obviously a non-starter.

A more viable solution is offering a way to pass in a session in via the context which you open and close as you see fit. Precedence would be:

1. Session in context
2. Driver in context
3. Driver passed in via constructor

I believe this would work as an alternative solution for your use case?

Kubera2017:
No, as we cannot rollback using 'session' object, main issue is 'transaction'. Basically my thought - all database queries should be executed in a single transaction (one user's request to API = one transaction). But sources of these queries during the request can be: the library query builder result, OGM usage, and this 'delegateToSchema'. The idea of 'delegateToSchema' explained more, you could check - https://graphaware.com/neo4j/2019/11/18/grandstack-tips-and-tricks.html ('Reusing generated resolvers')

darrellwarde:
Following up here, #953 will allow for the passing in of either driver, session _or_ transaction which should allow for this to be achieved.

Kubera2017:
Yes, exactly that I meant, thanks a lot

PR #953, opened 2022-02-09
Add option to use existing session and transaction

marius56782:
# Description

Proposed fix for #464.

# Issue

Closes:

- #464 
- #936

# Checklist


Hosch250 (approved):

packages/graphql/src/types.ts
```diff
@@ -42,6 +47,7 @@ export interface Context {
     jwt?: JwtPayload;
     auth?: AuthContext;
     queryOptions?: CypherQueryOptions;
+    executionContext?: ExecutionContext;
```
darrellwarde, line 50: The typing decisions which have been made here are a bit awkward - only one of `Driver`, `Session` and `Transaction` will be used, so perhaps this should be a union? `executionContext` also clashes with the `driver` field somewhat, so doing this would allow us to potentially depreciate the `driver` field in the next major release.

```suggestion
    executionContext?: Driver | Session | Transaction;
```

darrellwarde (commented):
Sorry it's taken a while to look at this. I would like to have a conversation about the API here - I think it could be clearer what happens with the `executionContext` field. It would then be great to see integration tests proving that drivers, sessions and transactions all work as intended, and also getting this written up in the documentation. Thanks for the contribution, I'm excited to see what we can do here!

darrellwarde:
Also, I don't seem to be able to find your signed CLA - would you mind doing this please? https://neo4j.com/developer/cla/#sign-cla

packages/graphql/src/types.ts
marius56782: Thanks for your reply! I agree that the typing is a bit awkward, and the API is a bit cryptic. 

The execute function uses more than one of them though (Driver, Session and Transaction). 

The driver is used [here](https://github.com/neo4j/graphql/blob/d4cba9a4e2bec99f4bb7b07e6dfd99acd2cace8e/packages/graphql/src/utils/execute.ts#L82-L88), but I am not sure if that part is actually important, or if it could be skipped in cases where a driver is not defined?

As it stands the Session is used to return a bookmark [here](https://github.com/neo4j/graphql/blob/d4cba9a4e2bec99f4bb7b07e6dfd99acd2cace8e/packages/graphql/src/utils/execute.ts#L124). With the new changes it is possible to specify an executionContext, so that the transaction will not be commited automatically. This means that the returned bookmark will not include the open transaction. I guess that can be a bit counterintuitive.

I can think of a couple of options for solving this:

-  We replace executionContext with something named just `transaction`, and then let the returned bookmark be null in cases where we only have a transaction and no session.
-  Or we can do it like you say. The API only lets you provide one of Driver, Session and Transaction, and then we just skip the parts of the code that requires a Session or Driver in the cases where they are not defined. This solution would also mean that the returned bookmark can be null.

I like the second option better, but it would also require changes to larger parts of the code which I am not that familiar with.

marius56782:
> Also, I don't seem to be able to find your signed CLA - would you mind doing this please? https://neo4j.com/developer/cla/#sign-cla

I have already done that.

packages/graphql/src/types.ts
darrellwarde: > The driver is used [here](https://github.com/neo4j/graphql/blob/d4cba9a4e2bec99f4bb7b07e6dfd99acd2cace8e/packages/graphql/src/utils/execute.ts#L82-L88), but I am not sure if that part is actually important, or if it could be skipped in cases where a driver is not defined?

No need to worry too much about that, the functionality is pretty flaky at the best of times anyway. We actually want to swap this out for putting the same information in the transaction metadata - feel free to add this in if you feel up to it!!

> As it stands the Session is used to return a bookmark [here](https://github.com/neo4j/graphql/blob/d4cba9a4e2bec99f4bb7b07e6dfd99acd2cace8e/packages/graphql/src/utils/execute.ts#L124). With the new changes it is possible to specify an executionContext, so that the transaction will not be commited automatically. This means that the returned bookmark will not include the open transaction. I guess that can be a bit counterintuitive.

We only really need to worry about returning the bookmark if we committed the transaction - otherwise it is the users responsibility to deal with the effects of causal clustering.

> * Or we can do it like you say. The API only lets you provide one of Driver, Session and Transaction, and then we just skip the parts of the code that requires a Session or Driver in the cases where they are not defined. This solution would also mean that the returned bookmark can be null.

I like this option, and I can think of very few downsides!! No issues making a currently non-nullable field nullable.

darrellwarde:
> > Also, I don't seem to be able to find your signed CLA - would you mind doing this please? https://neo4j.com/developer/cla/#sign-cla
> 
> I have already done that.

I see, seems there's a broader issue here with your Git configuration. If you check https://github.com/neo4j/graphql/pull/953/commits - you can see that you can't click your username to get through to your profile. This likely means your local Git configuration has not been set up with the email matching your GitHub profile: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address#setting-your-commit-email-address-in-git

packages/graphql/src/types.ts
marius56782: Ok, I have reimplemented it the way we discussed :+1:

packages/graphql/tests/integration/issues/464.int.test.ts
```diff
@@ -0,0 +1,172 @@
...
+                contextValue: { driver },
+            });
+
+            expect(books.data?.[typeBook.plural]).toEqual([{ id: bookId, name: bookName }]);
+        } finally {
+            await session.close();
+        }
+    });
+
+    // @TODO: more tests
```
darrellwarde, line 171: 👀  Do we need more tests here? I note there's no integration tests for using either a driver or a session in the `executionContext` field.

packages/graphql/src/utils/execute.ts
```diff
@@ -150,7 +183,9 @@ async function execute(input: {
 
         throw error;
     } finally {
-        await session.close();
+        if (executor.driver && executor.session) {
```
darrellwarde, line 186: Why checking for `executor.driver` here if this code is only closing the session?

packages/graphql/src/utils/execute.ts
```diff
@@ -88,40 +136,25 @@ async function execute(input: {
...
-                break;
-            // no default
-        }
+        const result: QueryResult | undefined = await executor.transaction.run(cypher, input.params);
 
         if (!result) {
             throw new Error("Unable to execute query against Neo4j database");
         }
 
+        if (executor.session) {
```
darrellwarde, line 145: Why checking for `executor.session` here if the conditional commits the transaction?

packages/graphql/src/utils/execute.ts
```diff
@@ -67,7 +62,60 @@ async function execute(input: {
...
+    context: Context;
+}): { driver?: Driver; session?: Session; transaction: Transaction } => {
+    if (!input.context.executionContext) {
+        const driver = input.context.driver;
+        const session = driver.session(getSessionParams(input));
+        const transaction = session.beginTransaction(getTransactionConfig());
+        return { driver, session, transaction };
+    }
+
+    const executionContextType = Object.getPrototypeOf(input.context.executionContext).constructor.name;
```
darrellwarde, line 89: See my comments below for using `instanceof`.

```suggestion
```

packages/graphql/src/utils/execute.ts
```diff
@@ -67,7 +62,60 @@ async function execute(input: {
...
+            type: "user-transpiled",
+        },
+    };
+}
+
+const getExecutor = (input: {
+    defaultAccessMode: SessionMode;
+    context: Context;
+}): { driver?: Driver; session?: Session; transaction: Transaction } => {
+    if (!input.context.executionContext) {
```
darrellwarde, line 82: Perhaps assign `executionContext` here to avoid constantly doing `input.context.executionContext`?

```suggestion
    const executionContext = input.context.executionContext;
    
    if (!executionContext) {
```

packages/graphql/src/utils/execute.ts
```diff
@@ -67,7 +62,60 @@ async function execute(input: {
...
+    if (executionContextType === "Driver") {
+        const driver = input.context.executionContext as Driver;
+        const session = driver.session(getSessionParams(input));
+        const transaction = session.beginTransaction(getTransactionConfig());
+        return { driver, session, transaction };
+    }
+
+    if (executionContextType === "Session") {
+        const session = input.context.executionContext as Session;
+        const transaction = session.beginTransaction(getTransactionConfig());
```
darrellwarde, line 100:
```suggestion
    if (executionContext instanceof Session) {
        const transaction = executionContext.beginTransaction(getTransactionConfig());
```

darrellwarde (changes requested):
I think the API is now looking good, just a few questions regarding the internals!

packages/graphql/tests/integration/issues/464.int.test.ts
marius56782: More tests are always good I guess. As you point out there should be tests for using driver and session, but for our team only transaction is necessary. That is also the issue, that this PR is targeted at.

packages/graphql/src/utils/execute.ts
marius56782: I realize this code is not very intuitive. We have two scenarios here:

1. If a session was provided via executionContext, we do not want to automatically close that session.
2. If no session was provided via executionContext, it means that the session was created by the execute function, and thus should be closed automatically.

If a driver was passed either via context.driver, or context.executionContext it means that the session was created by the execute function and should be automatically closed. There are probably better / more intuitive ways of achieving this.

packages/graphql/src/utils/execute.ts
marius56782: Same reason as above basically. If a session exists, it means that the transaction was opened by the execute function, and thus should also be closed by it.

packages/graphql/src/utils/execute.ts
marius56782: Yes, I think all of your suggestions here can be used :)

packages/graphql/src/utils/execute.ts
darrellwarde: I seeeee, may I suggest a boolean variable to describe this behaviour?

```suggestion
        const sessionCreatedByExecute = executor.driver && executor.session;
        if (sessionCreatedByExecute) {
```

packages/graphql/src/utils/execute.ts
darrellwarde: Makes sense! Perhaps a similar variable to what I suggested above?

packages/graphql/tests/integration/issues/464.int.test.ts
darrellwarde: Totally get that,  but if a PR is introducing new functionality then it all needs to be tested. Hopefully you can add them, but do give us a shout if you don't have time and we can get them added in.

packages/graphql/src/utils/execute.ts
marius56782: Sounds good, but we might want to cast it too? 

`Boolean(executor.driver && executor.session)`

packages/graphql/src/utils/execute.ts
marius56782: Yes, that would work :+1:

marius56782:
@darrellwarde, merging your suggestions seems to have caused some issues with the linting, so I am going to revert them.

darrellwarde:
> @darrellwarde, merging your suggestions seems to have caused some issues with the linting, so I am going to revert them.

Yes, the suggestions may not have been enough to get _everything_ working - what were the linting errors?

packages/graphql/src/utils/execute.ts
```diff
@@ -79,31 +79,29 @@ const getExecutor = (input: {
...
         const transaction = session.beginTransaction(getTransactionConfig());
         return { driver, session, transaction };
     }
 
-    const executionContextType = Object.getPrototypeOf(input.context.executionContext).constructor.name;
 
-    if (executionContextType === "Driver") {
-        const driver = input.context.executionContext as Driver;
-        const session = driver.session(getSessionParams(input));
+    if (executionContext instanceof Driver) {
```
marius56782, line 92: TS2693: 'Driver' only refers to a type, but is being used as a value here.

packages/graphql/src/utils/execute.ts
```diff
@@ -79,31 +79,29 @@ const getExecutor = (input: {
...
+        const session = executionContext.session(getSessionParams(input));
```
marius56782, line 93: TS2339: Property 'session' does not exist on type 'Driver | Session | Transaction'.   Property 'session' does not exist on type 'Session'.

marius56782:
> > @darrellwarde, merging your suggestions seems to have caused some issues with the linting, so I am going to revert them.
> 
> Yes, the suggestions may not have been enough to get _everything_ working - what were the linting errors?

Hmm, it seems like the reviewdog lets it pass, but I get errors in my IDE locally. Could be something with my local setup?

packages/graphql/src/utils/execute.ts
```diff
@@ -79,31 +79,29 @@ const getExecutor = (input: {
...
         const transaction = session.beginTransaction(getTransactionConfig());
         return { driver, session, transaction };
     }
 
-    if (executionContextType === "Session") {
-        const session = input.context.executionContext as Session;
-        const transaction = session.beginTransaction(getTransactionConfig());
+    if (executionContext instanceof Session) {
+        const transaction = executionContext.beginTransaction(getTransactionConfig());
         return { session, transaction };
```
marius56782, line 100: TS2552: Cannot find name 'session'. Did you mean 'Session'?

packages/graphql/src/utils/execute.ts
```diff
@@ -183,7 +181,8 @@ async function execute(input: {
 
         throw error;
     } finally {
-        if (executor.driver && executor.session) {
+        const sessionCreatedByExecute = executor.driver && executor.session;
+        if (sessionCreatedByExecute) {
             await executor.session.close();
```
marius56782, line 186: TS2532: Object is possibly 'undefined'.

packages/graphql/src/utils/execute.ts
```diff
@@ -79,31 +79,29 @@ const getExecutor = (input: {
...
-        const transaction = session.beginTransaction(getTransactionConfig());
+    if (executionContext instanceof Session) {
+        const transaction = executionContext.beginTransaction(getTransactionConfig());
         return { session, transaction };
     }
 
-    if (executionContextType === "Transaction") {
-        const transaction = input.context.executionContext as Transaction;
-        return { transaction };
+    if (executionContext instanceof Transaction) {
```
marius56782, line 103: Redundant 'instanceof' check: 'executionContext' has type 'Transaction' or an inheritor of 'Transaction'

darrellwarde:
Looks like something funky is going on in `tests/utils/builders/driver-builder.ts` - a slight type mismatch it seems.

Other than that, just get those remaining two tests added in and this is looking very good to go!

darrellwarde:
> Looks like something funky is going on in `tests/utils/builders/driver-builder.ts` - a slight type mismatch it seems.
> 
> Other than that, just get those remaining two tests added in and this is looking very good to go!

Actually, looks like a small issue with using `instanceof` in the output of the integration tests - this will surely be resolvable, using `instanceof` is definitely the way forward here.

darrellwarde:
> > > Looks like something funky is going on in `tests/utils/builders/driver-builder.ts` - a slight type mismatch it seems.
> > > Other than that, just get those remaining two tests added in and this is looking very good to go!
> > 
> > 
> > Actually, looks like a small issue with using `instanceof` in the output of the integration tests - this will surely be resolvable, using `instanceof` is definitely the way forward here.
> 
> How is it resolvable though?
> 
> Also the error in `tests/utils/builders/driver-builder.ts` is really weird. If you look at the code it is cast `as any`, but in the tests it fails on `as Driver`. Somehow the tests does not use the updated code?

Let me take a look tomorrow, I do want to get this merged this week. 🙂

marius56782:
> > > > Looks like something funky is going on in `tests/utils/builders/driver-builder.ts` - a slight type mismatch it seems.
> > > > Other than that, just get those remaining two tests added in and this is looking very good to go!
> > > 
> > > 
> > > Actually, looks like a small issue with using `instanceof` in the output of the integration tests - this will surely be resolvable, using `instanceof` is definitely the way forward here.
> > 
> > 
> > How is it resolvable though?
> > Also the error in `tests/utils/builders/driver-builder.ts` is really weird. If you look at the code it is cast `as any`, but in the tests it fails on `as Driver`. Somehow the tests does not use the updated code?
> 
> Let me take a look tomorrow, I do want to get this merged this week. slightly_smiling_face

I agree, but I am leaving on holiday though, so unfortunately I won't have a lot of time to look at this for the remainder of the week.

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-953.surge.sh

packages/graphql/src/utils/execute.ts
```diff
@@ -7,7 +7,7 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
+ *ß
```
marius56782, line 10: This looks out of place

marius56782, line 10: This looks out of place

packages/graphql/src/types.ts
```diff
@@ -51,6 +51,7 @@ export interface Context {
     plugins?: Neo4jGraphQLPlugins;
     jwt?: JwtPayload;
     subscriptionsEnabled: boolean;
+    executionContext: Driver | Session | Transaction;
```
marius56782, line 54: executionContext is required now, but driver is not required?

marius56782, line 54: executionContext is required now, but driver is not required?

darrellwarde: So `driver` can still be passed into the context, but internally it is all put into `executionContext` in `wrapper.ts` so that we don't have to account for both in the `execute` function. 🙂

marius56782: Ok, thanks for the explanation :)

packages/graphql/tests/utils/builders/driver-builder.ts
```diff
@@ -65,7 +72,7 @@ export class DriverBuilder extends Builder<Driver, Partial<Driver>> {
                     lastBookmark() {},
                 };
             },
-        } as Driver);
+        } as any);
```
angrykoala, line 75: Is there a chance of avoiding this any? Maybe we could find a more accurate type for this object?

marius56782: Does it matter though? It is just mocking used for testing?

angrykoala: It is not extremely important, but it is preferable to keep the types as close to the original as possible, mainly for ease of testing.

PR #956, opened 2022-02-09
Fix/872 Change ConnectOrCreateFieldInputOnCreate types

### LINKED ISSUE #872: *ConnectOrCreateFieldInputOnCreate and  *CreateFieldInput are completely identical
While harmonizing the logic for schema augmentation in java, I realized, that the types for `${prefix}ConnectOrCreateFieldInputOnCreate` and `${prefix}CreateFieldInput` are identical.

For example the following schema:

```graphql
type Movie {
  title: String!
  id: ID! @id
}

type Actor {
  name: String!
}
```

will generate the following types among others:

```graphql
input ActorMoviesConnectOrCreateFieldInputOnCreate {
  node: MovieCreateInput!
}

input ActorMoviesCreateFieldInput {
  node: MovieCreateInput!
}
```

And if there are relation properties like:

```graphql
type Movie {
  title: String!
  isan: String! @unique
}

type Actor {
  name: String!
  movies: [Movie!]! @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
}

interface ActedIn {
  screentime: Int!
  characterName: String
}
```

The following types are generated among others:

```graphql

input ActorMoviesConnectOrCreateFieldInputOnCreate {
  edge: ActedInCreateInput!
  node: MovieCreateInput!
}

input ActorMoviesCreateFieldInput {
  edge: ActedInCreateInput!
  node: MovieCreateInput!
}
```

angrykoala:
# Description

Fixes #872 by creating a new `...Node` field with only primitive fields for `connectOrCreate`

Related typedefs

```gql
input ActorMoviesConnectOrCreateFieldInputOnCreate {
  node: ActorMoviesConnectOrCreateFieldInputOnCreateNode!
  edge: ActedInCreateInput! # Note that this type has not change
}

input ActorMoviesConnectOrCreateFieldInputOnCreateNode {# including only primitive and scalar fields
  title: String 
  id: ID
}
```

Additionally, the functions `objectFieldsToCreateInputFields` and `objectFieldsToUpdateInputFields` added to remove duplication on field generation from nodes


packages/graphql/tests/schema/connect-or-create-id.test.ts
```diff
@@ -108,7 +108,12 @@ describe("connect or create with id", () => {
             }
 
             input ActorMoviesConnectOrCreateFieldInputOnCreate {
-              node: MovieCreateInput!
+              node: ActorMoviesConnectOrCreateFieldInputOnCreateNode!
```
darrellwarde, line 111: Just taking this single example - this feels like we're creating far too many types which are specific to this particular relationship, when this could really just be `MovieOnCreateInput`. I had this roughly working myself when I took a short look at this a while ago, so it should be feasible.

darrellwarde (commented):
Not a full review - but I have a question about one of the implementation details here.

packages/graphql/tests/schema/connect-or-create-id.test.ts
angrykoala: That's a good point, I'll take a look at that

packages/graphql/tests/schema/issues/872.test.ts
```diff
@@ -0,0 +1,611 @@
...
+ * limitations under the License.
+ */
+
+import { printSchemaWithDirectives } from "@graphql-tools/utils";
+import { lexicographicSortSchema } from "graphql/utilities";
+import { gql } from "apollo-server";
+import { Neo4jGraphQL } from "../../../src";
+
+describe("https://github.com/neo4j/graphql/issues/872", () => {
+    test("@deprecated directive should remain in output", () => {
```
darrellwarde, line 26: This description doesn't seem quite right! Also, is this test necessary, or will this case be covered elsewhere?

darrellwarde (approved):
With `872.test.ts` sorted, looks good to me!

packages/graphql/tests/schema/issues/872.test.ts
angrykoala: :facepalm:  good catch

darrellwarde:
> What do you mean `872.test.ts` sorted? Anything left to do?

Nope, just that bad description!! Good to go now. 🙂

darrellwarde (approved):

shordev:
Thanks for working on this!
Ran into this error recently.

danstarns (approved):
Nice refactor here too!

Issue #916, opened 2022-02-03
Extension of the @readonly directive to also prevent creation


AccsoSG:
**Is your feature request related to a problem? Please describe.**
Certain data should not be created via the GraphQL API, but only read out. Because of this, it would be great to extend the existing @readonly directive with the ability to also prevent node creation.

**Describe the solution you'd like**
Something like this: @readonly(enableCreation: false)

**Describe alternatives you've considered**
There is an option to do this using a custom cypher query. But I suppose this is not efficient since "apoc.cypher.runFirstColumn" is used for custom cypher queries.

darrellwarde:
Superseded by #1034.

PR #961, opened 2022-02-09
Extend @readonly to prevent creation

nelsonpecora:
# Description

Note: This PR is a redo of #950 from my personal account, so I can allow edits by maintainers 😅 

This PR addresses the feature request in #916, adding an `enableCreation` argument to the `@readonly` directive. If this argument is not set (or set to `true`), fields are added to the create input type as before:

```graphql
type SomeObj {
  id: ID!
  name: String @readonly
}

input SomeObjCreateInput {
  id: ID!
  name: String
}

input SomeObjUpdateInput {
  id: ID
}
```

If the argument is added and _explicitly set to false_, the field is omitted from both the create and update input types:

```graphql
type SomeObj {
  id: ID!
  name: String @readonly(enableCreation: false)
}

input SomeObjCreateInput {
  id: ID!
}

input SomeObjUpdateInput {
  id: ID
}
```

# Issue

This PR directly addresses #916. It adds schema and unit tests to cover the new use case, but I'm not entirely sure where to put integration or tck tests (or if it needs them). I'd definitely appreciate your inputs on that! ~Also, it seems like adding a description to the directive definition makes it show up in the documentation, is that accurate?~ Nevermind, I figured out the correct way to update the docs!

Closes:

- #916

# Checklist

### DIFF
--- a/docs/modules/ROOT/pages/type-definitions/access-control.adoc
+++ b/docs/modules/ROOT/pages/type-definitions/access-control.adoc
@@ -94,14 +94,16 @@ directive @ignore(
 [[type-definitions-access-control-readonly]]
 == `@readonly`
 
-The field will only feature in mutations for creating, and object types for querying, and will not be mutable after creation.
+The field will only feature in mutations for creating, and object types for querying, and will not be mutable after creation. If `enableCreation: false` is specified, the field will also not be included in mutations for creating.
 
 === Definition
 
 [source, graphql, indent=0]
 ----
-"""Instructs @neo4j/graphql to only include a field in generated input type for creating, and in the object type within which the directive is applied."""
-directive @readonly on FIELD_DEFINITION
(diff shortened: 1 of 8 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-961.surge.sh

darrellwarde:
Hey @nelsonpecora, we've just had a chat about this in the team. We love the concept, and the functionality is important, but we don't feel that adding an argument to `@readonly` is the right solution.

Ideally, we'd like to extend the `@exclude` directive so that it can also be used on fields as well as types, and this would allow for any combination of permissions a user may desire. For example, for this functionality, you could do:

```gql
type SomeObj {
  id: ID!
  name: String @exclude(operations: [CREATE, UPDATE, DELETE])
}

input SomeObjCreateInput {
  id: ID!
}

input SomeObjUpdateInput {
  id: ID
}
```

We're going to close this PR for now as we favour the above.

darrellwarde:
Please see #1034 for the feature request for extending `@exclude`.

PR #977, opened 2022-02-11
Migration guide 3.0.0

angrykoala:
# Description
Docs for migrating to 3.0 from 2.x


docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+----
+
+Now you'll need to do the following:
+
+[source, JavaScript, indent=0]
+----
+const neoSchema = new Neo4jGraphQL({ typeDefs, driver });
+const schema = await neoSchema.getSchema();
+const server = new ApolloServer({
+    schema: schema
```
tbwiss, line 33: nitpicking: couldn't this just be `schema`?

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+[source, graphql, indent=0]
+----
+type Movie {
+    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
+    director: Director @relationship(type: "DIRECTED", direction: IN)
+}
+----
+
+Note that any other notation, such as `[Actor]` or `[Actor!]` will **not** be valid. "One to one" relationships
+such as `director` remain unchanged.
```
tbwiss, line 52: .. such as `Director` remain ..

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+     },
+ });
+----
+
+== Types plurals changes
+To improve consistency, some automatically generated plurals (e.g. `createActors`) have changed. This may cause issues if
+your types use conventions such as `snake_case`.
+
+Because of this, you may find generated queries and mutations may have different names. If you encounter this problem,
+please update your clients to use the new query names or use the `plural` option in the xref::type-definitions/database-mapping.adoc#_plural[`@node directive`]
```
tbwiss, line 146: More of an accessibility note. The link to `@node directive` is very helpful but it wasn't very obvious to me that this is a link.

tbwiss (commented):
Nice one! Commented on some minor things.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: Yes :sweat_smile: , I decided to go with this as some people may not be completely familiar with object destructuring, but happy to change if you think it is cleaner that way

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: What alternative could we use? just "@node directive" without code blocks?

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
tbwiss: Yes I see. Good point! Let's opt for the most familiar version and leave it as is.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
tbwiss: The alternative without code blocks seems good. But on the other hand this may also just be some useless nitpicking from my side 😅 . We can leave it as is.

angrykoala: If you think it helps accessibility I'm happy to change it
```suggestion
please update your clients to use the new query names or use the `plural` option in the xref::type-definitions/database-mapping.adoc#_plural[@node directive]
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+Now you'll need to do the following:
+
+[source, JavaScript, indent=0]
+----
+const neoSchema = new Neo4jGraphQL({ typeDefs, driver });
+const schema = await neoSchema.getSchema();
+const server = new ApolloServer({
+    schema: schema
+});
+----
```
darrellwarde, line 35: Possibly also worth documenting this using `.then`? I imagine this will be appropriate for a lot of users.

```
----
const neoSchema = new Neo4jGraphQL({ typeDefs, driver });
neoSchema.getSchema().then((schema) => {
    schema: schema
  });
});
----
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+    usersAggregate {
+        count
+    }
+}
+----
+
+== Relationship filters
+`where` filters for relationship queries now explicitly state `ALL`, `NONE`, `SINGLE`, and `SOME` as part of filter name.
+
+Queries using old relationship filters, will now need to use `{relationship}_NONE`. For example:
```
darrellwarde, line 83:
```suggestion
Queries using old relationship filters, will now need to use `{relationship}_SOME`. For example:
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+And, instead of `_NOT`, `_NONE` should be used.
+
+NOTE: Old queries will still work in this release, but are marked as `@deprecated` and will not be available in the future.
+
+== `@ignore` directive renamed to `@computed`
+To better reflect its intended usage, the `@ignore` directive is now named `@computed`, behaviour is unchanged, so you just need to
+rename this directive in your schema.
+
+== Auth plugin system
+https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets[JSON Web Key Sets] support through `jwksEndpoint` option is no longer
```
darrellwarde, line 122: This is more than just `jwksEndpoint`? I would suggest this should have two subsections, one for JWT users and one for JWKS users.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+to force a custom plural value.
+
+== Types changes
+Some automatically generated types have changed to improve consistency.
+These should not require any changes from most developers, unless types names are directly used.
+
+If you are using auto-generated type names, please check the full list of xref::guides/v3-migration/generated-types.adoc[generated types changes].
+
+== Neo4j support
+Neo4j 4.1 is no longer supported in 3.0
```
darrellwarde, line 156:
```suggestion
Neo4j 4.1 is no longer supported in 3.0.0, inline with the [supported versions list](https://neo4j.com/developer/kb/neo4j-supported-versions/).
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+
+== GraphQL support
+GraphQL 15 is no longer supported, please migrate to GraphQL 16
```
darrellwarde, line 159: A bit more boilerplate, but I prefer something like this:

```suggestion
`graphql@^15.0.0` is no longer supported, please upgrade to `graphql@^16.0.0` using `npm` or the package manager of your choice.
```

darrellwarde (commented):
Something quite major which is missing is for users who use custom directives. Compare what is about to be published (https://neo4j-graphql-docs-977.surge.sh/graphql-manual/2.0/type-definitions/custom-directives/) to something like https://www.apollographql.com/docs/apollo-server/v2/schema/creating-directives/#implementing

I would say this is one of the biggest changes!

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+type Movie {
+    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
+    director: Director @relationship(type: "DIRECTED", direction: IN)
+}
+----
+
+Note that any other notation, such as `[Actor]` or `[Actor!]` will **not** be valid. "One to one" relationships
+such as `Director` remain unchanged.
+
+=== Relationship cardinality
```
wheresrhys, line 54: Are there any more details available of the effect this will have? e.g. will the server start to error if a rogue relationship is created in the DB? Can this functionality be turned off?

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: mmm, documenting both would be heavily redundant, so I guess it is promises vs async await

Thoughts?

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,159 @@
...
+== Relationship changes
+This release contains an overhaul of our relationship validations, this will require a few changes to the schema.
+
+=== Many to * relationships
+To improve consistency and validation, **all** "many to *" relationships need to be defined as _required_ in the schema:
+
+[source, graphql, indent=0]
+----
+type Movie {
+    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
```
wheresrhys, line 46: What will be the API behaviour if Movie does not have any relationships to Actors? I'm hoping this change doesn't mean there'll be an error.

darrellwarde: Behaviour will be identical - this always returned an empty List, and this new type definition constraint matches the possible return values.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
tbwiss: Does this come down to a matter a of (personal) preference? Then I vote for async await due to its slightly better readability IMO. But if the majority of our users are more familiar to promises then we should opt for that.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: Hi @wheresrhys, we will try to be as comprehensive with potential pitfalls as possible. As you mention, there is a change the server may return an error on **mutations**  if your database has rogue relationships. Please keep in mind that if that is the case, the current behaviour is already undetermined and **will** cause problems, at least for now we do not plan on making this optional, as the current behaviour is actually buggy.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: I agree with you, although I think most of our examples are already riddled with async/await. Thoughts on this @darrellwarde @danstarns ?

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: Agree, I think I'll wait until auth is merged to fix this, as we will probably want to link to the updated docs to avoid duplication

darrellwarde:
The more I read this, the more I think the "Generated Types" page is just overkill and it should just be on a single page.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+== How to upgrade
+Simply update `@neo4j/graphql` using npm or your package manager of choice:
+
+[source, bash, indent=0]
+----
+npm update @neo4j/graphql
+----
+
+== Async schema generation
```
darrellwarde, line 13:
```suggestion
== Asynchronous schema generation
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+Schema generation is now asynchronous. Instead of using the property `schema`, now the method `getSchema` will return the schema
+as a _Promise_. This means that creating a server now requires awaiting for that method:
```
darrellwarde, line 15:
```suggestion
as a `Promise`. This means that creating a server now requires awaiting for that method:
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+const neoSchema = new Neo4jGraphQL({ typeDefs, driver });
+neoSchema.getSchema().then((schema) => {
+    const server = new ApolloServer({
+        schema: schema
+    });
+});
+----
+
+== Relationship changes
+This release contains an overhaul of our relationship validations, this will require a few changes to the schema.
```
darrellwarde, line 39:
```suggestion
This release contains an overhaul of our relationship validations, which will require a few changes to the schema.
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+=== Many to * relationships
```
darrellwarde, line 41:
```suggestion
=== Many-to-* relationships
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+To improve consistency and validation, **all** "many to *" relationships need to be defined as _required_ in the schema:
```
darrellwarde, line 42:
```suggestion
To improve consistency and validation, **all** "many-to-*" relationships need to be defined as _required_ in the schema:
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+[source, graphql, indent=0]
+----
+type Movie {
+    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
+    director: Director @relationship(type: "DIRECTED", direction: IN)
+}
+----
+
+Note that any other notation, such as `[Actor]` or `[Actor!]` will **not** be valid. "One to one" relationships
```
darrellwarde, line 52:
```suggestion
Note that any other notation, such as `[Actor]` or `[Actor!]` will **not** be valid. "One-to-one" relationships
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+such as `Director` remain unchanged.
+
+=== Relationship cardinality
+Runtime checks for relationships number of connections have been rolled in with this release, this means that some
```
darrellwarde, line 56:
```suggestion
Runtime checks for "one-to*" relationships have been added in this release, ensuring that the correct number of relationships exist. This means that some
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+databases with inconsistent relationships between the schema definition and the actual data may now fail in some queries.
```
darrellwarde, line 57: Why would this have happened? It will be worthwhile giving some background as to why some databases might have ended up in this state.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+For those cases, please ensure that the database is following your schema definition or update the schema to reflect the
+actual existing relationships.
+
+== Count query no longer supported
+Queries using `count` at the root level are no longer supported. For example:
+[source, graphql, indent=0]
+----
+Query {
```
darrellwarde, line 65:
```suggestion
query {
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+    usersCount
+}
+----
+
+The same operation, can now be achieved with a xref::queries#_counting_using_aggregation[count aggregation] query:
+
+[source, graphql, indent=0]
+----
+Query {
```
darrellwarde, line 74:
```suggestion
query {
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+----
+
+== Relationship filters
+`where` filters for relationship queries now explicitly state `ALL`, `NONE`, `SINGLE`, and `SOME` as part of filter name.
+
+Queries using old relationship filters, will now need to use `{relationship}_SOME`. For example:
+
+[source, graphql, indent=0]
+----
+Query {
```
darrellwarde, line 88:
```suggestion
query {
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+    title
+  }
+}
+----
+
+Should be:
+
+[source, graphql, indent=0]
+----
+Query {
```
darrellwarde, line 103:
```suggestion
query {
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+== `@ignore` directive renamed to `@computed`
+To better reflect its intended usage, the `@ignore` directive is now named `@computed`, behaviour is unchanged, so you just need to
+rename this directive in your schema.
+
+== Auth plugin system
+Auth setup now relies on _plugins_ to setup the configuration. You'll need to install `@neo4j/graphql-plugin-auth` or a custom plugin
+as shown in the xref::auth/setup.adoc[auth setup page].
+
+=== JWT auth
```
darrellwarde, line 126: Worth showing a before and after?

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+            secret: "super-secret"
+        })
+    }
+});
+----
+
+
+=== JWKS decoding
+
+https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets[JSON Web Key Sets] support through `Neo4jGraphQLAuthJWKSPlugin`.
```
darrellwarde, line 146:
```suggestion
https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets[JSON Web Key Sets] are now supported through `Neo4jGraphQLAuthJWKSPlugin`.
```

darrellwarde (changes requested):

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: I can think of manual changes to the DB, bugs, or changes in typedefs as the main culprits, anything else we should list there?

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }) {
+    title
+  }
+}
+----
+
+And, instead of `_NOT`, `_NONE` should be used.
+
+NOTE: Old queries will still work in this release, but are marked as `@deprecated` and will not be available in the future.
```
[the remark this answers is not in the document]
angrykoala: mmm, I wasn't completely sure, but yeah, makes sense

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,217 @@
...
+sure to check the up-to-date documentation on xref::type-definitions/custom-directives.adoc[custom directives].
+
+== Types changes
+Some automatically generated types have changed to improve consistency.
+These should not require any changes from most developers, unless types names are directly used.
+
+Some automatically generated types have changed to improve consistency.
+These should not require any changes from the developer in most cases, unless in cases where types names are directly used.
+
+=== OnCreate Node Types
```
darrellwarde, line 185:
```suggestion
=== Removal of nested operation fields for `connectOrCreate`
```

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,217 @@
...
+Input types for `onCreate` in `connectOrCreate` mutations no longer accept relation fields.
```
darrellwarde, line 186:
```suggestion
Input types for `onCreate` in `connectOrCreate` operations no longer accept relationship fields. They were originally added in error and did not function as one would expect, so there is no regression in functionality.
```

darrellwarde (changes requested):

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
darrellwarde: More like, due to the fact that we never enforced this before, it's likely that through normal use of our library before this version, your database may have ended up in this state.

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+Note that any other notation, such as `[Actor]` or `[Actor!]` will **not** be valid. "One-to-one" relationships
+such as `Director` remain unchanged.
+
+=== Relationship cardinality
+Runtime checks for "one-to*" relationships have been added in this release, ensuring that the correct number of relationships exist. This means that some
+databases with inconsistent relationships between the schema definition and the actual data may now fail in some queries.
+This may have happened due to different reasons such as:
+* Direct changes in the database.
+* Changes to the Type Definitions, leading to an inconsistent database.
```
tbwiss, line 60: More a question here but do we want this to be a list with each element on a new line? If not, ignore this comment please. If so then it isn't the case just now: https://neo4j-graphql-docs-977.surge.sh/graphql-manual/3.0/guides/v3-migration/#_relationship_cardinality

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
```diff
@@ -0,0 +1,248 @@
...
+== Custom Directives
+Defining and applying custom directives has changed significantly, if you are using or plan to use custom directives, make
+sure to check the up-to-date documentation on xref::type-definitions/custom-directives.adoc[custom directives].
+
+== Types changes
+Some automatically generated types have changed to improve consistency.
+These should not require any changes from most developers, unless types names are directly used.
+
+Some automatically generated types have changed to improve consistency.
+These should not require any changes from the developer in most cases, unless in cases where types names are directly used.
```
tbwiss, line 214: These two blocks seem to have a lot of similar content. Shall we combine them?

danstarns (approved):

docs/modules/ROOT/pages/guides/v3-migration/index.adoc
angrykoala: Good catch. however, due to Darrell's comment this is not going to be a list, so problem solved XD

Issue #937, opened 2022-02-07
Renaming of @ignore directive


angrykoala:
As discussed in #842, the naming (and, potentially behaviour) of the `@ignore` directive is not intuitive enough. This issue exists to discuss the following:

* `@ignore` name (e.g. `@custom`, `@computed` ...)
* Optional argument naming (`dependsOn`) 
* Any behaviour improvements regarding projections

nelsonpecora:
FWIW, when I implemented a similar feature for another library, I ended up going with `@computed(from: ["list", "of", "fields"])`. Here's my full directive definition, if it's helpful when thinking about behavior improvements:

```
# Computed fields tell the query parser to fetch certain fields
# from the database when querying for data.
directive @computed(
  from: [String!] # field(s) to fetch from db instead of the api-layer field name
  terminating: Boolean # don't query the db for any further down in the tree (useful when using dataloaders to fetch external data)
) on FIELD_DEFINITION
```

As an example of the `terminating` arg, we have user data stored in a separate service, but link user ids to other records in our database. For a node like `{ id: "someid", createdById: 2 }` the SDL looks like:

```
createdBy: User! @computed(from: ["createdById"], terminating: true)
```

This tells our query parser to fetch `createdById` from the database. The fetching of actual user data (everything in the `User` type) happens in a dataloader in the `createdBy` field resolver.

Someone mentioned it in the other thread, but we also included (limited) support for deeper querysets by allowing dot notation in the `fields` argument, e.g. `@computed(from: ["label", "tags.title"])`. That functionality only ended up being used in a handful of places, but it was pretty helpful when we needed it for adding human-readable `name` fields to interfaces.

Hope this is helpful!

dmoree:
@oskarhane @darrellwarde @angrykoala 
The original attempt to modify the `@ignore` directive was to add an argument for a selection set. This comment (https://github.com/neo4j/graphql/pull/842#issuecomment-1022669890) outlined some initial hurdles. I believe I may have solved the problems presented there by rebuilding `graphql-parse-resolve-info` to suit the specific needs of the directive.

I'm aware that this is a wholly different direction, but the previous attempt only allowed for querying primitive fields. I believe that with this generalization a custom resolver will now have access to the entire graph using the source node as a root.

From #842 one could define a custom resolver with:
```graphql
type User {
    id: ID!
    firstName: String!
    lastName: String!
    fullName: String @ignore(dependsOn: ["firstName", "lastName"])
}
```
[...]

PR #979, opened 2022-02-11
Rename `@ignore` directive to `@computed`, and `dependsOn` argument to `from`

darrellwarde:
# Description

Renames `@ignore` to `@computed`. As far as we're aware, there are no other use cases of `@ignore` than to resolve them with a custom resolver. As such, this rename makes sense.

# Issue

#937


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-979.surge.sh

packages/graphql/src/types.ts
```diff
@@ -123,7 +123,7 @@ export interface BaseField {
     description?: string;
     readonly?: boolean;
     writeonly?: boolean;
-    ignored?: boolean;
+    // ignored?: boolean;
```
angrykoala, line 126: :eyes: 
```suggestion
```

angrykoala (approved):

packages/graphql/src/types.ts
darrellwarde: 😳

nelsonpecora:
Oh, I think @dmoree mentioned that `@computed` might have unintended connotations. Do you still feel strongly about that?

nelsonpecora:
@dmoree `@custom` makes sense to me! 👍🏼👍🏼

PR #983, opened 2022-02-14
Refactor: Use Subqueries

dmoree:
# Description

Refactor codebase to use subqueries where appropriate. Reduces code complexity and can perhaps more efficiently plan queries. For `query` fields (`read`), subqueries are projected before final return statement. For `mutation` fields, projection will work the same and any creation of nodes or relationships should probably be placed inside a [unit subquery](https://neo4j.com/docs/cypher-manual/current/clauses/call-subquery/#subquery-unit)

### Todo
- [x] Relationship fields
- [x] Connection fields
- [ ] Cypher fields - Top level
- [ ] Cypher fields - Field level (see #984)
- [ ] Unit subqueries in mutations

PR #984, opened 2022-02-14
Split: `@cypher` into `@computed` and `@related`

dmoree:
# Description

Building off of #983, it may be beneficial to split the `@cypher` directive into two distinct directives both of which require a `statement` argument: A `@computed` directive will return only scalar types and a `@related` directive will returned named types.

Subqueries can be deeply nested so it is important to identify the root of the query. Currently `this` can be used to refer to the root and is used inside of an `apoc.runFirstColumn` call that is placed inside of the projection (which takes care of the field name). Moving away from `apoc.runFirstColumn` and towards subqueries would require conditions on the statement passed to it.

For `@computed` the statement can reference the root of the query as `$$source` and return a value `AS $$field`. For `@related`, `$$source` likewise to reference the root and a `$$target` node to return.

Example:
##### Schema
```graphql
type Movie {
    id: ID!
    title: String!
    runtime: Int!
    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
    numberOfActors: Int!
        @computed(
            statement: """
            MATCH (actor:Actor)-[:ACTED_IN]->($$source)
            RETURN count(actor) AS $$field
            """
        )
    relatedActors: [Actor!]!
        @related(
            statement: """
            MATCH ($$target)-[:ACTED_IN]->($$source)
            RETURN $$target
            """
        )
}

type Actor {
    id: ID!
    name: String!
    movies: [Movie!]! @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
    totalScreenTime: Int!
        @computed(
            statement: """
            MATCH ($$source)-[r:ACTED_IN]->(:Movie)
            RETURN sum(r.screenTime) AS $$field
            """
        )
}

interface ActedIn @relationshipProperties {
    screenTime: Int!
}
```
##### Query
```graphql
query ($movieId: ID!) {
    movies(where: { id: $movieId }) {
        id
        numberOfActors
    }
}
```
##### Generated Output (edited for readability)
```
MATCH (this:Movie)
WHERE this.id = $this_id
CALL {
    WITH this
    MATCH (actor:Actor)-[:ACTED_IN]->(this)
    RETURN count(actor) AS this_numberOfActors
}
RETURN this { .id, numberOfActors: this_numberOfActors } as this
```

The similarly defined `@cypher` field will generate
```
MATCH (this:Movie)
WHERE this.id = $this_id
RETURN this { .id, numberOfActors:  apoc.cypher.runFirstColumn("MATCH (actor:Actor)-[:ACTED_IN]->(this) RETURN count(actor)", {this: this, auth: $auth}, false) } as this
```
##### Query
```graphql
query ($movieId: ID!) {
    movies(where: { id: $movieId }) {
        id
        relatedActors {
          id
          movies {
              id
              numberOfActors
          }
        }
    }
}
```
##### Generated Output (edited for readability)
```
MATCH (this:Movie)
WHERE this.id = $this_id
CALL {
    WITH this
    OPTIONAL MATCH (this)<-[:ACTED_IN]-(this_relatedActors:Actor)
    CALL {
        WITH this_relatedActors
        OPTIONAL MATCH (this_relatedActors)-[:ACTED_IN]->(this_relatedActors_movies:Movie)
        CALL {
            WITH this_relatedActors_movies
            MATCH (actor:Actor)-[:ACTED_IN]->(this_relatedActors_movies)
            RETURN count(actor) AS this_relatedActors_movies_numberOfActors
        }
        RETURN this_relatedActors_movies { .id, numberOfActors: this_relatedActors_movies_numberOfActors } AS this_relatedActors_movies
    }
    RETURN this_relatedActors { .id, movies: collect(this_relatedActors_movies) } AS this_relatedActors
}
RETURN this { .id, relatedActors: collect(this_relatedActors) } as this
```

The similarly defined `@cypher` field will generate
```
MATCH (this:Movie)
WHERE this.id = $this_id
RETURN this { .id, relatedActors: [this_relatedActors IN apoc.cypher.runFirstColumn("MATCH (this)<-[:ACTED_IN]-(actor:Actor)
RETURN actor", {this: this, auth: $auth}, true) | this_relatedActors { .id, movies: [ (this_relatedActors)-[:ACTED_IN]->(this_relatedActors_movies:Movie)   | this_relatedActors_movies { .id, numberOfActors:  apoc.cypher.runFirstColumn("MATCH (actor:Actor)-[:ACTED_IN]->(this)
RETURN count(actor)", {this: this_relatedActors_movies, auth: $auth}, false) } ] }] } as this
```


darrellwarde:
Hey @dmoree, we discussed this in the team this morning and have decided that we want to keep the `@cypher` directive intact as it is now, and infer any specific behaviour from the type of the field it is applied to. Additionally, we're happy with the new naming of the `@ignore` directive as `@computed`.

We realise you've implemented some of the changes for this here and we're sorry to close this PR with that being in the case. For future cases, we would really appreciate having a conversation about changes like these up front in an issue before any code changes are made. That way we can make an objective decision about what is being proposed.

Some of the changes here look very promising, especially the new `$$source`, `$$target` and `$$field` variables, which I would definitely like to see followed up in an issue for discussion.

As a heads up, we're now no longer accepting any breaking changes for the 3.0.0 release coming up, as we are now just around the corner from the release.

dmoree:
@darrellwarde 
No worries. I apologize for not getting back to you yesterday, but the discussion does seem better suited for an issue. The changes were minimal as the `@computed` field is a simple find and replace and the `@related` field just used the already existing relationship subquery from #983.

I can try to make the case stronger in a forthcoming issue, but I think at a fundamental level different return values should be treated differently. They will diverge at some point. Just quickly a `@computed` field should be available as a `sort` field and possibly as a `where` field both of which will be unavailable on a `@related` field. Conversely, a `@related` field should have the option to make it a connection field and as a top-level `query`neither make sense for `@computed`. Different requirements down the line will make it difficult to separate simply on returned value, but of course it could always be done.

PR #989, opened 2022-02-14
remove pending() from jest tests

angrykoala:
# Description
Latest versions of jest do not support the usage of `pending()`, currently used for dynamically skipping MULTIDB tests.

This PR removes its usage by simply returning the test (and thus, counting it as **passed**) along with a log noting the test is skipped. This behaviour is not ideal, but keeping jest up to date is worth this minor inconvenience.

In the future, we could improve this by setting an environment variable to skip tests without needing to reach out to the database in the test itself

Additionally, `delay` utils were added to simplify and solve an EsLint error in tests

### DIFF
--- a/examples/neo-push/server/package.json
+++ b/examples/neo-push/server/package.json
@@ -34,7 +34,7 @@
         "@types/node": "14.14.14",
         "@types/uuid": "8.3.0",
         "cross-env": "7.0.3",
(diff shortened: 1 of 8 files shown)


darrellwarde (approved):
This all looks like a broadly sensible workaround to me!

PR #990, opened 2022-02-14
Update dangle_underscore eslint rule

angrykoala:
# Description
This rule has been caused false negatives in eslint, as we **do** want some underscore variables (for now at least)

Additionally, a couple of low-hanging-fruit eslint errors fixed


darrellwarde (approved):
Looks good! Also see #991 for another rule change here.

PR #1024, opened 2022-02-21
Refactor: Centralise static GraphQL types outside of makeAugmentedSchema

darrellwarde:
# Description

Moves static GraphQL type definitions outside of the `makeAugmentedSchema` function, and centralises them in one directory.


packages/graphql/src/schema/types/scalars/Date.ts
```diff
@@ -49,3 +49,5 @@ export default new GraphQLScalarType({
         return neo4j.types.Date.fromStandardDate(new Date(ast.value));
     },
 });
+
+export { DateScalar as Date };
```
angrykoala, line 53: **question**: Maybe we want to use something different than `date` here? I know it is not as nice, but we should avoid name collisions with native types

packages/graphql/src/schema/types/scalars/DateTime.ts
```diff
@@ -29,7 +29,7 @@ export default new GraphQLScalarType({
         }
 
         if (isDateTime(outputValue as object)) {
-            return new Date((outputValue as typeof DateTime).toString()).toISOString();
+            return new Date((outputValue as typeof neo4j.types.DateTime).toString()).toISOString();
```
angrykoala, line 32: :clap:

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -708,17 +663,17 @@ function makeAugmentedSchema(
     if (pointInTypeDefs) {
         // Every field (apart from CRS) in Point needs a custom resolver
         // to deconstruct the point objects we fetch from the database
-        composer.createObjectTC(point.point);
-        composer.createInputTC(point.pointInput);
-        composer.createInputTC(point.pointDistance);
+        composer.createObjectTC(Point);
+        composer.createInputTC(PointInput);
+        composer.createInputTC(PointDistance);
```
angrykoala, line 668: :clap: :clap:

packages/graphql/src/schema/validation/validate-document.ts
```diff
@@ -80,9 +85,10 @@ function filterDocument(document: DocumentNode): DocumentNode {
         return fields?.filter((f) => {
             const type = getArgumentType(f.type);
```
angrykoala, line 86: **nitpick (blocking)**: Should `type` be `any` here? It feels like `getArgumentType` should return `string` (Noting this due to reviewdog)

packages/graphql/src/schema/types/objects/CartesianPoint.ts
```diff
@@ -0,0 +1,53 @@
...
+import { GraphQLFloat, GraphQLInt, GraphQLNonNull, GraphQLObjectType, GraphQLString } from "graphql";
+
+export const CartesianPoint = new GraphQLObjectType({
+    name: "CartesianPoint",
+    fields: {
+        x: {
+            type: new GraphQLNonNull(GraphQLFloat),
+            resolve: (source) => {
+                return source.point.x;
+            },
```
angrykoala, line 29: **question (non-blocking)**: I'm honestly not sure what the reasoning/advantage of using arrow functions here is :confused:. It feels like the function syntax is equivalent, slightly shorter and easier to read?

Happy to keep it this way if you disagree
```suggestion
            resolve(source) {
                return source.point.x;
            },
```

angrykoala (approved):
A couple of suggestions/questions, other than that looks really good

angrykoala (approved):

PR #1025, opened 2022-02-21
Refactor: Extract parsing of document and definition nodes into separate functions

darrellwarde:
# Description

Slowly moving functionality out of `makeAugmentedSchema`, this PR moves the early parsing of the type definitions into a document, and the subsequent parsing of that document into GraphQL definition nodes into separate functions.

Additionally, likely improve performance by moving several `filter` calls into one `reduce`.


packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -78,14 +67,21 @@ import getUniqueFields from "./get-unique-fields";
...
+import parseFulltextDirective from "./parse/parse-fulltext-directive";
 import { QueryOptionsDirective } from "../classes/QueryOptionsDirective";
+import { parseQueryOptionsDirective } from "./parse/parse-query-options-directive";
 
 function makeAugmentedSchema(
     typeDefs: TypeSource,
     { enableRegex, skipValidateTypeDefs }: { enableRegex?: boolean; skipValidateTypeDefs?: boolean } = {}
 ): { nodes: Node[]; relationships: Relationship[]; typeDefs: DocumentNode; resolvers: IResolvers } {
-    const document = mergeTypeDefs(Array.isArray(typeDefs) ? (typeDefs as string[]) : [typeDefs as string]);
+    const document = getDocument(typeDefs);
```
angrykoala, line 84: :clap:

packages/graphql/src/schema/get-definition-nodes.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export function getDefinitionNodes(document: DocumentNode): DefinitionNodes {
+    return document.definitions.reduce<DefinitionNodes>(
+        (definitionNodes, definition) => {
+            switch (definition.kind) {
+                case Kind.SCALAR_TYPE_DEFINITION:
+                    definitionNodes.scalarTypes.push(definition);
+                    break;
+                case Kind.OBJECT_TYPE_DEFINITION:
+                    if (!["Query", "Mutation", "Subscription"].includes(definition.name.value)) {
```
angrykoala, line 54: **nitpick**: Maybe `["Query", "Mutation", "Subscription"]` could be a constant, as it is used in some other places, but mainly to give a context on why these are not pushed to objectTypes, something like:
```suggestion
                    if (!nativeTopLevelTypes.includes(definition.name.value)) {
```
-----

**suggestion(non-blocking)**: Furthermore, a small function could make this really legible (you probably can think of better naming):

```typescript
function isNativeRootType(type: DefinitionNode): boolean
```

```suggestion
                    if (!isNativeRootType(definition)) {
```

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -168,37 +164,21 @@ function makeAugmentedSchema(
 
     const customResolvers = getCustomResolvers(document);
 
-    const scalars = document.definitions.filter((x) => x.kind === "ScalarTypeDefinition") as ScalarTypeDefinitionNode[];
-
-    const objectNodes = document.definitions.filter(
-        (x) => x.kind === "ObjectTypeDefinition" && !["Query", "Mutation", "Subscription"].includes(x.name.value)
-    ) as ObjectTypeDefinitionNode[];
+    const definitionNodes = getDefinitionNodes(document);
```
angrykoala, line 167: Another :clap:  for you

angrykoala (approved):
Approved, just a nitpick that I would like to get solved so it is :100:

packages/graphql/src/schema/get-definition-nodes.ts
darrellwarde: This is now done! For now, I've left it as a util function within the file as this is the only place where it is used. We can extract it out to a utils directory if and when it gets used elsewhere.

PR #1035, opened 2022-02-22
Extract Node generation into separate function

darrellwarde:
# Description

This is where the refactor of `makeAugmentedSchema` falls down a little - notice the more complex return value from this function. No optimisations done here, just a direct copy and paste out.


packages/graphql/src/schema/get-nodes.ts
```diff
@@ -0,0 +1,198 @@
+import { DirectiveNode, NamedTypeNode } from "graphql";
```
tbwiss, line 1: Just as a note, notice the failing ` license-header-check`

tbwiss (approved):
LGTM! just added a note.

PR #1042, opened 2022-02-23
RFC/performance testing

angrykoala:
# Description

RFC for performance testing

### DIFF
--- a/docs/rfcs/rfc-011-performance-testing.md
+++ b/docs/rfcs/rfc-011-performance-testing.md
@@ -1,0 +1,67 @@
+# Performance Testing
+
+## Problem
+To start work on performance improvements, we need a way of measuring these improvements, preferably automatically.
+
+## Proposed Solution
+The library has 3 possible areas where overall performance can be tested and improved. Each of these areas test a different part of the library, and will require a different strategy.
+
+* **Cypher Query**: Improvements on the generated Cypher queries.
+* **Server Execution**:  Improvements on resolvers and library code performance.
+* **Augmented Schema**: The generation of the augmented schema is a one-off execution at server startup, but due to the long execution of this process for large schemas, it may be worth optimising this process to speed up server startup.
+	* **Schema Size**: Another measurable metric is the total size of the augmented schema.
+
+## Cypher query performance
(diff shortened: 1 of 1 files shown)


docs/rfcs/rfc-011-performance-testing.md
```diff
@@ -0,0 +1,67 @@
...
+```typescript
+    const session = driver.session();
+    const result = await session.run("PROFILE MATCH(N) RETURN N");
+    const { dbHits, rows, pageCacheHitRatio } = result.summary.profile || {};
+    console.log("dbHits", dbHits);
+    console.log("rows", rows);
+    console.log("pageCacheHitRatio", pageCacheHitRatio); // Requires enterprise
+```
+
+`dbHits` and `rows` are only dependant on Neo4j version and the database. Both can easily be controlled in our tests, leading to consistent results in which we only test changes to our queries. Testing these is also faster and easier than time. The main pitfall is that in some cases may not be an accurate representation of real execution time (particularly when using [Apoc](https://neo4j.com/labs/apoc/)).
```
tbwiss, line 34: Nice! This really sounds like a good place to start IMO.

docs/rfcs/rfc-011-performance-testing.md
```diff
@@ -0,0 +1,67 @@
...
+
+> Note: Because the schema lives in memory under the GraphQL library internal representation, the metric of the raw schema may not be an accurate measure of the in-memory gains.
+## Challenges
+In addition to the challenges present in every testing strategy, we may find the following challenges when building any performance testing suite:
+* **Continuous Integration**: We should be able to measure and validate performance improvements, as well as detect noticeable drops in performance when developing new features.
+* **Benchmarks**: We need to find reasonable benchmarks and goals to hit in terms of performance (i.e. What is the performance we want to achieve for x feature?).
+* **Standard measures**: Any time-based measure is bound to hardware dependencies, for CI and long-term purposes, these measures should happen in a strictly controlled machine and standard benchmarking practices (e.g. average multiple runs) should be taken into consideration.
+
+## Risks and unknowns
+* Currently, `@neo4j/graphql` works on the principle of a single Cypher query per request, we need to validate if this is still the best way to achieve good performance.
```
tbwiss, line 67: **question**: Are there any security considerations we have to discuss or keep an eye out in the future?

angrykoala: Good point, I don't think there are any for the scope of this PR, as we only intend to benchmark, not change the queries.

Security should be discussed on each performance improvement as needed IMO

tbwiss (approved):

PR #1045, opened 2022-02-23
Create rfc-010-top-level-interfaces.md

darrellwarde:
# Description

RFC for top-level interface Query fields.

### DIFF
--- a/docs/rfcs/rfc-010-top-level-interfaces.md
+++ b/docs/rfcs/rfc-010-top-level-interfaces.md
@@ -1,0 +1,138 @@
+# Top-level Interfaces
+
+## Problem
+
+Interface types are currently only available to Query as part of a relationship field. 
+Our users would like to be able to query for Interface types directly from the root Query type.
+
+## Proposed Solution
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/rfc-010-top-level-interfaces.md
```diff
@@ -0,0 +1,138 @@
...
+  RETURN this { __resolveType: "Series", .title }
+}
+RETURN this
+```
+
+We can do some filtering for both types:
+
+```gql
+query {
+  productions(where: { _on: { Movie: { title: "Forrest Gump" }, Series: { title: "Star Wars: The Clone Wars" } } }) {
```
angrykoala, line 63: Would it be possible to filter directly on `Production`? , if so, an example would be useful

something a bit like:

```graphql
production(where: {title: "Start Wars" })
```

angrykoala (changes requested):
There is a case that I would like to either be documented or defined as "out of scope"

angrykoala (approved):

Issue #1049, opened 2022-02-24
Filtering by connections in where clauses cause server error when using interfaces


jnterry:
**Describe the bug**
The top level where filters with a relationship connection causes a server error if the relation is to an interface.

**Type definitions**

```graphql
interface Media {
	id      : ID! @id
	title   : String!
	likedBy : [Person!] @relationship(type: "LIKES", direction: IN)
	similar : [Media!] @cypher(statement: """
		MATCH (this)<-[:LIKES]-(:Person)-[:LIKES]->(other:Media)
		RETURN COLLECT(other { .*, __resolveType: apoc.coll.subtract(labels(other), ['Meda'])[0] })
	""")
}

type Book implements Media @node(additionalLabels: ["Media"]) {
	id      : ID!
	title   : String!
	likedBy : [Person!]
	similar : [Media!]

	pageCount: Int!
}

type Film implements Media @node(additionalLabels: ["Media"]) {
	id      : ID!
	title   : String!
	likedBy : [Person!]
	similar : [Media!]

	runTime: Int!
}

type Person {
	name   : String
	likes  : [Media!] @relationship(type: "LIKES", direction: OUT)
}
```

(schema also used in bug report #887)

**To Reproduce**

Seed the database with the following data:

```graphql
createPeople(input: [{
    name: "Bob",
    likes: { create: [
        { node: {Book: { title: "Harry Potter",      pageCount: 300, }}},
        { node: {Book: { title: "Lord of the Rings", pageCount: 400, }}}
    ]}
}]) {
    info { nodesCreated }
}
```

Run the following query:
```graphql
query {
  people (where: {
    likesConnection_SOME: {
      node: { title: "Harry Potter" }
    }
  }) { name }
}
```

The following server error is returned: 
 > Cannot read properties of undefined (reading 'name')

Looking at the code, I think this is because there does not exist a 'Node' instance for the interface, and hence this line evaluates to undefined:

https://github.com/neo4j/graphql/blob/ee0df92352e5fb8f0727608260c4bb5db30d7145/packages/graphql/src/translate/where/create-where-and-params.ts#L184

Reading the `refNode.name` therefore crashes:
https://github.com/neo4j/graphql/blob/ee0df92352e5fb8f0727608260c4bb5db30d7145/packages/graphql/src/translate/where/create-where-and-params.ts#L189

**Expected behavior**
The query should run successfully.

**System (please complete the following information):**

* OS: Debian 11
[...]

Issue #146, opened 2021-01-15
Unions / Interfaces at top level


smkhalsa:
As [mentioned](https://github.com/neo4j/graphql-tracker-temp/blob/master/introduction.md#top-level-unions), unions no longer work at the top level, and interfaces no longer work at the top level or on relationships.

This is a major issue for me, and I hope that we can find a way to maintain this functionality in the new library. I'd like to highlight several use cases that are relevant to me:

## 1. Unions for Search

Search is the canonical example for unions, and it's not clear how we can replicate this behavior without them.

```graphql
union SearchResult =
    Album
  | Song
  | Artist
  | UserPlaylist
  | FeaturedPlaylist
  | Podcast

type Query {
  Search(string: String!): [SearchResult!]!
    @cypher(
      statement: """
      CALL db.index.fulltext.queryNodes('names', $string+'~') YIELD node, score
      RETURN node
      """
    )
}
```

## 2. Interfaces on Relationships for Activity Feed

Assuming the below schema excerpt, we may want to query for all of a user's activity, ordered by `createdAt` time

```graphql
interface Activity {
  id: ID!
  createdAt: _Neo4jDateTime!
  user: User! @relation(name: "BY_USER", direction: "OUT")
}

type AudioListen implements Activity {
  id: ID!
  createdAt: _Neo4jDateTime!
  user: User! @relation(name: "BY_USER", direction: "OUT")
  audio: Audio! @relation(name: "AUDIO_LISTENED", direction: "OUT")
}

type VideoView implements Activity {
  id: ID!
  createdAt: _Neo4jDateTime!
  user: User! @relation(name: "BY_USER", direction: "OUT")
  video: Video! @relation(name: "VIDEO_VIEWED", direction: "OUT")
}

type User {
  id: ID!
  name: String
  createdAt: _Neo4jDateTime
  activity: [Activity!]! 
    @cypher(
      statement: """
        // get all activity, sorted by createdAt field
      """
    )
}
```

## 3. Top-Level Interface to Query Across Related Types

Assuming the below schema excerpt, we may wish to query across all `Audio` types.

```graphql
interface Audio {
  id: ID!
  name: String!
  url: String!
  listens: [AudioListen!]! @relation(name: "AUDIO_LISTENED", direction: "IN")
}

type Song implements Audio {
  id: ID!
  name: String!
  url: String!
  artists: [Artist!]! @relation(name: "BY_ARTIST", direction: "OUT")
[...]

PR #1056, opened 2022-02-25
WIP: Interface relation query fixes

jnterry:
# Description

Fixes a server error when filtering by connection fields in a where clause when the target of the relation is an interface.

# Issue

This is a proof of concept to resolve https://github.com/neo4j/graphql/issues/1049 and may well not be an acceptable final solution.

This PR makes interfaces more of a first-class citizen by passing information about them down through the `Context` object which may be a useful starting point to solve various other bugs querying interfaces (of which there are many).

Note that the generated cypher query tends to be slow if an interface has many implementors. I would be happy to attempt to optimize the query, add unit tests, docs etc if this approach of making interface information available in the `Context` seems sensible. I expect it would also be helpful in supporting feature requests such as https://github.com/neo4j/graphql/issues/146

# Checklist

### DIFF
--- a/packages/graphql/src/classes/Interface.ts
+++ b/packages/graphql/src/classes/Interface.ts
@@ -17,10 +17,14 @@
  * limitations under the License.
  */
 
-export { default as Node, NodeConstructor } from "./Node";
-export { default as Interface, InterfaceContructor } from "./Interface";
-export { default as Relationship } from "./Relationship";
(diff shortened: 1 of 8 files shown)


darrellwarde:
This seems to have now been inadvertently fixed elsewhere, so closing this and the related issue! 🙂

PR #1062, opened 2022-02-28
RFC/autogenerated-properties

angrykoala:


docs/rfcs/rfc-012-autogenerated-properties.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Proposed Solution
+The proposed solutions would add a new directive `@callback` an a new config option to `Neo4jGraphQL` (`callbacks`). This directive will mark some properties
+to be autogenerated by a callback function, passed to `Neo4jGraphQL`. The directive will have the following arguments:
+* `operations`: An array containing the oprations in which this callback will be executed and the result returned, possible values are `CREATE` and `UPDATE`.
+* `name`: Name of the callback to be executed, note that the name will have to match the named passed to `Neo4jGrapQL`.
+
+The new config option `callbacks` will accept an object with the callbacks to be used:
+```typescript
+type Neo4jGraphQLCallback<T extends string | number | Date | boolean> = (args: Record<string, any>) => T | undefined | null | Promise<T | undefined | null>;
```
darrellwarde, line 17: Shall we put the arguments in an object or not?

```suggestion
type Neo4jGraphQLCallback<T extends string | number | Date | boolean> = ({ args }: { args: Record<string, any>} ) => T | undefined | null | Promise<T | undefined | null>;
```

docs/rfcs/rfc-012-autogenerated-properties.md
```diff
@@ -0,0 +1,107 @@
...
+    },
+});
+```
+
+> Note that unlike [`@computed`](https://neo4j.com/docs/graphql-manual/current/custom-resolvers/#custom-resolvers-computed). 
+These properties will be stored in the database.
+
+## Risks
+
+* Using `@default` with `@callback`.
```
darrellwarde, line 97: We have a step of schema validation which allows us to check for "invalid directive combinations" - this is one of the combinations that should be added, also `@id` and `@timestamp` at the very least.

darrellwarde (commented):
Looks good! Just a couple of questions.

docs/rfcs/rfc-012-autogenerated-properties.md
angrykoala: mmm, I don't think so :thinking: , there will only be one argument for now (the type parameters) using an object is a bit unnecessarily complicated.

Any extra argument we may want to pass in the future could be passed as a separate argument (loosely reflecting resolvers). Am I missing something?

docs/rfcs/rfc-012-autogenerated-properties.md
```diff
@@ -0,0 +1,107 @@
+# Autogenerated Properties
+
+## Problem
+We want to allow for node properties to be generated automatically, via code, when creating (and possibly updating) nodes. 
+This would solve common problems such as custom ids or creation of slugs.
+
+Related Issue: [627](https://github.com/neo4j/graphql/issues/627)
+
+## Proposed Solution
+The proposed solutions would add a new directive `@callback` an a new config option to `Neo4jGraphQL` (`callbacks`). This directive will mark some properties
```
angrykoala, line 10: `@callback` directive name still feels a bit incomplete as it doesn't implies the data will be stored.
We may want to think a bit more in naming :confused:

PR #1099, opened 2022-03-08
Fix parse value not accepting Neo4J types which causes problems in OGM

mhlz:
# Description

In our project we have some custom mutations that accept some input types with Date and DateTime scalars. Eventually these will be passed in to the OGM to make changes to the database. This used to work just fine before 3.0.

3.0 introduced parseLiteral functions into the custom scalars, which means that strings the client gives will now be parsed to Date/DateTime objects. When passing these into the OGM however, parseValue gets called which currently only accepts strings and will throw an error. We could call toString on all Date/DateTime objects but this would make typing things in our custom resolvers very difficult because the Date scalar for example is now sometimes a string (when interacting with the OGM) and sometimes a Neo4JDate (when reading input types).

I changed parseValue to also accept Date/DateTime/Time/etc which fixes this issue.

### DIFF
--- a/packages/graphql/src/schema/types/scalars/Date.ts
+++ b/packages/graphql/src/schema/types/scalars/Date.ts
@@ -18,9 +18,9 @@
  */
 
 import { GraphQLError, GraphQLScalarType, Kind, ValueNode } from "graphql";
-import neo4j, { isDate } from "neo4j-driver";
+import neo4j, { isDate, Date as Neo4jDate } from "neo4j-driver";
 
-export const GraphQLDate = new GraphQLScalarType({
+export const GraphQLDate = new GraphQLScalarType<Neo4jDate<number>, string>({
     name: "Date",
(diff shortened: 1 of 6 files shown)


darrellwarde:
Hey @mhlz! Thanks for the contribution. Can you please ensure that we have edit access to your branch? I don't seem to be able to update it with changes from the base branch.

Secondly, could you please add integration tests for the cases which you have described above? 

As it stands, this breaks compatibility with `graphql@^15.0.0`, which whilst we don't officially support, we're trying to keep working for as long as possible. This is because of the generics added to `GraphQLScalarType`, and I'd like to see if we could find an alternative solution. Some tests will help us easily find what works and what doesn't.

darrellwarde:
This has been supersedes by #2503 which hopefully implements this in a non-breaking manner, and adds an integration test. Thanks!

PR #1108, opened 2022-03-09
turn unique types into classes instances

angrykoala:
# Description
This PR changes the unique types used for testing so these are now a class. All works exactly the same, however, the following are also valid:

```typescript
const typeMovie = new UniqueType("Movie"); // Same as generateUniqueType("Movie")

const typedefs = `type ${typeMovie}`; // Same as `type ${typeMovie.name}`

```


tbwiss (approved):
LGTM!

PR #1110, opened 2022-03-10
Fix 1073 connectOrCreate schema

### LINKED ISSUE #1073: connectOrCreate does not consider required! Fields
**Describe the bug**
It is possible to write a mutation that create nodes, that don´t follow the schema.

**Type definitions**
```
type foo {
  Name: String!
  bars: [bar!]! @relationship(type: "HAS", direction: OUT)
}

type bar {
  id: ID! @id
  Name: String!
  foo: foo @relationship(type: "HAS", direction: IN)
}

```

**To Reproduce**

1. run this mutation:
```
mutation {
  createFoos(
    input: {
      Name: "Hello World"
      bars: {
        connectOrCreate: {
          where: { node: { id: "bedd43c7-d3c4-456a-8361-3bd1bbc56afa" } }
          onCreate: { node: { id: "bedd43c7-d3c4-456a-8361-3bd1bbc56afa" } }
        }
      }
    }
  ) {
    info {
      nodesCreated
      relationshipsCreated
    }
    foos {
      Name
      bars {
        Name
      }
    }
  }
}
```
3. See error

```
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field bar.Name.",
....
```

**Expected behavior**
onCreate: {node.... should *require* a Name: String! input, as defined in typeDefs.

* OS: Windows]
* Version: @neo4j/graphql@3.0.1
* Node.js version: 16.6.1

angrykoala:
# Description
This PR fixes #1073 by properly marking the required fields in the connectOrCreate -> onCreateInput


packages/graphql/tests/integration/connect-or-create/connect-or-create-id.int.test.ts
```diff
@@ -154,16 +154,48 @@ describe("connectorcreate with @id", () => {
...
+                        onCreate: { node: { id: "myid", title: "The Terminal" } }
+                      }
+                    }
+                  }
+                ]
+              ) {
+                ${typeActor.plural} {
+                  name,
+                  movies {
+                      id,
```
darrellwarde, line 182:
```suggestion
                      id
```

darrellwarde (approved):
We should also check that if an autogenerated ID field is in the `where`, does the ID get overridden in `ON CREATE` with an autogenerated value.

angrykoala:
Tests updated, this pr does **not** cover newly found bug #1124 
A new skipped test has been added for this bug

darrellwarde (approved):
Broadly speaking, this looks good! I have a suggestion - should we make the `onCreate` input field non-nullable? That way, despite the `where`, we will always force our users to specify what the fields should be on creation, which will make it easier to change the implementation under the hood.

angrykoala:
:thinking: @darrellwarde onCreate is already required, or am I missing something?

darrellwarde:
> 🤔 @darrellwarde onCreate is already required, or am I missing something?

I might be losing my mind. 🙃

PR #1111, opened 2022-03-10
Subscriptions meta for update operations - nested update only

darrellwarde:


packages/graphql/src/translate/create-update-and-params.ts
```diff
@@ -428,6 +426,11 @@ function createUpdateAndParams({
             return res;
         }
 
+        // OLD PROPS
+        if (context.subscriptionsEnabled) {
+            res.strs.push(`WITH ${varName} { .* } AS oldProps, ${withVars.join(", ")}`);
+        }
```
darrellwarde, line 432:
```suggestion
        if (context.subscriptionsEnabled) {
            const oldProps = `WITH ${varName} { .* } AS oldProps, ${withVars.join(", ")}`;
            res.strs.push(oldProps);
        }
```

packages/graphql/src/translate/create-update-and-params.ts
```diff
@@ -475,7 +483,7 @@ function createUpdateAndParams({
                     skipIsAuthenticated: true,
                     context,
                     bind: { parentNode: node, varName, chainStr: param },
-                    escapeQuotes: Boolean(insideDoWhen),
+                    escapeQuotes: false,
```
darrellwarde, line 486:
```suggestion
```

packages/graphql/src/translate/utils/escape-query.ts
```diff
@@ -18,5 +18,6 @@
  */
 
 export function escapeQuery(query: string): string {
-    return query.replace(/("|')/g, "\\$1");
+    // return query.replace(/("|')/g, "\\$1");
```
darrellwarde, line 21:
```suggestion
    // TODO: Should single quotes be escaped?
    // return query.replace(/("|')/g, "\\$1");
```

packages/graphql/tests/integration/subscriptions/update/update.int.test.ts
```diff
@@ -0,0 +1,339 @@
...
+
+import { gql } from "apollo-server";
+import { graphql } from "graphql";
+import { Driver, Session } from "neo4j-driver";
+import { Neo4jGraphQL } from "../../../../src";
+import { generateUniqueType, UniqueType } from "../../../utils/graphql-types";
+import { TestSubscriptionsPlugin } from "../../../utils/TestSubscriptionPlugin";
+import neo4j from "../../neo4j";
+
+describe("Subscriptions delete", () => {
```
darrellwarde, line 28:
```suggestion
describe("Subscriptions update", () => {
```

PR #1112, opened 2022-03-10
Subscriptions: Nested create and delete under update Mutation

darrellwarde:


packages/graphql/tests/integration/subscriptions/update/update.int.test.ts
```diff
@@ -336,4 +336,333 @@ describe("Subscriptions update", () => {
...
+                    properties: {
+                        old: { name: "Arnold" },
+                        new: undefined,
+                    },
+                },
+            ])
+        );
+    });
+
+    test.only("update mutation with nested delete and create", async () => {
```
darrellwarde, line 471:
```suggestion
    test("update mutation with nested delete and create", async () => {
```

PR #1116, opened 2022-03-11
Add new multi DB error message, move to util function

darrellwarde:
# Description

5.0 tests failing with new error message when multi DB not supported.


packages/graphql/tests/utils/is-multi-db-unsupported-error.ts
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export function isMultiDbUnsupportedError(e: Error) {
+    if (
+        e.message.includes("This is an administration command and it should be executed against the system database") ||
+        e.message.includes(`Neo4jError: Unsupported administration command`) ||
```
tbwiss, line 23: **nitpicking (non-blocking)**: Shall we change that to a string (rather than a template string)? `

tbwiss (approved):
LGTM!

packages/graphql/tests/utils/is-multi-db-unsupported-error.ts
darrellwarde:
```suggestion
        e.message.includes("Neo4jError: Unsupported administration command") ||
```

tbwiss (approved):

PR #1143, opened 2022-03-16
Write details to transaction metadata instead of useragent

darrellwarde:
# Description

Overriding the driver useragent is not a reliable way of getting the details of the Neo4j GraphQL Library into the Neo4j query log. This instead writes those details to the transaction metadata, which are also available to view in logs.


tbwiss (approved):
Neat solution!
Tried it locally and checked the query.log file, the metadata is present.

PR #1146, opened 2022-03-17
Feat: Autogenerated Properties - callback directive

tbwiss:
# Description

With this PR the directive `@callback` is added as a new config option to `Neo4jGraphQL`.

See the details in the [RFC document](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-012-autogenerated-properties.md)

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-1146.surge.sh

darrellwarde:
Literally about to walk out of the office, but want to ask a quick question - why is `callbacks` added to the Cypher parameters given that these are JavaScript callbacks and not at all related to Cypher?

packages/graphql/src/types.ts
```diff
@@ -366,3 +375,11 @@ export interface JwtPayload {
     iat?: number | undefined;
     jti?: string | undefined;
 }
+
+export type CallbackReturnValue = string | number | boolean | undefined | null;
```
darrellwarde, line 379: What about dates and/or times? How about points?

packages/graphql/src/schema/get-nodes.ts
```diff
@@ -117,6 +117,7 @@ function getNodes(definitionNodes: DefinitionNodes): Nodes {
             objects: definitionNodes.objectTypes,
             scalars: definitionNodes.scalarTypes,
             unions: definitionNodes.unionTypes,
+            callbacks: options.callbacks,
```
darrellwarde, line 120: **question:** This feels a bit... Weird. You have the function `getObjFieldMeta` which previously accepted purely GraphQL types, and now we're passing in all of these callbacks. Feels like we're slightly polluting the interface, but unsure of a better way around this.

packages/graphql/src/schema/get-callback-meta.ts
```diff
@@ -0,0 +1,41 @@
...
+    const operationsArg = directive.arguments?.find((x) => x.name.value === "operations") as ArgumentNode;
+    const nameArg = directive.arguments?.find((x) => x.name.value === "name") as ArgumentNode;
+
+    const operationsList = operationsArg.value as ListValueNode;
+    const operations = operationsList.values.map((value) => (value as StringValueNode).value) as CallbackOperations[];
+    const name = (nameArg.value as StringValueNode).value;
+
+    if (typeof (callbacks || {})[name] !== "function") {
+        throw new Error(`Directive callback '${name}' must be of type function`);
+    }
```
darrellwarde, line 33: And here lies the root of the problem as to why `getObjFieldMeta` has been polluted with the `callbacks` arguments. I don't think this is really an issue for this PR, but I just wanted to raise it. I think we need to consider our architecture for directives - we really blur the lines between directive validation and directive parsing which is finally coming to bite us with things like this. I think we should sit down and consider what we do here.

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -88,7 +88,13 @@ function makeAugmentedSchema(
         enableRegex,
         skipValidateTypeDefs,
         generateSubscriptions,
-    }: { enableRegex?: boolean; skipValidateTypeDefs?: boolean; generateSubscriptions?: boolean } = {}
+        callbacks,
+    }: {
+        enableRegex?: boolean;
+        skipValidateTypeDefs?: boolean;
+        generateSubscriptions?: boolean;
+        callbacks?: Neo4jGraphQLCallbacks;
```
darrellwarde, line 96: Yeah, this being littered everywhere is definitely bugging me. The fact that it's both here and also in the context shows that we have some serious duplicity here. Perhaps we should have a discussion about this before merging?

packages/graphql/src/translate/create-create-and-params.test.ts
```diff
@@ -72,12 +73,13 @@ describe("createCreateAndParams", () => {
         const neoSchema: Neo4jGraphQL = {
             nodes: [node],
         };
-
+        // @ts-ignore
+        const context = { neoSchema } as Context;
```
darrellwarde, line 77: I realise not part of this PR, but should we use `ContextBuilder` here?

packages/graphql/src/translate/create-set-relationship-properties.ts
```diff
@@ -60,6 +63,22 @@ function createSetRelationshipProperties({
...
+
+        const paramName = `${varName}_${field.fieldName}_${field.callback?.name}`;
+
+        callbackBucket.addCallback({
+            functionName: field.callback?.name,
+            paramName,
+            parent: properties,
+        });
+
+        strs.push(`SET ${varName}.${field.dbPropertyName} = $callbacks.${paramName}`);
```
darrellwarde, line 79: This is repeated a lot in various files, should it be a shared function?

darrellwarde (commented):
I'm really sorry for this terrible quality review by me, it was done over several sessions and is very much the ramblings of a crazy man!! Happy to discuss any points. May delete a few comments when I go back through and read them. 😆

packages/graphql/src/translate/create-set-relationship-properties.ts
tbwiss: Moved it to a util function in https://github.com/neo4j/graphql/pull/1146/commits/9ae6ae67184430692b173e4e5708ece6855f39ee but I'm not sure this is a good place to put it..

Issue #1050, opened 2022-02-24
Exception when running a nested graphql query with @auth rules


dennisjlee:
**Describe the bug**
My team encountered a situation where a particular combination of @auth rules and querying through "fooConnection" relationships in GraphQL resulted in a Neo4j exception.

The exception would look like this: 
```
{"message":"Failed to invoke function `apoc.cypher.runFirstColumn`: Caused by: org.neo4j.exceptions.ParameterNotFoundException: Expected parameter(s): this_inboxes_message_attachments_auth_allow0_ownerId" ....
```

**To Reproduce**

I have a repo that reproduces the issue, with instructions: https://github.com/dennisjlee/neo4j-graphql-bug-repro

**Expected behavior**
The authorization checks should run correctly without causing any exceptions.

**System (please complete the following information):**

* OS: Windows
* Version: @neo4j/graphql@3.0.1, @neo4j/graphql-plugin-auth@1.0.0
* Node.js version: 16.14.0

**Additional context**

The underlying issue seems to be in `create-projection-and-params.ts` where the code creates the parameters for the `apoc.cypher.runFirstColumn` call. There are multiple auth-related parameters passed to the top-level query, but only one of them is passed along into the runFirstColumn call, resulting in the missing parameter exception. I have a fix that resolves the issue locally - will send a PR soon!

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @dennisjlee! :pray: We will now prioritise the bug and address it appropriately.

dennisjlee:
@angrykoala @tbwiss @darrellwarde  thank you for your help getting this merged. I saw that a new release 3.0.2 went out today but did not include this commit. Is there anything else we need to do in order to get this fix included in a release?

angrykoala:
Hi @dennisjlee 

Usually, patch release (like today) only contain non-breaking high priority bugfixes released from the `master` branch. This PR was made to `dev` and, sincerely, we didn't noticed it could have been ported to `master` so it could have been released along with the rest of the fixes. 

I've created a new pr (#1152) cherry-picking the merge commit into master so it can be released on the next patch release. Sadly, I cannot give you an exact date it will be released, but it should be in the next release, patch or minor

Sorry for the inconvenience, as we are still improving our workflow for patch releases :pray:

dennisjlee:
@angrykoala thank you for explaining! I appreciate the help.

PR #1152, opened 2022-03-18
Port PR #1051 to master

angrykoala:
The PR #1051 containing a bugfix for #1050 was made to `dev` instead of `master` so it wasn't released in `3.0.2`. This PR simply cherry picks that commit into master so it can be released on the next patch release:

Pass multiple connection parameters to runFirstColumn

### DIFF
--- a/packages/graphql/src/translate/create-projection-and-params.ts
+++ b/packages/graphql/src/translate/create-projection-and-params.ts
@@ -554,10 +554,12 @@ function createProjectionAndParams({
                 nodeVariable: varName,
             });
 
-            const connectionParamName = Object.keys(connection[1])[0];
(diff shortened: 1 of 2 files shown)


dennisjlee:
@angrykoala thank you! For future reference, should I send any PRs to target `master` instead of `dev`?

darrellwarde:
> @angrykoala thank you! For future reference, should I send any PRs to target `master` instead of `dev`?

Hey @dennisjlee, `dev` remains the default, merging to `master` is only for high priority bugs which we want to release as a hotfix. Normally these issues will be labelled with the "high priority" label and a comment will be made with instructions.

tbwiss (approved):

Issue #1150, opened 2022-03-18
UnionTypes with Authorization and ConnectionWhere generates illegal Cypher Query


tbwiss:
**Describe the bug**
The customer is using  @neoj4/graphql to provide multi-filter functionality to our users inside of a React application. A multi-filter is a list of filter functions (e.g. node.attribute = value) combined with operators like AND/OR. 

**Type definitions**
```gql
type Battery @exclude(operations: [CREATE, UPDATE, DELETE]) {
  batterySeriesNumber: String
  nameMigrated: String
}

extend type Battery @auth(rules: [{ isAuthenticated: true, roles: ["admin"] }])

extend type Battery implements BusinessObject {
  id: ID! @id(autogenerate: false)
  current: Boolean!
}

type CombustionEngine @exclude(operations: [CREATE, UPDATE, DELETE]) {
  power: Int
  torqueMaximum: Int
}

extend type CombustionEngine implements BusinessObject {
  id: ID! @id(autogenerate: false)
  current: Boolean!
}

type Drive @exclude(operations: [CREATE, UPDATE, DELETE]) {
  driveCompositions: [DriveComposition!]!
    @relationship(
      type: "CONSISTS_OF"
      properties: "RelationProps"
      direction: OUT
    )
}

extend type Drive implements BusinessObject {
  id: ID! @id(autogenerate: false)
  current: Boolean!
}

union DriveComponent = Battery | CombustionEngine

type DriveComposition @exclude(operations: [CREATE, UPDATE, DELETE]) {
  driveComponent: [DriveComponent!]!
    @relationship(type: "HAS", properties: "RelationProps", direction: OUT)
}

extend type DriveComposition implements BusinessObject {
  id: ID! @id(autogenerate: false)
  current: Boolean!
}

interface BusinessObject {
  id: ID! @id(autogenerate: false)
  current: Boolean!
}

interface RelationProps {
  current: Boolean!
}
```

**To Reproduce**

1. execute the following query
```gql
query getDrivesWithFilteredUnionType(
  $where: DriveWhere = { current: true }
  $connectionWhere: RelationPropsWhere = { current: true }
) {
  drives(where: $where) {
    current
    driveCompositionsConnection(where: { edge: $connectionWhere }) {
      edges {
        node {
          driveComponentConnection(
            where: {
              Battery: { edge: $connectionWhere }
              CombustionEngine: { edge: $connectionWhere }
            }
          ) {
            edges {
              current
              node {
                ... on Battery {
[...]

PR #1170, opened 2022-03-23
Fix: UnionTypes with Authorization and ConnectionWhere generates illegal Cypher Query

tbwiss:
# Description

In case of a connection-where, the `WHERE` clause was wrongly put after the `CALL` (rather than before) when an auth directive (allow) is present on a union type. See file `packages/graphql/src/translate/connection/create-connection-and-params.ts`.

# Issue

This is addressing a customer issue reported in https://github.com/neo4j/graphql/issues/1150

Closes:

- https://github.com/neo4j/graphql/issues/1150

# Checklist


darrellwarde (approved):
Looks like a good fix to me, nice work!! Merge and hotfix release perhaps? 👀

PR #1171, opened 2022-03-23
Finalising Subscriptions types

darrellwarde:


packages/graphql/src/schema/subscriptions/generate-subscription-where-type.ts
```diff
@@ -0,0 +1,42 @@
...
+        ...node.interfaceFields,
+        ...node.objectFields,
+        ...node.unionFields,
+        ...node.temporalFields,
+        ...node.pointFields,
+        ...node.computedFields,
+    ]);
+
+    return schemaComposer.createInputTC({
+        name: `${node.name}SubscriptionWhere`,
```
darrellwarde, line 39: I really don't like this type name...

packages/graphql/src/schema/subscriptions/generate-subscription-types.ts
```diff
@@ -31,47 +34,77 @@ export function generateSubscriptionTypes({
...
 
         const nodeCreatedEvent = schemaComposer.createObjectTC({
             name: `${node.name}CreatedEvent`,
             fields: {
-                [subscribeOperation.created]: composeNode,
+                event: {
+                    type: eventTypeEnum.NonNull,
+                    resolve: () => EventType.getValue("CREATE"),
+                },
+                [`created${node.name}`]: {
```
angrykoala, line 51: :thinking: if we use node.name here we will have problems with underscores right?

Maybe we should expose `node.singular` or something like that to provide a standard name here

packages/graphql/src/schema/subscriptions/generate-subscription-types.ts
```diff
@@ -31,47 +34,77 @@ export function generateSubscriptionTypes({
...
         const nodeUpdatedEvent = schemaComposer.createObjectTC({
             name: `${node.name}UpdatedEvent`,
             fields: {
-                [subscribeOperation.updated]: composeNode,
+                event: {
+                    type: eventTypeEnum.NonNull,
+                    resolve: () => EventType.getValue("UPDATE"),
+                },
+                previousState: {
+                    type: eventPayload.NonNull,
```
angrykoala, line 66: question: Will this work fine with nodes only having relationships?

packages/graphql/src/schema/subscriptions/generate-subscription-where-type.ts
angrykoala: Maybe with some ketchup and mustard?

angrykoala (changes requested):

angrykoala (dismissed):

packages/graphql/src/schema/subscriptions/generate-subscription-types.ts
darrellwarde: Following our conversation this morning, I believe the right approach is to not add the payload types if they have no actual value. This is now implemented.

PR #1174, opened 2022-03-24
apollo server and websocket client helpers for tests

angrykoala:
# Description

This PR adds the piping required for end to end tests, and adds a E2E test for subscriptions


darrellwarde:
Looks like `import/no-extraneous-dependencies` can be configured to ignore certain files rather than adding a load of inline comments: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md#options

packages/graphql/tests/e2e/subscriptions.e2e.test.ts
```diff
@@ -0,0 +1,126 @@
...
+         type ${typeMovie} {
+             title: String
+         }
+
+         type ${typeActor} {
+             name: String
+         }
+         `;
+
+        driver = neo4j.driver("bolt://localhost:7687", neo4j.auth.basic("neo4j", "dontpanic42"));
```
darrellwarde, line 49: This is very much catered to your environment - this should be configured the same way as integration tests.

packages/graphql/tests/e2e/subscriptions/update.e2e.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        await driver.close();
+        await wsClient.close();
+    });
+
+    test("update subscription", async () => {
+        const session = driver.session();
+        try{
+            await session.run(`CREATE (:${typeMovie.name} {title: "my-title"})`);
+        } finally {
+            session.close();
```
darrellwarde, line 70:
```suggestion
            await session.close();
```

darrellwarde (dismissed):
Looks good, I believe most of the E2E changes are in the "where branch" right?!

Issue #1189, opened 2022-03-28
Missing label check for fulltext search


Andy2003:
Given the schema ([taken from the tests](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/tck/tck-test-files/fulltext/match.test.ts#L31-L33)):

```graphql
type Movie @fulltext(indexes: [{name: "MovieTitle", fields: ["title"]}]) {
  title: String
}
```

and the request

```graphql
{
  movies(
    fulltext: {MovieTitle: {phrase: "something AND something"}}
    where: {title: "some-title"}
  ) {
    title
  }
}
```

the following cypher is generated:

```cypher
CALL db.index.fulltext.queryNodes(
    "MovieTitle",
    $this_fulltext_MovieTitle_phrase
) YIELD node as this, score as score
WHERE this.title = $this_title
RETURN this { .title } as this
```

I would expect  a label check, like in the [additional label tests](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/tck/tck-test-files/fulltext/node-labels.test.ts#L106):

```cypher
CALL db.index.fulltext.queryNodes(
    "MovieTitle",
    $this_fulltext_MovieTitle_phrase
) YIELD node as this, score as score
WHERE \\"Movie\\" IN labels(this)  AND this.title = $this_title
RETURN this { .title } as this
```

Additional Information:

* The index may contain other nodes not matching the ones queried
* The check for `additionalLabels` should be removed
https://github.com/neo4j/graphql/blob/f079969bcd5976cd4e92ae23d628e7136f234980/packages/graphql/src/translate/translate-top-level-match.ts#L65-L69

neo4j-team-graphql:
Many thanks for raising this bug report @Andy2003. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @Andy2003! :pray: We will now prioritise the bug and address it appropriately.

PR #1196, opened 2022-03-29
Remove broken and redundant full-text `defaultThreshold` and `score`

### LINKED ISSUE #1188: Wrong param for fulltext queries
given the schema ([taken from the tests](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/tck/tck-test-files/fulltext/score.test.ts#L27-L29)):

```graphql
type Movie @fulltext(indexes: [{name: "MovieTitle", fields: ["title"]}]) {
  title: String
}
```

and the query

```graphql
{
  movies(
    fulltext: {MovieTitle: {phrase: "something AND something", score_EQUAL: 1}}
  ) {
    title
  }
}
```

The library is currently generating the following cypher:

```cypher
CALL db.index.fulltext.queryNodes(
    "MovieTitle",
    $this_fulltext_MovieTitle_phrase
) YIELD node as this, score as score
WHERE score = this_fulltext_MovieTitle_score_EQUAL
RETURN this { .title } as this
```

The `this_fulltext_MovieTitle_score_EQUAL` should be a parameter (prefixed with `$`)

darrellwarde:
# Description

NOTE: This is technically a breaking change through the removal of a directive and a query argument. However, the two arguments are currently completely broken, and even if fixed, would be entirely useless due to the way they are implemented.

This branch removes `defaultThreshold` from the `@fulltext` directive and the `score` argument when querying a full-text index. If these were used, the parameter was not prepended with `$` as reported in #1188 and as such would bail out. Even if this were fixed, the functionality is completely useless. Due to the many decimal points returned by full-text searches (https://neo4j.com/docs/cypher-manual/current/indexes-for-full-text-search/#administration-indexes-fulltext-search-query), performing an equality check would likely never return any results. This should be reimplemented in the future as a `minimumScore`, as suggested in #1197.

Whilst I was here, added the label check as suggested in #1189.

# Issue

Closes #1188 and #1189.


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-1196.surge.sh

tbwiss (approved):
LGTM!
And the suggestion with using a `minimumScore` rather than a "score equal" seems plausible. Especially when looking at the cypher response for the fulltext query.

PR #1198, opened 2022-03-29
Add missing fields to `onCreate` input types

### LINKED ISSUE #1182: Missing scalar timestamps fields in  *OnCreateInput input type
Given the following schema ([from the tests](https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/schema/connect-or-create-id.test.ts#L412-L423))

```graphql
type Post {
    id: ID! @id(autogenerate: false)
    content: String!
    creator: User! @relationship(type: "HAS_POST", direction: IN)
    createdAt: DateTime!
}
type User {
    id: ID! @id(autogenerate: true)
    name: String!
}
```

The following type is generated:

```graphql
input PostOnCreateInput {
  content: String!
  id: ID!
}
```

I would expect the type `PostOnCreateInput` to contain the field `createdAt`, since this field is no auto generated field:

```graphql
input PostOnCreateInput {
  content: String!
  createdAt: DateTime!
  id: ID!
}
```

***Additional Info***

this is because of this filter: 

https://github.com/neo4j/graphql/blob/c60707d702daf3196a0bceba66969a05c9fe5427/packages/graphql/src/schema/to-compose.ts#L86-L89

which was introduced via #1073

darrellwarde:
# Description

Adds missing temporal and point fields to `onCreate` input types.

# Issue

Closes #1182


tbwiss (dismissed):
LGTM!

Issue #1130, opened 2022-03-15
Generate function on OGM fails when type begins with multiple uppercase letters 


litewarp:
**Describe the bug**

If you try and use the generate function from the OGM when you have a type that starts with multiple capital letters -- e.g., `SICIndustry` -- it will fail with the error `Cannot read property "split" of undefined`.  If you change the typeDefinition name to `SicIndustry`, there is no error.

**Types**

```graphql

type Company {
  name: String!
  industries: [SICIndustry!]! @relationship(type: "CONDUCTS_BUSINESS_IN", direction: OUT)
}

# <-- Changing the name to "SicIndustry" works
type SICIndustry {
  code: ID! @id(autogenerate: false)
  title: String!
  companies: [Company!]! @relationship(type: "CONDUCTS_BUSINESS_IN", direction: IN)
}

```

The debug trace points to the `createLines` function called by `createAggregationInput` function in the [generate.ts](https://github.com/neo4j/graphql/blob/dev/packages/ogm/src/generate.ts) file. 

**Screenshots**

Output from the error (I run generate as an api route I can call in nextjs during dev):

![image](https://user-images.githubusercontent.com/1900476/158323581-129ef82e-cca3-4828-bfad-68d540a75dbd.png)

**System (please complete the following information):**

* OS: macOs
* Version: 3.0.1
* Node.js version: lts/fermium

neo4j-team-graphql:
Many thanks for raising this bug report @litewarp. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @litewarp! :pray: We will now prioritise the bug and address it appropriately.

PR #1200, opened 2022-03-30
Fix/issue 1130

danstarns:
# Description

Use the correct plural name when referring to the existing type. Added some comments so people know what on earth is going on in this file! 

# Issue

#1130 

Closes:

#1130 
-

PR #1206, opened 2022-03-31
Fix/use __typename when looking for other types

danstarns:
# Description

Supersedes #1200 

Changes look at the `__typename` field on the generated types over the 'top level type name`, this is because GraphQL Codegen is not preserving the type names. I have raised an issue about that here https://github.com/dotansimha/graphql-code-generator/issues/7714

Some refactors to the aggregation types - centrally storing them in the `Node` class. 

# Issue

#1130 

Closes: #1130

### DIFF
--- a/packages/graphql/src/classes/Node.ts
+++ b/packages/graphql/src/classes/Node.ts
@@ -209,8 +209,8 @@ class Node extends GraphElement {
 
     public get aggregateTypeNames(): AggregateTypeNames {
         return {
-            selection: `${this.name}AggregateSelection`,
-            input: `${this.name}AggregateSelectionInput`,
+            selection: `${this.pascalCaseSingular}AggregateSelection`,
+            input: `${this.pascalCaseSingular}AggregateSelectionInput`,
         };
     }
 
--- a/packages/graphql/src/classes/Node.ts
+++ b/packages/graphql/src/classes/Node.ts
@@ -97,6 +97,11 @@ export type RootTypeFieldNames = {
     aggregate: string;
 };
 
+export type AggregateTypeNames = {
+    selection: string;
+    input: string;
+};
+
 export type MutationResponseTypeNames = {
     create: string;
     update: string;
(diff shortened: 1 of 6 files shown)


tbwiss (dismissed):
Looks good!
Left a minor comment also for this one.

packages/ogm/tests/issues/1130.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+describe("issues/1130", () => {
+    test("should re-create issue and return types without throwing", async () => {
+        const typeDefs = `
+            type Company {
+              name: String!
+              industries: [SICIndustry!]! @relationship(type: "CONDUCTS_BUSINESS_IN", direction: OUT)
+            }
+            
+            # <-- Changing the name to "SicIndustry" works
```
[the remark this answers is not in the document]
danstarns: I'll keep it just to copy exactly the reproduce steps!

packages/graphql/src/classes/Node.ts
```diff
@@ -202,6 +207,13 @@ class Node extends GraphElement {
         };
     }
 
+    public get aggregateTypeNames(): AggregateTypeNames {
+        return {
+            selection: `${this.name}AggregateSelection`,
+            input: `${this.name}AggregateSelectionInput`,
```
darrellwarde, line 213: This should probably use the pascal case singular for consistency - for instance we would want `AWSInstance` to have `AwsInstanceAggregateSelection` like we do elsewhere... I think. 😬 

```suggestion
            selection: `${this.pascalCaseSingular}AggregateSelection`,
            input: `${this.pascalCaseSingular}AggregateSelectionInput`,
```

darrellwarde (commented):
Almost LGTM!!! Just wondering whether we should be using the pascal case of the name rather than the name directly, like we do elsewhere.

packages/graphql/src/classes/Node.ts
danstarns: We should however I think doing so could cause unforeseen breaking changes to users with unusual type names such as the one demonstrated in this issue `SICIndustry`.

darrellwarde: Shall we just do it? Thoughts @angrykoala and @tbwiss?

tbwiss: I vote for consistency in this case. so pro pascalCaseSingular. but yes we might get some bug reports due to this breaking change.

Issue #847, opened 2022-01-21
Querying Multiple Interface Relations generates Invalid Cypher Query


jnterry:
**Describe the bug**

When a `type` contains multiple `@relationship`s to an `interface`, querying both fields at once results in an invalid cypher query being generated. Querying each field individually works as expected.

**Type definitions**

**NOTE**: This is the same schema that generates another (I think unrelated) bug: https://github.com/neo4j/graphql/issues/832

```graphql
interface Entity {
    id: String!
}

type Person implements Entity {
    id   : String! @unique
    name : String!
}

type Place implements Entity {
    id       : String! @unique
    location : Point!
}

type Interaction  {
    id       : ID! @id
    kind     : String!
    subjects : [Entity!]! @relationship(type: "ACTED_IN", direction: IN )
    objects  : [Entity!]! @relationship(type: "ACTED_IN", direction: OUT)
}
```

**To Reproduce**

Seed the database with some data using the following mutation:

```graphql
mutation {
  createPeople(input: [
    { id: "adam", name: "Adam" },
    { id: "eve",  name: "Eve" },
    { id: "cain", name: "Cain"},
  ]) {
    people {
      id
    }
  }
  createInteractions(input: [{
    subjects : { connect: { where: { node: { id_IN: ["adam", "eve"] }}}},
    kind     : "PARENT_OF",
    objects  : { connect: { where: { node: { id_IN: ["cain"] }}}}
  }]) {
    info { nodesCreated }
    interactions { id }
  }
}
```

Now execute the following query:

```graphql
query {
  interactions {
    id
    subjects { id }
    objects { id }
  }
}
```

The following error is produced - suggesting that the generated Cypher code is invalid:

```json
{
  "message": "Variable `subjects` not defined (line 22, column 38 (offset: 598))\n\"RETURN this { .id, subjects: collect(subjects), objects: collect(objects) } as this\"\n                                      ^",
  "locations": [{ "line": 2, "column": 3}],
  "path": ["interactions"]
}
```

Note that querying just `subjects { id }` or just `objects { id }` one at a time works as expected. The bug is only observed when querying both fields at once.

**Expected behavior**

Query should execute successfully.

**System (please complete the following information):**

OS: Linux (Debian 10)
Version: @neo4j/graphql@2.5.5, graphql@16.2.0, apollo-server@3.6.2
Node.js version: v14.15.0
Neo4j DB Version: 4.4.0

neo4j-team-graphql:
[...]

PR #1209, opened 2022-04-01
Fix: Querying Multiple Interface relations generates invalid Cypher Query

tbwiss:
# Description

The implemented solution carries over every variable defined in a prior subquery (in this test case `subjects`) onto the next subquery, to then be part of the last RETURN statement.

# Issue

The issue: #847

Closes:

- #847 

# Checklist


darrellwarde (approved):
This looks like it should work a treat to me!

Issue #1119, opened 2022-03-13
Unable to return multiple node types in a single query


danielfein:
I'm trying to return results from different node types in a single query:

```
        MATCH (this)-[a:WROTE]->(wrote:Post)
        WHERE a.date_added IS NOT NULL
        WITH COLLECT(wrote{ .*, date_added: a.date_added, typename: 'WROTE' }) as updates1, this
        MATCH (this)-[a:FOLLOWS]->(umb)
        WHERE (umb:User or umb:Movie or umb:Blog) AND a.date_added IS NOT NULL
        WITH updates1 + COLLECT(umb{ .*, date_added: a.date_added, typename: 'FOLLOWED' }) as allUpdates
        UNWIND allUpdates as update
        RETURN update
        ORDER BY update.date_added DESC
        LIMIT 5
```

And this works in the Neo4j browser, but it doesn't work in a GraphQL query. Where I'm using the Neo4j GraphQL library.

```
    updates: [PostMovieUser!]! @cypher(statement: """
        MATCH (this)-[a:WROTE]->(wrote:Post)
        ...
        (above query)
        ...
    """),
```

`union PostMovieUser = Post | Movie | User`

This is my error: 

```
"message": "Invalid input ']': expected \"NOT\" or an expression (line 12, column 79 (offset: 632))\n\"LIMIT 5\", {this: this, auth: $auth, cypherParams: $cypherParams}, false) |    ] } as this\"\n                                                                               ^",
```

neo4j-team-graphql:
Many thanks for raising this bug report @danielfein. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

tbwiss:
Hej @danielfein, thanks for reaching out!

Would you be able to add some more details, for instance the typeDefs, the cypher query you used to add nodes and relationships to the Neo4j database, and the query/mutation you are using (with the cypher statement)?

Thanks!

stkterry:
I'm experiencing the same issue.  I am new to GraphQL/Neo4j and might be missing something obvious.  
Anyway, I've provided a simplified schema that's causing the issue.

_This example is contrived without the extra bits but it fails with the same error regardless._
[...]

PR #1226, opened 2022-04-06
Fix: @cypher only returning __typename fails

tbwiss:
# Description

In the `@cypher` statement, if the return type is a union: we also return only the `__typename` if no (inline) fragment is specified in the result selection set. 

# Issue

https://github.com/neo4j/graphql/issues/1119

Closes:

- https://github.com/neo4j/graphql/issues/1119

# Checklist


darrellwarde (approved):
LGTM! 🏆

PR #1229, opened 2022-04-06
Rfc/Update subscriptions

angrykoala:
# Description
Updates subscriptiosn rfc to take into account advanced features and some pitfalls:
* Auth
* Complex filtering
* Connect/Disconnect events
* Plugin implementation
* Pubsub support
* Scalability considerations


docs/rfcs/rfc-007-subscriptions.md
```diff
@@ -8,42 +8,42 @@ Our users would like to use [GraphQL Subscriptions](https://graphql.org/blog/sub
...
-- Auth (on read)
+- Ability to filter which nodes are subscribed to (i.e. `where` clause).
+- Ability to horizontally scale - either now, or able to do so without breaking in the future.
+- Wherever possible, events are ordered.
+- Events are fired individually.
+- We won't mutate our users' databases - no metadata.
+- All events are sent (for example, if node created and then deleted, we get both events).
+- Database transactions must be successful - no optimisticness.
+- Garbage collection of old subscriptions.
+- Some form of auth validation.
```
tbwiss, line 23: How about adding this?
- Shall not break web browser support of the `@neo4j/graphql` lib.

docs/rfcs/rfc-007-subscriptions.md
```diff
@@ -209,39 +330,253 @@ class Neo4jGraphQLSubscriptionsAMQPPlugin extends Neo4jGraphQLSubscriptionsPlugi
...
+
+
+### Nested fields
+Due to subscriptions relying on the `meta` data returned by the queries, only top-level properties of the node are available to return or filter.
+
+To properly support nested queries, follow-up queries to the database, after the subscription has been triggered would be required. This may be not viable as each subscription would require an extra request, this not only put a huge load to the database, the subscription itself will take longer to fulfill, instead of close to real-time. Adding some extra metadata and caching subscriptions may help,
+
+Due to this, any nested field queries support should be opt-in.
+
+### Horizontal scalability
```
tbwiss, line 547: (It's not this line)
Just want to remind ourselves to think about any security considerations. For instance "How can a malicious user abuse this?" and in such a case even "How could we prevent that?"

docs/rfcs/rfc-007-subscriptions.md
angrykoala: Good point

docs/rfcs/rfc-007-subscriptions.md
angrykoala: Great point, I'll add a new section covering any security consideration

tbwiss (commented):
Nothing more to add from my side just now.
Good additions concerning security considerations!

darrellwarde (approved):
Happy to merge this as we need to get planning finalised, but the scope of this RFC is way beyond what we will achieve in the development phase. Let's pick out particular points to achieve in the Trello card.

cramatt:
@darrellwarde  @angrykoala  checking on availability of connect/ disconnect and found this MR, too. I see again that connect and disconnect have made it into the RFC - Any updates on development?

angrykoala:
Hi @cramatt Thanks for your interest

We are now working actively on connect/disconnect events (you can see some progress in PR #2233)

We found a couple of unexpected problems that we need to fix before connect/disconnect is ready, but we are making good progress and should be available in a few weeks.

Sorry it took us so long to get around this, this is a really tricky piece of API to design and implement :pray:

PR #1251, opened 2022-04-11
Fix: neo-push client and server

tbwiss:
This PR fixes the neo-push server and client. They should now run again locally.

**More importantly**:
There are several Dependabot alerts that should be resolved. For instance:
Upgrade ansi-html to version 0.0.8 or later
Upgrade node-forge to version 1.3.0 or later

It turns out that those two dependencies above are used in the outdated version of webpack-dev-server in the neo-push client. Bumping it gets rid of those two dependencies.


darrellwarde (approved):
Lovely, looks good to me!

PR #1274, opened 2022-04-16
Fix cypher on top level connections ignoring sort, after, and first arguments

### LINKED ISSUE #1273: Sort, first, and after arguments not correctly applied to root-level connection fields
**Describe the bug**

The sort, limit, and offset strings for top-level-connections get applied on the returned connection -- i.e. the object `{totalCount, edges}` and not the edges themselves. So the sort, offset, and limit arguments do nothing.

Also, the root-connection resolver uses the resolveTree args which don't get translated correctly when passed to the `createConnectionWithEdgeProperties` function. So 

These are probably oversights on my part while drafting the root-level connections PR, but I think things also changed slightly from the fix for #591.  I've fixed these and will be issuing a PR shortly.

To reproduce, run a query on a top level connection with first and sort arguments: 

**TypeDefs**

```graphql

type Movie {
  id: ID! @id
  title: String!
}

```

Create 20 movies, then run:

**Query**

```graphql

query {
  moviesConnection(first: 10, sort: [{title: ASC}]) {
    totalCount
    edges {
      node {
        title
      }
    }
    pageInfo {
      # <-- This will be false even though the totalCount returns 20 and the first argument is 10
      hasNextPage
    }
  }
}
```

**Cypher**

```
MATCH (this:Movie)
WITH COLLECT({ node: this { .title } }) as edges
RETURN { edges: edges, totalCount: size(edges) } as this
# <------- these arguments need to be applied before returning the connection
ORDER BY this.title ASC
LIMIT $this_limit
```

* OS: macOs
* Version: 3.1.0
* Node.js version: 14*

litewarp:
Modifies the projection in `translate-read` to return the correct data when sort, after, and first arguments are applied.

Fixes #1273

# Checklist


packages/graphql/tests/tck/tck-test-files/root-connection.test.ts
```diff
@@ -61,16 +67,141 @@ describe("Root Connection Query tests", () => {
...
+            }
+            CALL {
+            WITH this
+            MATCH (this)<-[this_acted_in_relationship:ACTED_IN]-(this_actor:Actor)
+            WITH collect({ node: { name: this_actor.name } }) AS edges
+            RETURN { edges: edges, totalCount: size(edges) } AS actorsConnection
+            }
+            WITH COLLECT({ node: this { .title, actorsConnection } }) as edges, totalCount
+            RETURN { edges: edges, totalCount: totalCount } as this"
+        `);
```
tbwiss, line 205: Let's also check the `result.params` here.

```suggestion
        `);
        expect(formatParams(result.params)).toMatchInlineSnapshot(`
            "{
                \\"this_limit\\": {
                    \\"low\\": 20,
                    \\"high\\": 0
                }
            }"
        `);
```

tbwiss (approved):
Thank you @litewarp, LGTM!
Added one suggestion to consider but otherwise I think it's good to go.

tbwiss:
@litewarp, would you be able to point this branch to `neo4j:master`? Then it can be part of the next (hotfix) release. Thank you!

tbwiss:
@litewarp, also make sure to update the base branch. There seem to be a lot of file changes now in the PR that are unrelated to your work.

litewarp:
@tbwiss My post-midnight attempt at rebasing this failed miserably which leads me to the conclusion that I will have to try again tomorrow. However, I would hate for my need for sleep to hold up this hotfix, especially since my git skills are less-than-ideal at the moment. So if you're looking to push today, and have the capacity to rebase the PR, please go ahead and fix. Otherwise I will give it another go tomorrow.

tbwiss:
@litewarp no problem at all! We are not planning to release today or tomorrow, so please let this not stop you from getting some rest!
Thank you very much!

litewarp:
@tbwiss should be good to go

tbwiss (approved):

PR #1282, opened 2022-04-20
Boyscouting: Custom grid for the Toolbox Editor view

tbwiss:
# Description

The main aim is to remove the dependency `"react-grid-resizable": "^1.0.0",` as it has a range of unresolved security issues in its dependencies.
With this we also gain more control over the editor view plus we can control the space between the editors better. That space now contains a svg to indicate that it's resizable.


packages/graphql-toolbox/src/components/editor/Editor.tsx
```diff
@@ -54,6 +54,7 @@ export const Editor = (props: Props) => {
     const [variableValues, setVariableValues] = useState("");
     const [output, setOutput] = useState("");
     const refForQueryEditorMirror = useRef<EditorFromTextArea | null>(null);
+    const isShowRightPanel = settings.isShowDocsDrawer || settings.isShowSettingsDrawer;
```
angrykoala, line 57: **nitpick not-blocking** I'm not sure about the name `is...`, I understand that this is a boolean to define whether or not should show the right panel, not whether or not is a thing. Maybe `showRightPanel` or `shouldShowRightPanel` is a bit more clear?

packages/graphql-toolbox/src/components/editor/utils.ts
```diff
@@ -69,3 +69,14 @@ export const handleEditorDisableState = (mirror: EditorFromTextArea | null, load
...
+export const isJsonString = (str: string | null | undefined): boolean => {
+    if (!str) return false;
+
+    try {
+        JSON.parse(str);
+    } catch (e) {
+        return false;
+    }
+    return true;
+};
```
angrykoala, line 82: **suggestion** :thinking: As I see, you use `.parse` to check if a variable is a json, and the run `json` again [here](https://github.com/neo4j/graphql/pull/1282/files#diff-475760ba8faf849a5b1fdcca13f67834d707b37914a49f593e7fd397a9dfd0e5R75)

Obviously, performance is not a huge concern, but maybe we can avoid parsing so many times by changing this into a custom parser function, something like:

```suggestion
export const safeParse = (str: string | null | undefined, default: Record<string, any>): Record<string,any> => {
    if (!str) return default;

    try {
        return JSON.parse(str);
    } catch (e) {
        return default;
    }
};
```

angrykoala (dismissed):
A couple of comments, non-blocking

Issue #1221, opened 2022-04-05
Missing filter with deep relations


tbwiss:
**Describe the bug**
A customer reported an issue where a filter is not applied on the generated Cypher query when the following condition is met:
- A relation to a child object (1) which also have an inner connection to another child object (2) is in place
- The filter is applied on an inner property of (2)

**Type definitions**
```gql
 type Series {
      id: ID! @id(autogenerate: false)
      current: Boolean!
      architecture: [MasterData!]! @relationship(type: "ARCHITECTURE", properties: "RelationProps", direction: OUT)
 }

 type NameDetails @exclude(operations: [CREATE, UPDATE, DELETE, READ]) {
      fullName: String!
 }
  
 interface RelationProps {
      current: Boolean!
 }

 type MasterData {
      id: ID! @id(autogenerate: false)
      current: Boolean!
      nameDetails: NameDetails @relationship(type: "HAS_NAME", properties: "RelationProps", direction: OUT)
 }
```

**To Reproduce**
- Add the following query variables:

```json
{
  "where": {
    "current": true,
    "architectureConnection_SINGLE": {
      "node": {
        "nameDetailsConnection": {
          "node": {
            "fullName": "MHA"
          }
        }
      }
    }
  },
  "connectionWhere": {
    "current": true
  }
}
```

- Execute the following mutation:

```gql
query getSeriesFilteredByArchitectureNameDetails(
  $where: SeriesWhere = { current: true }
  $connectionWhere: RelationPropsWhere = { current: true }
) {
  series(where: $where) {
    id
    architectureConnection(where: { edge: $connectionWhere }) {
      edges {
        node {
          nameDetailsConnection(where: { edge: $connectionWhere }) {
            edges {
              node {
                fullName
              }
            }
          }
        }
      }
    }
  }
}
```

**Expected result**

A single node is returned with the `fullName` property that matches the searched value ("MHA").

**Actual result**

All nodes are returned. If we take a look at the generated Cypher query it's apparent that the filter was not applied:

```cypher
MATCH (this:Series)
[...]

PR #1292, opened 2022-04-22
Fix: Missing filter with deep relations

tbwiss:
# Description

Note the `SINGLE` predicate in the `architectureConnection_SINGLE` connectionField. 
This solution checks, in a recursive manner, if any of the connectionFields contains the SINGLE predicate and if so applies the SINGLE predicate to each subsequent connection. Also it is
using the `expectMultipleValues` boolean of the [`apoc.cypher.runFirstColumn`](https://neo4j.com/labs/apoc/4.0/overview/apoc.cypher/apoc.cypher.runFirstColumn/) function to return the first value if there is a `SINGLE` predicate in any of the connectionFields.
This PR contains a test for two levels deep relations and one for three levels deep relations.

# Issue

 #1221

Closes:

- #1221

# Checklist


packages/graphql/src/translate/where/create-element-where-and-params.ts
```diff
@@ -190,15 +190,22 @@ function createElementWhereAndParams({
                         relationship,
                         relationshipVariable: `${collectedMap}.relationship`,
                         parameterPrefix: `${parameterPrefix}.${fieldName}`,
+                        listPredicates: [currentListPredicate, ...(listPredicates || [])],
```
angrykoala, line 193: **question** I'm not sure what this variable (`listPredicates` ) is, or how it works. Woudl appreciate a quick explanation :sweat_smile:

packages/graphql/src/translate/utils/apoc-run.ts
```diff
@@ -1,11 +1,37 @@
...
+    query: string,
+    params: Record<string, string> = {},
+    expectMultipleValues?: boolean | null
+): string {
     const serializedParams = stringifyObject(params);
     const escapedQuery = escapeQuery(query);
-    return `apoc.cypher.runFirstColumn("${escapedQuery}", ${serializedParams})`;
+    const hasExpectMultipleValues = expectMultipleValues === true || expectMultipleValues === false;
+    return `apoc.cypher.runFirstColumn("${escapedQuery}", ${serializedParams}${
+        hasExpectMultipleValues ? `, ${expectMultipleValues}` : ""
```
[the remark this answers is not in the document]
tbwiss: Good input! I'll make the changes to address this

packages/graphql/src/translate/where/create-element-where-and-params.ts
tbwiss: `listPredicates` will store all list predicates (SINGLE, ANY, NONE,..) while (recursively) translating the where clause of the query. It can then be used in `create-element-where-and-params` to decide which list predicate to use plus if we want to set the `expectMultipleValues` for `apoc.cypher.runFirstColumn` to `false`.
Maybe a better name would be suitable? I couldn't come up with anything better..

angrykoala: mmm, yeah, I think we could find a better name. but tbh I cannot think of much better name.

Maybe, for now, a comment explaining a bit what this is could help

angrykoala (dismissed):

packages/graphql/tests/integration/issues/1221.int.test.ts
```diff
@@ -0,0 +1,694 @@
...
+            variableValues,
+            contextValue: {
+                driver,
+            },
+        });
+
+        expect(res.errors).toBeUndefined();
+
+        expect(res.data).toEqual({
+            [testSeries.plural]: expect.arrayContaining([
```
darrellwarde, line 383: You should either remove this `expect.arrayContaining`, or add a length check for this field to check that the length is exact.

darrellwarde (approved):
Wow, a lot of tests for this! 🥵  A complicated issue to solve indeed. 

But this looks good, just sort the one issue I've pointed out in the tests and then I think this is good to merge!

packages/graphql/tests/integration/issues/1221.int.test.ts
tbwiss: 👍  Will add a length check

PR #1301, opened 2022-04-25
Feature/top level unions

darrellwarde:
Adds Unions to the root Query. 

Given the typeDefs: 

```gql
type Movie {
    title: String!
}

type Series {
    name: String!
}

union Production = Movie | Series
```

You can now query: 

```gql
query {
    productions(where?) {
        ... on Movie {
            title
        }
        ... on Series {
            name
        }
    }
}
````

Based on RFC https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-009-top-level-unions.md

## Notes on implementation
Opted to reuse the `translateRead` function and use individual subqueries over the `union` statement as a subquery is more flexible than a union! There is an extra wrapping `CALL` statement and this is so if one union member MATCH is null the entire query won't exit!

### CODE REVIEW

**Comment by darrellwarde** [comment]
Previous comments in https://github.com/neo4j/graphql/pull/1215

File: packages/graphql/src/classes/AbstractGraphElement.ts [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface AbstractGraphElementConstructor {
+    name: string;
+    description?: string;
+}
+
+export abstract class AbstractGraphElement {
```
angrykoala, line 25: This naming is extremely confusing.

Considering `Union` is the only class that uses this, I think we should get rid of this for now

packages/graphql/src/schema/get-unions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { UnionTypeDefinitionNode } from "graphql";
+import { Node } from "../classes";
+import { Union } from "../classes/Union";
+
+type Unions = {
+    unions: Union[];
+};
+
+function getUnions(unionTypes: UnionTypeDefinitionNode[], options: { nodes: Node[] }): Unions {
```
angrykoala, line 28: Why not returning directly `Union[]`?

```suggestion
function getUnions(unionTypes: UnionTypeDefinitionNode[], options: { nodes: Node[] }): Union[] {
```

packages/graphql/src/schema/resolvers/top-level-union-read.ts
```diff
@@ -0,0 +1,55 @@
...
+import { GraphQLResolveInfo } from "graphql";
+import translateToplevelUnionRead from "../../translate/translate-top-level-union-read";
+import { Context } from "../../types";
+import { execute } from "../../utils";
+import getNeo4jResolveTree from "../../utils/get-neo4j-resolve-tree";
+import { Union } from "../../classes/Union";
+
+export default function topLevelUnionReadResolver({ union }: { union: Union }) {
+    async function resolve(_root: any, args: any, _context: unknown, info: GraphQLResolveInfo) {
+        const context = _context as Context;
```
angrykoala, line 29: Do we need this casting?

```suggestion
    async function resolve(_root: any, args: any, context: Context, info: GraphQLResolveInfo) {
```

packages/graphql/src/schema/resolvers/top-level-union-read.ts
```diff
@@ -0,0 +1,55 @@
...
+            cypher,
+            params,
+            defaultAccessMode: "READ",
+            context,
+        });
+
+        const record = executeResult.records[0];
+        const result = record && Object.entries(record).flatMap(([, members]) => members);
+
+        return result || [];
```
angrykoala, line 45: The types here are a bit wonky, `record` is understood as always truthy, making the rest of the checks non obvious.

I think it will be easier (and closer to the types) if we ensure record is an object since the beggining:

```suggestion
        const record = executeResult.records[0] || {};
        const result = Object.entries(record).flatMap(([, members]) => members);

        return result;
```

packages/graphql/src/translate/translate-read.ts
```diff
@@ -30,14 +30,20 @@ import translateTopLevelMatch from "./translate-top-level-match";
...
     node,
     context,
+    overrideVarName,
+    resolveType,
+    whereInput,
     isRootConnectionField,
 }: {
     context: Context;
     node: Node;
+    overrideVarName?: string;
```
angrykoala, line 40: maybe use a `varName` parameter with default to `"this"` instead?

packages/graphql/src/translate/translate-top-level-union-read.ts
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Union } from "../classes/Union";
+import Node from "../classes/Node";
+import { Context } from "../types";
+import translateRead from "./translate-read";
+
+function translateToplevelUnionRead({ union, context }: { union: Union; context: Context }): [string, any] {
```
angrykoala, line 25:
```suggestion
function translateToplevelUnionRead({ union, context }: { union: Union; context: Context }): [string, Record<string, any>] {
```

packages/graphql/src/translate/translate-top-level-union-read.ts
```diff
@@ -0,0 +1,63 @@
...
+    const returnEntries: string[] = [];
+    let cypherParams = {};
+
+    Object.entries(context.resolveTree.fieldsByTypeName).forEach((entry, i) => {
+        if (entry[0] === union.name) {
+            return;
+        }
+
+        const toNode = context.nodes.find((n) => n.name === entry[0]) as Node;
+        const whereInput = context.resolveTree.args?.where?.[toNode.name] as Record<string, unknown>;
```
angrykoala, line 36: to make sure the types match correctly, we should add `undefined` to these types
```suggestion
        const whereInput = context.resolveTree.args?.where?.[toNode.name] as Record<string, unknown> | undefined;
```

angrykoala (changes requested):

krasish:
Any chance that this feature request will be implemented soon? 👀

darrellwarde:
This has sadly fallen so far out-of-date that we have no option but to close it for now. 😞

PR #1353, opened 2022-05-04
Create rfc-018-mathematical-operators.md

MacondoExpress:
# Description
Contains first draft of rfc-018-mathematical-operators.


docs/rfcs/rfc-018-mathematical-operators.md
```diff
@@ -0,0 +1,100 @@
...
+  update: {
+    counter: { inc: 1 },
+  }
+)
+```
+### Technical considerations
+#### Inconsistent design
+The proposed solution implements a design different from similar features like [https://neo4j.com/docs/graphql/manual/current/filtering/](https://neo4j.com/docs/graphql/manual/current/filtering/),
+and increase the learning curve of the library.
+
```
angrykoala, line 77: Another consideration to keep in mind, is the feasibility of having a scalar and object union type.
i.e., the same input property to support both `counter: 5` and `counter: { INC: 1}`

angrykoala:
Personally, I agree that the alternative solution is nicer, but I'm afraid that it will probably not be feasible, thoughts @darrellwarde @MacondoExpress ?

docs/rfcs/rfc-018-mathematical-operators.md
MacondoExpress: I Agree with that, the whole solution stands on the assumption that the workaround of creating a custom scalar type to be used as a wrapper in the augmented schema is achievable. I add this as a consideration to improve the visibility.

MacondoExpress: Please, let me know if this could be solved by 99701e156480bd92348ed987ab2f12df7611ba18 @angrykoala

docs/rfcs/rfc-018-mathematical-operators.md
```diff
@@ -0,0 +1,126 @@
...
+- `_MULTIPLY`
+- `_DIVIDE`
+
+For instance `counter_ADD` or `counter_SUBTRACT`.
+These properties should be available for all the following types:
+- Int
+- Float
+- BigInt
+
+### Usage Examples
```
tbwiss, line 24: Just for completion's sake (and a bit of nitpicking on my end): Shall we add the typeDefs here as well?

docs/rfcs/rfc-018-mathematical-operators.md
```diff
@@ -0,0 +1,126 @@
...
+As described in the issue #677, if we have to do a mathematical operation using the GraphQL library, 
+the common approach is to query the database and then perform the mutation with the final property value.
+The former approach has several downsides:
+- It requires multiple roundtrips to the database.
+- It's not concurrency safe.
+- It requires more code than necessary in the client.
+
+## Proposed Solution
+Add new input properties that solve atomic operations like:
+- `_ADD`
```
tbwiss, line 13: Oh, it changes from `INC` to `ADD` since yesterday or not? I'm okay with both, but kind of prefer the shortest spelling/reading that is possible.

tbwiss (commented):
Good one @MacondoExpress!

docs/rfcs/rfc-018-mathematical-operators.md
angrykoala: @darrellwarde knows more about graphql limitations, but I think this is not doable, at least not with quite a good bit of hacks

docs/rfcs/rfc-018-mathematical-operators.md
MacondoExpress: Yes, I changed yesterday here 08a80e2c486611ccf062508ff14531fd4903eb64.
I like both long/short versions, Although I think that the DIV operator will be confused for an Integer Division operator with this name, that could be the case anyway. I think @darrellwarde prefer the long version, any takes on that @angrykoala ?

docs/rfcs/rfc-018-mathematical-operators.md
MacondoExpress: I agree :)

MacondoExpress:
Let's assume the  following type definitions:
```graphql
type Node {
  counter: Int
}
```
Assume that the Node with id "e9bc687a-efd1-419d-b208" has `counter` equal to `5`.
Then consider the following mutation:
```graphql
updateNodes(
  where: { id: "e9bc687a-efd1-419d-b208" }
  update: {
    counter_DIVIDE: 2
  }
)
```
At this point, the Node with id "e9bc687a-efd1-419d-b208" should have `counter` equal to 2.5, but now `counter` is of the type Float rather than Int.
Any thoughts on this scenario? @darrellwarde @angrykoala @tbwiss

darrellwarde:
> Let's assume the following type definitions:
> 
> ```graphql
> type Node {
>   counter: Int
> }
> ```
> 
> Assume that the Node with id "e9bc687a-efd1-419d-b208" has `counter` equal to `5`. Then consider the following mutation:
> 
> ```graphql
> updateNodes(
>   where: { id: "e9bc687a-efd1-419d-b208" }
>   update: {
>     counter_DIVIDE: 2
>   }
> )
> ```
> 
> At this point, the Node with id "e9bc687a-efd1-419d-b208" should have `counter` equal to 2.5, but now `counter` is of the type Float rather than Int. Any thoughts on this scenario? @darrellwarde @angrykoala @tbwiss

Oh yeah, this is an interesting scenario! I guess the options are either throw an error (this is difficult, because you will have to do this from Cypher) or round to the nearest whole number... Also not nice.

The _easiest_ option is shall we just leave division out at this stage for integers?

docs/rfcs/rfc-018-mathematical-operators.md
darrellwarde: I think the full words for these is appropriate. I think shortening is good if there is widely accepted convention for it (for example, `_LT`, `_GTE`), but the full words should be used otherwise (for example, `_CONTAINS`, `_STARTS_WITH`). I think these operators fall into the latter group.

docs/rfcs/rfc-018-mathematical-operators.md
```diff
@@ -0,0 +1,126 @@
...
+```graphql
+updateNodes(
+  where: { id: "e9bc687a-efd1-419d-b208" }
+  update: {
+    counter: 10,
+    counter_ADD: 1,
+  }
+)
+```
+`counter` is ambiguous, and a strategy for this situation should be defined.
```
darrellwarde, line 46: Let's be very specific here and specify that we will throw an error.

darrellwarde (changes requested):
Only a small new comment from me, but I will reply with my thoughts to the others!

darrellwarde:
> Personally, I agree that the alternative solution is nicer, but I'm afraid that it will probably not be feasible, thoughts @darrellwarde @MacondoExpress ?

I'm fairly sure the only technically feasible option is the first which is presented.

Even if the alternative solution is possible through some custom scalar, it will feel very unfamiliar for users of GraphQL, so in my opinion should be avoided.

docs/rfcs/rfc-018-mathematical-operators.md
MacondoExpress: Done!

MacondoExpress:
> > Let's assume the following type definitions:
> > ```graphql
> > type Node {
> >   counter: Int
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Assume that the Node with id "e9bc687a-efd1-419d-b208" has `counter` equal to `5`. Then consider the following mutation:
> > ```graphql
> > updateNodes(
> >   where: { id: "e9bc687a-efd1-419d-b208" }
> >   update: {
> >     counter_DIVIDE: 2
> >   }
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > At this point, the Node with id "e9bc687a-efd1-419d-b208" should have `counter` equal to 2.5, but now `counter` is of the type Float rather than Int. Any thoughts on this scenario? @darrellwarde @angrykoala @tbwiss
> 
> Oh yeah, this is an interesting scenario! I guess the options are either throw an error (this is difficult, because you will have to do this from Cypher) or round to the nearest whole number... Also not nice.
> 
> The _easiest_ option is shall we just leave division out at this stage for integers?

I tested the current behavior of the library in the case that a float is returned to a property defined as Int in the type definitions.
What I discovered is that the following error is raised: `GraphQLError: Int cannot represent non-integer value`.
From this starting point, I would say that the round approach is not viable without changing the current behavior of the library.
One approach could be to implement both the DIVISION operator and the INTEGER DIVISION operator, in this case, the Int/BigInt properties could access only the INTEGER DIVISION operator and the Float properties can access both.
Just for visibility, Cypher/Neo4j uses the `/` token as an Integer division if both parameters are of type Integer and the division if at least one is an of type Float,
```cypher
RETURN 10/20 // 0
RETURN 10.0/20 // 0.5
RETURN 10/20.0 // 0.5
RETURN 10.0/20.0 // 0.5
```
or to not implement either of the two.

MacondoExpress:
@darrellwarde  @angrykoala  @tbwiss  there is anything that we should add before merge this?

darrellwarde:
How about for `Int` and `BigInt`:

- `INCREMENT`
- `DECREMENT`

And for `Float`:

- `_ADD`
- `_SUBTRACT`
- `_MULTIPLY`
- `_DIVIDE`

And let's document how this should behave if the mathematical operator performed results in integer overflow.

MacondoExpress:
Anymore for anymore?

docs/rfcs/rfc-018-mathematical-operators.md
```diff
@@ -0,0 +1,167 @@
...
+- `_ADD`
+- `_SUBTRACT`
+- `_MULTIPLY`
+- `_DIVIDE`
+
+available for Float fields.
+
+For instance `counter_ADD` or `counter_SUBTRACT`.
+These properties should be available for all the following types:
+- Int
```
tbwiss, line 27: Just so I don't get it the wrong way around. If I read the text here in the proposed solution I get the understanding  that Int and BigInt should have the operators `_ADD`, `_SUBTRACT`, `_INCREMENT` and `_DECREMENT`?

tbwiss (commented):
Just one last question, then it's good to go from my side.

docs/rfcs/rfc-018-mathematical-operators.md
MacondoExpress: I meant that Int and BigInt fields have only _INCREMENT, _DECREMENT, and  Float fields have only _ADD, _SUBTRACT, _MULTIPLY, _DIVIDE.
Do you think that we can reword this differently to make it more clear?

tbwiss: How about just removing these lines https://github.com/neo4j/graphql/pull/1353/files#diff-dfb3da75e39cfefef29c376e7e57172cbaa318b10748bb7174192cd9cb8d00dbR25-R29 then? Then it only contains the information you wrote in the comment above. Wdyt?

docs/rfcs/rfc-018-mathematical-operators.md
```diff
@@ -0,0 +1,167 @@
...
+```
+`counter` is ambiguous, and an Error will be raised.
+
+#### Schema size
+The proposed solution could increase the size of the augmented schema noticeably.
+
+#### Overflow
+In cases where a mathematical operation leads to an overflow, then an exception must be raised.
+
+## Alternative Solution
```
angrykoala, line 71: We should move this to discarded solutions

docs/rfcs/rfc-018-mathematical-operators.md
MacondoExpress: It makes perfectly sense, thank you for the suggestion!

docs/rfcs/rfc-018-mathematical-operators.md
MacondoExpress: Done

MacondoExpress:
Just added one last consideration about the Infinity type. I put divisions by zero as out of scope as well.

darrellwarde (approved):
This now looks good to me, I think we can start implementing this!

tbwiss (approved):

Issue #1348, opened 2022-05-03
Missing results when querying interface relationships


AccsoSG:
**Describe the bug**
When querying interface relationships, result nodes that do not have such a relationship are filtered out

**Type definitions**

```gql
interface Product {
    productTitle: String!
    releatsTo: [Product!]!
}

type Series implements Product {
    productTitle: String!
    releatsTo: [Product!]!  @relationship(type: "RELATES_TO", direction: OUT, queryDirection: DEFAULT_UNDIRECTED)

    seasons: [Season!]!
}

type Season implements Product {
    productTitle: String!
    releatsTo: [Product!]!  @relationship(type: "RELATES_TO", direction: OUT, queryDirection: DEFAULT_UNDIRECTED)

    seasonNumber: Int
    episodes: [ProgrammeItem!]!
}

type ProgrammeItem implements Product {
    productTitle: String!
    releatsTo: [Product!]!  @relationship(type: "RELATES_TO", direction: OUT, queryDirection: DEFAULT_UNDIRECTED)

    episodeNumber: Int
}
```

**To Reproduce**

1. Run a server with the following code...
2. Execute the following Mutation...

```gql
mutation {
  createProgrammeItems(input: [
    {
      productTitle: "TestEpisode1",
      episodeNumber: 1
    },
    {
      productTitle: "TestEpisode2",
      episodeNumber: 2
    },
    {
      productTitle: "TestFilm1"
    }
  ]) {
    programmeItems {
      productTitle
      episodeNumber
    }
  }
}
```

```gql
mutation {
  updateProgrammeItems(where: {productTitle: "TestFilm1"}, connect: {releatsTo: {where: {node: {productTitle: "TestEpisode1"}}}}) {
    programmeItems {
      productTitle
      episodeNumber
      releatsTo {
        __typename
        productTitle
      }
    }
  }
}
```

4. Then run the following Query...
Query 1:

```gql
query {
  programmeItems {
    productTitle
    episodeNumber
  }
}
```

Result 1:
```json
{
  "data": {
    "programmeItems": [
      {
        "productTitle": "TestEpisode1",
        "episodeNumber": 1
      },
      {
        "productTitle": "TestEpisode2",
        "episodeNumber": 2
      },
      {
        "productTitle": "TestFilm1",
        "episodeNumber": null
      }
    ]
  }
}
```

Query 2:
```gql
query {
  programmeItems {
    productTitle
    episodeNumber
    releatsTo {
      __typename
      productTitle
[...]

PR #1356, opened 2022-05-05
Fix Issue 1348: Missing results when querying interface relationships

tbwiss:
# Description

Wrap the subqueries call (The UNIONs) in an additional CALL to prevent altering the original input (the `this` variable).

# Issue

#1348

Closes:

- #1348

# Checklist


angrykoala (approved):
mmm, It is not too nice to have 2 nested `CALLS` directly next to each other, but I can't think of a better alternative to fix this easily :thinking:

tbwiss:
I agree @angrykoala ! I was testing different ways to solve this in cypher but came up short with anything better than the proposed solution here in the PR.

PR #1361, opened 2022-05-05
Add the ability to utilise context values in callbacks

### LINKED ISSUE #1358: Provide graphql context in @callback directives
It is more related to a use case which I am not able to implement right now.

E.g.: I want to automatically populate a field called `modified by`. The needed information should already be available inside the graphql `context`. However currently only the parent- or root-node is given inside a callback function of the `@callback` directive.

I'd like to have the current graphql `context` as an additional argument to the callback function used by the `@callback` directive.

I don't see any other way with the current implementation. The only other solution I've considered is to atleast enforce the correct value with the usage of the `@auth(bind)` directive.
However the client would then be in need to send the correct value which I don't see as a good solution.

**Additional context**
Providing the current `context` inside the callback function would enhance the callback directive with additional capabilities:
- as opposed to the `@computed` directive one is now able to precompute *advanced* field values instead of computing them at query time
- in combination with the `@private` directive the storage of arbitrary values that should only be available in the backend is possible

darrellwarde:
# Description

Allows you to use the GraphQL context in callbacks, for example:

```gql
const modifiedByCallback = async (_parent, _args, context) => {
    return context.username;
}

new Neo4jGraphQL({
    typeDefs,
    driver,
    config: {
        callbacks: {
            modifiedBy: modifiedByCallback
        }
    }
})
```

# Issue

Closes #1358 

# Checklist

### DIFF
--- a/packages/graphql/src/classes/CallbackBucket.ts
+++ b/packages/graphql/src/classes/CallbackBucket.ts
@@ -47,9 +47,11 @@ export class CallbackBucket {
         await Promise.all(
             this.callbacks.map(async (cb) => {
                 const callbackFunction = (this.context?.callbacks as Neo4jGraphQLCallbacks)[cb.functionName] as (
-                    args?: Record<string, unknown>
+                    parent?: Record<string, unknown>,
+                    _args?: Record<string, unknown>,
+                    context?: Record<string, unknown>
                 ) => Promise<any>;
-                const param = await callbackFunction(cb.parent);
+                const param = await callbackFunction(cb.parent, {}, this.context);
 
                 if (param === undefined) {
                     cypher = cypher
--- a/docs/modules/ROOT/pages/type-definitions/autogeneration.adoc
+++ b/docs/modules/ROOT/pages/type-definitions/autogeneration.adoc
@@ -143,7 +143,7 @@ directive @callback(
 
 === Usage
 
-Type Definitions: 
+Type Definitions:
 
 [source, graphql, indent=0]
 ----
@@ -153,7 +153,7 @@ type Product {
 }
 ----
 
-Schema Construction: 
+Schema Construction:
 
 > Note that the callback is asynchronous!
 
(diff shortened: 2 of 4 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-1361.surge.sh

packages/graphql/src/classes/CallbackBucket.ts
```diff
@@ -47,9 +47,11 @@ export class CallbackBucket {
         await Promise.all(
             this.callbacks.map(async (cb) => {
                 const callbackFunction = (this.context?.callbacks as Neo4jGraphQLCallbacks)[cb.functionName] as (
-                    args?: Record<string, unknown>
+                    parent?: Record<string, unknown>,
+                    _args?: Record<string, unknown>,
```
tbwiss, line 51: Sry for not knowing this, does the leading underscore mean/indicate that the variable is unused?

tbwiss (dismissed):
LGTM!

packages/graphql/src/classes/CallbackBucket.ts
angrykoala: Usually that shuts up our eslint rule of "unused variable" as it is understood that it is required for the function to work, but not used inside. Not sure in this case

darrellwarde: I can just call this `args` if you like - I did this more for us just to show that it's unused.

angrykoala (approved):

AnsonHwang86:
Could you also add the ability to return the value of all field along with the added/updated value?  such as
```
"products": [
				{
					"listings": [
						{
							"id": "8264b791-4sd8-411d-ba32-39f6277a8605",
							"sku": "ssd",
							"sellingPlatform": {
								"sellingPlatformCode": "XX"
							}
                                                 ... all other fields in the node
						}
					]
				}
			]
```
My use case is to concat value between different fields.

tbwiss:
Hi @AnsonHwang86! Thanks for reaching out!
Would you be able to raise a [feature request](https://github.com/neo4j/graphql/issues/new/choose) for this (as we already closed the PR/work for this issue)? You may reference to this PR here in your feature request.
This would help us greatly to keep things in order and we can prioritize it accordingly.
Thank you very much!

PR #1371, opened 2022-05-09
Subscriptions authentication

angrykoala:
# Description

Adds authentication checks to subscriptions, throwing an error to connect if not authenticated.

Support for `isAuthenticated` and `allowUnauthenticated`

This PR depends on https://github.com/neo4j/graphql/pull/1350


packages/graphql/src/types.ts
```diff
@@ -70,10 +70,10 @@ export interface AuthRule extends BaseAuthRule {
     operations?: AuthOperations[];
 }
 
-export type Auth = {
+export interface Auth {
```
angrykoala, line 73: Changed to interface to support implementing it in classes

packages/graphql/src/utils/get-token.ts
```diff
@@ -63,3 +62,11 @@ export function getToken(context: Context): string | undefined {
 
     return token;
 }
+
+export function parseBearerToken(bearerAuth: string): string {
+    const token = bearerAuth.split("Bearer ")[1];
+    if (!token) {
+        debug("Authorization header was not in expected format 'Bearer <token>'");
```
darrellwarde, line 69: **question:** surely this is just going to break down the line if we don't throw and error here?

packages/graphql/tests/e2e/subscriptions/auth/authentication.int.test.ts
```diff
@@ -0,0 +1,583 @@
...
+        let server: TestGraphQLServer;
+        let wsClient: WebSocketTestClient;
+
+        beforeAll(async () => {
+            const typeDefs = `
+            type ${typeMovie} {
+                title: String!
+            }
+
+            extend type ${typeMovie} @auth(rules: [{ isAuthenticated: true, allowUnauthenticated: true, operations: [SUBSCRIBE] }])
```
darrellwarde, line 484: Perhaps out of the scope of this PR, but I feel like combining `isAuthenticated` and `allowUnauthenticated` shouldn't be allowed, and `allowUnauthenticated` should only be allowed to be combined with _authorization_ rules. We should discuss this!

darrellwarde (dismissed):
Sorry for the slow review, this looks good! I think the error message perhaps needs to be a bit clearer, but other than that, just ESLint violations to address.

packages/graphql/src/utils/get-token.ts
angrykoala: Possibly, this logic is actually the same as our current token logic for normal requests (same file, line 58).

I'm not sure what the fallback is for or its implications, but I think that falls under authentication improvements task we have pending to do

packages/graphql/tests/e2e/subscriptions/auth/authentication.int.test.ts
angrykoala: We need to discuss, isn't the point of allow unauthenticated to support whitelisting?

PR #1379, opened 2022-05-09
Add changesets configuration

darrellwarde:
# Description

I propose using [`changesets`](https://github.com/changesets/changesets) to manage our versioning, changelogs and releases. 

You can read more about it in the repository above, but essentially, all it will require is for every branch which contains work which should be released, you run the `changeset` command in the package to be released, which will prompt you for the severity of the version bump, and a changelog entry. At release time, you run `changeset version` which will apply _all_ of the version bumps, and then `changeset publish` which will publish _all_ packages which have been bumped. As you can probably guess, that means we will need just one release workflow to handle _all_ of our releases whenever we decide to release. It should also lead to more readable changelogs because they would have been written with a bit more thought. We will also no longer need to apply labels to PRs. And there's a nice little GitHub bot that can be enabled to prompt for `changesets` entries. 

No rush to merge this, it will require some GitHub Actions changes to move it all into action. Just want to awareness about my intentions and request comments! 🙂


tbwiss:
Sounds like a very handy tool! I was thinking that it's probably easy to forget to add a changelog entry to the branches we need it but the Github bot can help there. I'll have to do some more reading to understand the changeset tool a bit better.

angrykoala (dismissed):

PR #1410, opened 2022-05-11
Subscription roles

angrykoala:
# Description

This PR adds roles behaviour to `auth` directive for subscriptions.

This PR is dependant on #1371


packages/graphql/src/schema/resolvers/subscriptions/subscribe.ts
```diff
@@ -42,6 +40,18 @@ type SubscriptionArgs = {
...
 export function generateSubscribeMethod(node: Node, type: "create" | "update" | "delete") {
     return (_root: any, args: SubscriptionArgs, context: SubscriptionContext): AsyncIterator<[SubscriptionsEvent]> => {
+        if (node.auth) {
+            const authRules = node.auth.getRules(["SUBSCRIBE"]);
+            for (const rule of authRules) {
+                if (!SubscriptionAuth.validateAuthenticationRule(rule, context)) {
+                    throw new Error("Error, request not authenticated");
+                }
+                if (!SubscriptionAuth.validateRolesRule(rule, context)) {
+                    throw new Error("Error, request not authorized");
```
darrellwarde, line 50: **question:** I think safe to perhaps add a bit more detail to this error message, possibly which role was missing?

darrellwarde (approved):
This one looks good to me, just one question about the error message!

packages/graphql/src/schema/resolvers/subscriptions/subscribe.ts
angrykoala: I'm hesitant to add extra information, we haven't deeply thought on the security implications of this :/

PR #1415, opened 2022-05-12
Fix serialization for `[Int!]!`

darrellwarde:
# Description

Prior to this fix, returning `[Int!]!` would result in an error.


packages/graphql/tests/integration/scalars.init.test.ts
```diff
@@ -157,4 +158,43 @@ describe("scalars", () => {
             await session.close();
         }
     });
+
+    test("should serialize a list of integers correctly", async () => {
+        const type = generateUniqueType("Type");
+
+        const typeDefs = `
+            type ${type.name} {
+              integers: [Int!]!
```
angrykoala, line 167: **questions** Is `[Float!]!` tested as well?

packages/graphql/src/schema/resolvers/field/id.ts
```diff
@@ -18,17 +18,34 @@
...
+        return value.map((v) => {
+            if (isNeoInt(v)) {
+                return v.toNumber();
+            }
+
+            if (typeof v === "number") {
+                return v.toString(10);
+            }
+
+            return v;
```
angrykoala, line 38: Consider extracting this into a separate function to remove duplicate code, so `idResolver` looks something like:

```ts
const value = defaultFieldResolver(source, args, context, info);
if (Array.isArray(value)){
  return value.map((v)=> numberResolver(v));
}
return numberResolver(v)
```

angrykoala (approved):
A suggestion to be fixed in idResolver

PR #1427, opened 2022-05-13
Neo place demo

angrykoala:


examples/neo-place/client/qr.html
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
```
tbwiss, line 1: question: Do we still need this file? I assume the `index.html` is enough? The QR code is not really needed (anymore)?

examples/neo-place/client/qr.html
```diff
@@ -0,0 +1,68 @@
...
+<html>
```
tbwiss, line 2: (It's not here)
question: Shall we also delete the QR code png? Do we still need it?

examples/neo-place/secrets.yaml
```diff
@@ -0,0 +1,4 @@
+env_variables:
```
tbwiss, line 1: question: Is there any security downside if we include this file?

examples/neo-place/README.md
```diff
@@ -0,0 +1,25 @@
+# Neo/Place
+
+## Getting Started
+
+1. Setup a .env file with the neo4j credentials:
```
tbwiss, line 5: I think it would be good to mention briefly that there are two options, AMQP and local single instance. And maybe a few words on how to set those two up. I know it's partly mentioned already but I think we can make it a bit easier and clearer for the people who want to try this super cool application themselves. Wdyt?

examples/README.md
```diff
@@ -4,3 +4,4 @@ List of example applications to help you get started with Neo4j + GraphQL.
 
 1. [Migration](./migration) - A small schema which was used in a blog to demonstrate migrating from `neo4j/graphql-js`
 2. [Neo Push](./neo-push) - Example blog site built with Neo4j + GraphQL.
+3. [Neo Place](./neo-place) - Subscriptions Demo.
```
tbwiss, line 7: Wdyt?
```suggestion
3. [neo/place](./neo-place) - Subscriptions Demo.
```

tbwiss (commented):
Nice!!
Left just a few comments.

examples/neo-place/secrets.yaml
angrykoala: yes there are, and my bad for committing this, I'll change the keys and remove this file

examples/neo-place/client/qr.html
tbwiss: Fair enough!
One small thing we could do is adding a comment on the top of this file saying that this file is meant for (live) demonstration purposes or similar. But it also works without.

PR #1475, opened 2022-05-27
Subscription auth where

angrykoala:
This PR adds support for `where` argument for subscriptions.
The behaviour will be the same as for where parameter (i.e. no nested where or operations)

`where` supports `jwt` params.

Behaviour between roles and where is still untested


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-1475.surge.sh

docs/modules/ROOT/pages/auth/subscriptions.adoc
```diff
@@ -23,7 +23,10 @@ Roles can be set for subscriptions. Only requests matching the roles set will be
 NOTE: Not Supported
 
 == Where
-NOTE: Not Supported
+If the rule `where` is set. The provided arguments will be used on xref::subscriptions/filtering.adoc[subscriptions filtering], regardless
```
mjfwebb, line 26:
```suggestion
If the rule `where` is set the provided arguments will be used on xref::subscriptions/filtering.adoc[subscriptions filtering], regardless
```

docs/modules/ROOT/pages/subscriptions/filtering.adoc
```diff
@@ -0,0 +1,39 @@
...
+        title
+    }
+}
+```
+
+This way, only newly created movies with the genre `"Drama"` will trigger events to this subscription.
+
+
+=== Events
+Filtering behaves slightly differently depending on the event type:
```
tbwiss, line 35:
```suggestion
Filtering behaves slightly different depending on the event type:
```

packages/graphql/tests/e2e/subscriptions/auth/where.int.test.ts
```diff
@@ -0,0 +1,217 @@
...
+                        name: "arthur",
+                        surname: "Conan Doyle",
+                    },
+                },
+            },
+        ]);
+        expect(wsClient.errors).toEqual([]);
+    });
+
+    test("where option in auth takes preference over where arg", async () => {
```
tbwiss, line 164: I assume this should be `precedence``
```suggestion
    test("where option in auth takes precedence over where arg", async () => {
```

tbwiss (dismissed):
LGTM @angrykoala!

packages/graphql/src/schema/resolvers/subscriptions/subscribe.ts
```diff
@@ -50,14 +52,28 @@ export function generateSubscribeMethod(node: Node, type: "create" | "update" |
...
+            const whereParams = { ...args.where, ...authWhere };
+            const matchWhereFilter = subscriptionWhere(whereParams, data[0]);
+            const matchUpdateDiffFilter = updateDiffFilter(data[0]);
+
+            return isSubscribedType && matchWhereFilter && matchUpdateDiffFilter;
         });
     };
 }
+
+function getAuthWhereParameters(authRules: AuthRule[], context: SubscriptionContext): Record<string, any> {
```
mjfwebb, line 73: Is it possible to change the function return type so it doesn't use `any`?

packages/graphql/tests/e2e/subscriptions/auth/roles.int.test.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import { Driver } from "neo4j-driver";
-import supertest, { Response } from "supertest";
+import { Response } from "supertest";
```
mjfwebb, line 21:
```suggestion
import type { Response } from "supertest";
```

Can we specify it as a type import?

mjfwebb (commented):
Just a couple of leading questions and they apply to all similar cases. No changes are requested, but I think we would benefit greatly from reducing the use of `any`.

angrykoala:
Branch closed as it is stale

PR #1502, opened 2022-06-01
Fix: Lint issue in GHA workflow for changelog

tbwiss:
# Description

GitHub Actions / actionlint has found a lint issue: https://github.com/neo4j/graphql/runs/6684365189
It is present since a while back.


mjfwebb (approved):
LGTM :)

PR #1504, opened 2022-06-01
Fix: SonarCloud workflow

tbwiss:
# Description

Our SonarCloud tests are currently failing.
See the error message here: https://github.com/neo4j/graphql/runs/6684360781?check_suite_focus=true#step:5:27
Extended the `graphql-plugin-auth` entry with its subdirectory, `plugins/graphql-plugin-auth`, so that the work directory (or projectBaseDir in the workflow file) is specified correctly.


.github/workflows/sonarcloud.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Uppercase package name
         run: |
           lowercase_package=${{ matrix.package }}
-          lowercase_package=$(echo "$lowercase_package" | tr - _)
+          lowercase_package=$(echo "$lowercase_package" | tr - _ | tr -d plugins/)
```
mjfwebb, line 33: Using `tr` to replace a word might have unintended consequences, I recommend changing this to `sed` instead.

mjfwebb (approved):
Nice one :)

Issue #1430, opened 2022-05-14
Mutation on Interface fail to keep one to one constraint


AnsonHwang86:
**Describe the bug**
Mutation on Interface fail to keep one to one constraint

**Type definitions**
```
type ABCE{
  id:ID @id
  name: String
  interface: InterfaceMom @relationship(type:"HAS_INTERFACE", direction:OUT)
}

interface InterfaceMom {
  id:ID @id
  name:String
}

type ChildOne implements InterfaceMom {
  id:ID @id
  name:String
  feathur: String
}

type ChildTwo implements InterfaceMom {
  id:ID @id
  name:String
  sth: String
}
```
**To Reproduce**

1. Run a server with the following code...

2. Execute the following Mutation...

```
mutation createAbces{
	createAbces(input:[
		{
			
			interface:{
				
				create:{
					node:{
						ChildOne:{
							name:"childone name"
						}
						ChildTwo:{
							name:"childtwo name"
						}
					}
				}
			}
		}
	]){
		abces{
			id
			name
			interface{
				id
				name
				__typename
			}
		}
	}
}
```
3. Then run the following Query...

```
{
	"data": {
		"abces": [
			{
				"id": "23a9cdc4-bf9c-46bf-a51d-343d9c7ef3ab",
				"name": null,
				"interface": {
					"id": "2c50f399-1660-4d2d-83c9-f266bd1a0a5f",
					"name": "childone name1",
					"__typename": "ChildOne"
				}
			},
			{
				"id": "23a9cdc4-bf9c-46bf-a51d-343d9c7ef3ab",
				"name": null,
				"interface": {
					"id": "05ced8da-14c9-4639-8be9-062734c262d8",
					"name": "childtwo name1",
					"__typename": "ChildTwo"
				}
			}
		]
	}
}
```

The return data contain two ABCE nodes with same ID. 

**Expected behavior**
It should throw an error, beacause the relationship between ABCE and InterfaceMom is one to one as in scheme.

If I continue to operate update mutation as following

```
mutation ddfs{
	updateAbces(where:{	id:"23a9cdc4-bf9c-46bf-a51d-343d9c7ef3ab"}
		create:{
			interface:{
				node:{ChildOne:{
					name:"childone name2"
				}}
			}
		}
	){
		abces{
			id
			interface{
				id
				name
				__typename
			}
		}
	}
}
```

[...]

PR #1506, opened 2022-06-01
Fix: Mutation on Interface fail to keep one to one constraint

tbwiss:
# Description

This fix adds two validations for the situations outlined in #1430.
- Create: For a one-to-one relationship it should throw an error if more than one node wants to be created: `packages/graphql/src/translate/create-create-and-params.ts`.
- Update: Should not allow adding/creating a second node to an existing one-to-one relationship: `packages/graphql/src/translate/translate-update.ts`.

The wording of the error message can definitely be discussed!

# Issue

Addresses: #1430 

Closes:

- #1430 

# Checklist


darrellwarde (changes requested):
I _think_ there might still be a few cases where this falls over:

- `connect` operations
- `create` operations within a nested update - does the Cypher validation also need go in `create-update-and-params.ts` for the nested case?

We can either address these now or raise them as separate bug reports.

tbwiss:
@darrellwarde with https://github.com/neo4j/graphql/pull/1506/commits/63b830efa169fcdf6cfa53531890eaf42b57f609 I added some additional tests for `connect` and `create`. I'm however not sure if this is the best solution to solve it thou,

packages/graphql/src/translate/create-create-and-params.ts
```diff
@@ -89,6 +89,20 @@ function createCreateAndParams({
...
+                    const createNodeInputIsOfTypeRefNode = !!v.create.node?.[refNode.name];
+                    const createNodeInputKeys = createNodeInputIsOfTypeRefNode
+                        ? Object.keys((v.create.node as any[]) || [])
+                        : [];
+                    const isCreatingMultipleNodesForOneToOneRel = !isInterfaceAnArray && createNodeInputKeys.length > 1;
+                    if (isCreatingMultipleNodesForOneToOneRel) {
+                        throw new Error(
+                            `Relation field "${
+                                relationField.interface?.dbPropertyName || relationField.interface?.fieldName
+                            }" cannot have more than one node linked`
```
darrellwarde, line 102: Just a suggestion about error message wording here, would be good to also have the typename in the error message. 🙂 

```suggestion
                            `Relationship field "${node.name}.${
                                relationField.interface?.dbPropertyName || relationField.interface?.fieldName
                            }" cannot have more than one node linked`
```

darrellwarde (approved):
Apart from my small comment around the error message wording, this is looking really great! Super well tested! 🏆

PR #1513, opened 2022-06-02
RFC: Array methods

mjfwebb:
# Description

Adds an RFC for array methods


tbwiss (commented):
This looks pretty good @mjfwebb!
Added one thing that I thought of.

docs/rfcs/rfc-020-array-methods.md
```diff
@@ -0,0 +1,86 @@
...
+    info {
+      poppedElements: ["Emil"]
+    }
+  }
+}
+```
+
+We could perform a PUSH operation like so, where the update parameter is an array containing the values to be pushed to the end of the array:
+
+```
```
MacondoExpress, line 52:
```suggestion
```graphql
```
Just the formatter/prettifier :)

docs/rfcs/rfc-020-array-methods.md
```diff
@@ -0,0 +1,86 @@
...
+      middleNames
+    }
+  }
+}
+```
+
+### Technical considerations
+
+In an update mutation when the POP method is not used, the `info` object will contain the field `poppedElements` with a value of `[]`.
+
```
MacondoExpress, line 69: A couple of things that I would add as technical considerations are:

- What happens when used on an optional/not defined field.
- What happens when the same query contains another operator/setter for the same field(I would suggest raising an error there!)

MacondoExpress:
Added a couple of comments, but it looks good 🎉 !

docs/rfcs/rfc-020-array-methods.md
mjfwebb: I think I've now addressed your points here @MacondoExpress. If you agree I'll go ahead and resolve this comment.

MacondoExpress: Absolutely! good joob!

tbwiss:
IMHO this looks like a good version of the RFC to get started with the implementation!
Let's discuss this later today and also keep this PR open for visibility. We can also get rid of the draft state of this PR.

MacondoExpress (dismissed):
LGTM 🎉 !

angrykoala:
Considering the feature is already in progress/finished, maybe we should merge this

MacondoExpress (approved):

Issue #1535, opened 2022-06-07
Aliases in query fails when interfaces are used in schema


megatrond:
**Describe the bug**
If using aliases in the graphql query, on a field that points to an interface, the lib generates invalid cypher and throws an error

**Type definitions**
```graphql
type Tenant {
    id: ID! @id
    name: String!
    events: [Event!]! @relationship(type: "HOSTED_BY", direction: IN)
    fooBars: [FooBar!]! @relationship(type: "HAS_FOOBARS", direction: OUT)
}

interface Event {
    id: ID!
    title: String
    beginsAt: DateTime!
}

type Screening implements Event {
    id: ID! @id
    title: String
    beginsAt: DateTime!
}

type Booking implements Event {
    id: ID!
    title: String
    beginsAt: DateTime!
    duration: Int!
}

type FooBar {
    id: ID! @id
    name: String!
}
```

**To Reproduce**

Running the following query:

```graphql
query { 
  tenants {
    id
    name
    events232: events {
      id  
    }
  }
}
```

This instead returns an error:

```
Variable `events232` not defined (line 16, column 38 (offset: 396))\n\"RETURN this { .id, .name, events232: events232 } as this\"
```

**System (please complete the following information):**

* OS: macOS
* Version: @neo4j/graphql@3.2.0,
* Node.js version: 14

**Additional context**
If querying in the same way for `fooBars` this error is not thrown.

neo4j-team-graphql:
Many thanks for raising this bug report @megatrond. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @megatrond! :pray: We will now prioritise the bug and address it appropriately.

megatrond:
There's a similar, but distinct, bug to this, also involving interfaces (but not aliases).

Do you prefer I file a second issue, or add the schema+query in this issue?

tbwiss:
@megatrond, raising a second issue/bug report would be great. Thank you!

tbwiss:
Debugging this

With alias present:
```
Query:
{
  tenants {
    id
[...]

PR #1556, opened 2022-06-10
Fix: Aliases in query fails when interfaces are used in schema

tbwiss:
# Description

Use the field's name, if present, rather than the field's alias as the data value, see `packages/graphql/src/translate/create-projection-and-params.ts`.

# Issue

Regarding: #1535

Closes:

- #1535

# Checklist

### DIFF
--- a/packages/graphql/src/translate/create-projection-and-params.ts
+++ b/packages/graphql/src/translate/create-projection-and-params.ts
@@ -403,7 +403,7 @@ function createProjectionAndParams({
                     }
                 }
 
-                res.projection.push(`${field.alias}: ${field.alias}${offsetLimitStr}`);
+                res.projection.push(`${field.alias}: ${field.name || field.alias}${offsetLimitStr}`);
 
                 return res;
             }
--- a/packages/graphql/tests/integration/issues/1535.int.test.ts
+++ b/packages/graphql/tests/integration/issues/1535.int.test.ts
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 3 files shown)


packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -403,7 +403,7 @@ function createProjectionAndParams({
                     }
                 }
 
-                res.projection.push(`${field.alias}: ${field.alias}${offsetLimitStr}`);
+                res.projection.push(`${field.alias}: ${field.name || field.alias}${offsetLimitStr}`);
```
darrellwarde, line 406: This is interesting, because `field.alias` defaults to `field.name` if an alias has not been set! 😕

darrellwarde (dismissed):
Slightly confused by this one, but it clearly works as per the tests!

packages/graphql/src/translate/create-projection-and-params.ts
darrellwarde: Actually, let's think about this. As far as I'm aware, `field.name` is _always_ populated - so the `||` here will never be reached - it might be worthwhile looking into this a bit more.

tbwiss: Yes, good point. Let me have a closer look at this.

tbwiss: You're right @darrellwarde in regards to `field.alias` defaulting to `field.name` if there is no alias specified!
I dropped the fallback to `field.alias` in https://github.com/neo4j/graphql/pull/1556/commits/5d5a0def5cf8a9d5577ebe43f1456372b4ddf723.
Reason: In the case of the result projection, we always use the `field.name` as the _value_ since we use the `field.name` as the variable name in the foregoing `CALL subqueries`, see the `events` variable in the [tck test](https://github.com/neo4j/graphql/pull/1556/files#diff-4d3c520cf33cd25706cdc78896581b8e1bd9985be6ebc8dd73e947f9144f7af1R90-R96).

darrellwarde (approved):
The simplest change in recorded history, nice tests! 🏅

PR #1563, opened 2022-06-13
Fix: Correctly translate Relay IDs when Id Field is an Int

### LINKED ISSUE #1560: Relay Ids not translating properly when underlying field is an integer
**Describe the bug**

If using the relay generated `id` in a `where` argument, and the underlying id field is an integer, the db does not find the node. This is likely due to the id being translated as a string rather an integer. 

```graphql
type Organization {
  dbId: Int! @id(autogenerate: false, global:true) @alias(property: "id")
  name: String!
}


# create an entry 

mutation {
  createOrganizations(input: [{ dbId: 12345, name: "test" }]) {
    organizations {
       # <----- make sure and request the relay id in the response
       id
       dbId
       name
     }
  }
}

# then query for the organization using the return relayId

query Organizations {
  organizations(where: { node: { id:  "id-from-above" } }) {
     id
     dbId
     name
  }
}

```

**Screenshots**

The resulting cypher renders the param as a string rather an integer, so the lookup fails. Screenshot is an example from a project -- the "affinityId" is an Int field (like the dbId above)

image

* OS: mac
* Version:  3.3.0
* Node.js version:  lts/fermium

litewarp:
# Description

Fixes #1560 - now the fromGlobalId function converts the string back to an int if that is the underlying Id field type.

Please merge asap!

# Checklist

- [X] Documentation has been updated


tbwiss (approved):
LGTM!

PR #1583, opened 2022-06-15
Remove support for neo4j 4.2.*

angrykoala:
# Description
This PR removes tests for Neo4j 4.2.* and updates documentation

PR #1584, opened 2022-06-15
Fix update subscriptions dev

angrykoala:
This fix removes all references to meta in the updates query itself. If subscriptions are set, it will wrap the whole block in a call statement, with the pre and post assignments of the meta variables

This PR depends on #1583


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-1584.surge.sh

packages/graphql/tests/tck/tck-test-files/subscriptions/update.test.ts
```diff
@@ -73,7 +73,12 @@ describe("Subscriptions metadata on update", () => {
             MATCH (this:Movie)
             WHERE this.id = $this_id
             WITH this { .* } AS oldProps, this, meta
-            SET this.id = $this_update_id
+            CALL {
+            	WITH *
+            	SET this.id = $this_update_id
+            	RETURN meta as update_meta
+            }
+            WITH *, update_meta as meta
```
darrellwarde, line 81: Is there any value in this? Could it not have just been:

```
SET
SET
SET
WITH this, meta + { ...meta }
```

I guess I'm asking, was this possible without the subquery?

darrellwarde (approved):
It looks good, I just have a question around the choice of using a subquery here, but it's an uneducated question!

packages/graphql/tests/tck/tck-test-files/subscriptions/update.test.ts
angrykoala: The problem without subquery is that oldProps will pollute the nested queries, the elegant thing about `CALL { WITH *` is that it completely encapsulates subqueries

darrellwarde: I have no major objection, just a curiosity thing!

PR #1586, opened 2022-06-15
Subscriptions fixes

angrykoala:
# Description
This PR sets the maximum event listeners to infinite in subscriptions plugins to avoid warning on events leak (this is expected behaviour). Proper cleanup tests added as well


packages/graphql/tests/e2e/subscriptions/single-instance-plugin.e2e.test.ts
```diff
@@ -0,0 +1,174 @@
...
+        await wsClient2.close();
+    });
+
+    afterAll(async () => {
+        await server.close();
+        await driver.close();
+    });
+
+    // NOTE: This test **may** be flaky, if so, feel free to remove it
+    test("listeners are properly triggered and cleaned up", async () => {
```
tbwiss, line 89: question: Would adding a delay of 100ms before checking the listenerCount make the test less flaky?

packages/graphql/tests/e2e/subscriptions/single-instance-plugin.e2e.test.ts
```diff
@@ -0,0 +1,174 @@
...
+                [typeMovie.operations.subscribe.created]: {
+                    [typeMovie.fieldNames.subscriptions.created]: { title: "movie2" },
+                    event: "CREATE",
+                    timestamp: expect.any(Number),
+                },
+            },
+        ]);
+    });
+
+    async function createMovie(title: string): Promise<Response> {
```
tbwiss, line 157: question (absolutely non-blocking): Do you think there is any value in reusing this function for all the E2E subscription tests (delete, update, create)? It seems to be present in most of the E2E tests in a very similar form.

packages/graphql/tests/e2e/subscriptions/single-instance-plugin.e2e.test.ts
```diff
@@ -0,0 +1,174 @@
...
+            {
+                [typeMovie.operations.subscribe.created]: {
+                    [typeMovie.operations.subscribe.payload.created]: { title: "movie1" },
+                    event: "CREATE",
+                    timestamp: expect.any(Number),
+                },
+            },
+            {
+                [typeMovie.operations.subscribe.created]: {
+                    [typeMovie.fieldNames.subscriptions.created]: { title: "movie2" },
```
tbwiss, line 149: question: Shouldn't this also be `[typeMovie.operations.subscribe.payload.created]: `?

tbwiss (approved):
LGTM @angrykoala!

packages/graphql/tests/e2e/subscriptions/single-instance-plugin.e2e.test.ts
angrykoala: The flakiness should be fixed by the 50ms delay that is already in place (the comment refers to those). In my tests, I didn't see any problem in the first check

packages/graphql/tests/e2e/subscriptions/single-instance-plugin.e2e.test.ts
angrykoala: I thought about that, but in the end this is a concrete query, dependant on the test, it already reuses all the "make a query" logic, so I think is better to leave it as-is for now

packages/graphql/tests/e2e/subscriptions/single-instance-plugin.e2e.test.ts
angrykoala: Good question. mmm, I think we have a bit of duplicate code here. Both seem to be exactly the same. I think the alternative you propose is better, so I will remove the "fieldNames" property and change all references

PR #1587, opened 2022-06-15
Compatibility checks which are compatible with 5.0

darrellwarde:
# Description

Removes the usage of `dbms.functions()` and `dbms.procedures()` - note, this is dependent on #1583!

Broken out into three separate calls to the database - `SHOW FUNCTIONS` and `SHOW PROCEDURES` can not be combined into Cypher queries like their predecessors. However, this is done with 3 parallel calls.


tbwiss (approved):
LGTM!
The Neo4j 4.2 tests are failing as expected.

PR #1595, opened 2022-06-16
Upgrade TS config base to `@tsconfig-node14`

darrellwarde:
# Description

As per title with deprecation of Node.js 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
         "prepare": "husky install"
     },
     "devDependencies": {
-        "@tsconfig/node12": "1.0.11",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
         "prepare": "husky install"
     },
     "devDependencies": {
-        "@tsconfig/node12": "1.0.11",
+        "@tsconfig/node14": "^1.0.3",
```
tbwiss, line 27: question (non-blocking): Do we need the caret(^) in this case here? All other (dev) dependencies do not have one.

tbwiss (approved):
👍

PR #1603, opened 2022-06-20
Switch multi DB checks to `driver.supportsMultiDb()` to avoid our own parsing of errors

darrellwarde:
# Description

As per the title, continued work towards 5.0 compatibility.


tbwiss (commented):
The code looks good @darrellwarde!
The only issue is that on Neo4j 4.3 `point.distance()` is an unknown function (it's working on 4.4).

tbwiss (dismissed):
LGTM!

packages/graphql/tests/integration/directives/unique.int.test.ts
```diff
@@ -24,36 +24,28 @@ import { gql } from "apollo-server";
 import neo4j from "../neo4j";
 import { Neo4jGraphQL } from "../../../src/classes";
 import { generateUniqueType } from "../../utils/graphql-types";
```
darrellwarde, line 26:
```suggestion
import { generateUniqueType } from "../../utils/graphql-types";
import { delay } from "../../../src/utils/utils";
```

PR #1610, opened 2022-06-21
Fix: GHA doc and UI publish

tbwiss:
# Description

Cleanup PR number usage in docs and UI workflows to follow the commonly used `github.event.pull_request.number`.

### DIFF
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -49,7 +49,7 @@
         "@neo4j/graphql-plugin-auth": "^1.0.0",
         "@types/deep-equal": "1.0.1",
         "@types/is-uuid": "1.0.0",
-        "@types/jest": "28.1.2",
+        "@types/jest": "28.1.3",
         "@types/jsonwebtoken": "8.5.8",
         "@types/node": "16.11.41",
         "@types/pluralize": "0.0.29",
--- a/docs/rfcs/rfc-006-auth-plugin.md
+++ b/docs/rfcs/rfc-006-auth-plugin.md
@@ -80,7 +80,7 @@ const neoSchema = new Neo4jGraphQL({
     typeDefs,
     driver,
     plugins: {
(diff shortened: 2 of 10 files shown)


packages/graphql/package.json
```diff
@@ -99,4 +99,4 @@
         "graphql": "^16.0.0",
         "neo4j-driver": "^4.1.0"
     }
-}
+}
```
mjfwebb, line 102: I think that this change can be removed

.github/workflows/ui-stage-build.yml
```diff
@@ -19,10 +19,12 @@ jobs:
         with:
           node-version: lts/*
       - name: UI build
+        env:
+          PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
         run: |
           yarn
           yarn build
-          echo ${{ github.event.number }} > ./dist/prnumber
+          echo $PULL_REQUEST_NUMBER > ./build/prnumber
```
mjfwebb, line 27: The command probably needs to use `dist` instead of `build`

mjfwebb (approved):
I think this looks good!

.github/workflows/docs-build.yml
```diff
@@ -20,10 +20,12 @@ jobs:
         with:
           node-version: lts/*
       - name: Build documentation
+        env:
+          PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
```
darrellwarde, line 24: Does loading it into an environment variable here offer any value, given that this can be modified by external contributors anyway?

darrellwarde (approved):
It looks good, but I wonder whether the modifications to the build workflows are actually necessary.

.github/workflows/docs-build.yml
mjfwebb: I think you're probably right here, but it could be good to start the practice of always using `env` to automatically interpolate input values.

PR #1620, opened 2022-06-22
Feat: Jest shard integration tests

mjfwebb:
# Description

Add sharding to integration tests to the pull-requests workflow.


tbwiss (approved):

.github/workflows/pull-requests.yml
```diff
@@ -196,6 +196,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        shard: [1, 2, 3, 4]
```
darrellwarde, line 199: Good to use the same list formatting in our workflows. Also, worth keeping shard configuration in one place rather than shard number being up here and maximum number of shards being on the run line?

```suggestion
        shard: 
          - 1/4
          - 2/4
          - 3/4
          - 4/4
```

.github/workflows/pull-requests.yml
```diff
@@ -233,7 +234,7 @@ jobs:
         run: yarn up --exact graphql@${{ matrix.graphql-version }}
       - name: Run @neo4j/graphql integration tests and move coverage
         run: |
-          yarn test:int --coverage
+          yarn test:int --shard=${{ matrix.shard }}/4 --coverage
```
darrellwarde, line 237: With the suggestion from above - what do you think?

```suggestion
          yarn test:int --shard=${{ matrix.shard }} --coverage
```

darrellwarde (approved):
It looks good, but I have a small suggestion for a tweak! Have you tested that this works locally? 👀

darrellwarde:
Also, I have approved this in principal, but it doesn't seem to be working... 👀

mjfwebb:
> Also, I have approved this in principal, but it doesn't seem to be working... eyes
> 
> EDIT: Accidentally hit close, sorry! laughing

I've had to rewrite it for the cases where we don't have enough tests to divide them into four shards (for example, two tests can't be divided into four), so instead each package has specified sharding.

tbwiss (approved):
Good changes @mjfwebb!

Issue #517, opened 2021-09-28
Using "_NOT" on an empty relationship in where clause returns false, leaving out the queried node that doesn't have any relationships


jrsperry:
The "_NOT" operator on relationships requires the relationship to exist. Using "_NOT" in a where clause on a relationship therefore leaves out the object you are querying for. 

For example, you have a graph with 2 people, 1 of which has a child. When you look for people who don't have a child with a particular name, people who don't have any children are always excluded from the result set. Looking at the cypher executed in the logs this appears to occur because the cypher predicate "NONE" is used, which requires the relationship to exist. Not sure if this is a cypher or graphql bug then. 

neo4j version: 4.3.2
neo4j-graphql version: 2.2.0

Steps to reproduce: 

Create graph: 

```
create (p:Person {name: "Josh"})-[:HAS_CHILD]->(c:Child {name: "Bob"})
create (p2:Person {name: "Sue"})
```

Graphql Schema: 

```graphql
type Person {
  name: String
  children: [Child] @relationship(type: "HAS_CHILD", direction: OUT)
}

type Child {
  name: String
}
```

Graphql query, I would expect to see 2 persons returned, instead only 1 is returned (the one with a child). 

```graphql
{
  people(where: {children_NOT: {name: "Adam"}}){
    name
  }
}
```

![image](https://user-images.githubusercontent.com/43385427/135132322-1fe2751f-0839-45ed-ab6b-9f6707fc4c72.png)

But when you query for people, without a filter, 2 people are return, and an empty array on the "children" property. 

![image](https://user-images.githubusercontent.com/43385427/135132502-90b864b4-6f2d-4969-8fb0-5b9c93d3c564.png)

PR #1629, opened 2022-06-23
Fix: indexes not being used when filtering on properties of related nodes

### LINKED ISSUE #1628: Filtering on properties of related nodes does not utilise indexes
**Describe the bug**

Reported internally on Trello.

darrellwarde:
# Description

Using list predicates (`any()`, etc.) does not trigger the use of indexes for filtering, which makes filtering very slow.

This PR completely changes the filtering against related node properties primarily by trying to use the `EXISTS` clause where it is possible to do so. Where this is not possible, `size()` has been used, but this should be changed when filtering is no longer performed in projections.

# Issue

Closes #1628.

Please find details on internal Trello card.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -105,10 +105,15 @@ jobs:
 
     strategy:
       matrix:
-        package:
-          - graphql
-          - ogm
-          - introspector
+        packages:
+          [
+            { package: "ogm", shard: 1/1 },
+            { package: "introspector", shard: 1/1 },
+            { package: "graphql", shard: 1/4 },
+            { package: "graphql", shard: 2/4 },
+            { package: "graphql", shard: 3/4 },
+            { package: "graphql", shard: 4/4 },
+          ]
         neo4j-version:
           - 4.3.2
           - 4.3
@@ -142,9 +147,9 @@ jobs:
         run: yarn up --exact graphql@${{ matrix.graphql-version }}
       - name: Run @neo4j/graphql integration tests and move coverage
         run: |
-          yarn test:int --coverage
+          yarn test:int --shard=${{ matrix.packages.shard }} --coverage
           mv coverage coverage-${{ matrix.neo4j-version }}
-        working-directory: packages/${{ matrix.package }}
+        working-directory: packages/${{ matrix.packages.package }}
         env:
           NEO_USER: neo4j
           NEO_PASSWORD: testpassword
@@ -152,8 +157,8 @@ jobs:
       - name: Archive coverage report
         uses: actions/upload-artifact@v3
         with:
-          name: integration-coverage-${{ matrix.package }}
-          path: packages/${{ matrix.package }}/coverage-${{ matrix.neo4j-version }}/
+          name: integration-coverage-${{ matrix.packages.package }}
+          path: packages/${{ matrix.packages.package }}/coverage-${{ matrix.neo4j-version }}/
 
   toolbox-tests:
     uses: ./.github/workflows/reusable-toolbox-tests.yml
@@ -201,10 +206,15 @@ jobs:
 
     strategy:
       matrix:
-        package:
-          - graphql
-          - ogm
-          - introspector
+        packages:
+          [
+            { package: "ogm", shard: 1/1 },
+            { package: "introspector", shard: 1/1 },
+            { package: "graphql", shard: 1/4 },
+            { package: "graphql", shard: 2/4 },
+            { package: "graphql", shard: 3/4 },
+            { package: "graphql", shard: 4/4 },
+          ]
         aura-instance:
           - free
           - professional
@@ -227,9 +237,9 @@ jobs:
           echo "AURA_PASSWORD_SECRET=AURA_${lowercase_instance^^}_PASSWORD" >>"${GITHUB_ENV}"
       - name: Run @neo4j/graphql integration tests
         run: |
-          yarn test:int --coverage
+          yarn test:int --shard=${{ matrix.packages.shard }} --coverage
           mv coverage coverage-aura-${{ matrix.aura-instance }}
-        working-directory: packages/${{ matrix.package }}
+        working-directory: packages/${{ matrix.packages.package }}
         env:
           NEO_USER: neo4j
           NEO_PASSWORD: ${{ secrets[env.AURA_PASSWORD_SECRET] }}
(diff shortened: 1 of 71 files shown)


packages/graphql/src/translate/where/create-where-and-params.ts
```diff
@@ -227,16 +218,29 @@ function createWhereAndParams({
                     whereInput: entry[1] as any,
                     context,
                     node: refNode,
-                    nodeVariable: `${collectedMap}.node`,
+                    nodeVariable: thisParam,
                     relationship,
-                    relationshipVariable: `${collectedMap}.relationship`,
+                    relationshipVariable: relationshipVariable,
```
darrellwarde, line 223:
```suggestion
                    relationshipVariable,
```

tbwiss (commented):
NIce changes @darrellwarde! Left a few comments to consider

packages/graphql/tests/tck/tck-test-files/issues/index-issues.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        const result = await translateQuery(neoSchema, query);
+
+        expect(formatCypher(result.cypher)).toMatchInlineSnapshot(`
+            "MATCH (this:\`frbr__Work\`:\`Resource\`)
+            WHERE EXISTS { (this)-[:dcterms__title]->(this_dcterms__title:\`dcterms_title\`:\`property\`) WHERE this_dcterms__title.value CONTAINS $this_dcterms__title_value_CONTAINS }
+            RETURN this { iri: this.uri, dcterms__title: [ (this)-[:dcterms__title]->(this_dcterms__title:\`dcterms_title\`:\`property\`)  WHERE this_dcterms__title.value CONTAINS $this_dcterms__title_value_CONTAINS | this_dcterms__title { .value } ] } as this
+            LIMIT $this_limit"
+        `);
+    });
```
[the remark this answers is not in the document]
darrellwarde: Good spot, thanks!

packages/graphql/tests/tck/tck-test-files/advanced-filtering.test.ts
```diff
@@ -730,12 +730,15 @@ describe("Cypher Advanced Filtering", () => {
...
             expect(formatCypher(result.cypher)).toMatchInlineSnapshot(`
                 "MATCH (this:Movie)
-                WHERE exists((this)-[:IN_GENRE]->(:Genre)) AND any(this_genresConnection_Genre_map IN [(this)-[this_genresConnection_Genre_MovieGenresRelationship:IN_GENRE]->(this_genresConnection_Genre:Genre)  | { node: this_genresConnection_Genre, relationship: this_genresConnection_Genre_MovieGenresRelationship } ] WHERE this_genresConnection_Genre_map.node.name = $this_movies.where.genresConnection.node.name)
+                WHERE size([(this)-[this_genresConnection_Genre_MovieGenresRelationship:IN_GENRE]->(this_genresConnection_Genre:Genre) WHERE this_genresConnection_Genre.name = $this_movies.where.genresConnection.node.name | 1]) > 0
                 RETURN this { .actorCount } as this"
             `);
 
             expect(formatParams(result.params)).toMatchInlineSnapshot(`
                 "{
+                    \\"node\\": {
```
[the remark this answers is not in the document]
darrellwarde: That is a great spot, I'll look into this tomorrow!

packages/graphql/src/translate/where/create-element-where-and-params.ts
```diff
@@ -110,22 +111,13 @@ function createElementWhereAndParams({
...
 
+                const matchPattern = `(${varName})${inStr}${relTypeStr}${outStr}(${relatedNodeVariable}${labels})`;
+
                 if (value === null) {
-                    res.clauses.push(
-                        `${isNot ? "" : "NOT "}exists((${varName})${inStr}${relTypeStr}${outStr}(:${
-                            relationField.typeMeta.name
-                        }))`
-                    );
+                    res.clauses.push(`${isNot ? "" : "NOT "}EXISTS { ${matchPattern} }`);
```
angrykoala, line 117: Nice

packages/graphql/src/translate/where/create-where-and-params.ts
```diff
@@ -227,16 +218,29 @@ function createWhereAndParams({
...
-                resultStr += connectionWhere[0];
-                resultStr += ")"; // close ALL
-                res.clauses.push(resultStr);
+                if (connectionWhere[0]) {
+                    const clause = listPredicateToSizeFunction(listPredicate, matchPattern, connectionWhere[0]);
+                    res.clauses.push(clause);
+                    res.params = { ...res.params, ...connectionWhere[1] };
+                }
+
+                // TODO: Change to EXISTS when filtering no longer in projection
```
angrykoala, line 234: **question**
Is it worth keeping this commented code?

packages/graphql/src/translate/where/list-predicate-to-clause.ts
```diff
@@ -0,0 +1,45 @@
...
+            clause = `EXISTS { ${matchPattern} WHERE ${where} }`;
+            break;
+        case "none":
+            clause = `NOT EXISTS { ${matchPattern} WHERE ${where} }`;
+            break;
+        case "single":
+            // Sadly no expressive way to check for single match
+            clause = `size([${matchPattern} WHERE ${where} | 1]) = 1`;
+            break;
+        // no default
```
angrykoala, line 41: Maybe we should add a default with a throw?
Just in case `listPredicate` is not on of the returned values, to have a fast fail

angrykoala (approved):

packages/graphql/src/translate/where/create-where-and-params.ts
darrellwarde: My mistake!

packages/graphql/src/translate/where/list-predicate-to-clause.ts
darrellwarde: I just really hate doing this because the `ListPredicate` type only specifies these four values - it may as well be a string if we add a default case. I believe this is just an ESLint thing, I would rather have a discussion about the value of the rule.

tbwiss (approved):

jrsperry:
This is very nice news to me. I had this issue opened a while back where I felt that _NOT (and more modernly _NONE) should return true if that relation doesn't exist. By switching to the EXISTS predicate instead of ANY or NONE on the filter the logic has changed and now would return true if the relationship is empty. 

Here's a link to the original issue: https://github.com/neo4j/graphql/issues/517

PR #1679, opened 2022-07-05
Remove runFirstColumn from aggregation strings

angrykoala:
This PR change removes runFirstColumn from aggregations of strings, going from:
```
MATCH (this:Movie)
RETURN { title: { shortest:
reduce(shortest = collect(this.title)[0], current IN collect(this.title) | apoc.cypher.runFirstColumn("
	RETURN
	CASE size(current) < size(shortest)
	WHEN true THEN current
	ELSE shortest
	END AS result
", { current: current, shortest: shortest }, false))
  }
}
```
to 
```
MATCH (this:Movie)
RETURN { title: { shortest:
reduce(shortest = collect(this.title)[0], current IN collect(this.title) | 
	CASE size(current) < size(shortest)
	WHEN true THEN current
	ELSE shortest
	END
	)
}} as this
```

This change also improves performance on aggregations queries noticeable. performance stats updated


MacondoExpress:
LGTM!

MacondoExpress (approved):

MacondoExpress (approved):

darrellwarde (approved):
Excellent change, love these quick wins!

PR #1682, opened 2022-07-06
Update performance stats

angrykoala:
This PR updates performance stats generated by `yarn performance -u`


angrykoala:
This is only a stats file, will merge without review

PR #1684, opened 2022-07-06
Add support for cypher queries in performance tests

angrykoala:
# Description
This PR allows running raw cypher queries in the performance test environment using `yarn performance --cypher`. This is useful for comparison with our queries mainly


tbwiss (approved):
LGTM @angrykoala!
Just one comment to consider

PR #1689, opened 2022-07-06
Fix: database execution is not using transaction functions

### LINKED ISSUE #1688: Database execution is not using transaction functions
**Describe the bug**

Raised by a customer, details in Trello.

darrellwarde:
# Description

#953 inadvertently removed a useful feature provided by using the driver's transaction function, which retry a query a number of times in the event of execution failure. This PR reintroduced this functionality, with a refactor breaking out execution code into an `Executor` class.

This should be reviewed, merged and released with fairly high priority.

# Issue

Closes #1688.


packages/graphql/src/classes/Executor.ts
```diff
@@ -0,0 +1,215 @@
+/*
```
tbwiss, line 1: question: Do we have to make any updates/additions to the documentation?

packages/graphql/src/classes/Executor.ts
```diff
@@ -0,0 +1,215 @@
...
+                const result = await this.sessionRun(query, parameters, defaultAccessMode, session);
+                await session.close();
+                return result;
+            }
+
+            if (isSessionLike(this.executionContext)) {
+                return await this.sessionRun(query, parameters, defaultAccessMode, this.executionContext);
+            }
+
+            return await this.transactionRun(query, parameters, this.executionContext);
```
tbwiss, line 116: question: what about calling `transaction.commit()` after `transaction.run()`? Is that still needed or is that a thing of the past?
(this maybe shows more that I'm not quite sure how it all works but I thought of bringing it up anyway)

packages/graphql/src/classes/Executor.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export type ExecutorConstructorParam = {
+    executionContext: ExecutionContext;
+    auth: AuthContext;
+    queryOptions?: CypherQueryOptions;
+    database?: string;
+    bookmarks?: string | string[];
+};
+
+export class Executor {
```
tbwiss, line 82: Nice refactor!

packages/graphql/src/classes/Executor.ts
darrellwarde: We do not - these changes were documented in #953, and this is just fixing some of the underlying implementation. 🙂

packages/graphql/src/classes/Executor.ts
darrellwarde: So this is a case of when a user has passed us their own transaction, which they want to be in control of - so they will have to perform the commit after the execution is complete.

packages/graphql/src/classes/Executor.ts
```diff
@@ -0,0 +1,215 @@
...
+        return {
+            metadata: {
+                app,
+                type: "user-transpiled",
+            },
+        };
+    }
+
+    private async sessionRun(query: string, parameters, defaultAccessMode, session): Promise<QueryResult> {
+        const result = await session[`${defaultAccessMode.toLowerCase()}Transaction`](
```
angrykoala, line 193: **nitpick, non-blocking** for readability purposes, maybe extract `${defaultAccessMode.toLowerCase()}Transaction` into a separate varaible "transactionName" or something like that

angrykoala (approved):
Nice refactor

packages/graphql/src/classes/Executor.ts
darrellwarde: Good shout!

PR #1694, opened 2022-07-07
RFC String comparators and Optional features

MacondoExpress:
# Description
RFC for the String comparator operators and for the Optional Features


docs/rfcs/rfc-019-optional-features.md
```diff
@@ -0,0 +1,68 @@
...
+The Config solution is to extend the `Neo4jGraphQLConfig` to support the list of features to be included with the `features` attribute.
+
+### Usage Examples
+```javascript
+  const neoSchema = new Neo4jGraphQL({
+        config: {
+            enableDebug: true,
+            features: {
+              stringComparator: { enabled: true, settings: {...} }
+            }
```
darrellwarde, line 17: I was personally thinking something more along the lines of:

```suggestion
            features: {
              filters: {
                String: {
                  GT: true,
                  LT: true,
                }
              }
            }
```

The would make it significantly more extensible, and more expressive in my opinion, for example:

```
            features: {
              filters: {
                String: {
                  GT: true,
                  LT: true,
                  CONTAINS: false
                }
                Point: {
                  DISTANCE: false
                }
              }
            }
```

MacondoExpress: Thank you for the suggestion! I really like the idea to have the common interface `filters` in the settings, so that could be used also retroactively with the current operators. I will not include that level of details like `GT: true`, as then the implementation will have to add a check for any comparators used, rather than using a more generic check. In my opinion, the cases where a user needs `GT` but not `LT` are rare, and even more considering that in a case when they discover later that they need that, they'll have to regenerate the schema without being able to modify it at runtime. 
What about something like this:
```javascript
features: {
  filters: {
    String: {
      COMPARATORS: true,
      STRING_FUNCTIONS: false
    }
    Point: {
      DISTANCE: false
    }
  }
}
```
Thoughts?

tbwiss: The granularity is a good question. I agree with Simone, it seems unlikely that a user wants to allow `LT` but not for instance `LTE` or `GT`. The middle ground would be for instance a `COMPARATORS_LT` and `COMPARATORS_GT` setting but that probably does not add too much.

angrykoala: While I agree on those points, it is important to take into account that a user may not intuitively know what a "Comparator" or a "String_function" is

Using the same names as the query itself makes it obvious. Do you want to use GT? enable the GT filter.

docs/rfcs/rfc-020-string-comparator-operators.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+that could be then translated to a Cypher like:
+```cypher
+MATCH(m:Movie)
+WHERE m.title STARTS WITH "The Matrix" AND m.title > "The Matrix"
+RETURN m.title 
+```
+
+Assuming that `The Matrix`, `The Matrix 2`, `The Matrix 3` are presents in the Neo4j instance then we can expect that `The Matrix 2` and `The Matrix 3` are returned.
```
angrykoala, line 45: Maybe change this example to the real names
The Matrix Reloaded and The Matrix Revolutions. Just to make it clear that `>` is nothing related to numbers (2,3)

angrykoala (commented):
Looking good to me

docs/rfcs/rfc-020-string-comparator-operators.md
MacondoExpress: Good call!

docs/rfcs/rfc-019-optional-features.md
MacondoExpress: Let's move forward with the @darrellwarde suggestion so!

tbwiss (dismissed):

docs/rfcs/rfc-019-optional-features.md
```diff
@@ -0,0 +1,82 @@
...
+```javascript
+  const neoSchema = new Neo4jGraphQL({
+        config: {
+            enableDebug: true,
+            features: {
+                filters: {
+                    String: {
+                        GT: true,
+                        LT: false,
+                    }
```
darrellwarde, line 24: So I have another suggestion here as I understand that this may be _too_ granular.

How about a config item `comparators` which contains each comparator for strings.

```suggestion
                    String: {
                        comparators: {
                            GT: true,
                            LT: false,
                        }
                    }
```

But then also allow it to accept a boolean value, which would allow you to disable all comparators. Could be the best of both worlds?

```
                    String: {
                        comparators: false
                    }
```

What do people think?

darrellwarde (commented):
I have another suggestion which could bring the best of both worlds of granularity and brevity - thoughts?

darrellwarde:
[[S4][Levelset - OREM Construction Job Radar] GraphQL v3.2.3 Request to Support ( GTE, GT, LTE, LT ) Operators For String Fields](https://trello.com/c/uRWUptV2/58-s4levelset-orem-construction-job-radar-graphql-v323-request-to-support-gte-gt-lte-lt-operators-for-string-fields)

PR #1698, opened 2022-07-07
Set `importsNotUsedAsValues` to `error` and fix

darrellwarde:
# Description

As per the title, this will make it clear whether an import is being used as a type or a value in a file.


neo4j-team-graphql:
Looks like you've updated the UI!

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1698.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1698.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

darrellwarde:
Single reviewdog failure here is a false negative, for some reason ESLint doesn't seem to pick up the TypeScript paths configuration.

And the license header check seems to be all over the place here - the headers are certainly present.

A lot of files, but all just switching over to only importing the values when necessary.

tbwiss:
@darrellwarde I had that before with the license check failing, the check sometimes finds files from way back that do not have a license header. In this case it's these two:
```
packages/graphql/src/schema/parse/parse-fulltext-directive.ts
packages/graphql/src/utils/context-parser.ts
```

darrellwarde:
> @darrellwarde I had that before with the license check failing, the check sometimes finds files from way back that do not have a license header. In this case it's these two:
> 
> ```
> packages/graphql/src/schema/parse/parse-fulltext-directive.ts
> packages/graphql/src/utils/context-parser.ts
> ```

Fixed! Thanks for the pointer.

PR #1713, opened 2022-07-08
Add performance tests for create and unwind

angrykoala:
# Description

This PR add performance tests for batch create requests


packages/graphql/tests/performance/performance.json
```diff
@@ -187,10 +215,10 @@
             "maxRows": 35,
             "dbHits": 214,
             "cache": {
-                "hits": 25,
+                "hits": 31,
                 "misses": 0
             },
-            "time": 19
+            "time": 6
```
tbwiss, line 221: question: We don't store the cypher test query results right? Is that something we should consider?

tbwiss (approved):
LGTM!

packages/graphql/tests/performance/performance.json
angrykoala: not for now, it can easily be added if we agree we need it

PR #1714, opened 2022-07-08
RFC: global authentication (RFC-23)

tbwiss:
# Description

RFC for global authentication.


docs/rfcs/rfc-021-global-authentication.md
```diff
@@ -0,0 +1,66 @@
...
+        globalAuthentication: true,
+    },
+    typeDefs: typeDefs,
+    driver,
+    plugins: {
+        auth: new Neo4jGraphQLAuthJWTPlugin({
+            secret: "1234",
+            noVerify: false,
+        }),
+    },
```
darrellwarde, line 25: Just a quick parting comment on this before I head off on vacation - I believe that the configuration for global authentication should be in the auth plugin, and not in the library. It depends on JWT decoding and verification, which is a plugin feature.

darrellwarde (changes requested):
A quick comment here!

docs/rfcs/rfc-021-global-authentication.md
tbwiss: Really good point @darrellwarde! Will make the necessary adjustments

tbwiss:
@darrellwarde after today's talk (we go for the proposed solution), is this ready to be merged?

darrellwarde (dismissed):
Indeed, I think this is good to merge for now!

PR #1730, opened 2022-07-11
Refactor Aura tests in GHA, add tests for 5.0 against `dev`

darrellwarde:
# Description

Instead of having specific instance details in the reusable workflow for testing Aura, this expects the URI and password to passed in as secrets, so you call the workflow as many times as you require for testing against Aura. This is helpful because for instance, failing against 5.0 shouldn't be a release blocker.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -37,7 +37,26 @@ jobs:
   toolbox-tests:
     uses: ./.github/workflows/reusable-toolbox-tests.yml
 
-  aura-tests:
+  aura-free-tests:
+    if: github.event.pull_request.merged == true
     uses: ./.github/workflows/reusable-aura-tests.yml
-    secrets: inherit
-    concurrency: aura
+    secrets:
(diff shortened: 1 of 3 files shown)


mjfwebb (approved):
This is great. I like the cleanup in .github/workflows/reusable-aura-tests.yml especially!

.github/workflows/dev.yml
```diff
@@ -37,7 +37,26 @@ jobs:
   toolbox-tests:
     uses: ./.github/workflows/reusable-toolbox-tests.yml
 
-  aura-tests:
+  aura-free-tests:
+    if: github.event.pull_request.merged == true
```
tbwiss, line 41: question: Is this "PR got merged" check really required for the `dev` workflow?

tbwiss (approved):
LGTM! Left a comment to consider.

.github/workflows/dev.yml
darrellwarde: Great spot, good old bit of copy and paste. Fixed!

PR #1746, opened 2022-07-13
Cypher builder refactor

angrykoala:
# Description
Refactor cypher builder and uses it for top-level-match


packages/graphql/src/translate/cypher-builder/clauses/Return.ts
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+import { Clause } from "./Clause";
+import type { NodeRef } from "../variables/NodeRef";
+import type { CypherEnvironment } from "../Environment";
+import { isString } from "../../../utils/utils";
+import type { Expr } from "../types";
+
+// TODO: make projection more elegant
```
mjfwebb, line 26: I like TODOs, but this comment isn't very helpful. Instead, I think that if there is a feeling that something can be improved in this code then writing more specifically about what those improvements would be will be more helpful.

packages/graphql/src/translate/cypher-builder/Exists.ts
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { CypherEnvironment } from "./Environment";
+import { CypherASTNode } from "./CypherASTNode";
+import { padBlock } from "./utils";
+import type { Clause } from "./clauses/Clause";
+
+/** Note: This is not a proper clause, I have no idea what this is */
```
mjfwebb, line 25: Should this comment have been left in?

packages/graphql/src/translate/cypher-builder/operations/boolean.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    public getCypher(env: CypherEnvironment): string {
+        const childStr = this.child.getCypher(env);
+        return `${this.operator} ${childStr}`;
+    }
+}
+
+export function and(left: BooleanOpChild, right: BooleanOpChild, ...extra: BooleanOpChild[]): BooleanOp;
+export function and(...ops: BooleanOpChild[]): BooleanOpChild | BooleanOp | undefined;
+export function and(...ops: BooleanOpChild[]): BooleanOp | BooleanOpChild | undefined {
```
mjfwebb, line 71: I think these two function signatures are identical but probably don't get flagged as such because of the union in the return types being in different orders.

packages/graphql/src/translate/cypher-builder/operations/boolean.ts
```diff
@@ -0,0 +1,93 @@
...
+    return op1;
+}
+
+export function not(child: BooleanOpChild): BooleanOp {
+    return new NotOp(child);
+}
+
+export function or(left: BooleanOpChild, right: BooleanOpChild, ...extra: BooleanOpChild[]): BooleanOp;
+export function or(...ops: BooleanOpChild[]): BooleanOpChild | BooleanOp | undefined;
+export function or(...ops: BooleanOpChild[]): BooleanOp | BooleanOpChild | undefined {
```
mjfwebb, line 86: Same as in `and`: I think these two function signatures are identical but probably don't get flagged as such because of the union in the return types being in different orders.

mjfwebb (commented):
I had a first look over, and I have just some general feedback (meta-commentary, one might say).

angrykoala:
Closed in favor of #1769

MacondoExpress (approved):

Issue #1683, opened 2022-07-06
Querying connections when no relationships exist nullifies parent result set in 3.5.0


wheresrhys:
**Describe the bug**
When querying connections if no relationship corresponding to the connection exists in the underlying data then the parent record is not retrieved either and an empty array is returned. This appears to be a regression in 3.5.0

**Type definitions**
Relevant part:

```
type System {
    code: String!
    updatesData: [GovernedData!]! 
}
type GovernedData {
    code: String!
    updatedBy: [System!]!
}

```

**To Reproduce**

1. Create a System record with no relationship
2. Run a server with @neo4j/graphql v3.5.0 installed
3. Then run the following Query...
```
{
   systems {
       code
       updatesDataConnection {edges{node{code}}}
   }
}
```
**Expected behavior**
Expect the result `{systems: [{code: 'blah', updatesDataConnection: {edges: []}}]}`

Received: `{systems: []}`

**System (please complete the following information):**

* OS: osx and linux
* Version: @neo4j/graphql@3.5.0 (have verified the bug is not present in 3.4.0)
* Node.js version: 14.x
* neo4j version: 4.3.9

**Additional context**
- I've also verified that this happens with many-to-one relationships
- Only happens when using the Connection. Does not happen with e,g. `{systems { updatesData {code }}}`

neo4j-team-graphql:
Many thanks for raising this bug report @wheresrhys. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @wheresrhys! :pray: We will now prioritise the bug and address it appropriately.

litewarp:
I'm having this issue as well -- queries that had been working are no longer returning data. It has something to do with the cypher projected and (I think) more specifically with the connection subquery.

angrykoala:
Thanks for that research @litewarp 
As you say, it is a bug introduced when fixing the ordering of nested connection fields.

tonxxd:
Hi, is the fix already deployed? I still see the same issue in `v3.5.0` but not in `v3.2.3` that i was using before.

PR #1757, opened 2022-07-14
Fix collection of edges in connection

angrykoala:
# Description

Fixes a connections regression

# Issue

Closes: #1683


wheresrhys:
Thanks for addressing this so quickly

Quick question - does the fix retain the changes that were made before which improved the previous performance problems?

PR #1769, opened 2022-07-18
Cypher builder

angrykoala:
# Description
Refactor cypher builder and uses it for top-level-match

Review note: Most files are tck tests, feel free to ignore all of those.


MacondoExpress (approved):

packages/graphql/src/translate/where/property-operations/create-comparison-operation.ts
```diff
@@ -0,0 +1,120 @@
...
+            return CypherBuilder.contains(property, param);
+        case "IN":
+        case "NOT_IN":
+            return CypherBuilder.in(property, param);
+        case "INCLUDES":
+        case "NOT_INCLUDES":
+            return CypherBuilder.in(param, property);
+        case "EQ":
+        case "NOT":
+            return CypherBuilder.eq(property, param);
```
darrellwarde, line 116: **question:** This is interesting, we're moving away from `x != y` to `NOT x = y`?

packages/graphql/src/translate/create-connect-or-create-and-params.ts
```diff
@@ -52,6 +52,8 @@ export function createConnectOrCreateAndParams({
     context: Context;
     withVars: string[];
 }): CypherBuilder.CypherResult {
+    const withVarsVariables = withVars.map((name) => new CypherBuilder.RawVariable(name));
```
darrellwarde, line 55: **question:** I remember us having a conversation about parameters vs variables and I thought we agreed on variables would always be "raw" (i.e. directly in the Cypher). Do we need the "Raw" prefix here?

packages/graphql/src/translate/create-connect-or-create-and-params.ts
```diff
@@ -88,7 +96,7 @@ function createConnectOrCreatePartialStatement({
     relationField: RelationField;
     refNode: Node;
     context: Context;
-}): CypherBuilder.Query {
```
darrellwarde, line 91: **question:** What was wrong with "Query" out of interest?

packages/graphql/src/translate/where/property-operations/create-point-comparison-operation.ts
```diff
@@ -0,0 +1,83 @@
...
+            return CypherBuilder.in(CypherBuilder.point(param), propertyRefOrCoalesce);
+        default:
+            throw new Error(`Invalid operator ${operator}`);
+    }
+}
+
+function createPointListComprehension(param: CypherBuilder.Param): CypherBuilder.ListComprehension {
+    const comprehensionVar = new CypherBuilder.Variable();
+    const mapPoint = CypherBuilder.point(comprehensionVar);
+    return new CypherBuilder.ListComprehension(comprehensionVar, param, undefined, mapPoint);
```
darrellwarde, line 74: **suggestion:** This might be a good opportunity to put `Point` types directly into the Cypher parameters as the native `point` types, rather than doing comprehensions in the Cypher. I think this would go a long way in aligning the spatial types with the other types.

darrellwarde (approved):
There's a lot to review here - I primarily have a lot of questions with one major suggestion (not related to this PR though). Nice work, the code is looking a lot more readable!

packages/graphql/src/translate/translate-top-level-match.ts
```diff
@@ -39,50 +39,41 @@ function translateTopLevelMatch({
...
     const whereInput = resolveTree.args.where as GraphQLWhereArg;
     const fulltextInput = (resolveTree.args.fulltext || {}) as Record<string, { phrase: string }>;
-    const whereStrs: string[] = [];
 
-    if (!Object.entries(fulltextInput).length) {
-        cyphers.push(`MATCH (${varName}${node.getLabelString(context)})`);
-    } else {
+    const matchNode = new CypherBuilder.NamedNode(varName, { labels: node.getLabels(context) });
+
+    let matchQuery: CypherBuilder.Match<CypherBuilder.Node> | CypherBuilder.db.FullTextQueryNodes =
```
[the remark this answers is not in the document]
angrykoala: While I'm happy to rename the internals (move these to a folder called procedures etc..), I think the exposed API is the easiest, as someone trying to write `CALL db.index.fulltext` will not necessarily know that those are called "procedures" in CypherBuilder.

For future `dbms` procedures I would put them inside `CypherBuilder.dbms` to keep a close match to the generated cypher

packages/graphql/src/translate/translate-top-level-match.ts
```diff
@@ -39,50 +39,41 @@ function translateTopLevelMatch({
...
-        if (node.nodeDirective?.additionalLabels?.length) {
-            node.getLabels(context).forEach((label) => {
-                whereStrs.push(`"${label}" IN labels(${varName})`);
-            });
-        } else {
-            whereStrs.push(`"${node.getMainLabel()}" IN labels(${varName})`);
-        }
+        const labelsChecks = node.getLabels(context).map((label) => {
+            return CypherBuilder.in(new CypherBuilder.Literal(`"${label}"`), CypherBuilder.labels(matchNode));
+        });
```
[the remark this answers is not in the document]
angrykoala: I'm pretty sure the above if...else is redundant, as the node will already contain its labels, whether is one or more than one

packages/graphql/src/translate/where/property-operations/create-comparison-operation.ts
angrykoala: I may be wrong, but I think we have never had != for this use case https://github.com/neo4j/graphql/blob/dev/packages/graphql/tests/tck/tck-test-files/advanced-filtering.test.ts#L121

packages/graphql/src/translate/create-connect-or-create-and-params.ts
angrykoala: Here RawVariable refers to a variable that already has a name, maybe "NamedVariable" is a better name for this.

The difference is:
```
new CypherBuilder.Variable(); // Will get translated to a unique var0
new CypherBuilder.Variable("my-var"); // Will get translated to my-var
```

This class is purely for compatibility reasons.

packages/graphql/src/translate/create-connect-or-create-and-params.ts
angrykoala: Query is not really accurate, as in Cypher terms what we are building is a `Clause`.

I also think is it a much better name, as it avoid ambigüity with the several `query` things that we have in the project

darrellwarde: Whilst this is true, I've never once said to myself "time to open up Neo4j Browser and write a quick Cypher clause!" - is it not worthwhile using phrasing that we are all familiar with in the Cypher builder?

angrykoala: They are not the same, a query is a full input you throw to Neo4j, a Clause is one type of statement supported by Cypher.

So, technically speaking, all clauses can be queries, but the inverse is not necessarily true. Talking about CypherBuilder, a query would be the whole tree (basically, what happens when you call `.build`. But each AST node of the tree is either a Clause or any other of the elements that compose a Cypher Query. So, I think it is way more accurate to say that `Cypher.Match` is a clause than a query. 

It is true, though, that the definition of clause is stretched a bit in the CypherBuilder. `CompositeClause` (`CypherBuilder.concat`) is no longer a clause but can be treated as one, as for simplicity's sake it is classed as such.

We could have 2 levels of semantics here (Clause and Query) or a type named Query that works as some kind of syntax sugar, but tbh considering how overused and meaningless the word query is already in the codebase, it feels redundant. Because of that, I decided to only name `query` the actual string result of the CypherBuilder after `.build` is executed

darrellwarde: Perhaps we shouldn't be talking in terms of clauses and queries at all for the "top-level" `CypherBuilder` object given that it isn't a query until `.build()` is called?

What if we had a `CypherBuilder.Instance` which contains the `.build()` method which then returns a query? I feel like perhaps we trying to apply an abstraction to something where it doesn't really fit.

packages/graphql/src/translate/where/property-operations/create-point-comparison-operation.ts
darrellwarde: One for later!

PR #1773, opened 2022-07-18
RFC 019, 020 initial implementation

MacondoExpress:
# Description
Add comparator operators to String type and introduce it as an optional feature for strings. 
Described in more detail in RFC 019 and RFC 020.

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-1773.surge.sh

.changeset/string-comparators-changeset.md
```diff
@@ -0,0 +1,7 @@
+---
+"graphql-manual": patch
+"@neo4j/graphql": patch
```
angrykoala, line 3: I think this should be a minor release, as it is adding some features

packages/graphql/tests/integration/advanced-filtering.int.test.ts
```diff
@@ -680,6 +680,288 @@ describe("Advanced Filtering", () => {
...
+                            CREATE (:${movieType.name} {title: $matrix})
+                            CREATE (:${movieType.name} {title: $matrixReloaded})
+                            CREATE (:${movieType.name} {title: $matrixRevolutions})
+                        `,
+                    { matrix, matrixReloaded,  matrixRevolutions }
+                );
+
+                const query = `
+                            {
+                                ${movieType.plural}(where: { title_GT: "${matrix}" }) {
```
angrykoala, line 726: I may be missing another test case, but just to ensure this work as expected, it would be good to add a film with a smaller title (e.g "IT") so we can confirm that this filter is removing that one

packages/graphql/tests/integration/advanced-filtering.int.test.ts
```diff
@@ -680,6 +680,288 @@ describe("Advanced Filtering", () => {
...
+                            CREATE (:${movieType.name} {title: $matrix})
+                            CREATE (:${movieType.name} {title: $matrixReloaded})
+                            CREATE (:${movieType.name} {title: $matrixRevolutions})
+                        `,
+                    { matrix, matrixReloaded,  matrixRevolutions }
+                );
+
+                const query = `
+                            {
+                                ${movieType.plural}(where: { title_LT: "${matrixRevolutions}" }) {
```
angrykoala, line 796: Same as before, but with a title longer (e.g. "The Englishman who Went up a Hill but Came down a Mountain")

angrykoala (approved):
LGTM :+1:

darrellwarde:
[[S4][Levelset - OREM Construction Job Radar] GraphQL v3.2.3 Request to Support ( GTE, GT, LTE, LT ) Operators For String Fields](https://trello.com/c/uRWUptV2/58-s4levelset-orem-construction-job-radar-graphql-v323-request-to-support-gte-gt-lte-lt-operators-for-string-fields)

.changeset/string-comparators-changeset.md
MacondoExpress: I agree!

packages/graphql/tests/integration/advanced-filtering.int.test.ts
angrykoala: Oh, sorry, for some reason in my head I though it was length and not alphabetical order.

My point still is the same (only use something like aaa, or whatever) just to make sure that at least a field not greater than is filtered

MacondoExpress: No problem!

packages/graphql/tests/integration/optional-features-int.test.ts
```diff
@@ -0,0 +1,136 @@
+/*
```
darrellwarde, line 1: **suggestion:** I would say this file is completely unnecessary, as this is totally covered by [packages/graphql/tests/schema/string-comparators.test.ts](https://github.com/neo4j/graphql/pull/1773/files#diff-dd5f7d162b2d2fa7ee4b30ea50e73942b783c8cba16dd3e4244143fbb9dd013f)! 🙂

packages/graphql/tests/integration/advanced-filtering.int.test.ts
```diff
@@ -680,6 +680,299 @@ describe("Advanced Filtering", () => {
...
+                                LT: true,
+                                GT: true,
+                                LTE: true,
+                                GTE: true,
+                            },
+                        },
+                    },
+                },
+                typeDefs,
+            });
```
darrellwarde, line 709: **suggestion:** I think this configuration is a level too deep! What if it could be like this? I think `features` sits nicely next to `plugins`. 

```suggestion
            const neoSchema = new Neo4jGraphQL({
                features: {
                    filters: {
                        String: {
                            LT: true,
                            GT: true,
                            LTE: true,
                            GTE: true,
                        },
                    },
                },
                typeDefs,
            });
```

darrellwarde (changes requested):
This looks great, super fast work! 🥇 

I have a couple of changes suggested, and have a think about my suggestion for the structure of the constructor!

packages/graphql/tests/integration/advanced-filtering.int.test.ts
MacondoExpress: Could it worth remove the features level and let the config be:
```javascript
config: {
  filters: {
      String: {
          LT: true,
          GT: true,
          LTE: true,
          GTE: true
      }
  }
}
```
thoughts?

packages/graphql/tests/integration/optional-features-int.test.ts
MacondoExpress: I think you're right 😄

packages/graphql/tests/integration/advanced-filtering.int.test.ts
MacondoExpress: No problem, it makes sense to me as well! I'm going to move this outside the configuration file 😄

PR #1780, opened 2022-07-19
fixed: an EXISTS clause is used in a RETURN clause where it is not valid

### LINKED ISSUE #1779: An `EXISTS` clause is used in a `RETURN` clause where it is not valid
**Describe the bug**
An `EXISTS` clause is used in a `RETURN` clause where it is not valid.

**Type definitions**

```gql
type Person {
    name: String
    age: Int
    attends: [School!]! @relationship(type: "attends", direction: OUT)
}

type School {
    name: String
    students: [Person!]! @relationship(type: "attends", direction: IN)
}
```

**To Reproduce**

1. Execute the following query:

```gql
{
  people{
    name
    attends(where: {students_ALL: {age_GT: 23}}){
      name
    }
  }
}
```

2. See the error `The EXISTS subclause is not valid inside a WITH or RETURN clause.`

**Expected behavior**

No error to be thrown.

* Version: `@neo4j/graphql@3.5.0`

darrellwarde:
# Description

Reported on Discord, `EXISTS` was being used in a `RETURN` clause where it is not valid.

This PR fixes this by switching to the `listPredicateToSizeFunction` to generate the clause. It's a shame, because we're now only using `EXISTS` in one place, which is the better solution. We need to switch from doing our projections of relationships in the `RETURN` clause itself to get around this.

# Issue

Closes #1779

### DIFF
--- a/packages/graphql/tests/integration/issues/1779.int.test.ts
+++ b/packages/graphql/tests/integration/issues/1779.int.test.ts
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


packages/graphql/tests/integration/issues/1779.int.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        CREATE (personD:${personType.name} { name: "D", age: 25 })-[:ATTENDS]->(schoolYoung)
+    `;
+
+        const session = await neo4j.getSession();
+
+        try {
+            await session.run(cypher);
+        } finally {
+            await session.close();
+        }
```
angrykoala, line 71: **non-blocking**

Consider moving `neo4j.getSession` and `session.close` to `beforeEach` and `afterEach` so we don't need the try...finally in the test case

angrykoala (approved):

PR #1787, opened 2022-07-19
Wraps translation in CypherBuilder

angrykoala:
# Description

This PR wraps all the logic in `translate-x` inside a `CypherBuilder.RawCypher` element, to provide the basic framework to start integrating the pieces into a single Cypher AST and facilitate extracting subqueries.

None of the internal logic has been changed, and there are only some minor changes to generated parameters  

Depends on #1769


darrellwarde (dismissed):
Good initiative to wrap the old stuff up!

PR #1790, opened 2022-07-21
Cypher Builder return and extra functions

angrykoala:
# Description

This PR adds the plumbing required in the cypher builder to support aggregations queries to remove `runFirstColumn` in these.

Additionally, this PR introduces `mixins` in the CypherBuilder to reduce code duplicity between clauses


packages/graphql/src/translate/cypher-builder/examples/create.test.ts
```diff
@@ -21,31 +21,27 @@ import * as CypherBuilder from "../CypherBuilder";
 
 describe("CypherBuilder Create Examples", () => {
     test("Create a movie", () => {
-        const titleParam = new CypherBuilder.Param("The Matrix");
```
MacondoExpress, line 24: I think that the test was better before with both the `title` and the `released` parameter!

packages/graphql/src/translate/cypher-builder/examples/match.test.ts
```diff
@@ -41,13 +41,13 @@ describe("CypherBuilder Match Examples", () => {
...
             .and(movieNode, { released: releasedParam })
-            .return(movieNode, ["title", "released"], "movie");
+            .return(movieNode.property("title"), [movieNode.property("released"), "year"]);
 
         const queryResult = matchQuery.build();
         expect(queryResult.cypher).toMatchInlineSnapshot(`
             "MATCH (this1:\`Person\`)-[this0:ACTED_IN]->(this2:\`Movie\`)
             WHERE (this1.name = $param0 AND this2.released = $param1)
-            RETURN this2 {.title, .released} AS movie"
+            RETURN this2.title, this2.released AS year"
```
MacondoExpress, line 50: Is it still possible to obtain a projection `movie` with both the `title` and `year` as before?

MacondoExpress:
It looks really good to me! I left some comments where I have some doubts🥇

MacondoExpress (approved):

PR #1796, opened 2022-07-22
Fix: move `ORDER BY`, `SKIP` and `LIMIT` to as early as possible in Cypher

### LINKED ISSUE #1760: Ordering, skipping and limiting happening late in Cypher causing poor performance
**Describe the bug**

Customer issue reported internally.

darrellwarde:
# Description

Previously, `ORDER BY`, `SKIP` and `LIMIT` were happening as the very last step in our generated Cypher. This had the side effect of _every single_ relationship from the source node being projected, which resulted in very poor performance.

This moves these clauses to right after the initial matching and filtering of the source node, so that subsequent steps are only executed against required nodes.

# Issue

Closes #1760

# Performance

The following query can be added to the performance tests to validate the effectiveness of this work:

```gql
{
  movies(options: { sort: { title: ASC }, limit: 10 }) {
    title
    actorsConnection {
      edges {
        node {
          name
        }
      }
    }
    directorsConnection {
      edges {
        node {
          name
        }
      }
    }
  }
}

```


angrykoala:
For context on the aforementioned query, the performance increase of the example goes from 801 dbHits to 274, testing database size is not big enough to give conclusive time improvements

packages/graphql/tests/integration/issues/1760.int.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        `;
+        const neoGraphql = new Neo4jGraphQL({ typeDefs, driver });
+        schema = await neoGraphql.getSchema();
+    });
+
+    afterAll(async () => {
+        await driver.close();
+    });
+
+    test("provides query does not result in an error", async () => {
```
angrykoala, line 79:
```suggestion
    test("provided query does not result in an error", async () => {
```

packages/graphql/src/translate/translate-read.ts
```diff
@@ -189,70 +191,51 @@ function translateRootField({
...
+
+    const cypher: string[] = [
+        subStr.matchAndWhereStr,
+        ...(sortOffsetLimit.length > 1 ? sortOffsetLimit : []),
+        subStr.authStr,
+        ...withStrs,
+        ...subStr.connectionStrs,
+        ...subStr.interfaceStrs,
+        ...returnStrs,
+    ];
```
angrykoala, line 237: :clap: Much simpler logic now

angrykoala (dismissed):

PR #1803, opened 2022-07-25
RFC Database info

MacondoExpress:
# Description
Contains details of the RFC Database info

### DIFF
--- a/docs/rfcs/rfc-021-database-info.md
+++ b/docs/rfcs/rfc-021-database-info.md
@@ -2,9 +2,9 @@
 
 ## Problem
 Providing compatibility with multiple Neo4j versions comes with some difficulty.
-Different Neo4j versions require different approaches to obtain the same result optimally. Without knowing where the query will be executed during the query generation, it's not possible to change the query accordingly.
+Different Neo4j versions require different Cypher queries to obtain the same result. Without knowing where the query will be executed during the query generation, it's not possible to change the query accordingly.
 
-As a use case, let's consider the following example:
+For instance, let's consider the following example:
 ```cypher
 RETURN distance(point({x:10, y:5}), point({x:10, y:10}))
 ```
@@ -16,7 +16,7 @@ RETURN point.distance(point({x:10, y:5}), point({x:10, y:10}))
 Which can be executed against a `Neo4j 5.x` instance.
 
 ## Proposed Solution
-From version `Neo4j 3.5` it is available the procedure `dbms.components` can be used as:
+From version `Neo4j 3.5` the procedure `dbms.components` can be used as:
 ```cypher
 call dbms.components() yield name, versions, edition unwind versions as version return name, version, edition;
 ```
@@ -30,7 +30,7 @@ with this information available it's possible to write an optimal query.
 To make this available to the resolvers, it is possible to add a new class `Neo4jDatabaseInfo` to the GraphQL `Context`. It is then possible to compare the version with the [`ServerInfo`](https://neo4j.com/docs/api/javascript-driver/current/class/lib6/result-summary.js~ServerInfo.html) version in the driver response.
 
 An essential implementation of the `Neo4jDatabaseInfo`: 
-```javascript
+```typescript
 interface Neo4jVersion {
   major: number;
   minor: number;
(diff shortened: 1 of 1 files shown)


docs/rfcs/rfc-021-database-info.md
```diff
@@ -0,0 +1,117 @@
+# Database info
+
+## Problem
+Providing compatibility with multiple Neo4j versions comes with some difficulty.
+Different Neo4j versions require different approaches to obtain the same result optimally. Without knowing where the query will be executed during the query generation, it's not possible to change the query accordingly.
```
angrykoala, line 5:
```suggestion
Different Neo4j versions require different Cypher queries to obtain the same result. Without knowing where the query will be executed during the query generation, it's not possible to change the query accordingly.
```

docs/rfcs/rfc-021-database-info.md
```diff
@@ -0,0 +1,117 @@
...
+
+As a use case, let's consider the following example:
```
angrykoala, line 7:
```suggestion
For instance, let's consider the following example:
```

docs/rfcs/rfc-021-database-info.md
```diff
@@ -0,0 +1,117 @@
...
+```cypher
+RETURN distance(point({x:10, y:5}), point({x:10, y:10}))
+```
+Which produces `5.0` using `Neo4j 4.x`, it will result in an error if used against a `Neo4j 5.x` instance.
+The same query could be written as:
```
angrykoala, line 12: For `5.x` the same query should be written instead as:

docs/rfcs/rfc-021-database-info.md
```diff
@@ -0,0 +1,117 @@
...
+```cypher
+RETURN point.distance(point({x:10, y:5}), point({x:10, y:10}))
+```
+Which can be executed against a `Neo4j 5.x` instance.
+
+## Proposed Solution
+From version `Neo4j 3.5` it is available the procedure `dbms.components` can be used as:
```
angrykoala, line 19:
```suggestion
From version `Neo4j 3.5` the procedure `dbms.components` can be used as:
```

docs/rfcs/rfc-021-database-info.md
```diff
@@ -0,0 +1,117 @@
...
+
+|  name            |  version         |  edition         |
+|  --------------  |  --------------  |  --------------  |
+|  "Neo4j Kernel"  |  "4.4.5"         |  "enterprise"    |
+
+with this information available it's possible to write an optimal query.
+To make this available to the resolvers, it is possible to add a new class `Neo4jDatabaseInfo` to the GraphQL `Context`. It is then possible to compare the version with the [`ServerInfo`](https://neo4j.com/docs/api/javascript-driver/current/class/lib6/result-summary.js~ServerInfo.html) version in the driver response.
+
+An essential implementation of the `Neo4jDatabaseInfo`: 
+```javascript
```
angrykoala, line 33: This code is actually typescript :)

```suggestion
```typescript
```

docs/rfcs/rfc-021-database-info.md
```diff
@@ -0,0 +1,117 @@
...
+
+  constructor(version: Neo4jVersion, edition: string) {
+    this.version = version;
+    this.edition = edition;
+  }
+}
+```
+
+### Application Flow
+```mermaid
```
angrykoala, line 51: Interesting, I never used mermaid for real

docs/rfcs/rfc-021-database-info.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### Roundtrips to the Database
+It will be better to avoid unnecessary roundtrips for every request that will create a network and performance overhead.
+
+#### Database updates
+The implementation needs to handle the case where the database version is updated while the Neo4j Graphql library is running.
+
+#### Context defined driver
+The implementation needs to handle the case where the driver configuration is passed in a GraphQL Context.
+
```
angrykoala, line 109: We should support that the user specifies a version in the configuration, in that case we will ignore the first round-trip and just use that version

angrykoala (changes requested):
Looks good, just a few minor improvements on some details

angrykoala (approved):

PR #1822, opened 2022-07-29
Database info

MacondoExpress:
# Description

Provide a mechanism to access the Neo4j version before generating the query.
It fixes also the compatibility with Neo4j 5.x by supporting the syntax `point.distance`.

# Checklist


packages/graphql/src/schema/resolvers/wrapper.ts
```diff
@@ -105,10 +118,45 @@ export const wrapResolver =
             executorConstructorParam.bookmarks = context.driverConfig?.bookmarks;
         }
 
+        if (!neo4jDatabaseInfo?.version) {
```
darrellwarde, line 121: Surely this is going to _always_ be undefined, because it is only defined on line 56. I believe this database version fetching will happen on every execution with this current implementation.

packages/graphql/src/schema/resolvers/wrapper.ts
```diff
@@ -105,10 +118,45 @@ export const wrapResolver =
...
+            );
+            const rawRow = dbmsComponentsQueryResult?.records[0];
+            const [rawVersion, edition] = rawRow;
+            neo4jDatabaseInfo = new Neo4jDatabaseInfo(rawVersion as string, edition as Neo4jEdition);
+        }
+        executorConstructorParam.neo4jDatabaseInfo = neo4jDatabaseInfo;
         context.executor = new Executor(executorConstructorParam);
+        context.neo4jDatabaseInfo = neo4jDatabaseInfo;
+
+        return versionMismatchHandler(next)(root, args, context, info);
```
darrellwarde, line 137: It feels slightly strange for the return clause of the `wrapResolver` to be an error handler. Could this not have been caught and then handled, or alternatively, handled within the executor?

packages/graphql/src/classes/Executor.ts
```diff
@@ -189,13 +197,28 @@ export class Executor {
...
 
     private async sessionRun(query: string, parameters, defaultAccessMode, session): Promise<QueryResult> {
         const transactionType = `${defaultAccessMode.toLowerCase()}Transaction`;
-        const result = await session[transactionType](
-            (transaction: Transaction) => this.transactionRun(query, parameters, transaction),
-            this.getTransactionConfig()
-        );
+        const result = await session[transactionType](async (transaction: Transaction) => {
+            const txResult = await this.transactionRun(query, parameters, transaction);
+            this.throwIfVersionMismatch(txResult.summary.server);
```
darrellwarde, line 219: I really think this is unnecessary - I don't think we should be worried if we get to this stage because the Cypher is likely to have executed without error.

darrellwarde (changes requested):
A few thoughts on this!

packages/graphql/src/schema/resolvers/wrapper.ts
darrellwarde: Misunderstanding here!!

packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -56,6 +57,35 @@ export interface Neo4jGraphQLConfig {
     callbacks?: Neo4jGraphQLCallbacks;
 }
 
+export interface Neo4jVersion {
+    major: number;
+    minor: number;
+}
```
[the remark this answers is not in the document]
MacondoExpress: Not any particular reason!

packages/graphql/src/classes/Executor.ts
MacondoExpress: My idea was around the scenario when the version is automatically detected and later the database is changed or updated, in that case, we could rollback the transaction and catch the error at the level when the resolved is invoked, change the memorized version to the new one and re-generate the query using the newer version.
We could think of another mechanism to automatically update the version anyway. 
I think the rollback mechanism will become necessary if we're going to adopt a 'retry' handler, if not, we could remove both in my opinion!
Happy to remove both, we could instead just print a warning like "different Neo4j version detected".

neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1822.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1822.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/tests/utils/builders/driver-builder.ts
```diff
@@ -76,9 +77,44 @@ export class DriverBuilder extends Builder<Driver, Partial<Driver>> {
...
+                            return "";
+                        },
+                    },
+                },
+            };
+        }
+
+        mockFunc.calls = calls;
+        return mockFunc;
+    } */
```
angrykoala, line 98: :upside_down_face:

packages/graphql/tests/utils/builders/driver-builder.ts
```diff
@@ -76,9 +77,44 @@ export class DriverBuilder extends Builder<Driver, Partial<Driver>> {
...
     private createRunMock(): RunFunction {
         const calls: Array<any> = [];
         function mockFunc(...params) {
+            // this is needed as the first query could be the DB version check query
+            if (params && params[0] && params[0] === DBMS_COMPONENTS_QUERY) {
```
angrykoala, line 103:
```suggestion
            if (params?.[0] === DBMS_COMPONENTS_QUERY) {
```

packages/graphql/src/classes/utils/verify-version.ts
```diff
@@ -17,45 +17,26 @@
...
-import type { Session } from "neo4j-driver";
-import semver from "semver";
 import { MIN_VERSIONS } from "../../constants";
+import type { Neo4jDatabaseInfo } from "../Neo4jDatabaseInfo";
 
-export async function verifyVersion(sessionFactory: () => Session): Promise<void> {
-    const session = sessionFactory();
+export function verifyVersion(dbInfo: Neo4jDatabaseInfo): void {
+    if (!dbInfo.toString().includes("aura")) {
+        const minimumVersions = MIN_VERSIONS.find(({ majorMinor }) => dbInfo.toString().startsWith(majorMinor));
```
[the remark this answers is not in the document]
darrellwarde: I'm not tempted to fix this now - this isn't a new check I've introduced, purely refactoring the existing code to use the new class. In the interest of getting the 5.0 compatibility stuff merged, save for another time?

PR #1845, opened 2022-08-04
Create rfc-22-projection-subquery.md

angrykoala:
# Description
Move projections to subqueries

### DIFF
--- a/docs/rfcs/rfc-22-projection-subquery.md
+++ b/docs/rfcs/rfc-22-projection-subquery.md
@@ -1,0 +1,180 @@
+# Example Title
+
+## Problem
+Currently, projection and nested filtering is done in the `RETURN` statement, through [Pattern Comprehension](https://neo4j.com/docs/cypher-manual/current/syntax/lists/#cypher-pattern-comprehension). 
+For instance:
+
+**Normal query**
+```graphql
+query Query {
+  movies {
+    title
+  }
+}
+```
+
+```cypher
+MATCH (this:`Movie`)
+RETURN this { .title } as this
+```
+
+**Nested query with filters**
+
+```graphql
+query Query {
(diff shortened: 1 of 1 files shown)


docs/rfcs/rfc-22-projection-subquery.md
```diff
@@ -0,0 +1,180 @@
+# Example Title
```
tbwiss, line 1: Would be good to replace this.

docs/rfcs/rfc-22-projection-subquery.md
```diff
@@ -0,0 +1,180 @@
...
+      WITH this_actors
+      MATCH(this_actors)-[:DIRECTED]->(this_movies:Movie)
+      WHERE m.released > 1999
+      RETURN collect(this_movies {.title}) as this_movies
+    }
+    RETURN collect(this_actors {.name, directed: this_movies}) as this_actors
+}
+RETURN this { .title, actors: this_actors}
+```
+
```
tbwiss, line 116: Which alternative are we using now? `OPTIONAL MATCH` or `CALL` subqueries? Will it be a mix?

tbwiss (commented):
Thanks for the detailed explanation and examples. Left a comment which you may want to consider.

docs/rfcs/rfc-22-projection-subquery.md
angrykoala: I just updated the RFC, we will go with `CALL` as it is the one already used in some places, as well as the easier one.
I will keep the optional match as an alternative solution, as it may lead to performance improvements in the future

darrellwarde (approved):
This looks good! Let's definitely go for the subqueries approach now, and then take a look at the alternatives in bulk if we feel like we are experiencing performance issues.

PR #1858, opened 2022-08-05
Release tweaks following first hotfix

darrellwarde:
# Description

A couple of changes which came out of the woodwork as part of the hotfix release:

- The GraphQL Toolbox should be bumped whenever we bump the library - this will lead onto more reliable releases
- A branch should be created from `master` before creating a PR into `dev` following a hotfix release - this will avoid corrupting the commit history of `master`.


tbwiss (approved):
LGTM!

PR #1863, opened 2022-08-08
Fix some flaky tests

darrellwarde:
# Description

Beginning to fix some flaky tests, this PR mainly focusses on array matching and using `expect.toIncludeSameMembers()`.


tbwiss (approved):
Nice clean up @darrellwarde!
Interesting to see that roughly only half of the tests which used `arrayContaining` also checked the length of the array.. Luckily that is not needed anymore with `toIncludeSameMembers`.

PR #1870, opened 2022-08-09
Add tests to run nightly against the nightly build of Neo4j

darrellwarde:
# Description

This new workflow pulls down the nightly build of Neo4j from a private ECR registry.

I _think_ this will work, how I expect some rustiness from this at first.


.github/workflows/reusable-integration-tests-on-prem-nightly.yml
```diff
@@ -0,0 +1,93 @@
+name: "Integration tests (on-prem, nightly build)"
```
tbwiss, line 1: Could we clarify that it's the nightly build of Neo4j that we run the tests against? Otherwise it might be a bit confusing since there is also an `integration-tests-on-prem` workflow. Wdyt?

tbwiss (approved):
LGTM!

PR #1873, opened 2022-08-10
feat: Toolbox restructure after UX feedback

tbwiss:
# Description

This PR contains the following changes:
- Prompt to introspect type definitions on connection
- Restructuring of the page
  - Type definitions/query selector component moves to the top (forming a second "top bar")
  - Introspect button to be inside type definitions
  - Back to "Prettify" text in the button
  - Beta tag, use a blue label
  - Move the documentation sidebar to also be on the left, toggled by a slider in the Explorer
- Tweak the "Add new Query/Mutation" button in the Explorer
- E2E tests to cover the above-mentioned features/extensions


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1873.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1873.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

tbwiss:
@darrellwarde there is still no changeset entry required for the toolbox right?

darrellwarde:
> 

I think we should actually!! Because then we can produce a changelog, trigger off that, and then deploy the Toolbox.

darrellwarde:
This all looks amazing!! I'm assuming there's no way to move the docs toggle more inline with the "Explorer" heading, and then possible anchor both to the top? I'm assuming the heading is built into the explorer component and isn't editable?

Screenshot 2022-08-10 at 15 27 33

Amazing work getting the "Add new Query/Mutation" button working!! 🏅

tbwiss:
> This all looks amazing!! I'm assuming there's no way to move the docs toggle more inline with the "Explorer" heading, and then possible anchor both to the top? I'm assuming the heading is built into the explorer component and isn't editable?
> 
> Screenshot 2022-08-10 at 15 27 33
> 
> Amazing work getting the "Add new Query/Mutation" button working!! 🏅

To anchor both to the top would be great yet sadly the Explorer component does not offer this (or a lot of other) flexibility... Exactly, this is all pre-defined and there is no way to edit it. So the only way would be to fix the position of the switch (show Docs) yet then as soon as the Explorer component displays a lot of content (and therefore a scroll bar is required) the fixed position becomes a problem. The solution in the image above is the best alternative I could come up with so far. Happy to hear other suggestions!

darrellwarde (approved):
Looks great!

PR #1893, opened 2022-08-12
Add order example to rfc

angrykoala:
# Description
Updates rfc on subquery projection

### DIFF
--- a/docs/rfcs/rfc-022-projection-subquery.md
+++ b/docs/rfcs/rfc-022-projection-subquery.md
@@ -104,6 +104,39 @@ RETURN this { .title, actors: this_actors}
 
 ### Examples
 
+#### With Order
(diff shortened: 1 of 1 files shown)


docs/rfcs/rfc-022-projection-subquery.md
```diff
@@ -152,6 +185,8 @@ CALL {
 RETURN this { .released, actorsConnection } as this
 ```
 
+Projections insider connections, however, may required nested CALL statements
```
tbwiss, line 188: Wdyt?
```suggestion
Projections inside connections, however, may require nested CALL statements
```

tbwiss (approved):
LGTM!
Great effort with getting rid of this apoc method dependency here!

PR #1918, opened 2022-08-16
Subquery projection

angrykoala:
# Description
Implements [rfc 22](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-022-projection-subquery.md)
Moves projections from the `RETURN` statements to separate subqueries (`CALL`) 

No changes in performance noticed during testing


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1918.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1918.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/src/translate/translate-top-level-cypher.ts
```diff
@@ -0,0 +1,225 @@
+/*
```
mjfwebb, line 1: I think this file's readability (and subsequent ease of collaboration) would benefit from a few things:

- Declaring variables only just before usage. For example, if a variable is used only once in an if statement, I think it would be easier to read the file in a general sense if they were placed together.
- Being broken up into smaller functions. For example the two blocks `if (referenceNode)` and `if (referenceUnion) {`

packages/graphql/src/translate/projection/elements/create-projection-subquery.ts
```diff
@@ -0,0 +1,182 @@
...
+    });
+
+    if (preAuth) {
+        const allowAuth = new CypherBuilder.apoc.ValidatePredicate(CypherBuilder.not(preAuth), AUTH_FORBIDDEN_ERROR);
+        subqueryMatch.and(allowAuth);
+    }
+
+    if (authValidateStrs?.length) {
+        const authValidateStatements = authValidateStrs.map((str) => new CypherBuilder.RawCypher(str));
+        const authValidatePredicate = CypherBuilder.and(...authValidateStatements) as CypherBuilder.Predicate;
```
mjfwebb, line 122: Is there a reason `CypherBuilder.and()` isn't able to return the type `CypherBuilder.Predicate`?

tbwiss (commented):
Would be great if you could go through the comments you made (TODO, note, etc) and see if some of them can be removed.

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -467,40 +470,25 @@ export default function createProjectionAndParams({
...
-            const nodeWhereAndParams = createNodeWhereAndParams({
+            const parentNode = new CypherBuilder.NamedNode(chainStr || varName); // TODO: improve this
+
+            const direction = getRelationshipDirection(relationField, field.args);
+            const subquery = createProjectionSubquery({
+                parentNode,
                 whereInput,
-                varName: `${varName}_${alias}`,
-                node: referenceNode,
+                node: referenceNode || node, // Why || node?
```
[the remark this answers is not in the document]
angrykoala: I'm questioning my sanity at this point

packages/graphql/src/translate/projection/elements/create-projection-subquery.ts
```diff
@@ -0,0 +1,182 @@
...
+    });
+
+    const relationship = new CypherBuilder.Relationship({
+        source: parentNode,
+        target: targetNode,
+        type: relationField.type,
+    });
+    if (relationshipDirection === "IN") {
+        relationship.reverse();
+    }
```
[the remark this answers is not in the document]
angrykoala: Yes, it may be, and we may end up adding it, but this whole direction logic is a bit cumbersome and we may be able to find a better way of doing this. For now we need to keep this logic here until we refactor more of the old way of doing things to cypher builder

packages/graphql/src/translate/translate-top-level-cypher.ts
angrykoala: I completely agree with you, but keep in mind that this is a block of code that was moved from the cypher resolver into a separate function (it is not new code, the diff is misleading). Refactor this further is a bit of a scope creep for this task, although I'm happy to implement some quick wins if you see them

packages/graphql/src/translate/projection/elements/create-projection-subquery.ts
angrykoala: `CypherBuilder.and` returns `Predicate | undefined` as the typings system is not clever enough to detect that there are elements in the `authValidateStatements` array (checked by the previous if)

packages/graphql/tests/integration/limit.int.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        schema = await neoGraphql.getSchema();
+    });
+
+    beforeEach(async () => {
+        session = await neo4j.getSession();
+    });
+
+    afterEach(async () => {
+        await session.close();
+    });
```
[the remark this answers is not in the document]
angrykoala: mmm, we could, but I think is worthwhile to have this any way, to keep consistency and avoid race conditions when a new test is added

packages/graphql/src/translate/projection/elements/create-projection-subquery.ts
```diff
@@ -0,0 +1,182 @@
...
+    alias: string; // TODO: this should be output variable instead
+    nestedProjection: string;
+    nestedSubqueries: CypherBuilder.Clause[];
+    relationField: RelationField;
+    relationshipDirection: RelationshipDirection;
+    optionsInput: GraphQLOptionsArg;
+    authValidateStrs: string[] | undefined;
+}): {
+    subquery: CypherBuilder.Clause;
+    projectionColumn: CypherBuilder.ProjectionColumn[];
```
darrellwarde, line 55: I've maybe just not seen this before, but "projection column" seems like a very odd name to me!

darrellwarde (approved):
These PRs are difficult to review because there are just so many changes, but it looks good! 😃

packages/graphql/src/translate/projection/elements/create-projection-subquery.ts
angrykoala: It is the best name I could think for this entity based on [Cypher docs](https://neo4j.com/docs/cypher-manual/current/clauses/return/#return-column-alias)

webberwang:
We've encountered a regression bug and have narrowed down the reason to upgrading these packages

```
 "@neo4j/graphql": "3.6.3",
 "@neo4j/graphql-ogm": "3.6.3",
 ```

to 

```
 "@neo4j/graphql": "3.7.0",
 "@neo4j/graphql-ogm": "3.7.0",
 ```

It involves a query with a relationship, it's a bit complex to describe the bug fully, but we'll look to provide some examples.

PR #1919, opened 2022-08-16
fix: Toolbox Safari support

tbwiss:
# Description

Currently, the Toolbox can not be used in Safari because of a lack of support for [lookbehind](https://caniuse.com/js-regexp-lookbehind).
This PR splits the Duration regex, which is using two lookbehinds, into several smaller regexes.
~~Also adding `webkit` to the devices we run our E2E tests against.~~
Included one small styling fix that was noticeable when using the Toolbox on Safari.


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1919.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1919.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/src/graphql/scalars/Duration.ts
```diff
@@ -39,41 +46,34 @@ export const SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR;
...
+        hours: hourUnit = 0,
+        minutes: minuteUnit = 0,
+        seconds: secondUnit = 0,
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     } = match.groups!;
 
-    // NOTE: xUnit and xDT cannot both be nonzero by construction => (xUnit + xDT) = xUnit | xDT | 0
-    const years = +yearUnit + +yearDT;
-    const months = +monthUnit + +monthDT;
+    const years = +yearUnit;
```
angrykoala, line 70: Maybe we can get rid of the leading sign?

```suggestion
    const years = yearUnit;
```

In fact, I think we may be able to get rid of this assignments all together

packages/graphql/src/graphql/scalars/Duration.ts
```diff
@@ -130,7 +130,11 @@ export const GraphQLDuration = new GraphQLScalarType({
             return value.toString();
         }
 
-        if (!DURATION_REGEX.test(value)) {
+        if (
+            !DURATION_REGEX_ISO.test(value) ||
+            !DURATION_REGEX_WITH_DELIMITERS.test(value) ||
+            !!DURATION_REGEX_NO_DELIMITERS.test(value)
```
angrykoala: I think there may be an extra `!` here?
```suggestion
            !DURATION_REGEX_NO_DELIMITERS.test(value)
```

angrykoala (approved):

packages/graphql/src/graphql/scalars/Duration.ts
tbwiss: Yes it was! Had to refactor this entire check here though

packages/graphql/src/graphql/scalars/Duration.ts
tbwiss: That leading `+` is to cast a string to a number. That was the best way I could come up so far. Happy to hear a better approach!

darrellwarde (approved):
I think this looks good! Thankfully the unit tests for `Duration` means we can have quite high confidence here.

PR #1920, opened 2022-08-16
Actions triggered off GitHub releases (Toolbox deploy and Trello actions)

darrellwarde:
# Description

To really start rounding off some of the work achieved with changesets, trigger some actions based off releases being created:

- Deploy the latest version of the Toolbox
- In Trello, create a list for the release, containing a card with the changelog, attached to the release


.github/workflows/release-trello-actions.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  create-trello-list:
+    runs-on: ubuntu-latest
+
+    environment: trello
+
+    steps:
+      - name: Create list
+        run: |
+          list=$( curl --request POST --url "https://api.trello.com/1/lists?name=Done%20-%20${{ github.event.release.name }}&idBoard=${{ secrets.TEAM_GRAPHQL_BOARD }}&key=${{ secrets.API_KEY }}&token=${{ secrets.API_TOKEN }}&pos=bottom" | jq .id )
```
tbwiss, line 17: question (non-blocking): Wdyt of appending `TRELLO_` to the secret keys/names, aka `TRELLO_API_KEY` and `TRELLO_API_TOKEN`? if one looks at the secrets not noticing (or knowing) that the environment is `trello` then it's hard to figure out what for instance `API_KEY` stands for.

tbwiss (dismissed):
Great! Thanks @darrellwarde!

.github/workflows/toolbox-deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    environment: aws
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: refs/tags/${{ github.event.release.tag }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
```
mjfwebb, line 22: I wonder if it might be better to specify a version here. My thinking is that it could possibly break in the future with a new node version, and there might even be caching benefits if we specify a version.

.github/workflows/toolbox-deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Install dependencies
+        run: yarn
+      - name: Build graphql-toolbox
+        run: yarn build
```
mjfwebb, line 26: Not entirely sure, but it could possibly be better to add the `--production` flag: `yarn build --production`

.github/workflows/release-trello-actions.yml
darrellwarde: Good shout, done!

.github/workflows/toolbox-deploy.yml
darrellwarde: Possibly, however, if we're going to do this, we should do it across the board, so maybe not one for this PR?

.github/workflows/toolbox-deploy.yml
darrellwarde: This is already implied, as `build` is a script in `package.json`: `cross-env NODE_ENV=production webpack --config webpack.config.js`

darrellwarde:
Just modified this PR to also create cards in Trello for new feature requests, which closes one of the loops of us successfully managing a backlog and roadmap.

PR #1925, opened 2022-08-17
feat: global authentication

tbwiss:
This is based on the [global authentication RFC](https://github.com/neo4j/graphql/blob/dev/docs/rfcs/rfc-023-global-authentication.md).

Please have a (close) look at the documentation and comment if content is missing or needs improvement.

Note: To verify if a user is (globally) authenticated the existence of a valid JWT token is checked. Checking for the existence of for instance at least one role present in the JWT was not implemented as implementations/requirements may be different depending on the (customer/user) use cases. Happy to discuss this though!


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-1925.surge.sh

packages/graphql/src/schema/resolvers/wrapper.ts
```diff
@@ -147,3 +151,15 @@ async function decodeToken(
...
+    context: SubscriptionContext | Context,
+    plugin: Neo4jGraphQLAuthPlugin | undefined
+): void {
+    const isGlobalAuthenticationEnabled = plugin?.getGlobalAuthenticationEnabled();
+    if (isGlobalAuthenticationEnabled) {
+        if (!context.jwt) {
+            throw new Neo4jGraphQLAuthenticationError("Enabled global authentication requires a valid JWT token");
+        }
+    }
+}
```
angrykoala, line 165: **Suggestion - Non-blocking**: Could we move this to a different file? just to avoid increasing the size of `wrapper.ts`

packages/graphql/tests/integration/directives/auth/jwks-endpoint.int.test.ts
```diff
@@ -32,7 +32,7 @@ import jwksRsa from "jwks-rsa";
 import Neo4j from "../../neo4j";
 import { Neo4jGraphQL } from "../../../../src/classes";
 
-describe("https://github.com/neo4j/graphql/issues/564", () => {
+describe("auth/jwks-endpoint", () => {
```
angrykoala, line 35: :clap:

packages/graphql/tests/integration/rfcs/global-authentication-jwks-plugin.int.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    };
+};
+
+const tearDown = async ({ jwksMock, server }) => {
+    await server.close();
+    await jwksMock.stop();
+};
+
+const createApp = ({ jwksUri }) => {
+    const app = new Koa();
```
angrykoala, line 187: Are we using a full Koa server for this test?
Maybe this should be an `e2e` test instead?

A testing apollo server already exists for this kind of test with `supertest` you could use that or move this koa setup to a separate setup file there.

packages/plugins/graphql-plugin-auth/src/Neo4jGraphQLAuthJWKSPlugin.ts
```diff
@@ -65,6 +68,10 @@ class Neo4jGraphQLAuthJWKSPlugin {
         return result;
     }
 
+    public getGlobalAuthenticationEnabled(): boolean {
```
angrykoala, line 71: Maybe `isGlobalAuthenticationEnabled` is a better name?

packages/plugins/graphql-plugin-auth/src/Neo4jGraphQLAuthJWTPlugin.ts
```diff
@@ -65,6 +74,10 @@ class Neo4jGraphQLAuthJWTPlugin {
         return result;
     }
     /* eslint-enable @typescript-eslint/require-await */
+
+    public getGlobalAuthenticationEnabled(): boolean {
+        return this.globalAuthentication;
+    }
```
angrykoala, line 80: Mmm, I'm not too fond of this duplicate code between both plugins. Maybe we could create a new abstract class `AuthPlugin` inherited by these 2 plugins so we can share this method and property?

angrykoala (changes requested):
Looking good, just a few concerns on testing and the plugin API

packages/graphql/src/schema/resolvers/wrapper.ts
tbwiss: Sure!

packages/graphql/tests/integration/rfcs/global-authentication-jwks-plugin.int.test.ts
tbwiss: I'll try to turn this into an `e2e` test instead.

packages/plugins/graphql-plugin-auth/src/Neo4jGraphQLAuthJWTPlugin.ts
tbwiss: Good point!

angrykoala (dismissed):

packages/graphql/src/schema/resolvers/wrapper-utils.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function verifyGlobalAuthentication(
+    context: SubscriptionContext | Context,
+    plugin: Neo4jGraphQLAuthPlugin | undefined
+): void {
+    const isGlobalAuthenticationEnabled = plugin?.isGlobalAuthenticationEnabled();
+    if (isGlobalAuthenticationEnabled) {
+        if (!context.jwt) {
+            throw new Neo4jGraphQLAuthenticationError("Enabled global authentication requires a valid JWT token");
```
darrellwarde, line 45: This gives away an implementation detail - this should probably just be "Unauthenticated".

darrellwarde (changes requested):
This looks good, I do just have a couple of observations!

packages/plugins/graphql-plugin-auth/src/AuthPlugin.ts
```diff
@@ -0,0 +1,30 @@
...
+export abstract class AuthPlugin {
+    private globalAuthentication: boolean;
+
+    constructor(globalAuthentication: boolean | undefined) {
+        this.globalAuthentication = globalAuthentication || false;
+    }
+
+    public isGlobalAuthenticationEnabled(): boolean {
+        return this.globalAuthentication;
+    }
```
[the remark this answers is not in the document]
tbwiss: Fair enough comment. Looks like my inner Java self took over for a minute and wanted to do it by the books. I think it's a good idea to refactor this into using a member, just like `rolesPath`.

darrellwarde (approved):
LGTM!

PR #1929, opened 2022-08-18
Replace Apoc Full with Apoc Core in tests

darrellwarde:
# Description

As per description - Apoc Full is and will not be fully supported moving forwards.


angrykoala (approved):

darrellwarde:
Not doing this in the end - I've been informed that `apoc-core` wasn't a flag back in 4.3.2 which we test against - maybe we will revisit this when we remove 4.3 support at the end of the year.

PR #1931, opened 2022-08-18
Refactor point element into cypherBuilder

angrykoala:
# Description
Refactors create-point-element to cypherBuilder. It also adds tests to the builder and `CASE` statements


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1931.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1931.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/src/translate/cypher-builder/clauses/Clause.ts
```diff
@@ -40,3 +40,17 @@ export abstract class Clause extends CypherASTNode {
         return new CypherEnvironment(prefix);
     }
 }
+
+/** For testing purposes only */
+export class TestClause extends Clause {
```
darrellwarde, line 45: Do we want this in our "production code"?

packages/graphql/src/translate/projection/elements/create-point-element.test.ts
```diff
@@ -58,9 +58,12 @@ describe("createPointElement", () => {
...
         });
 
-        expect(element).toMatchInlineSnapshot(
-            `"point: (CASE WHEN this.point IS NOT NULL THEN { point: this.point, crs: this.point.crs } ELSE NULL END)"`
-        );
+        expect(element).toMatchInlineSnapshot(`
+            "point: (CASE
+                WHEN this.point IS NOT NULL THEN { point: this.point, crs: this.point.crs }
+                ELSE NULL
+            END)"
```
darrellwarde, line 65: Whilst this formatting makes sense when the `CASE` is an individual clause, I think it looks terrible in the context of a projection.

darrellwarde (dismissed):
This looks great, just a couple of points!

packages/graphql/src/translate/cypher-builder/clauses/Clause.ts
angrykoala: mmm, the only alternative I can think of is creating a new "tests" folder in cypher-builder, but it will only contain these utils classes :confused:

packages/graphql/src/translate/projection/elements/create-point-element.test.ts
angrykoala: mmm, I agree, but it is worth adding logic to change how CASE is formatted?
Between the 2 options, I feel this is the better one, if we were to use it in several places

packages/graphql/src/translate/cypher-builder/clauses/Clause.ts
angrykoala: I ended up moving it to a separate file inside `utils/` at least that way is not in the same place

PR #1934, opened 2022-08-18
Subquery union projection

angrykoala:
# Description
Removes `apoc.coll.sortMulti` in favour of ORDER BY in a subquery

This PR depends on #1918


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1934.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1934.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/src/translate/create-auth-and-params.ts
```diff
@@ -61,8 +61,53 @@ export function createAuthAndParams({
     bind?: Bind;
     where?: { varName: string; chainStr?: string; node: Node };
 }): [string, any] {
```
mjfwebb, line 63: Does this actually return `any`, or is it an object that can be typed? I'm just looking at the early return added with if `(!authPredicate) return ["", {}];`

packages/graphql/src/translate/create-interface-projection-and-params.ts
```diff
@@ -206,37 +209,56 @@ function createInterfaceProjectionAndParams({
...
-    if (field.typeMeta.array) {
-        interfaceProjection = [
-            `WITH ${fullWithVars.join(", ")}`,
-            "CALL {",
-            ...interfaceProjection,
-            `RETURN collect(${field.fieldName}) AS ${field.fieldName}`,
-            "}",
-        ];
+    const optionsInput = resolveTree.args.options as GraphQLOptionsArg | undefined;
+    let withClause: CypherBuilder.With | undefined;
```
mjfwebb, line 222: Might be nicer to make `.With` return undefined directly if it's falsy.

packages/graphql/src/translate/projection/elements/create-projection-subquery.ts
```diff
@@ -0,0 +1,154 @@
+/*
```
mjfwebb, line 1: Similarly to the other PR I believe this file could benefit from a little re-arranging.

.changeset/silent-turtles-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neo4j/graphql": patch
+---
+
+Remove apoc.coll.sortMulti
```
tbwiss, line 5: I think this might need a bit of an extension/tweak. The PR title suggests something else than the changeset message here. But yes removing `apoc.coll.sortMulti` is definitely part of this PR.

tbwiss (approved):
LGTM! Great to remove yet another apoc depedency.

packages/graphql/src/translate/create-interface-projection-and-params.ts
angrykoala: Mmm, I think that would be a bit of an antipattern, as we are defining "with" depending on the options, which may be unrelated to `.with`.

We may improve this a bit if we create a new function with this content and return `With | undefined` but I feel that's a bit of overdoing at this stage

.changeset/silent-turtles-doubt.md
angrykoala: mmm, well, this PR does more changes, but I don't think any of those are really worth mentioning in the changelog

Issue #1937, opened 2022-08-19
Performance regression in relationship queries


angrykoala:
**Description**
There is a performance regression after changes made after `3.4.0`.

With the following typedefs:
```graphql
    type Person {
        name: String!
        born: Int!
        movies: [Movie!]! @relationship(type: "ACTED_IN", direction: OUT)
    }

    type Movie {
        id: ID!
        title: String!
        tagline: String
        actors: [Person!]! @relationship(type: "ACTED_IN", direction: IN)
    }
```

The following query:
```graphql
query DeeplyNestedWithStringFilters_only {
    people(
        where: {
            name_STARTS_WITH: "T"
            movies_SOME: {
                title_CONTAINS: "i"
                actors_SOME: {
                    name_CONTAINS: "i"
                    movies_SOME: {
                        title_NOT: "non-existant"
                    }
                }
            }
        }
    ) {
        name
    }
}
```

In 3.4.0, this query performs with `~692 dbhits`, and  `<100ms` in the benchmarks, generating the following cypher:
```cypher
 MATCH (this:Person)
WHERE this.name STARTS WITH $this_name_STARTS_WITH AND EXISTS((this)-[:ACTED_IN]->(:Movie)) AND ANY(this_movies_SOME IN [(this)-[:ACTED_IN]->(this_movies_SOME:Movie) | this_movies_SOME] WHERE this_movies_SOME.title CONTAINS $this_movies_SOME_title_CONTAINS AND EXISTS((this_movies_SOME)<-[:ACTED_IN]-(:Person)) AND ANY(this_movies_SOME_actors_SOME IN [(this_movies_SOME)<-[:ACTED_IN]-(this_movies_SOME_actors_SOME:Person) | this_movies_SOME_actors_SOME] WHERE this_movies_SOME_actors_SOME.name CONTAINS $this_movies_SOME_actors_SOME_name_CONTAINS AND EXISTS((this_movies_SOME_actors_SOME)-[:ACTED_IN]->(:Movie)) AND ANY(this_movies_SOME_actors_SOME_movies_SOME IN [(this_movies_SOME_actors_SOME)-[:ACTED_IN]->(this_movies_SOME_actors_SOME_movies_SOME:Movie) | this_movies_SOME_actors_SOME_movies_SOME] WHERE (NOT this_movies_SOME_actors_SOME_movies_SOME.title = $this_movies_SOME_actors_SOME_movies_SOME_title_NOT))))
RETURN this { .name } as this
```

[...]

PR #1938, opened 2022-08-19
Add performance test for array filtering

angrykoala:
# Description
This tests #1937


angrykoala:
Will merge directly, as it is only a performance test query

PR #1955, opened 2022-08-22
Fix label injection

angrykoala:
# Description
This PR properly fixes the escape of node labels to avoid cypher injection


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1955.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1955.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/tests/integration/issues/label-cypher-injection.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    let neo4j: Neo4j;
+    let session: Session;
+
+    beforeAll(async () => {
+        neo4j = new Neo4j();
+        driver = await neo4j.getDriver();
+    });
+
+    beforeEach(async () => {
+        session = await neo4j.getSession({ defaultAccessMode: "WRITE" });
```
tbwiss, line 41: question (non-blocking): Is the `session` really needed for this particular test case? (I've never used a label from the context or the JWT token)

packages/graphql/tests/integration/issues/label-cypher-injection.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        const res = await graphql({
+            schema,
+            source: query,
+            contextValue: neo4j.getContextValuesWithBookmarks(session.lastBookmark(), {
+                label: "Movie\\u0060) MATCH",
+            }),
+        });
+
+        expect(res.errors).toBeUndefined();
```
tbwiss, line 83: question (non-blocking): Is it worthwhile to check if `data` contains an empty `movies` array?

tbwiss (approved):
LGTM!

packages/graphql/tests/integration/issues/label-cypher-injection.test.ts
angrykoala: Well, this is just to pass the graphql test with the correct bookmark. I think it is the best way to avoid race conditions.

packages/graphql/tests/integration/issues/label-cypher-injection.test.ts
angrykoala: Not really, this test is only to check if an injection happens

PR #1968, opened 2022-08-24
Unpin peerDependencies

angrykoala:
# Description
Unpins peer dependencies to the latest major release and updates `yarn.lock`


.changeset/fresh-avocados-divide.md
```diff
@@ -0,0 +1,7 @@
+---
+"@neo4j/graphql": patch
+"@neo4j/introspector": patch
```
darrellwarde, line 3: We will get away without releasing the Introspector as this was not released following the accidental pinning. 🙂 

```suggestion
```

darrellwarde (changes requested):
The Introspector shouldn't be released following this. Also, is the intention to also unpin all of the _dependencies_ of the GraphQL Library and OGM? They will also be problematic.

tbwiss:
@angrykoala what about the package tests? they also got pinned with #1914

packages/graphql/package.json
```diff
@@ -99,7 +99,7 @@
         "semver": "7.3.7"
```
tbwiss, line 99: question (non-blocking): What about the dependencies? They were not pinned before, https://github.com/neo4j/graphql/pull/1914/files#diff-aa225a5d16977aaf2faf11389b30a7277a5fdd8a2e36d63bf0625a337bdea6b2L90-L99

packages/ogm/package.json
```diff
@@ -39,8 +39,8 @@
         "prettier": "2.7.1"
```
tbwiss, line 39: question (non-blocking): Same question here, What about the dependencies?

tbwiss (commented):
Just have 2 questions but I think after that it's good to go!

packages/graphql/package.json
angrykoala: Ups, good catch, forgot to push those

tbwiss (dismissed):
LGTM now!

PR #1970, opened 2022-08-24
boyscouting: Toolbox login updates

tbwiss:
As a result of the guidance/direction from the security team regarding storage of the database credentials on the frontend, the Toolbox, it's needed that we remove the password from the stored data. We can and still do store the database username and uri.

This change means that the user has to re-login on each page (re)load. That's obviously not super user-friendly, therefore it's recommended to use a browser vault to store the password and use the auto-prefill functionality for the password login field.

If one opens the Toolbox via Neo4j Desktop, as a Neo4j Desktop app, the login still happens automatically as before, no change in functionality.


packages/graphql-toolbox/src/contexts/auth.tsx
```diff
@@ -79,13 +80,16 @@ export function AuthProvider(props: any) {
...
+            const storedConnectionUrl = Storage.retrieve(LOCAL_STATE_CONNECTION_URL);
+            if (storedConnectionUrl && storedConnectionUsername) {
+                loginPayload = {
+                    username: storedConnectionUsername,
+                    url: storedConnectionUrl,
+                };
             }
         }
-        if (loginPayload && value && !value.driver) {
+        if (loginPayload && loginPayload.password && value && !value.driver) {
```
mjfwebb, line 92: Can you use optional chaining here?

```suggestion
        if (loginPayload?.password && value && !value.driver) {
```

darrellwarde (approved):
Sad, but LGTM!

PR #1971, opened 2022-08-24
Subquery unions

angrykoala:
# Description
This PR moves nested unions into subqueries instead of comprehensions


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1971.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1971.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/tests/tck/tck-test-files/union.test.ts
```diff
@@ -54,7 +54,114 @@ describe("Cypher Union", () => {
...
+                    WITH this_search_0  { __resolveType: \\"Genre\\",  .name } AS this_search_0
+                    RETURN collect(this_search_0) AS this_search_0
+                }
+                CALL {
+                    WITH this
+                    MATCH (this)-[thisthis1:SEARCH]->(this_search_1:\`Movie\`)
+                    WITH this_search_1  { __resolveType: \\"Movie\\",  .title } AS this_search_1
+                    RETURN collect(this_search_1) AS this_search_1
+                }
+                WITH this_search_0 + this_search_1 AS thisvar2
```
darrellwarde, line 95: For this whole union piece, I have to ask why the approach as per below wasn't taken? It feels like Cypher subquery unions are the perfect solution here (and what we use elsewhere). This would also mean you can fully get rid of the unique (0, 1, etc.) variables in each subquery, which I would guess massively simplifies the solution.

```suggestion
                CALL {
                    WITH this
                    MATCH (this)-[thisthis0:SEARCH]->(this_search_0:\`Genre\`)
                    WHERE apoc.util.validatePredicate(NOT ((this_search_0.name IS NOT NULL AND this_search_0.name = $thisparam0)), \\"@neo4j/graphql/FORBIDDEN\\", [0])
                    WITH this_search_0  { __resolveType: \\"Genre\\",  .name } AS this_search_0
                    RETURN collect(this_search_0) AS this_search
                UNION
                    WITH this
                    MATCH (this)-[thisthis1:SEARCH]->(this_search_1:\`Movie\`)
                    WITH this_search_1  { __resolveType: \\"Movie\\",  .title } AS this_search_1
                    RETURN collect(this_search_1) AS this_search
                }
```

darrellwarde (commented):
I'm going to continue reviewing, but wanted to leave this one off comment as food for thought!

packages/graphql/src/translate/projection/subquery/collect-union-subqueries-results.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!isArray) {
+        returnProjection = CypherBuilder.head(returnProjection);
+    }
+
+    const returnClause = new CypherBuilder.Return([returnProjection, resultVariable]);
+
+    return CypherBuilder.concat(withCollectClause, unwindClause, withSortClause, returnClause);
+}
+
+function createWithSortClause(variable: CypherBuilder.Variable, optionsInput: GraphQLOptionsArg): CypherBuilder.With {
```
darrellwarde, line 52: This applies `ORDER BY`, `SKIP` and `LIMIT` right? Worth making the name a bit more descriptive if that is indeed the case?

```suggestion
function createWithSortAndPaginationClauses(variable: CypherBuilder.Variable, optionsInput: GraphQLOptionsArg): CypherBuilder.With {
```

darrellwarde (commented):
It all looks good to me! My own query is really the big one above, I definitely think we should be using `UNION` for things like this.

packages/graphql/src/translate/projection/subquery/collect-union-subqueries-results.ts
angrykoala: 100% agree, for some reason my brain refused to find the word `pagination` when writing this method

packages/graphql/tests/tck/tck-test-files/union.test.ts
angrykoala: That's a good point, I already considered doing this, but I found a few roadblocks when adding sorting and discarded the idea. However, after doing a bit more manual testing it is indeed doable (although, you still need to add a second CALL for the sorting)

All tests seem to be working after this change, so I'll update this PR for using unions

angrykoala: Done, please, review latest changes

packages/graphql/tests/tck/tck-test-files/union.test.ts
```diff
@@ -54,7 +54,108 @@ describe("Cypher Union", () => {
...
+            "MATCH (this:\`Movie\`)
+            CALL {
+                WITH this
+                CALL {
+                    WITH this
+                    MATCH (this)-[thisthis0:SEARCH]->(this_search:\`Genre\`)
+                    WHERE apoc.util.validatePredicate(NOT ((this_search.name IS NOT NULL AND this_search.name = $thisparam0)), \\"@neo4j/graphql/FORBIDDEN\\", [0])
+                    WITH this_search  { __resolveType: \\"Genre\\",  .name } AS this_search
+                    RETURN this_search AS this_search
+                    UNION
```
darrellwarde, line 88: The only suggestion I have now is if it is possible, to dedent the `UNION` by one tab. But realise this might be easier said than done!

```suggestion
                UNION
```

darrellwarde (approved):
Following the change to using a single subquery for unions, this looks great!

Intrigued to hear what the issues were which you mentioned you having with ordering when using a single subquery.

packages/graphql/tests/tck/tck-test-files/union.test.ts
angrykoala: Are you thinking of dedenting union or indenting the queries?
It is doable, but I think it doesn't make much sense, all the examples on Cypher docs are at the same indentation, and if you think about it, union may happen at root level:
```
MATCH(n)
RETURN n
UNION
MATCH(p)
RETURN p as n
```
So you wouldn't be able to dedent anything here :thinking: 

Another option, if you want to signify the diff of the union, is to actually indent union, although it may be a bit weird

Issue #1933, opened 2022-08-18
Sum aggregate on where unexpected behaviour


Hellenic:
**Describe the bug**
When using sum aggregate on where filter, I would expect it to filter based on the same values returned by the same sum aggregate. Currently this does not seem to be the case, and behaviour is similar to the term used without `_SUM`

**Type definitions**
Below truncated version of relevant types

```gql
type Employee {
  employeeId: ID! @id(autogenerate: false)
  firstName: String! @readonly
  lastName: String @readonly
  projects: [Project!]!
    @relationship(
      type: "PARTICIPATES"
      direction: OUT
      properties: "EmployeeParticipationProperties"
    )
}

interface EmployeeParticipationProperties @relationshipProperties {
  allocation: Float
}

type Project {
  projectId: ID! @id(autogenerate: false)
  name: String! @readonly
  description: String
  employees: [Employee!]!
    @relationship(
      type: "PARTICIPATES"
      direction: IN
      properties: "EmployeeParticipationProperties"
    )
  }
```

**To Reproduce**
See expected behaviour and the screenshots below. 

**Expected behavior**
I would expect the screenshot with `allocation_SUM_LTE: 25` to have returned nothing, as sum for each project allocation is over 25 (either 40 or 60 in the screenshot). `allocation_LTE: 25` works correctly.

By copy-pasting the generated cypher queries to Neo4j Browser I got the same results, so I'd guess the generated cypher is incorrect.

**Screenshots**

`allocation_LTE: 25`
![allocation_LTE](https://user-images.githubusercontent.com/1734471/185398187-2c1d5d34-087e-4109-85ab-f2ab5a51d44b.png)
Above screenshot generated this Cypher:
```
Cypher:
MATCH (this:`Employee`)
WHERE apoc.cypher.runFirstColumn(" MATCH (this)-[aggr_edge:PARTICIPATES]->(aggr_node:Project)
RETURN aggr_edge.allocation <= $aggr_edge_allocation_LTE
", { this: this, aggr_edge_allocation_LTE: $aggr_edge_allocation_LTE }, false )
RETURN this { .employeeId, .firstName, .lastName, projectsAggregate: { count: size([(this)-[this_projectsAggregate_this1:PARTICIPATES]->(this_projectsAggregate_this0:`Project`) | this_projectsAggregate_this0]), edge: { allocation: head(apoc.cypher.runFirstColumn("MATCH (this)-[r:PARTICIPATES]->(n:Project)    
[...]

PR #1986, opened 2022-08-26
fix: Sum aggregate on where unexpected behaviour

tbwiss:
The sum calculation was executed as part of the `WITH` statement, it should however be executed as part of the `RETURN` statement. With this change, the sum is correctly calculated over _all_ matching edges rather than as it was before when the sum was calculated _per_ edge. That led to a sum value that was equal to the property value, `allocation` in this case.

The, correct, behavior already exists for the AVG operator, see https://github.com/neo4j/graphql/blob/dev/packages/graphql/src/translate/create-aggregate-where-and-params.ts#L130.

Closes: #1933

# Checklist


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-1986.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-1986.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

mjfwebb (approved):
Looks good to me :+1:

PR #2002, opened 2022-08-30
fix: Remove "Math.floor(Math.random() * Math.random())" from tests

tbwiss:
Based on a Trello card with the following content:
```
From @angrykoala:

"I'm not sure what is the advantage of using random data here 😕, I think random data adds unnecessary complexity to a test.

In fact, I think this setup is already problematic, as Math.random() gives a value between 0 and 0.99999..., multiplied with another Math.random we will always get a value between 0 and 0.999..., using .floor means that someInt1, someInt2, someInt3 are always 0

"IMO, random data in tests make repeatability and readability harder (+ prone to bugs such as this one), without really giving any wider range of testing (with exceptions, of course), at best you get the same test as hard-coded data, at worst you get a test that randomly fails."
```

Addressed in this PR:

- Use the `faker` lib to generate random integers where appropriate.
- Replaced the usage of random integers with hard-coded values for all the "average" and "sum" integration tests.
- Remove any occurrence of `console.log`.


angrykoala:
:clap:

angrykoala (approved):

PR #2003, opened 2022-08-30
Cypher builder refactor

angrykoala:
# Description
Minor refactor of CypherBuilder:
* Remove some unnecessary classes
* Simplifies constructor of clauses
* Moves expressions into a separate folder


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-2003.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-2003.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/src/translate/cypher-builder/variables/Literal.ts
```diff
@@ -18,22 +18,35 @@
...
 /** Represents a literal value, it is not a variable */
-export class Literal<T = LiteralValues | LiteralValues[]> {
+export class Literal<T extends LiteralValue = any> implements CypherCompilable {
     public value: T;
 
     constructor(value: T) {
         this.value = value;
     }
 
     public getCypher(_env: CypherEnvironment): string {
```
tbwiss, line 33: question (non-block): What is the `_env` for?

packages/graphql/src/translate/cypher-builder/variables/Literal.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        expect(queryResult.cypher).toMatchInlineSnapshot(`"5"`);
+    });
+
+    it("Serialize array", () => {
+        const literal = new CypherBuilder.Literal(["hello", 5, "hello"]);
+
+        const testClause = new TestClause(literal);
+
+        const queryResult = testClause.build();
+        expect(queryResult.cypher).toMatchInlineSnapshot(`"[\\"hello\\", 5, \\"hello\\"]"`);
```
tbwiss, line 57: question (non-blocking): Doesn't the `5` also need to be in quotes `"5"`? (It's the case on line 48)

packages/graphql/src/translate/where/property-operations/create-global-node-operation.ts
```diff
@@ -37,7 +37,7 @@ export function createGlobalNodeOperation({
     if (coalesceValue) {
         idProperty = CypherBuilder.coalesce(
             idProperty as CypherBuilder.PropertyRef,
-            new CypherBuilder.Literal(coalesceValue)
+            new CypherBuilder.RawCypher(`${coalesceValue}`) // TODO: move into CypherBuilder.literal
```
tbwiss, line 40: question (non-blocking): move it as part of this PR or a future PR?

packages/graphql/src/translate/where/property-operations/create-where-property-operation.ts
```diff
@@ -83,7 +83,7 @@ export function createWherePropertyOperation({
         if (coalesceValue) {
             propertyRef = CypherBuilder.coalesce(
                 propertyRef as CypherBuilder.PropertyRef,
-                new CypherBuilder.Literal(coalesceValue)
+                new CypherBuilder.RawCypher(`${coalesceValue}`) // TODO: move into CypherBuilder.literal
```
tbwiss, line 86: question (non-blocking): move it as part of this PR or a future PR?

tbwiss (approved):
LGTM overall! Left a few comments to consider.

packages/graphql/src/translate/cypher-builder/variables/Literal.test.ts
angrykoala: Numbers and booleans are not quoted, as those are directly translated in cypher.

E.g.
```
MATCH(m:Movie)
WHERE m.title = "Matrix"
AND m.released = 1999
```

packages/graphql/src/translate/where/property-operations/create-global-node-operation.ts
angrykoala: Right now coalesceValue is processed outside of the cypher builder, once we moved everything to cypher builder we should be able to get rid of that pre processing

packages/graphql/src/translate/cypher-builder/variables/Literal.ts
angrykoala: It is part of the expected interface for an element that can be compiled into cypher. Is the environment that holds reference to the variables.

In this case it is not needed, but because all of these instances will be processed the same way, I added to keep consistency

packages/graphql/src/translate/cypher-builder/variables/Literal.test.ts
MacondoExpress: I think that it's not the case in the line 48, the cypher generated is a string but `5` is not, if it was a string it would have been wrapped as `"\\"5\\""` as for the test line 30.

MacondoExpress (dismissed):
LGTM!

PR #2015, opened 2022-09-01
feat: GHA - code scanning with Github CodeQL

tbwiss:
Testing the CodeQL GHA with this PR to gather information on how it works, how long it runs, how the results look like, etc.

Enabling the [CodeQL analysis](https://github.com/github/codeql-action) was promoted by the security team.

Read more about the used [CodeQL Github action](https://github.com/github/codeql-action) and for a deep dive read about [configuring code scanning](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning).

Notes:
- ~~Long execution time when specifying a CodeQL config file (canceled the workflow after 60mins)~~
- ~~Without the CodeQL config file an analysis runs ca. 10-12 mins~~
- ~~Our (valid!) practice for tests (int and tck) with storing a string with the "secret" when using the Auth plugin is causing 1k+ 'critical' alerts. These can be safely ignored. We may have to manually dismiss all those alerts. These are test files, the alerts can be ignored.~~
- ~~If we run all packages in parallel with a matrix, the execution time is between 3-4 mins for each of the packages. A matrix is not needed as we won't need to build the application.~~
- The code scanning takes around 2-3 minutes
- If the analysis yields any 'critical' alerts the workflow will fail in the "Code scanning results / CodeQL" step. 
- The alerts are also listed in the 'Security' tab under 'code scanning alerts'

Going forward:
- ~~Question: Shall we execute it as part of the PR workflow or as a nightly execution?~~
- ~~An idea could be to run this on the `dev` branch as a nightly (or early morning) GHA action and have the results posted in the team-graphql-feed Slack channel~~
- We should include the CodeQL analysis as part of the pull request workflow


.github/workflows/reusable-codeql-analysis.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          node-version: lts/*
+          cache: yarn
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+      - name: Install dependencies
+        run: yarn
+      - name: Build graphql
+        run: yarn build
+        working-directory: packages/graphql
```
mjfwebb, line 22: Only for the graphql package?

tbwiss: To start with yes. I first want to see how it all works before expanding it to other packages

darrellwarde (approved):

PR #2009, opened 2022-08-31
Fix 1536

angrykoala:
# Description

This PR fixes #1536 by adding the interface subqueries into the projection subquery flow

PR #2016, opened 2022-09-01
feat: additional interface tests

tbwiss:
Adding additional interface tests after the changes made in #2009.
Specifically to check for the ordering of the `@auth` directive code (the `apoc.util.validate()`) in relation to the subquery `CALL`s.


.changeset/dirty-crews-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neo4j/graphql": patch
+---
+
+feat: add additional interface tests to verify the ordering of clauses
```
angrykoala, line 5: mmm, I don't think that new tests should be part of the changelog :thinking:

angrykoala (commented):
Looks good, just a comment regarding changeset

.changeset/dirty-crews-invite.md
tbwiss: Good one! Will remove the changelog

PR #2019, opened 2022-09-01
Clean cypher-where-predicate logic

angrykoala:
# Description
This PR cleans a bit the logic regarding creating where predicates


packages/graphql/src/translate/where/create-where-predicate.ts
```diff
@@ -37,46 +36,25 @@ export function createCypherWherePredicate({
...
 }): CypherBuilder.Predicate | undefined {
-    const mappedProperties = mapPropertiesToOperators({
-        whereInput,
-        targetElement,
-        element,
-        context,
+    const whereFields = Object.entries(whereInput);
+
+    const predicates = whereFields.map(([key, value]): CypherBuilder.Predicate | undefined => {
+        if (key === "OR") {
```
MacondoExpress, line 42: The function `mapPropertiesToPredicates` could potentially be removed by refactoring the following: 
```suggestion
        if (key === "OR" || key === "AND") {
            const nested = value.map((v) => {
                return createWherePredicate({ whereInput: v, element, targetElement, context });
            }
            if (key === "OR") {
                return CypherBuilder.or(...nested);
            }
            return CypherBuilder.and(...nested);
    });
```

MacondoExpress (dismissed):
LGTM, just left a not blocking suggestion 👍

packages/graphql/src/translate/where/create-where-predicate.ts
angrykoala: That's a great point, I'll make this change.
In the last commit, I also further refactored this by extracting the logic inside `if (key === "OR" || key === "AND") {` to a separate function. Let me know what you think of that

MacondoExpress: It looks good to me!

PR #2026, opened 2022-09-02
feat: add rate-limiting to neo-push server

tbwiss:
This popped up during the code scanning.

Info:
```
HTTP request handlers should not perform expensive operations such as accessing the file system, executing an operating
system command or interacting with a database without limiting the rate at which requests are accepted. Otherwise, the 
application becomes vulnerable to denial-of-service attacks where an attacker can cause the application to crash or become 
unresponsive by issuing a large number of requests at the same time.
```

Since neo-push is an example project it's not critical to address this yet it's good practice to follow the guidance even for an example project.

### DIFF
--- a/examples/neo-push/server/package.json
+++ b/examples/neo-push/server/package.json
@@ -5,7 +5,7 @@
     "description": "",
     "main": "./dist/src/index.js",
     "scripts": {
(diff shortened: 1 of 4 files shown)


angrykoala:
It is great to see some love put into our examples, thanks @tbwiss !

angrykoala (approved):

PR #2028, opened 2022-09-02
fix: remove unsused script and stylesheets links in neo-push client

tbwiss:
This popped up in the code scanning (low severity).

Info:
```
Including a resource from an untrusted source or using an untrusted channel may allow an attacker to include
arbitrary code in the response. When including an external resource (for example, a script element or an iframe element)
on a page, it is important to ensure that the received data is not malicious.
```

Can't find any trace that these stylesheets and the imported `highlight.js` library are still used.


mjfwebb (approved):
LGTM :+1:

PR #2030, opened 2022-09-02
Move interface projection to cypher builder

angrykoala:
# Description

This PR refactors interface projections so they are generated with CypherBuilder. This allows to remove duplicate code regarding filtering


tbwiss (commented):
LGTM, nothing really to add apart from maybe double-checking if the comments on your new code are (still) needed.

tbwiss (approved):
LGTM!

packages/graphql/tests/tck/directives/auth/projection-interface-relationships.test.ts
```diff
@@ -97,12 +97,12 @@ describe("Auth projections for interface relationship fields", () => {
...
                 WITH this
-                MATCH (this)-[:ACTED_IN]->(this_Movie:Movie)
+                MATCH (this)-[thisthis0:ACTED_IN]->(this_Movie:\`Movie\`)
                 RETURN { __resolveType: \\"Movie\\", runtime: this_Movie.runtime, title: this_Movie.title } AS actedIn
                 UNION
                 WITH this
-                MATCH (this)-[:ACTED_IN]->(this_Series:Series)
-                CALL apoc.util.validate(NOT ((this_Series.episodes IS NOT NULL AND this_Series.episodes = $this_Seriesauth_param0)), \\"@neo4j/graphql/FORBIDDEN\\", [0])
+                MATCH (this)-[thisthis1:ACTED_IN]->(this_Series:\`Series\`)
+                WHERE apoc.util.validatePredicate(NOT ((this_Series.episodes IS NOT NULL AND this_Series.episodes = $thisparam0)), \\"@neo4j/graphql/FORBIDDEN\\", [0])
```
darrellwarde, line 105: Yeah this is super slick, I love it! 👏

packages/graphql/src/translate/create-interface-projection-and-params.ts
```diff
@@ -19,185 +19,51 @@
...
+import { getRelationshipDirection } from "../utils/get-relationship-direction";
 import * as CypherBuilder from "./cypher-builder/CypherBuilder";
 import { addSortAndLimitOptionsToClause } from "./projection/subquery/add-sort-and-limit-to-clause";
 import { compileCypherIfExists } from "./cypher-builder/utils/utils";
+import type { Node } from "../classes";
+import { createWherePredicate } from "./where/create-where-predicate";
+import { AUTH_FORBIDDEN_ERROR } from "../constants";
 
-function createInterfaceProjectionAndParams({
+export default function createInterfaceProjectionAndParams({
```
darrellwarde, line 37: Any reason for the default export? It feels to me like we're moving away from this direction in most cases.

packages/graphql/tests/tck/directives/auth/arguments/where/interface-relationships/implementation-where.test.ts
```diff
@@ -168,12 +168,12 @@ describe("Cypher Auth Where", () => {
...
                 WITH this
-                MATCH (this)-[:HAS_CONTENT]->(this_Comment:Comment)
+                MATCH (this)-[thisthis0:HAS_CONTENT]->(this_Comment:\`Comment\`)
                 RETURN { __resolveType: \\"Comment\\" } AS content
                 UNION
                 WITH this
-                MATCH (this)-[:HAS_CONTENT]->(this_Post:Post)
-                WHERE (exists((this_Post)<-[:HAS_CONTENT]-(:\`User\`)) AND all(auth_this0 IN [(this_Post)<-[:HAS_CONTENT]-(auth_this0:\`User\`) | auth_this0] WHERE (auth_this0.id IS NOT NULL AND auth_this0.id = $this_Postauth_param0)))
+                MATCH (this)-[thisthis1:HAS_CONTENT]->(this_Post:\`Post\`)
+                WHERE (exists((this_Post:\`Post\`)<-[:HAS_CONTENT]-(:\`User\`)) AND all(thisthis2 IN [(this_Post:\`Post\`)<-[:HAS_CONTENT]-(thisthis2:\`User\`) | thisthis2] WHERE (thisthis2.id IS NOT NULL AND thisthis2.id = $thisparam0)))
```
darrellwarde, line 176: I know we're currently suffering from this bug with `EXISTS`, but I feel like we should shortly (or maybe even already) be in a position where we can move to using `EXISTS` clauses across the board! 🥳

darrellwarde (approved):
This looks really good, some nice improvements on top of the refactor! It might even address the bug I have fixed in my PR, time will tell when I merge!

packages/graphql/src/translate/create-interface-projection-and-params.ts
angrykoala: Oh, it is just because that was what were being exported, and to reduce the scope of the refactor I kept it as such

Issue #1287, opened 2022-04-21
Querying for connections fails when using interfaces in type definition


megatrond:
**Describe the bug**
When using the `where: connection` filter option, and the connection is to an interface, the query fails

**Type definitions**
```graphql
type Screening {
    id: ID! @id
    title: String
    beginsAt: DateTime!
    movie: ScreenableMeta! @relationship(type: "SCREENS_MOVIE", direction: OUT)
}

interface ScreenableMeta {
    id: ID! @id
    spokenLanguage: String!
    subtitlesLanguage: String!
    premiere: DateTime!
    locale: Locale!
}

type NorwegianScreenableMeta implements ScreenableMeta {
    id: ID! @id
    spokenLanguage: String!
    subtitlesLanguage: String!
    premiere: DateTime!
    locale: Locale!
    ediNumber: String!
}
```
and query:

```graphql
query queryScreenings {
  screenings(
    where: {
      movieConnection: {
        node: {
          id: "52a26830-86f8-47b5-bcb5-4ebbf1df2a01"
        }
      }
    }
  ) {
    beginsAt
    movie {
      id
    }
  }
}
```

**To Reproduce**

1. Given the schema above and query above
2. The server returns an `INTERNAL_SERVER_ERROR` with the following stacktrace

```json
{
  "errors": [
    {
      "message": "Cannot read property 'name' of undefined",
      "locations": [
        {
          "line": 6,
          "column": 3
        }
      ],
      "path": [
        "screenings"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: Cannot read property 'name' of undefined",
            "    at /Users/trond/projects/tenants/api/node_modules/@neo4j/graphql/dist/translate/where/create-where-and-params.js:138:55",
            "    at Array.forEach (<anonymous>)",
            "    at reducer (/Users/trond/projects/tenants/api/node_modules/@neo4j/graphql/dist/translate/where/create-where-and-params.js:135:41)",
            "    at Array.reduce (<anonymous>)",
            "    at createWhereAndParams (/Users/trond/projects/tenants/api/node_modules/@neo4j/graphql/dist/translate/where/create-where-and-params.js:193:60)",
            "    at translateTopLevelMatch (/Users/trond/projects/tenants/api/node_modules/@neo4j/graphql/dist/translate/translate-top-level-match.js:76:61)",
[...]

PR #2038, opened 2022-09-05
Add integration test for #1287

angrykoala:
# Description
Issue #1287 was fixed in `3.6.0` however, it was missing an integration test to avoid regression


tbwiss (approved):
LGTM!

PR #2039, opened 2022-09-05
fix: docs - include Toolbox page in content navigation

tbwiss:
The Toolbox page is not part of the content navigation so if one opens the Toolbox page the navigation in the left-hand menu disappears.


docs/modules/ROOT/pages/index.adoc
```diff
@@ -18,7 +18,7 @@ ifdef::backend-pdf[]
 Documentation license: <<license, Creative Commons 4.0>>
 endif::[]
 
-> This is the documentation for the Neo4j GraphQL Library version 3.0, authored by the Neo4j GraphQL Team.
+> This is the documentation for the Neo4j GraphQL Library version 3.7, authored by the Neo4j GraphQL Team.
```
tbwiss, line 21: We clearly missed updating this many times over.. Is it worth stating the version?

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-2039.surge.sh

docs/modules/ROOT/pages/index.adoc
mjfwebb: It would be great to do a find and replace in a workflow build step if the version can be obtained from somewhere else (package.json?)

darrellwarde: I honestly think we should just get rid of the reference to the version here.

```suggestion
> This is the documentation for the Neo4j GraphQL Library, authored by the Neo4j GraphQL Team.
```

PR #2042, opened 2022-09-06
Increase reviewdog memory limit to 4096 GB

mjfwebb:
There is an issue where reviewdog/action-actionlint is running out of memory in the JS heap


tbwiss (approved):
LGTM!

PR #2043, opened 2022-09-06
Set node_options in reviewdog inputs

mjfwebb:
Specify the node_options as input to reviewdog instead of as env


tbwiss (approved):
Good one!

PR #2045, opened 2022-09-06
Tests to ensure duplicate nodes not created, fix for interface projection

### LINKED ISSUE #832: Duplicate Nodes Created with Nested 'Connect' when using Interfaces
**Describe the bug**
When using the autogenerated `create` mutations, with `connect` on nested relationship fields - if the type the relationship refers to is an interface rather than type, and you are attempting to create multiple nodes in a single call, spurious duplicate nodes are created.

**Type definitions**

**Note:** This is the same schema that generates another (I think unrelated) bug: https://github.com/neo4j/graphql/issues/847

```gql
interface Entity {
    id: String!
}

type Person implements Entity {
    id   : String! @unique
    name : String!
}

type Place implements Entity {
    id       : String! @unique
    location : Point!
}

type Interaction  {
    id       : ID! @id
    kind     : String!
    subjects : [Entity!]! @relationship(type: "ACTED_IN", direction: IN )
    objects  : [Entity!]! @relationship(type: "ACTED_IN", direction: OUT)
}
```

**To Reproduce**

Run the following mutation against the graphql API using an empty neo4j database:

```graphql
mutation {
  createPeople(input: [
    { id: "adam", name: "Adam" },
    { id: "eve",  name: "Eve" },
    { id: "cain", name: "Cain"},
    { id: "abel", name: "Abel" }, 
  ]) {
    people {
      id
    }
  }
  createInteractions(input: [{
    subjects : { connect: { where: { node: { id_IN: ["adam", "eve"] }}}},
    kind     : "PARENT_OF",
    objects  : { connect: { where: { node: { id_IN: ["cain"] }}}}
  },{
    subjects : { connect: { where: { node: { id_IN: ["adam", "eve"] }}}},
    kind     : "PARENT_OF",
    objects  : { connect: { where: { node: { id_IN: ["abel"] }}}}
  }]) {
    info { nodesCreated }
    interactions { id }
  }
}
```

**Expected behavior**

`createInteractions` should create exactly 2 nodes - instead it creates 3 nodes.

The output `createInterations.interactions` list contains just 2 ID strings - however `info.nodesCreated` is set to 3 - and examining the graph shows indeed that an extra node has been incorrectly created.

Note that if the two nodes are created with 2 separate calls to `createInteractions` (with an input array of length 1 in both cases), the correct behavior is observed. Additionally, modifying the type definitions on the `subjects` and `objects` relationship fields to be `Person` rather than `Entity` results in the correct behavior. Neither of these workarounds should be required.

**Screenshots**

Incorrect resultant graph structure from above mutation:
image

Expected Graph Structure:
image

-   OS: Linux (Debian 10)
-   Version: @neo4j/graphql@2.5.5, graphql@16.2.0, apollo-server@3.6.2
-   Node.js version: v14.15.0
-  Neo4j DB Version: 4.4.0

darrellwarde:
# Description

This PR adds tests for #832... I'm quite confused about the situation. I could have sworn I was able to reproduce the bug, but now I have a variety of tests where I cannot reproduce the behaviour.

However, one thing that is fixed is that selecting interface relationships following a batch create will now not result in an error, by adding a unique prefix to each relationship return.

Additionally, addresses a couple of other small things:

- Pins `jest` to `>=28.0.0 <29.0.0-0` because `ts-jest` is not yet compatible
- Renames a couple of test files to end with `.int.test.ts`

# Issue

Closes #832.


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-2045.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-2045.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

tbwiss (approved):
LGTM!

PR #2052, opened 2022-09-07
Remove extra labels in Auth

angrykoala:
# Description
Removes labels that are not needed and may lead to bugs in auth generation


darrellwarde (approved):
LGTM!

PR #2061, opened 2022-09-08
Further fixes for 5.0.0 compatibility

darrellwarde:
# Description

There is still one issue left which I am working through, but this PR:

- Allows the `neo4j-driver@5.0.0` to be used as a peer dependency
- Generates `CREATE CONSTRAINT` calls dynamically based on database version
- Removes some use of index functions from tests

### DIFF
--- a/packages/graphql/tests/integration/directives/fulltext.int.test.ts
+++ b/packages/graphql/tests/integration/directives/fulltext.int.test.ts
@@ -106,7 +106,7 @@ describe("@fulltext directive", () => {
         const session = driver.session({ database: databaseName });
 
         const cypher = `
-            CALL db.indexes() yield
+            SHOW INDEXES yield
                 name AS name,
                 type AS type,
                 entityType AS entityType,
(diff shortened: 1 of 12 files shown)


packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -570,7 +564,7 @@ describe("@fulltext directive", () => {
         const session = driver.session({ database: databaseName });
 
         const cypher = `
-            CALL db.indexes() yield
+            SHOW INDEXES yield
```
MacondoExpress, line 567: Do we still have to support the 4.1 version?
I think that `SHOW INDEXES` is available only from the 4.2 
https://neo4j.com/docs/cypher-manual/4.2/deprecations-additions-removals-compatibility/

darrellwarde: Nope, only 4.3 onwards. 🙂

MacondoExpress: Cool!

MacondoExpress (approved):
LGTM!!

PR #2065, opened 2022-09-09
Fix: duplicate nodes created if creating following the connection of interface relationships

### LINKED ISSUE #832: Duplicate Nodes Created with Nested 'Connect' when using Interfaces

darrellwarde:
# Description

So it turns out that having `UNION` between our connect, update and disconnect operations for interface relationships affects cardinality, which can cause subsequent commands to be executed twice. This simply removed the `UNION`, making them independent subqueries.

# Issue

Closes #832


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-2065.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-2065.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/src/translate/create-connect-and-params.ts
```diff
@@ -436,7 +436,7 @@ function createConnectAndParams({
                     res.params = { ...res.params, ...subquery.params };
                 }
             });
-            res.connects.push(subqueries.join("\nUNION\n"));
+            res.connects.push(subqueries.join("\n}\nCALL {"));
```
tbwiss, line 439: question (non-blocking): Why is this not also `\n}\nCALL {\n\t` like in the 'disconnect' and 'update' file?
I also see some tck tests where the `CALL` and `WITH ...` are in the same line rather than the `WITH ...` being on a new line.

tbwiss (approved):
Just have one comment but apart from that this is good to go IMO!

PR #2068, opened 2022-09-09
Add optional `EXPLAIN` to `TCK` tests

darrellwarde:
# Description

Adds an optional `EXPLAIN` to TCK tests via the `VERIFY_TCK` environment variable (set to anything) - verifies the validity of the generated Cypher.


angrykoala (approved):

.github/workflows/reusable-api-library-tests.yml
```diff
@@ -45,10 +45,10 @@ jobs:
       - name: Overwrite GraphQL version
         run: yarn up --exact graphql@${{ matrix.graphql-version }}
       - name: Run TCK tests
-        run: yarn --cwd packages/graphql run test:tck --coverage
+        run: yarn --cwd packages/graphql run test tck --coverage
```
mjfwebb, line 48: Is this an intentional change?

angrykoala:
I think I found the leak issue. Basically we were creating a neo4j driver without closing it when the TCK test where ran with `VERIFY_TCK`. I changed it so the driver is created and destroyed on each `translate` operation. We can probably improve this with global hooks to avoid re-creating the driver so many times, but at least this seems to fix the memory leak

darrellwarde:
> I think I found the leak issue. Basically we were creating a neo4j driver without closing it when the TCK test where ran with `VERIFY_TCK`. I changed it so the driver is created and destroyed on each `translate` operation. We can probably improve this with global hooks to avoid re-creating the driver so many times, but at least this seems to fix the memory leak

Hm, I thought because I had defined the `neo4j` variable outside of the function, that this wouldn't be the case! Turns out I may have been wrong. 🙃

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-2068.surge.sh

PR #2081, opened 2022-09-12
Fulltext index compatibility for ID fields

### LINKED ISSUE #2080: Fulltext-Index for properties of type "ID"
I want to define a fulltext index on several properties. One of its properties is of type ID. The following exception is thrown when starting the server:
````
Error: Node 'ProgrammeItem' @fulltext index contains invalid index 'ProgrammeItemSearchIndex' cannot use find String field 'id'
````

When creating a fulltext index, handle properties of type ID the same as properties of type String.

**Additional context**
````gql
type ProgrammeItem {
  id: ID!
  name: String!
  productTitle: String!
  originalTitle: String
  workingTitle: String!
}

extend type ProgrammeItem
  @fulltext(
    indexes: [
      {
        name: "ProgrammeItemSearchIndex"
        fields: ["id", "name", "productTitle", "originalTitle", "workingTitle"]
      }
    ]
  )
````

### PR DESCRIPTION
**Pull request by darrellwarde**
# Description

Allows the `@fulltext` index to be used to specify `ID` fields, as these are always serialised as strings in the database.

# Issue

Closes #2080

### CODE REVIEW

**Review by tbwiss — APPROVED** [review]
LGTM!

File: packages/graphql/tests/integration/directives/fulltext.int.test.ts [hunk 1]
```diff
@@ -647,4 +647,99 @@ describe("@fulltext directive", () => {
...
+            expect(record.type).toBe("FULLTEXT");
+            expect(record.entityType).toBe("NODE");
+            expect(record.labelsOrTypes).toEqual([type.name]);
+            expect(record.properties).toEqual(["id"]);
+
+            await session.run(`
+                CREATE (:${type.name} { id: "${id}" })
+            `);
+        } finally {
+            await session.close();
```
angrykoala, line 719: I got the feeling that we are testing 2 different things here. The first one is checking for the fulltext index to exists, while the second is an actual fulltext graphql query.
Could it be worth splitting this into 2 separate test cases?

angrykoala (approved):
One comment regarding testing

packages/graphql/tests/integration/directives/fulltext.int.test.ts
darrellwarde: I guess it could be, but it follows the pattern of every single other test in this file (I literally copied and pasted), and if I'm being quite honest, I don't have the energy to update every single test! 😆

PR #2098, opened 2022-09-15
Refactor connections into cypher builder

angrykoala:
# Description
Refactors `createConnectionAndParams` into using CypherBuilder using projection subqueries


packages/graphql/tests/tck/issues/1760.test.ts
```diff
@@ -125,48 +125,48 @@ describe("https://github.com/neo4j/graphql/issues/1760", () => {
         expect(formatCypher(result.cypher)).toMatchInlineSnapshot(`
             "MATCH (this:\`ApplicationVariant\`)
             WHERE this.current = $param0
-            WITH *,  apoc.cypher.runFirstColumnSingle(\\"MATCH (this)<-[:HAS_BASE]-(n:BaseObject) RETURN n.id\\", {this: this, auth: $auth}) AS relatedId
```
darrellwarde, line 128: From what I can tell, this PR has broken a key performance improvement for a customer, where we evaluate the sorting of a custom Cypher field at the _beginning_ of execution. In this test you can see that this has once again been moved down to the end of the Cypher query.

packages/graphql/src/translate/cypher-builder/clauses/Clause.ts
```diff
@@ -40,4 +42,10 @@ export abstract class Clause extends CypherASTNode {
     private getEnv(prefix?: string | EnvPrefix): CypherEnvironment {
         return new CypherEnvironment(prefix);
     }
+
+    /** Custom log for console.log */
+    [util.inspect.custom](depth: number, opts: InspectOptions): string {
+        const cypher = padBlock(this.build().cypher);
+        return `<Clause ${this.constructor.name}> """\n${cypher}\n"""`;
+    }
```
darrellwarde, line 50: Somewhat thrown off by this - can you explain how this works and where it is used?

darrellwarde (changes requested):
From what I can tell, new code looks good!

However, we have a regression in here which will need addressing before we merge. 😢

packages/graphql/src/translate/cypher-builder/expressions/procedures/apoc/date.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import type { CypherEnvironment } from "../../../Environment";
+import { CypherASTNode } from "../../../CypherASTNode";
+import type { PropertyRef, Variable } from "../../../CypherBuilder";
+
+export class ConvertFormat extends CypherASTNode {
+    private temporalParam: Variable | PropertyRef;
+    private currentFormat: string;
+    private convertTo: string;
```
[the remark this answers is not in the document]
angrykoala: mm I don't think so. `convertTo` supports a wide array of formats as well as some custom date templates,  so an enum would not be enough I'm afraid (https://neo4j.com/labs/apoc/4.4/overview/apoc.date/apoc.date.convertFormat/)

packages/graphql/src/translate/cypher-builder/expressions/procedures/apoc/apoc.ts
```diff
@@ -17,5 +17,16 @@
...
+
+export type ApocPredicate = ValidatePredicate;
+export type ApocExpr = RunFirstColumn | ConvertFormat;
+
 export { RunFirstColumn } from "./RunFirstColumn";
 export { ValidatePredicate } from "./ValidatePredicate";
+
+export const date = {
+    convertFormat,
+};
```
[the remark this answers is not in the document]
angrykoala: That's true, but if you are ok with that I'll prefer to defer that change for the overall improvements of the CypherBuilder, to keep the scope of this PR to a minimum

packages/graphql/src/translate/cypher-builder/clauses/Clause.ts
angrykoala: :magic_wand: Magic

This is similar to the `toString` method, but it is basically to customize how this class is shown in `console.log`

The idea is that, right now, for debugging, if you do:
```ts
const match = new CypherBuilder.Match(..)
console.log(match)
```

You get  a really unhelpful representation of the class, with this method, it will return the actual cypher that this CypherBuilder statement will generate:
```
console.log(match)
<Clause Match>
    MATCH (this0:Actor)-[:ACTED_IN]->(this1:Movie)
```

packages/graphql/tests/tck/issues/1760.test.ts
angrykoala: Good point. I'll write a quick perf. benchmark to see if I can reproduce the problem

packages/graphql/src/translate/translate-read.ts
```diff
@@ -196,8 +196,8 @@ function translateRootField({
 
     const cypher: string[] = [
         subStr.matchAndWhereStr,
-        subStr.projectionSubqueries,
         ...(sortOffsetLimit.length > 1 ? sortOffsetLimit : []),
+        subStr.projectionSubqueries,
```
darrellwarde, line 200: This is all it was in the end?! 😆 At least it was an easy fix!

darrellwarde (approved):

packages/graphql/src/translate/translate-read.ts
angrykoala: Not only that, this seems to have fixed other related issues with `@cypher`

PR #2121, opened 2022-09-21
RFC: Authentication and Authorization re-architecture

darrellwarde:
Summary of proposed changes:

- Breaking out the existing `@auth` functionality into two new directives, `@authentication` and `@authorization`
- Expanding the filtering capabilities available, by implementing the new filtering proposals from #2730
- Introducing a new directive, `@accessControl`, which deals with access to the root-level operations on types
- Offering some options on how we might want to implementing authorization on Subscription events
- A method for our users to define their JWT payload structure, so that they can perform matches on any JWT claim
- Move the functionality back to be a core feature of the library, through the use of the `jose` browser-compatible library

I realise that this is a lot to process, but I think it is sensible to present this work as whole rather than as several different proposals. Happy to take feedback on how this information might be better presented.


docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,859 @@
...
+  allow: [AccessControlAllowRule!]
+) on OBJECT | FIELD_DEFINITION | INTERFACE | SCHEMA
+```
+
+The proposal is that this will _only_ be in relation to top-level operations, and not when performing a nested operation on a type. Authorization rules as described below should be used for this use case.
+
+Giving an example of using the above, to prevent a `User` type having its delete Mutation called by anyone but an administrator, one could do:
+
+```gql
+type User @accessControl(allow: [{ operations: [DELETE], where: { jwtPayload: { roles: { single: "ADMIN" } } } }]) {
```
MacondoExpress, line 168: What are the limitations that block us to have it as part of the `@authorization` directive?

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,859 @@
...
+  READ
+  CREATE
+  UPDATE
+  DELETE
+  SUBSCRIBE
+}
+
+input AccessControlAllowRule {
+  operations: [AccessControlOperation!]! = [READ, CREATE, UPDATE, DELETE, SUBSCRIBE]
+  requireAuthentication: Boolean! = true
```
MacondoExpress, line 154: Is it possible to have unauthenticated users and still use @accessControl?

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,859 @@
...
+
+In the new solution, the following type definitions would be strictly equivalent:
+
+```gql
+type User
+  @authorization(
+    filter: [
+      {
+        operations: [READ]
+        requireAuthentication: false
```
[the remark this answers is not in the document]
authorisation: Woops...

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,859 @@
...
+type AuthConfig = {
+  secret: string | JWKS
+  verify: boolean;
+  jwtPayload: Schema;
+}
+```
+
+`AuthConfig.secret` will assume symmetric type if passed a string, or other types by passing in different objects. This will allow us to use the same configuration for a variety of secret types. `jose` also supports SPKI encoded RSA keys.
+
+`AuthConfig.verify` will be `true` by default, but can be set to `false` if the desired behaviour is to decode only.
```
Liam-Doodson, line 31: Can we consider defining `verify`/`jwtPayload` on a schema directive so that this information is defined in the same place it is relevant?

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,859 @@
...
+}
+```
+
+This would perform a check of the JWT claim at the beginning of the resolver to check that the user has the appropriate role.
+
+#### Authorization
+
+The `@authorization` directive will not have a static definition, as the definition will be different depending on which location it has been applied in.
+
+The usage of the `@authorization` directive implies that authentication is required.
```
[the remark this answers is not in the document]
darrellwarde: This is correct - authorization implies authentication, but this requirement can be conditionally disabled by setting the `requireAuthentication` to false... It's pretty messy. 😢

docs/rfcs/rfc-034-auth.md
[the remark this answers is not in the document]
darrellwarde: As above, this is due to the need to disable authentication conditionally for some specific _authorization_ rules - there's really no way around this sadly.

docs/rfcs/rfc-034-auth.md
darrellwarde: No limitations, apart from I think it will be confusing because this is assessed only once, on entry to the relevant resolver. Authorization is assessed in a nested manner, whenever you hit a specified. I have actually added an alternative suggestion for how we might do this with limited input ourselves, based on what you said earlier!

docs/rfcs/rfc-034-auth.md
darrellwarde: Yes - perhaps they only want to check that a value exists in the context, for instance.

docs/rfcs/rfc-034-auth.md
darrellwarde: I'm not sure about this... `jwtPayload` has some nice TypeScript types which we could leverage - I think this is just too complex to put into type definitions.

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,889 @@
...
+
+* Each lifecycle hook has different operation compatibility:
+  * Filter:
+    * READ
+    * UPDATE
+    * DELETE
+    * CREATE_RELATIONSHIP
+    * DELETE_RELATIONSHIP
+  * Pre validate:
+    * READ
```
Liam-Doodson, line 338: Is it worth considering allowing `validate.pre`/`filter` rules to be applied for CREATE operations? If not it is impossible to prevent nested data from being accessed in CREATE mutations

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,889 @@
...
+  users(where: { name: "Bob" }) {
+    id
+    name
+  }
+}
+```
+
+This will generate Cypher and parameters along the lines of the following:
+
+```cypher
```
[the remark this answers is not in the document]
darrellwarde: Indeed, I think this is an example of exactly what we were discussing earlier, where we need to produce checks right before data return to ensure users have access!! Good find.

docs/rfcs/rfc-034-auth.md
darrellwarde: Even with this being the case, `CREATE` does not imply `READ` - if a user has a rule for `CREATE` _and_ `READ`, then we will apply the filter and pre-validate rules before returning data. If they have a rule only for `CREATE`, then what is documented here applies.

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,889 @@
...
+MATCH (this:User)
+WHERE this.name = $name
```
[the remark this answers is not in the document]
darrellwarde: I think this should be the intention, I don't see the need for a mixed economy.

For some reason, I can't even imagine the kind of field-level authorization rules people might want to apply!! Why would they not just define it at the object level? 🤔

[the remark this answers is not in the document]
MacondoExpress: At this moment the field level implies that the authorization is applied only if the field is impacted, for instance, I can read the user but if I'm trying to access his very secret property an error is raised.

docs/rfcs/rfc-034-auth.md
```diff
@@ -0,0 +1,917 @@
...
+  roles: StringListWhere
+}
+
+input UserWhere {
+  OR: [UserWhere!]
+  AND: [UserWhere!]
+  NOT: UserWhere
+  id: StringWhere
+  name: StringWhere
+}
```
a-alle, line 290: would be interesting to see an example where the `User` type has a relationship

PR #2136, opened 2022-09-26
RFC-028: Surface executed cypher

tbwiss:
This is still WIP!

Show the executed cypher and its params for each query.

### DIFF
--- a/docs/rfcs/rfc-028-executed-cypher.md
+++ b/docs/rfcs/rfc-028-executed-cypher.md
@@ -24,7 +24,7 @@ Shall we have the ability to do `dry-run`s? That is not possible at the moment. 
 
 Expose a hook/callback which is called on each execution of a query. Similar to a logger component like the [log-level lib](https://www.npmjs.com/package/loglevel).
 
-The drawback is that linking a GraphQL query execution to a cypher query "callback" can be tricky and potentially be out of sync.
+The drawback is that linking a specific GraphQL query execution to a cypher query "callback" requires those to either be in sync or use a timestap or id for correct ordering/association.
 
 ## Usage Example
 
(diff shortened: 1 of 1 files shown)


helenesanche:
**Idea for this feature**
The GraphQL Toolbox in its current shape is a great tool for first time users to explore the library features, explore queries, debugging, but once people have gained a better understanding and confidence in using the library they might not need the Toolbox anymore. 
By adding a way to surface the cypher translation of their GraphQL queries we can _**retain**_ **_more users_** who might be interested in understanding why some queries are slow, and by giving them additional information we can create a user journey where a developer uses the Toolbox to debug a problem or a slow query, then they reach out to us on Discord by sending a link to the Toolbox. This was we can make the Toolbox part of the developer experience of Neo4j GraphQL community.

helenesanche:
This looks great to me, thanks @tbwiss ! 
One addition I suggest is to show a one-time pop-up to users who login after we've implemented the feature and haven't previously login, so we can educate them on the new changes. 
The pop-up can say something along the lines of:
`"You can see the cypher executed by enabling this new feature in Settings" `

tbwiss:
@helenesanche Having a thought about this one-time pop-up. We can leverage Canny (which we already have in place), or more precise the [Canny changelog widget](https://developers.canny.io/install/changelog), for that. The integration of that changelog widget with the Toolbox should not be challenging and since we already use Canny we can leverage it for one more task.
The content of the changelog widget can be drafted, published and removed via the Canny UI and without having to change any code in the Toolbox. Needless to say it also helps for future "releases" and "feature notifications".
What do you think?

tbwiss:
Closing this as we need to re-think the approach here. The new approach will be handled in a new RFC

PR #2168, opened 2022-10-03
Update changesets config.json

darrellwarde:
# Description

An update to changesets means that private packages can now be tagged - this should fully automate the Toolbox deployment flow! 🎉 

Depends on dependency update in #2163.


tbwiss (approved):
Nice 🎉 !

PR #2169, opened 2022-10-03
RFC: Subscribe to connections

angrykoala:
# Description
This RFC describes a proposal on subscribing to nodes connected


docs/rfcs/rfc-007-subscriptions.md
```diff
@@ -276,6 +276,19 @@ UNWIND meta AS m
...
 ```
 
+##### Examples of subscriptions to relationships
+Get the source and target nodes of a relationship in subscription:
+
+
+ ```graphql   
+ personConnected {  // Non-directed, another event will be triggered for movies
+    to { Movie | Actor } // Target node, union is the list of all types related to a person
+    from { Actor }
```
darrellwarde, line 286: An initial thought I had following yesterday's conversation is that `to` and `from` imply direction. May we be better with `connectedNode` and perhaps the name of the node, `person`?

angrykoala:
Will merge as work on this RFC has already started. Any further refinement can be done in a separate PR

PR #2174, opened 2022-10-04
RFC: Apollo Federation

darrellwarde:


docs/rfcs/rfc-028-federation.md
```diff
@@ -0,0 +1,84 @@
+# Federation
+
```
angrykoala, line 2: A brief introdcution of the use case and why it is needed could be useful here

darrellwarde:
This is very outdated with implementation being quite different - closing.

Issue #2147, opened 2022-09-28
Sort priority is not maintained between edge and node sort fields


angrykoala:
**Describe the bug**
The following 2 queries, return connection nested fields, ordered by both, edge and node fields, but the order is not being affected in the resulting cypher:

```graphql
query MoviesSortedByRelationship {
  movies {
    actorsConnection(sort: [{edge: {year: DESC}}, {node: {name: ASC}}]) {
      edges {
        year
        node {
          name
        }
      }
    }
  }
}
```

```graphql
query MoviesSortedByRelationship {
  movies {
    actorsConnection(sort: [{node: {name: ASC}},{edge: {year: DESC}}]) {
      edges {
        year
        node {
          name
        }
      }
    }
  }
}
```

However, both queries yield the same following Cypher:
```cypher
MATCH (this:`Movie`)
CALL {
    WITH this
    MATCH (this)<-[this_connection_actorsConnectionthis0:ACTED_IN]-(this_Person:`Person`)
    WITH this_connection_actorsConnectionthis0, this_Person
    ORDER BY this_connection_actorsConnectionthis0.year DESC, this_Person.name ASC
    WITH { year: this_connection_actorsConnectionthis0.year, node: { name: this_Person.name } } AS edge
    WITH collect(edge) AS edges
    WITH edges, size(edges) AS totalCount
    UNWIND edges AS edge
    WITH edge, totalCount
    ORDER BY edge.year DESC, edge.node.name ASC
    WITH collect(edge) AS edges, totalCount
    RETURN { edges: edges, totalCount: totalCount } AS actorsConnection
}
RETURN this { actorsConnection: actorsConnection } as this
```

**Expected result**
The `ORDER BY` priority should be maintained as defined in the `sort` input field

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @angrykoala! :pray: We will now prioritise the bug and address it appropriately.

neo4j-team-graphql:
Many thanks for raising this bug report @angrykoala. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

Issue #1575, opened 2022-06-14
It's possible to alias to the same name as an existing field in relationship properties


darrellwarde:
**Describe the bug**

It's possible to alias to the same name as an existing field

**Type definitions**

The following should really throw an error. 

```graphql
type Foo {
  point: Point
  geo_point: Point @alias(property: "point")
}
```

neo4j-team-graphql:
Many thanks for raising this bug report @darrellwarde. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @darrellwarde! :pray: We will now prioritise the bug and address it appropriately.

a-alle:
Node properties fixed in https://github.com/neo4j/graphql/pull/2200.
Relationship properties still TODO.

angrykoala:
This has already been fixed, some tests confirming this behaviour in #5794

PR #2180, opened 2022-10-05
Fix sort priority not maintained for edge vs node sorting fields

a-alle:
# Description

 Sort priority for edge vs node sorting fields maintained in the order specified on query creation
Additionally, `first` on connection field taken into account when connected type uses the `@queryOptions` directive with a default and/or a max argument.

# Issue

Closes https://github.com/neo4j/graphql/issues/2147
Closes https://github.com/neo4j/graphql/issues/1575

# Checklist


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-2180.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-2180.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/tests/integration/relationship-properties/read.int.test.ts
```diff
@@ -19,69 +19,73 @@
...
-        type Movie {
-            title: String!
-            actors: [Actor!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: IN)
-        }
-
-        type Actor {
-            name: String!
-            movies: [Movie!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: OUT)
-        }
+    let typeMovie: UniqueType;
```
angrykoala, line 36: :clap:

packages/graphql/src/translate/connection-clause/create-sort-and-limit.ts
```diff
@@ -61,17 +61,20 @@ export function createSortAndLimitProjection({
...
-    });
-
-    addSortAndLimitOptionsToClause({
-        optionsInput: { sort: [nodeSortFields] },
-        target: nodeRef,
-        projectionClause: withStatement,
+    nodeAndEdgeSortFields.forEach((sortField) => {
+        const [nodeOrEdge, sortKeyAndValue] = Object.entries(sortField)[0];
+        addSortAndLimitOptionsToClause({
+            optionsInput: { sort: [sortKeyAndValue], limit: firstArg, offset },
```
angrykoala, line 67: :thinking: 
There is something weird here, we are passing the limit values, but after this loop we are using addSortAndLimitOptionsTOClause again with these values. Maybe we can remove them from here?

packages/graphql/src/translate/projection/subquery/add-sort-and-limit-to-clause.ts
```diff
@@ -27,13 +27,13 @@ export function addSortAndLimitOptionsToClause({
...
     optionsInput: GraphQLOptionsArg;
-    target: CypherBuilder.Variable | CypherBuilder.PropertyRef;
+    target?: CypherBuilder.Variable | CypherBuilder.PropertyRef;
     projectionClause: CypherBuilder.Return | CypherBuilder.With;
 }): void {
     if (optionsInput.sort) {
         const orderByParams = createOrderByParams({
             optionsInput,
-            target, // This works because targetNode uses alias
+            target: target as CypherBuilder.Variable | CypherBuilder.PropertyRef, // This works because targetNode uses alias
```
angrykoala, line 36: What happens if `target` is undefined?

angrykoala (changes requested):
Looking good only the target comment is blocking, everything else is minor comments

packages/graphql/src/translate/connection-clause/create-sort-and-limit.ts
a-alle: So the two cases here are:
1. inside `nodeAndEdgeSortFields.forEach`: for queries of the format `actorsConnection(first: 5, sort: [{ node: { name: ASC } }])` where limit is combined with sorting args - excluding the limit values from here results in pagination not taking place
2. after `nodeAndEdgeSortFields `: when `limit` is present but sorting on edge/nodes is not - nodeAndEdgeSortFields will be an empty `[]` and the `.forEach` never takes place to add the `limit` in cypher

packages/graphql/src/translate/connection-clause/get-sort-fields.ts
```diff
@@ -18,25 +18,21 @@
...
 
-/** Returns sort fields for connections, split between edge and node */
-export function getSortFields(resolveTree: ResolveTree): SortFields {
-    const sortInput = (resolveTree.args.sort ?? []) as ConnectionSortArg[];
-
-    return {
-        edge: getSortFieldsByElement(sortInput, "edge"),
-        node: getSortFieldsByElement(sortInput, "node"),
-    };
+/** Returns sort fields for connections, of format [{ ['edge' | 'node']: { [key]: [value] } }] */
```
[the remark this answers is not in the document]
a-alle: yes that's correct, what I wanted to highlight is that either an `edge` or a `node` is present, but not both

packages/graphql/src/translate/connection-clause/create-sort-and-limit.ts
angrykoala: Fair enough, thanks for the clarification

packages/graphql/src/translate/connection-clause/get-sort-fields.ts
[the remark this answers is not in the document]
angrykoala: Maybe we could improve typings to reflect that?

```ts
type SortFields = { edge: GraphQLSortArg } | { node: GraphQLSortArg }
```

Again, just a non-blocking nitpick, in case you think is worth it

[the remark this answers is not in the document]
a-alle: sounds good

angrykoala:
LGTM

PR #2221, opened 2022-10-12
Add before and after each hooks to performance tests

angrykoala:
# Description
Fix mutation performance tests setup by adding afterEach hooks with a resetDB.

Additional tests for delete added


Liam-Doodson (approved):
Looks good to me!

PR #2222, opened 2022-10-12
Remove flatten and map.values apoc procedures

angrykoala:
# Description
This PR removes `apoc.coll.flatten` from the constants list and removes `apoc.map.values` in favour of `[k in keys(value) | value[k]][0]`


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-2222.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-2222.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql/src/translate/translate-top-level-cypher.ts
```diff
@@ -172,8 +172,8 @@ export function translateTopLevelCypher({
     } else {
         cypherStrs.push(`
             CALL apoc.cypher.doIt("${statement}", ${apocParamsStr}) YIELD value
-            WITH apoc.map.values(value, [keys(value)[0]])[0] AS this
-        `);
+            WITH [k in keys(value) | value[k]][0] as this
```
Liam-Doodson, line 175: Not blocking but is the convention not for `as` to be upper case?

Liam-Doodson (dismissed):
Looks good to me other than the non-blocking comment!

packages/graphql/src/translate/translate-top-level-cypher.ts
angrykoala: Good catch
```suggestion
            WITH [k in keys(value) | value[k]][0] AS this
```

PR #2223, opened 2022-10-12
Remove foreach

angrykoala:
# Description
Foreach is currently used for connections, and it is limiting our ability to use arbitrary Cypher to generate the subscription metadata, as well as return the data from the removed elements.

This PR  aims to remove `FOREACH` in favour of `CALL ... UNWIND`


darrellwarde (approved):
Looks good, a nice improvement I think!

PR #2051, opened 2022-09-07
RFC-026: GraphQL Toolbox tracking

tbwiss:
RFC to add tracking to the GraphQL Toolbox.

PR #2248, opened 2022-10-18
feat: Toolbox product usage tracking UX part

tbwiss:
RFC: https://github.com/neo4j/graphql/pull/2051

This is the first part of the Toolbox product usage tracking task. It contains:
- An alert (using the neo4h NDL component) to highlight that product usage tracking is enabled by default
- A setting option to enable/disable product usage tracking at any time
- The basic setup for invoking Segment analytics and sending product usage tracking events to Segment
- A way to provide the keys for the two Segment sources (PROD and DEV) both for local usage as well as in GHA
- E2E test to verify the enabling/disabling of the product usage tracking

NOTE:
Until this point, only a _dummy_ product usage tracking event has been set up, when one clicks the "star" to store the type definitions as a favorite, to verify if the event gets sent to Segment.
-> Both the DEV and the PROD Segment source receive the (_dummy_) product usage tracking events, this has been verified.

The rest of the product usage tracking events will be added in the next PR.


neo4j-team-graphql:

Check out your changes at
 http://neo4j-graphql-toolbox-stage-2248.surge.sh (=> for local development; use for instance [Neo4j Desktop](https://neo4j.com/download/))
or
 https://neo4j-graphql-toolbox-stage-2248.surge.sh (=> use to connect to an [Aura Free](https://neo4j.com/cloud/aura-free/) dbms)

packages/graphql-toolbox/package.json
```diff
@@ -45,6 +45,7 @@
         "@neo4j/introspector": "^1.0.2",
         "codemirror": "^5.65.8",
         "codemirror-graphql": "^2.0.0",
+        "dotenv-webpack": "^8.0.1",
```
mjfwebb, line 48: Another candidate for pinned package versions perhaps

packages/graphql-toolbox/src/utils/tracking.ts
```diff
@@ -0,0 +1,43 @@
...
+        const trackingConsent = Storage.retrieve(LOCAL_STATE_ENABLE_PRODUCT_USAGE_TRACKING);
+
+        if (trackingConsent !== "true") return;
+        if (!window.analytics || !window.analytics.track) return;
+
+        const enrichedEventProperties = {
+            graphQLToolboxVersion: process.env.VERSION,
+            neo4jGraphQLLibraryVersion: process.env.NEO4J_GRAPHQL_VERSION,
+            ...eventProperties,
+        };
```
mjfwebb, line 38:
```suggestion
        const enrichedEventProperties = {
            ...eventProperties,
            graphQLToolboxVersion: process.env.VERSION,
            neo4jGraphQLLibraryVersion: process.env.NEO4J_GRAPHQL_VERSION, 
        };
```

Just a thought about the order of the spreading here: I prefer spreading first so defaults get overwritten by anything more specifically added after. It may not functionally change anything here at all but it feels like it could be a nice habit to get into.

mjfwebb (changes requested):

packages/graphql-toolbox/src/utils/tracking.ts
tbwiss: Good point! I moved the spreading to the bottom to prevent it from being overwritten but I suppose it would be ok if we actually could overwrite the default. changing this

mjfwebb (approved):
Great work :+1:

Issue #2269, opened 2022-10-20
Incorrect nodes created in nested update operation with interfaces


angrykoala:
**Describe the bug**
Given the following type definitions:

```graphql
    type Movie {
        title: String!
        reviewers: [Reviewer!]! @relationship(type: "REVIEWED", properties: "Review", direction: IN)
    }

    interface Review @relationshipProperties {
        score: Int!
    }

    type Person implements Reviewer {
        name: String!
        reputation: Int!
        id: Int @unique
        reviewerId: Int @unique
    }

    type Influencer implements Reviewer {
        reputation: Int!
        url: String!
        reviewerId: Int
    }

    interface Reviewer {
        reputation: Int!
        reviewerId: Int
    }
```

With the following database setup:
```
CREATE(:Movie {title: "John Wick"})
```

And the following mutation:
```graphql
mutation UpdateMovies {
  updateMovies(
    where: { title: "John Wick" }
    update: {
      reviewers: [
        {
          create: [
            {
              edge: { score: 10 }
              node: { Person: { reputation: 100, name: "Ana" } }
            }
          ]
        }
      ]
    }
  ) {
    movies {
      title
    }
  }
}
```

The expected behaviour is for a new node `Person` with the name `Ana` to be created and connected to the movie with the `REVIEWED` type in the relationship.

Currently, 2 new unnamed nodes are created, with the label `Movie`, connected to the movie.

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @angrykoala! :pray: We will now prioritise the bug and address it appropriately.

neo4j-team-graphql:
Many thanks for raising this bug report @angrykoala. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

Issue #2249, opened 2022-10-18
Update with create on an interface type fails 


a-alle:
**Describe the bug**

When performing an update with create on an interface type, it fails because it attempts to set a property with a Map (of the interface's type's properties).

**Type definitions**

```gql
            type Movie {
                title: String!
                actors: [Actor!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: IN)
                directors: [Director!]! @relationship(type: "DIRECTED", properties: "Directed", direction: IN)
                reviewers: [Reviewer!]! @relationship(type: "REVIEWED", properties: "Review", direction: IN)
                imdbId: Int @unique
            }
            
            type Actor {
                name: String!
                movies: [Movie!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: OUT)
            }
            
            interface ActedIn @relationshipProperties {
                screenTime: Int!
            }
            
            interface Directed @relationshipProperties {
                year: Int!
            }
            
            interface Review {
                score: Int!
            }
        
            type Person implements Reviewer {
                name: String!
                reputation: Int!
            }
            
            type Influencer implements Reviewer {
                reputation: Int!
                url: String!
            }
            
            union Director = Person | Actor
            
            interface Reviewer {
                reputation: Int!

            }
```

**To Reproduce**

Executing the following mutation:
```gql
{
                query: `
                    mutation {
                        updateMovies(
                                where: {
                                  title: "John Wick"
                                },
                                update: {
                                    reviewers: [{
                                      create: [{
                                        edge: {
                                            score: 10
                                        },
                                        node: {
                                            Person: {
                                                reputation: 100,
                                                name: "Ana"
                                            }
                                        }
                                      }]
                                    }]
                                }
                        ) {
                            movies {
                                title
                            }
                        }
                    }
                `,
            }
```
results in a 200 response, with the errors object containing a Neo4jError: 
`"Property values can only be of primitive types or arrays thereof. Encountered: Map{name -> String(\"Ana\"), reputation -> Long(100)}."`

**Expected behavior**

Similar to a create instead of an update, each property should be set independently.
The difference lies in that `create-update-and-params` calls `createCreateAndParams` with `input: create**.node**` which no longer qualifies as a relation inside `createCreateAndParams`,
while `create-create-and-params` doesn't.

neo4j-team-graphql:
[...]

PR #2273, opened 2022-10-21
2269 incorrect nodes created in nested update operation with interfaces

angrykoala:
# Description
This PR fixes nested creation with interfaces #2269 and #2249


MacondoExpress (approved):
LGTM!

Issue #1197, opened 2022-03-29
Add ability to specify minimum score for fulltext index search


darrellwarde:
**Is your feature request related to a problem? Please describe.**

Full-text searches can return a lot of results, and filtering on the search score can help minimise the results.

**Describe the solution you'd like**

It should be possible to set a default value for this, and also override it per query.

For instance, this could be applied to a type like so:

```gql
type Movie @fulltext(name: "MovieTitle", fields: ["title"], defaultMinimumScore: 1) {
  title: String!
}
```

And also when querying like so:

```gql
query {
  movies(fulltext: { MovieTitle: { phrase: "Forrest", minimumScore: 2 } }) {
    title
  }
}
```

EDIT: another option is allowing minimum and maximum:

```gql
query {
  movies(fulltext: { MovieTitle: { phrase: "Forrest", score: { min: 2, max: 50 } } }) {
    title
  }
}
```

**Additional context**

An attempt at this functionality was removed in #1196, because the implementation was flawed in that it checked for an exact score match rather than a minimum.

trondaal:
What about just adding the score returned from db.index.fulltext to the returned node in some way? Access to the score value would enable any user defined logic to be implemented.

darrellwarde:
> What about just adding the score returned from db.index.fulltext to the returned node in some way? Access to the score value would enable any user defined logic to be implemented.

I think that's a good idea in principle, but I think it will be a very messy addition to just add it into the returned node. I've raised an issue #1373 to capture how we might do this in future.

darrellwarde:
This feature is now supported if you use the root-level queries generated for full-text indexes.

PR #2286, opened 2022-10-25
Full-text improvements

### LINKED ISSUE #1373: Return score for fulltext index search
It would be good if users could access the score for fulltext index search results. However, this will not nicely fit into the current API, as it would just need to be added to the node type, which is a hacky fix. 

We could potentially get around this by using a return payload as we do with Mutations. For instance, for a type `Movie`:

```gql
type Movie @fulltext(indexes: [{ name: "MovieTitle", fields: ["title"] }]) {
  title: String!
}

type QueryInfo {
  fulltextScore: Float
}

type MoviesQueryResponse {
  info: QueryInfo!
  movies: [Movie!]!
}

type Query {
  movies(fulltext: MovieFulltext, where: MovieWhere, options: MovieOptions): MoviesQueryResponse!
}
```

Liam-Doodson:
# Description

Created a new top-level query for full-text indexes. Added the full-text score to the results of this new query and made it possible to filter, sort and paginate the results.

Deprecated the existing full-text queries.

# Issue

Closes  #1373 & #1197

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-2286.surge.sh

packages/cypher-builder/src/expressions/procedures/db.ts
```diff
@@ -35,14 +35,15 @@ export class FullTextQueryNodes extends Clause {
     private targetNode: NodeRef;
     private indexName: string;
     private phrase: Variable;
-
+    private scoreVar: Variable | undefined;
     private whereClause: Where | undefined;
 
-    constructor(targetNode: NodeRef, indexName: string, phrase: Variable, parent?: Clause) {
+    constructor(targetNode: NodeRef, indexName: string, phrase: Variable, parent?: Clause, scoreVar?: Variable) {
```
angrykoala, line 41: Usually, parent is better as the last parameter.

```suggestion
    constructor(targetNode: NodeRef, indexName: string, phrase: Variable, scoreVar?: Variable, parent?: Clause) {
```

packages/cypher-builder/src/expressions/procedures/db.ts
```diff
@@ -58,14 +59,20 @@ export class FullTextQueryNodes extends Clause {
 
     public getCypher(env: CypherEnvironment): string {
         const targetId = env.getReferenceId(this.targetNode);
+        let scoreYield = "";
+        if (this.scoreVar) {
+            const scoreId = env.getReferenceId(this.scoreVar);
+            scoreYield = `, score AS ${scoreId}`;
+        }
```
angrykoala, line 66: We have a neat helper for this:
```typescript
const scoreYield = compileCypherIfExists(this.scoreVar, env, { prefix: ", score AS "});
```

This will already handle the logic of pretending `score AS` only if `scoreVar` is defined.

Just for consistency sake, the line above (`env.getReferenceId`) should be:
```ts
const targetId = this.targetNode.getCypher(env);
```

packages/cypher-builder/src/expressions/procedures/db.ts
```diff
@@ -58,14 +59,20 @@ export class FullTextQueryNodes extends Clause {
...
+
+        const textSearchStr = `CALL db.index.fulltext.queryNodes(
+            "${this.indexName}",
+            ${this.phrase.getCypher(env)}
+        ) YIELD node AS ${targetId}${scoreYield}`;
```
angrykoala, line 71: After some discussion, Darrell and I decided to go with a one-liner for this, so:

```suggestion
        const textSearchStr = `CALL db.index.fulltext.queryNodes("${this.indexName}", ${this.phrase.getCypher(env)}) YIELD node AS ${targetId}${scoreYield}`;
```

packages/cypher-builder/src/index.ts
```diff
@@ -19,6 +19,4 @@
 
 import * as Cypher from "./Cypher";
 
-export * from "./Cypher";
```
angrykoala, line 22: :thinking: 
**question:** What is the reasoning for this change?

I'm not an expect on creating package for typescript so I may be wrong here, but this export is intended to allow for javascript based exports:
```js
const Cypher = require('cypher-builder')
```

As well as destructured exports:

```ts
import { Variable } from "cypher-builder";
```

Both of which cannot be made easily (afaik) with a default export. So by exporting both we get the best of both worlds

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -657,10 +663,66 @@ function makeAugmentedSchema(
...
+            const fulltextResultDescription = `The result of a fulltext search on an index of ${node.name}`;
+            const fulltextWhereDescription = `The input for filtering a fulltext query on an index of ${node.name}`;
+            const fulltextSortDescription = `The input for sorting a fulltext query on an index of ${node.name}`;
+
             composer.createInputTC({
                 name: `${node.name}Fulltext`,
                 fields,
             });
+
+            if (!fulltextScoreWhereCreated) {
```
angrykoala, line 675: **Non-blocking**
**If possible and easy** Consider moving the fulltext types to a separate function, to avoid increasing the size of `makeAugmentedSchema`

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -408,3 +421,48 @@ const generateMissingOrAliasedRequiredFields = ({
...
+}: {
+    resolveTree: ResolveTree;
+    node: Node;
+    context: Context;
+    chainStr?: string;
+    varName: string;
+    literalElements?: boolean;
+    resolveType?: boolean;
+}): ProjectionResult {
+    if (!resolveTree.fieldsByTypeName[node.fulltextTypeNames.result][lowerFirst(node.name)]) {
```
angrykoala, line 442: Question: Do we need to use `lowerFirst` here? Feels a bit flaky to depend on that logic to find the correct nodeResolveTree

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -408,3 +421,48 @@ const generateMissingOrAliasedRequiredFields = ({
...
+            params: {},
+            meta: {},
+            subqueries: [],
+            subqueriesBeforeSort: [],
+        };
+    }
+
+    const nodeResolveTree = {
+        ...resolveTree.fieldsByTypeName[node.fulltextTypeNames.result][lowerFirst(node.name)],
+    };
```
angrykoala, line 454: Do we need to clone this? Is it supposed to be modified in the method `createProjectionAndParams`?

packages/graphql/src/translate/create-projection-and-params.ts
```diff
@@ -408,3 +421,48 @@ const generateMissingOrAliasedRequiredFields = ({
...
+
+    const nodeContext = { ...context };
+    nodeContext.fulltextIndex = false;
```
angrykoala, line 457: **Nitpick**

```suggestion
    const nodeContext = { ...context, fulltextIndex: false };
```

packages/graphql/tests/tck/sort.test.ts
```diff
@@ -233,19 +233,19 @@ describe("Cypher sort tests", () => {
             WHERE this.title = $param0
             WITH *
             ORDER BY this.id DESC, this.title ASC
-            SKIP $this_offset
-            LIMIT $this_limit
+            SKIP $param1
```
angrykoala, line 236: Nice!

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +95,2324 @@ describe("@fulltext directive", () => {
 
         await driver.close();
     });
-
-    test("should create index if it doesn't exist and then query using the index", async () => {
+    describe("Functionality", () => {
```
angrykoala, line 98: :thinking: This feels like a too generic description to be really useful.

Consider wether we can find a better name, or maybe this should not even be a `describe` block

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +95,2324 @@ describe("@fulltext directive", () => {
...
+        const movie2 = {
+            title: "Another Title",
+            description: "this is a description",
+            released: 2002,
+        };
+
+        beforeEach(async () => {
+            personType = generateUniqueType("Person");
+            movieType = generateUniqueType("Movie");
+            queryType = `${personType.plural}Fulltext${upperFirst(personType.name)}Index`;
```
angrykoala, line 139: The unique type class have an `operations` object to generate the top-level operations so we have an unified place to generate these instead of having this logic scattered around. Consider moving this to that class so we can have something like:
```suggestion
            queryType = personType.operations.fulltext;
```

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +95,2324 @@ describe("@fulltext directive", () => {
...
+                contextValue: {
+                    driver,
+                    driverConfig: { database: databaseName },
+                },
+            });
+
+            expect(gqlResult.errors).toBeFalsy();
+            expect((gqlResult.data?.[queryType] as any[])[0][personTypeLowerFirst]).toBeUndefined();
+            expect((gqlResult.data?.[queryType] as any[])[1][personTypeLowerFirst]).toBeUndefined();
+            expect((gqlResult.data?.[queryType] as any[])[2][personTypeLowerFirst]).toBeUndefined();
```
angrykoala, line 308: It may be easier to directly test the whole data array, instead of going one by one.

Using something like `toIncludeSameMembers`

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +95,2324 @@ describe("@fulltext directive", () => {
...
+            const gqlResult = await graphql({
+                schema: generatedSchema,
+                source: query,
+                contextValue: {
+                    driver,
+                    driverConfig: { database: databaseName },
+                },
+            });
+
+            expect(gqlResult.errors).toBeDefined();
```
angrykoala, line 331: Consider validating the exact error that is thrown here

packages/graphql/tests/integration/deprecated/fulltext.int.test.ts
```diff
@@ -0,0 +1,745 @@
...
+        const neoSchema = new Neo4jGraphQL({ typeDefs });
+        await neoSchema.getSchema();
+
+        const session = driver.session({ database: databaseName });
+
+        try {
+            await session.run(
+                [`CREATE FULLTEXT INDEX ${indexName}`, `IF NOT EXISTS FOR (n:${type.name})`, `ON EACH [n.title]`].join(
+                    " "
+                )
```
angrykoala, line 634: Is there a reason for having this as an array with join?

packages/graphql/src/translate/translate-top-level-match.ts
```diff
@@ -20,47 +20,51 @@
...
+import { SCORE_FIELD } from "../graphql/directives/fulltext";
 
 export function translateTopLevelMatch({
+    matchNode,
     node,
     context,
     varName,
     operation,
 }: {
+    matchNode?: Cypher.NamedNode;
```
angrykoala, line 35: It is not a good idea to expect a `NamedNode`.

This function should not really care on wether or not a node is named or not, as it is the same to the queryBuilder and it will make our future refactors harder, as we are trying to really less on explicitly named variables in Cypher

angrykoala: Also, are we really using this variable? I don't see where we pass this :sweat_smile:

packages/graphql/src/translate/translate-top-level-match.ts
```diff
@@ -20,47 +20,51 @@
...
     varName: string;
     operation: AuthOperations;
 }): Cypher.Match | Cypher.db.FullTextQueryNodes {
     const { resolveTree } = context;
-    const whereInput = resolveTree.args.where as GraphQLWhereArg;
     const fulltextInput = (resolveTree.args.fulltext || {}) as Record<string, { phrase: string }>;
-
-    const matchNode = new Cypher.NamedNode(varName, { labels: node.getLabels(context) });
-
+    matchNode = matchNode || new Cypher.NamedNode(varName, { labels: node.getLabels(context) });
```
angrykoala, line 62: I don't really like this logic, if both `matchNode` and varName are `defined` now we don't have a source of truth for our name.

Ideally, we wouldn't need to use varName here and we would only use the given node.

If that is not possible, it is probably better to revert to the previous behaviour or accept `varName` as a `Cypher.Node | string`

packages/graphql/src/translate/translate-top-level-match.ts
```diff
@@ -70,21 +74,24 @@ export function createMatchClause({
         matchQuery = new Cypher.db.FullTextQueryNodes(matchNode, indexName, phraseParam);
 
         const labelsChecks = node.getLabels(context).map((label) => {
-            return Cypher.in(new Cypher.Literal(label), Cypher.labels(matchNode));
+            return Cypher.in(new Cypher.Literal(label), Cypher.labels(matchNode as Cypher.NamedNode));
```
angrykoala, line 77: Lets try to avoid casting :smiley:

packages/graphql/src/translate/translate-top-level-match.ts
```diff
@@ -106,3 +113,41 @@ export function createMatchClause({
...
+        matchNode,
+        context.fulltextIndex.name,
+        phraseParam,
+        undefined,
+        scoreVar
+    );
+
+    const labelsChecks = node.getLabels(context).map((label) => {
+        return Cypher.in(new Cypher.Literal(label), Cypher.labels(matchNode));
+    });
```
angrykoala, line 136: Alternatively:

```suggestion
    const expectedLabels = node.getLabels(context);
    const labelsChecks=matchNode.hasLabels(...expectedLabels);
```

This is a relatively new feature of the cypherBuilder, but should work fine and much easier :)

packages/graphql/src/translate/translate-top-level-match.ts
```diff
@@ -106,3 +113,41 @@ export function createMatchClause({
...
+    if (whereInput?.[SCORE_FIELD]) {
+        if (whereInput[SCORE_FIELD].min || whereInput[SCORE_FIELD].min === 0) {
+            const scoreMinOp = Cypher.gte(scoreVar, new Cypher.Param(whereInput[SCORE_FIELD].min));
+            if (scoreMinOp) matchQuery.where(scoreMinOp);
+        }
+        if (whereInput[SCORE_FIELD].max || whereInput[SCORE_FIELD].max === 0) {
+            const scoreMaxOp = Cypher.lte(scoreVar, new Cypher.Param(whereInput[SCORE_FIELD].max));
+            if (scoreMaxOp) matchQuery.where(scoreMaxOp);
+        }
+    }
```
angrykoala, line 147: Nice usage of the builder here :+1:

packages/graphql/src/translate/translate-read.ts
```diff
@@ -118,28 +143,38 @@ export function translateRead({
...
+        if (!projectionSubqueries.empty && hasOrdering) {
+            addSortAndLimitOptionsToClause({
+                optionsInput,
+                target: matchNode,
+                projectionClause: returnClause,
+                node,
+                context,
+            });
+        }
+
```
angrykoala, line 177: I see you've been making changes to our order translation.

We should run the performance tests before emerging this PR to make sure that it has not been affected (we can check at that together)

angrykoala (changes requested):
Really comprehensive job here :+1: 

A few comments, mainly on some decisions regarding code organization and a few tips on utils and cypher builder

packages/graphql/src/translate/create-projection-and-params.ts
Liam-Doodson: Yeah it's needed because the query field this is looking for has been created with `lowerFirst`

packages/graphql/tests/integration/directives/fulltext.int.test.ts
Liam-Doodson: Also the order does matter for these tests as they are supposed to be ordered by score DESC by default so it's safe to assert these individually

packages/graphql/tests/integration/deprecated/fulltext.int.test.ts
Liam-Doodson: I don't know the reason for this. These are the old full-text tests and I just threw them in a deprecated folder to be deleted when the deprecated version is removed in 4.0.

As they're going to be deleted soon anyway I don't think it's worth changing this

packages/cypher-builder/src/index.ts
Liam-Doodson: Yes you're right, I was having problems importing cypher builder when I first merged the changes and this was one of the things I tried to get it working as VSCode didn't like this line but I meant to put it back to how it was!

packages/graphql/src/schema/make-augmented-schema.ts
Liam-Doodson: Yeah I'll do this. I did actually consider it because I also think it'd be nicer but seeing as it hadn't been done anywhere else I wanted to avoid stylistic changes

packages/graphql/src/translate/create-projection-and-params.ts
Liam-Doodson: Okay my bad, turns out I can just use `node.singular` instead

packages/graphql/tests/integration/directives/fulltext.int.test.ts
Liam-Doodson: The describe block is mainly being used for efficiency so that the before each is run only for tests that need it. I've renamed this to Query tests so hopefully that should be less generic

packages/graphql/tests/integration/directives/fulltext.int.test.ts
Liam-Doodson: Unfortunately, this depends on the index name. These are defined separately and there can be several for each type/node so it's not really possible to do it like that

packages/graphql/src/translate/translate-top-level-match.ts
Liam-Doodson: I've changed this to accept a `Node` not just a `NamedNode`. The variable itself is used a lot throughout `createMatchClause`. It was originally only assigned within `createMatchClause` but I needed to be able to reference the same variable within translate-read

packages/graphql/src/translate/translate-top-level-match.ts
Liam-Doodson: This was just because typescript wasn't picking up that it can't be undefined at this point. A small refactor has removed the need for this now though 😄

packages/graphql/src/translate/translate-top-level-match.ts
Liam-Doodson: I've refactored this so matchNode is always passed and is the only value that is used (not `varName`)

packages/graphql/tests/schema/fulltext.test.ts
```diff
@@ -168,6 +174,8 @@ describe("@fulltext schema", () => {
               movies(fulltext: MovieFulltext, options: MovieOptions, where: MovieWhere): [Movie!]!
```
darrellwarde, line 174: Now, if I'm being honest I'm not 100% what the language allows, but if we could do the following it would be amazing:

```suggestion
              movies(fulltext: MovieFulltext @deprecated(reason: "Use root Query level full-text fields instead"), options: MovieOptions, where: MovieWhere): [Movie!]!
```

And then also mark the `MovieFulltext` type as deprecated as well:

```gql
input MovieFulltext @deprecated(reason: "Use root Query level full-text fields instead") {
    MovieDescription: MovieMovieDescriptionFulltext
    MovieTitle: MovieMovieTitleFulltext
}
```

If you want to get _super_ creative:

```gql
input MovieFulltext @deprecated(reason: "Use root Query level full-text fields instead") {
    MovieDescription: MovieMovieDescriptionFulltext @deprecated(reason: "Use Query.moviesFulltextMovieDescription instead")
    MovieTitle: MovieMovieTitleFulltext @deprecated(reason: "Use Query.moviesFulltextMovieTitle instead")
}
```

That last one is probably more trouble than it's worth!

Perhaps also worth deprecating the other input types:

```gql
input MovieMovieDescriptionFulltext @deprecated(reason: "Use root Query level full-text fields instead") {
    phrase: String!
}

input MovieMovieTitleFulltext @deprecated(reason: "Use root Query level full-text fields instead") {
    phrase: String!
}
```

Choose a reason you feel is appropriate! 🙂 Also, unsure how much of the above is even possible, GraphQL can be a bit picky.

packages/graphql/src/schema/resolvers/query/fulltext.ts
```diff
@@ -0,0 +1,69 @@
...
+import getNeo4jResolveTree from "../../../utils/get-neo4j-resolve-tree";
+import Cypher from "@neo4j/cypher-builder";
+
+export function fulltextResolver(
+    { node }: { node: Node },
+    index: FulltextIndex
+): ObjectTypeComposerFieldConfigDefinition<any, any, any> {
+    async function resolve(_root: any, args: any, _context: unknown, info: GraphQLResolveInfo) {
+        const context = createFulltextContext(index, args, _context, info);
+        const { cypher, params } = translateRead({ context, node }, node.singular);
```
darrellwarde, line 35: Could this full-text stuff now be seen as independent enough that it could have its own translate function? Thoughts @angrykoala?

packages/graphql/src/translate/translate-read.ts
```diff
@@ -148,70 +182,14 @@ export function translateRead({
...
-
-    if (hasLimit) {
-        params[`${varName}_limit`] = limit;
-        sortOffsetLimit.push(`LIMIT $${varName}_limit`);
-    }
-
-    return new Cypher.RawCypher(() => {
-        return [sortOffsetLimit.join("\n"), params];
-    });
-}
```
darrellwarde, line 213: 👏

packages/graphql/src/graphql/directives/fulltext.ts
```diff
@@ -43,6 +45,9 @@ export const fulltextDirective = new GraphQLDirective({
                             fields: {
                                 type: new GraphQLNonNull(new GraphQLList(GraphQLString)),
                             },
+                            queryName: {
+                                type: GraphQLString,
+                            }
```
darrellwarde, line 50: Hmmmmm, just had a thought of could we add `indexName` now, make it take precedence over `name` and mark `name` as deprecated. Means the only thing to do for 4.0 will be remove it, and some will already be migrated!

```suggestion
                            queryName: {
                                type: GraphQLString,
                            }
                            indexName: {
                                type: GraphQLString,
                            }

```

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +94,2415 @@ describe("@fulltext directive", () => {
...
+                contextValue: {
+                    driver,
+                    driverConfig: { database: databaseName },
+                },
+            });
+
+            expect((gqlResult.errors as any[])[0].message).toBe(
+                `Field "${queryType}" argument "phrase" of type "String!" is required, but it was not provided.`
+            );
+        });
```
darrellwarde, line 354: I'm sorry to say it, but these tests could really all be removed - they're just testing the GraphQL query language which I don't really think we should be doing in our integration tests.

```suggestion

```

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +94,2415 @@ describe("@fulltext directive", () => {
...
+                    driverConfig: { database: databaseName },
+                },
+            });
+
+            expect(gqlResult.errors).toBeFalsy();
+            expect((gqlResult.data?.[queryType] as any[])[0][personTypeLowerFirst].name).toBe(person1.name);
+            expect((gqlResult.data?.[queryType] as any[])[0][SCORE_FIELD]).toBeNumber();
+            expect(gqlResult.data?.[queryType] as any[]).toBeArrayOfSize(1);
+        });
+        test("Filters score with max score of 0", async () => {
```
darrellwarde, line 560: Just an opinionated stylistic comment for future reference, it would be nice to see a bit of whitespace between tests to make them just a bit easier to read!

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +94,2415 @@ describe("@fulltext directive", () => {
...
+
+            expect(gqlResult.errors).toBeFalsy();
+            expect((gqlResult.data?.[queryType] as any[])[0][personTypeLowerFirst].name).toBe(person3.name);
+            expect((gqlResult.data?.[queryType] as any[])[1][personTypeLowerFirst].name).toBe(person2.name);
+            expect((gqlResult.data?.[queryType] as any[])[2][personTypeLowerFirst].name).toBe(person1.name);
+            expect((gqlResult.data?.[queryType] as any[])[0][SCORE_FIELD]).toBeNumber();
+            expect((gqlResult.data?.[queryType] as any[])[1][SCORE_FIELD]).toBeNumber();
+            expect((gqlResult.data?.[queryType] as any[])[2][SCORE_FIELD]).toBeNumber();
+        });
+        test("Unordered sorting", async () => {
```
darrellwarde, line 838: I'm unsure what this is trying to test, given that sort arguments are provided?! The `born` sort should be applied first, and then the `name` sort.

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +94,2415 @@ describe("@fulltext directive", () => {
...
+                contextValue: {
+                    driver,
+                    driverConfig: { database: databaseName },
+                },
+            });
+
+            expect(gqlResult.errors).toBeFalsy();
+            expect(gqlResult.data?.[queryType]).toBeArrayOfSize(2);
+        });
+        test("Offesetting is possible", async () => {
```
darrellwarde, line 1134:
```suggestion
        test("Offsetting is possible", async () => {
```

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +94,2415 @@ describe("@fulltext directive", () => {
...
+            });
+
+            expect(gqlResult.errors).toBeFalsy();
+            expect(gqlResult.data?.[queryType]).toEqual([
+                {
+                    [personTypeLowerFirst]: person3,
+                },
+            ]);
+        });
+        test("Combined limiting and offesetting is possible", async () => {
```
darrellwarde, line 1161:
```suggestion
        test("Combined limiting and offsetting is possible", async () => {
```

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +94,2415 @@ describe("@fulltext directive", () => {
...
+                contextValue: {
+                    driver,
+                    driverConfig: { database: databaseName },
+                },
+            });
+
+            expect((gqlResult.errors as any[])[0].message).toStartWith(
+                `Unknown argument "${invalidArgument}" on field`
+            );
+        });
```
darrellwarde, line 1286: These are really just testing GraphQL validation, I would remove them.

```suggestion

```

packages/graphql/tests/integration/directives/fulltext.int.test.ts
```diff
@@ -74,672 +94,2415 @@ describe("@fulltext directive", () => {
...
+                    [personTypeLowerFirst]: {
+                        name: person1.name,
+                    },
+                },
+                {
+                    [personTypeLowerFirst]: {
+                        name: person2.name,
+                    },
+                },
+            ]);
```
darrellwarde, line 1323: Just to confirm, this is the correct order? The score sort is applied within the Cypher in this scenario?

packages/graphql/src/translate/projection/subquery/add-sort-and-limit-to-clause.ts
```diff
@@ -63,14 +75,40 @@ export function addSortAndLimitOptionsToClause({
...
+    const orderList = sortOptions.flatMap(
+        (arg: GraphQLSortArg | NestedGraphQLSortArg): Array<[string, "ASC" | "DESC"]> => {
+            if (
+                context?.fulltextIndex &&
+                node?.name &&
+                arg[lowerFirst(node.name)] &&
+                typeof arg[lowerFirst(node.name)] === "object"
+            ) {
+                return Object.entries(arg[lowerFirst(node.name)] as GraphQLSortArg);
+            }
```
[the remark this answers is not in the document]
Liam-Doodson: I think I've got this as good as it's going to be able to get at the moment. I've been more specific about what is passed now as it was mainly just a field name & cypher builder variable that was needed from the context & node.

Unfortunately, the varName and cypher fields from the node still need to stay. However, they shouldn't need to be here and can be replaced once nested `@cypher` fields are translated using cypher builder

darrellwarde (changes requested):
Ignoring the number of comments made, this is great work! It's great to see you digging into the Cypher Builder already.

A bit of general feedback on your integration tests is it would be great to see them broken out into separate files a little (filtering, sorting, auth, etc.) - it's quite a lot to process to have them all in one file! Also a little _less_ defensive, we don't really need to test against things that the GraphQL query language gives for free (missing arguments, wrong types, etc.).

Also be careful with our naming conventions - as a general rule, type names should always be PascalCase and arguments, camelCase. 

But looking really great, and super well tested! 👏

packages/graphql/tests/schema/fulltext.test.ts
Liam-Doodson: Unfortunately, I believe I tried most of these when I was working on this PR. The `@deprecated` directive can only be used on fields (but not input fields) so doesn't work anywhere we can put it.

The old implementation for full-text just used the normal return object so it's not even possible to deprecate the fields on there as it would deprecate them for all queries (not just full-text)

packages/graphql/tests/integration/directives/fulltext.int.test.ts
Liam-Doodson: Yes, this is the correct order

packages/graphql/tests/integration/directives/fulltext.int.test.ts
Liam-Doodson: Yeah I wrote I test for both ordered/unordered sorting e.g. (`sort: { score: ASC }` vs `sort: [{ score: ASC }]`. I did actually run into errors on only these tests when working on this PR so I think they should definitely be kept

packages/graphql/src/schema/resolvers/query/fulltext.ts
Liam-Doodson: I did consider doing that but I barely made any changes to translate-read that were specifically for translating full-text. The changes to translate-read are almost all refactoring the sorting to use CypherBuilder. Most of the changes for full-text were needed within the functions called by translate-read. 

It's obviously possible to try and make those changes at a higher level (e.g. in a `translate-fulltext` function), but most of them depended on data that is only available within the helper functions (e.g. a field/argument name). As a result, it felt like that approach would have led to me writing lots of functions that only had minor changes compared to existing ones

packages/graphql/src/schema/resolvers/query/fulltext.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+function createFulltextContext(index: FulltextIndex, args: any, _context: unknown, info: GraphQLResolveInfo): Context {
+    const context = _context as Context;
+    context.resolveTree = getNeo4jResolveTree(info, { args });
+    context.resolveTree.args.options = {
+        sort: context.resolveTree.args.sort,
+        limit: context.resolveTree.args.limit,
+        offset: context.resolveTree.args.offset,
+    };
```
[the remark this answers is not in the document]
Liam-Doodson: I don't think I understand this comment?

Is it not cleaner to do it here? Literally, every helper function expects these arguments to be in this format so why not do it just once in a function where it's clear that it's just doing this for full-text?

packages/graphql/src/schema/augment/fulltext.ts
```diff
@@ -0,0 +1,103 @@
...
+ * limitations under the License.
+ */
+
+import type { SchemaComposer } from "graphql-compose";
+import type { Node } from "../../classes";
+import { fulltextResolver } from "../resolvers/query/fulltext";
+import { upperFirst } from "../../utils/upper-first";
+import { SCORE_FIELD } from "../../graphql/directives/fulltext";
+
+const fulltextScoreWhereType = "FulltextScoreWhere";
```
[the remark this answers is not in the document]
Liam-Doodson: I think you're right, it makes more sense for this to be generic so it can be used elsewhere in the future

packages/graphql/tests/schema/fulltext.test.ts
darrellwarde: Damn, I see this now, that's disappointing! Seems like a shortcoming of the directive to me...

packages/graphql/tests/integration/directives/fulltext.int.test.ts
darrellwarde: Good catch, super thorough testing!

packages/graphql/src/schema/resolvers/query/fulltext.ts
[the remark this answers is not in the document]
darrellwarde: Well the resolve tree is kind of "sacred" - it's the representation of the incoming GraphQL query. It feels like we're having to "hack" it because we're using the `translateRead` function. It's not a massive problem, just feels a little bit dirty, we should perhaps look at removing these helper functions reliance on the resolve tree itself in future.

docs/modules/ROOT/pages/type-definitions/indexes-and-constraints.adoc
```diff
@@ -100,6 +100,124 @@ CALL db.index.fulltext.createNodeIndex("ProductName", ["Product"], ["name"])
...
+}
+
+"""The result of a fulltext search on an index of Product"""
+type productFulltextResult {
+  score: Float
+  product: Product
+}
+
+"""The input for filtering a fulltext query on an index of Product"""
+input productFulltextWhere {
```
darrellwarde, line 120: This should now be upper case I believe!

```suggestion
input ProductFulltextWhere {
```

docs/modules/ROOT/pages/type-definitions/indexes-and-constraints.adoc
```diff
@@ -100,6 +100,124 @@ CALL db.index.fulltext.createNodeIndex("ProductName", ["Product"], ["name"])
...
+  score: FulltextScoreWhere
+  product: ProductWhere
+}
+
+"""The input for sorting a fulltext query on an index of Product"""
+input productFulltextSort {
```
darrellwarde, line 126:
```suggestion
input ProductFulltextSort {
```

docs/modules/ROOT/pages/type-definitions/indexes-and-constraints.adoc
```diff
@@ -100,6 +100,124 @@ CALL db.index.fulltext.createNodeIndex("ProductName", ["Product"], ["name"])
...
+  score: SortDirection
+  product: ProductSort
+}
+
+"""The input for filtering the score of a fulltext search"""
+input FulltextScoreWhere {
```
darrellwarde, line 132: And also this change if we're going to take this approach?

```suggestion
input FloatWhere {
```

docs/modules/ROOT/pages/type-definitions/indexes-and-constraints.adoc
```diff
@@ -100,6 +100,124 @@ CALL db.index.fulltext.createNodeIndex("ProductName", ["Product"], ["name"])
...
+
+[source, graphql, indent=0]
+----
+type Query {
+    productsFulltextProductName(phrase: String!, where: productFulltextWhere, sort: [productFulltextSort!],
+    limit: Int, offset: Int): [productFulltextResult!]!
+}
+
+"""The result of a fulltext search on an index of Product"""
+type productFulltextResult {
```
darrellwarde, line 114:
```suggestion
type ProductFulltextResult {
```

docs/modules/ROOT/pages/type-definitions/indexes-and-constraints.adoc
```diff
@@ -100,6 +100,124 @@ CALL db.index.fulltext.createNodeIndex("ProductName", ["Product"], ["name"])
...
 === Using the `@fulltext` directive
 
+For every index specified, a new top level query will be generated by the library. For example, for the type definitions above,
+the following query and types would be generated:
+
+[source, graphql, indent=0]
+----
+type Query {
+    productsFulltextProductName(phrase: String!, where: productFulltextWhere, sort: [productFulltextSort!],
+    limit: Int, offset: Int): [productFulltextResult!]!
```
darrellwarde, line 110:
```suggestion
    productsFulltextProductName(phrase: String!, where: ProductFulltextWhere, sort: [ProductFulltextSort!],
    limit: Int, offset: Int): [ProductFulltextResult!]!
```

docs/modules/ROOT/pages/type-definitions/indexes-and-constraints.adoc
```diff
@@ -100,6 +100,124 @@ CALL db.index.fulltext.createNodeIndex("ProductName", ["Product"], ["name"])
...
+}
+----
+
+This would then produce the following top-level query:
+
+[source, graphql, indent=0]
+----
+type Query {
+    CustomProductFulltextQuery(phrase: String!, where: productFulltextWhere, sort: [productFulltextSort!],
+    limit: Int, offset: Int): [productFulltextResult!]!
```
darrellwarde, line 198:
```suggestion
    CustomProductFulltextQuery(phrase: String!, where: ProductFulltextWhere, sort: [ProductFulltextSort!],
    limit: Int, offset: Int): [ProductFulltextResult!]!
```

packages/graphql/src/graphql/directives/fulltext.ts
```diff
@@ -43,6 +45,13 @@ export const fulltextDirective = new GraphQLDirective({
...
```
darrellwarde, line 50: Oops, small syntax issue here!

```suggestion
                            },
```

packages/graphql/src/schema/augment/fulltext.ts
```diff
@@ -0,0 +1,93 @@
...
+                [SCORE_FIELD]: FloatWhere.name,
+                [node.singular]: nodeWhereTypeName,
+            },
+        });
+
+        composer.createObjectTC({
+            name: node.fulltextTypeNames.result,
+            description: fulltextResultDescription,
+            fields: {
+                [SCORE_FIELD]: "Float!",
```
darrellwarde, line 78: Just a personal preference when using `graphql-compose` - I prefer to use type objects here, for example, this would be `new GraphQLNonNull(GraphQLFloat)`. I think it just makes it easier to find all of the instances of a particular type in our schema augmentation, but it all works the same!

Same for our own generated types when we use them elsewhere, best to reference the objects themselves because then if we change the name or something, it doesn't all break.

darrellwarde (changes requested):
This looks good! I've just pointed out a couple of docs changes that need making. Is it just the parsing of `indexName` that needs adding now? That was possibly a slightly scope-creepy suggestion of me! 😬

darrellwarde (approved):
I think we might be there! Really nice work on this piece, it's super thorough - although there were quite a few comments made, they were all specific and very often stylistic. On to the next new thing!

AccsoSG:
@Liam-Doodson Looks good, but what happened to the aggregate functions? How can I output the count of the result set? This is an essential function for a full-text search

Liam-Doodson:
> @Liam-Doodson Looks good, but what happened to the aggregate functions? How can I output the count of the result set? This is an essential function for a full-text search
> 
> #2681

Hi @AccsoSG! We're currently reviewing our whole API ahead of 4.0.0 and we are considering the issue that you have mentioned as part of that work. We aim to make sure this is still supported in 4.0.0

PR #2314, opened 2022-10-28
Add 5 community/enterprise tests, shard TCK tests, remove unused

darrellwarde:


angrykoala (dismissed):
:clap:

darrellwarde:
Community tests failing, will remove to unblock merge, and then raise another PR just adding community.

packages/graphql/tests/tck/connections/filtering/node/points.test.ts
```diff
@@ -59,6 +59,13 @@ describe("Cypher -> Connections -> Filtering -> Node -> Points", () => {
     });
 
     test("DISTANCE", async () => {
+        let verifyTCK;
+
+        if (process.env.VERIFY_TCK) {
```
MacondoExpress, line 64: Would it be better to accept an option's argument in the translateQuery function and delegate the client the control of this behavior?
For instance:
```typescript
function translateQuery(    
    neoSchema: Neo4jGraphQL,
    query: DocumentNode,
    options?: {
        req?: IncomingMessage;
        variableValues?: Record<string, any>;
        neo4jVersion?: string;
        contextValues?: Record<string, any>;
        verifyTCK?:   boolean
    }
) {
    const verifyTCK = options?.verifyTCK ?? process.env.VERIFY_TCK;
}
```
So that previous behavior is maintained, but for this case, it will be:
```typescript
const result = await translateQuery(neoSchema, query, {
    req,
    verifyTCK: false,
});
```

darrellwarde: Potentially! But that was added prior to this PR, and I'm keen to get this merged as I've had absolute nightmares getting it green - I'll note this down to perhaps look at during this cool-down.

PR #2358, opened 2022-11-03
boyscouting: Toolbox use window Neo4jDesktopAPI

tbwiss:
The Neo4j Desktop exposed a GraphQL API to fetch for instance dbms information which is now deprecated.
We used it to safely fetch the dbms connection credentials when running the Toolbox from Neo4j Desktop. This allowed us to "auto login" the user when starting the Toolbox from Neo4j Desktop.

This updated solution uses the `context` that is available in the `window` object, more accurately `window.neo4jDesktopApi` instead. With this approach, we can get the connection credentials as before and use them to "auto login".

The login logic otherwise is untouched.


neo4j-team-graphql:

Check out your changes at
 https://neo4j-graphql-toolbox-stage-2358.surge.sh

mjfwebb (approved):
Nice one

PR #2361, opened 2022-11-03
Rfc connect in update mutation

Liam-Doodson:
# Description

RFC for new connection behaviour in update mutations


docs/rfcs/rfc-029-connect-in-update-mutations.md
```diff
@@ -0,0 +1,281 @@
+# Connect on Update
```
darrellwarde, line 1: Perhaps a better name for the RFC (and the file) is something like "Connect if a relationship already exists"?

docs/rfcs/rfc-029-connect-in-update-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+                            id: "2"
+                        }
+                    },
+                    edge: {
+                        type: "newType2",
+                        startDate: "123"
+                    }
+                }
+            ]
+        }
```
darrellwarde, line 146: Right, some suggestions here. Where you currently have `operation` there would be no granularity for each connection within the array. So see the below example where I have moved `operation` into each element, and I also propose making it a boolean that I believe it will only ever have two values. I've made a suggestion below, but I'm open to alternatives here!

```suggestion
        connect: {
            sponsor: [
                {
                    where: {
                        node: {
                            id: "2"
                        }
                    },
                    edge: {
                        type: "newType2",
                        startDate: "123"
                    },
                    connectIfExists: true
                },
                {
                    where: {
                        node: {
                            id: "4"
                        }
                    },
                    edge: {
                        type: "newType2",
                        startDate: "123"
                    },
                    connectIfExists: false
                },
            ]
        }
```

When implementing this work, I would split it into two - firstly do this part, the Mutation arguments, and then move onto the directive argument, both merged in individual PRs.

docs/rfcs/rfc-029-connect-in-update-mutations.md
```diff
@@ -0,0 +1,281 @@
...
+
+#### `@relationship` definition
+
+Specifying the `defaultUpdateOperation` on the `@relationship` directive:
+
+```gql
+type Client {
+    id: String!
+    login: String!
+    sponsor: [Client!]! @relationship(type: "HAS_SPONSOR", properties: "HasSponsor", direction: OUT, defaultUpdateOperation: "CREATE")
```
darrellwarde, line 111: I would change up the argument name here to make it clear it is for connect operations. Once again, I have played with having it as a boolean, I'm open to a discussion!

```suggestion
    sponsor: [Client!]! @relationship(type: "HAS_SPONSOR", properties: "HasSponsor", direction: OUT, connectIfExists: true)
```

darrellwarde (changes requested):
A couple of moderately large suggestions for the API here, but otherwise looks good!

docs/rfcs/rfc-029-connect-in-update-mutations.md
Liam-Doodson: I definitely agree that the new position is better and that an enum probably wasn't necessary as I can't think of other behaviours that could be added in the future.

However, I don't know if I like `connectIfExists` as I feel like it's not clear that it would create duplicate connections. Could `newConnectionIfExists` work instead?

PR #2369, opened 2022-11-04
RFC-remove-runFirstColumn-from-aggregate-where

Liam-Doodson:
# Description

RFC-for removing `runFirstColumn` from aggregate where


docs/rfcs/rfc-030-remove-runFirstColumn-from-aggregate-where.md
```diff
@@ -0,0 +1,70 @@
...
+```cypher
+MATCH(this:Post)
+CALL {
+    WITH this
+    MATCH(this)<-[:LIKES]-(u:User)
+    RETURN count(u) as agg_count
+}
+WITH *
+WHERE agg_count = 10
+RETURN this { .content } AS this
```
MacondoExpress, line 61: Not blocking, 
I would suggest keeping the name of the cypher variable more dependent on the context used so that a future implementation that uses this RFC as documentation could be less prone to introduce a variable name collision.

```suggestion
MATCH(this:Post)
CALL {
    WITH this
    MATCH(this)<-[:LIKES]-(u:User)
    RETURN count(u) as thisLikesAggregateCount
}
WITH *
WHERE thisLikesAggregateCount = 10
RETURN this { .content } AS this
```

MacondoExpress (approved):
Left a small comment, but overall looks good to me!

MacondoExpress (approved):

PR #2411, opened 2022-11-14
Add schema generation benchmark

angrykoala:
# Description
Adds performance tests on `getSchema` (i.e. server startup) by running `yarn performance --schema`

### DIFF
--- a/packages/graphql/tests/performance/graphql/batch-create/batch-create-interface.graphql
+++ b/packages/graphql/tests/performance/graphql/batch-create/batch-create-interface.graphql
@@ -1,4 +1,4 @@
-mutation BatchCreateInterface {
+mutation BatchCreateInterface_skip {
     createPeople(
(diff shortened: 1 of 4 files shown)


packages/graphql/tests/performance/graphql/batch-create/batch-create-interface.graphql
```diff
@@ -1,4 +1,4 @@
-mutation BatchCreateInterface {
+mutation BatchCreateInterface_skip {
```
darrellwarde, line 1: This is meant to be skipped?

angrykoala: Yes, this one was added after Simone changes, but it is really slow, making the perf. tests unbearably slow. Once we address that improvement we can remove the skip

PR #2417, opened 2022-11-15
RFC - OGM Select

danstarns:
No more string selection set.


angrykoala (approved):

angrykoala:
This is a good idea, we may want to discuss the some details of the API like naming (`select` vs `selectionSet`) but the principle looks good

PR #2427, opened 2022-11-16
Cypher builder string functions

angrykoala:
# Description
Exposes the `Function` class for arbitrary functions and add the string functions

### DIFF
--- a/packages/cypher-builder/src/Cypher.ts
+++ b/packages/cypher-builder/src/Cypher.ts
@@ -75,6 +75,8 @@ export {
 export { plus, minus } from "./expressions/operations/math";
 
 // --Functions
+export { CypherFunction as Function } from "./expressions/functions/CypherFunction";
+
 export {
     coalesce,
     point,
@@ -93,6 +95,9 @@ export {
     sum,
     randomUUID,
 } from "./expressions/functions/CypherFunction";
+
+export * from "./expressions/functions/StringFunctions";
+
 export * from "./expressions/functions/ListFunctions";
 export { any, all, exists, single } from "./expressions/functions/PredicateFunctions";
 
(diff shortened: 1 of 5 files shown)


packages/cypher-builder/src/Cypher.ts
```diff
@@ -93,6 +95,9 @@ export {
     sum,
     randomUUID,
 } from "./expressions/functions/CypherFunction";
+
+export * from "./expressions/functions/StringFunctions";
```
MacondoExpress, line 99: Good that we start to categorize them!

MacondoExpress (commented):
I think you forgot to include the CypherFunction class in the commit!

angrykoala:
I'm not sure what do you mean. The Function class is now exported normally, instead of as a type

packages/cypher-builder/src/expressions/functions/StringFunctions.ts
```diff
@@ -0,0 +1,75 @@
...
+export function rTrim(original: Expr): CypherFunction {
+    return new CypherFunction("rTrim", [original]);
+}
+
+export function split(original: Expr, delimiter: Expr): CypherFunction {
+    return new CypherFunction("split", [original, delimiter]);
+}
+
+export function substring(original: Expr, start: Expr, length?: Expr): CypherFunction {
+    return new CypherFunction("substring", filterTruthy([original, start, length]));
```
darrellwarde, line 54: Why the `filterTruthy` here? Could result in some unexpected behaviour no?

darrellwarde (approved):
LGTM! Just one question.

packages/cypher-builder/src/expressions/functions/StringFunctions.ts
angrykoala: The main reason is the `length` argument being optional. At the moment, CypherFunction class do not accept undefined parameters.

The only unexpected behaviour would be if a JavaScript user passes undefined in the first 2 parameters. But that would generate errors anyway

MacondoExpress:
Don't worry I forgot that was actually the current name of the Function class, so I thought it was a newer class 🤦

PR #2429, opened 2022-11-16
Add reduce Function

angrykoala:
# Description
Adds [`reduce`](https://neo4j.com/docs/cypher-manual/current/functions/list/#functions-reduce) function


a-alle (approved):
nice!!

PR #2449, opened 2022-11-21
docs: Passing Neo4j database info version

tbwiss:
This allows a user to use this functionality, setting the `neo4jDatabaseInfo` version to X.X, without using the full `Neo4jDatabaseInfo` class (which is not exported by the library).


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-2449.surge.sh

MacondoExpress:
The codebase needs to be changed to support the new documentation approach! 
In particular, the Neo4jDatabaseInfo has to be constructed anyway (just not by the user, but directly by the library).

tbwiss:
Alright! Then I'll close this PR.

PR #2274, opened 2022-10-21
2267 field sort order is incorrect when a relationship field is also queried

angrykoala:
# Description
Fix incorrect ordering on nested fields with interfaces #2267 

To fix this, an extra `ORDER BY` statement is added at the end of the read queries when there are subqueries

Issue #2267, opened 2022-10-20
Field sort order is incorrect when a relationship field is also queried


vpctorr:
**Describe the bug**
A sort order defined on a query is incorrect when a relationship field is also queried. If the relationship field returns a result, its parent will end up after the results that return an empty relationship field.

**To Reproduce**
The following query
```
  places(options: { sort: { displayName: ASC } }) {
    displayName
  }
```
yields
```
{
  "data": {
    "places": [
      { "displayName": "786 – Burger & brasserie" },
      { "displayName": "8 à Huit" },
      **{ "displayName": "9ème Sauvage" },**
      { "displayName": "A One Shop" },
      { "displayName": "A.P.C. Surplus" },
      { "displayName": "A2Pas" }
    ]
  }
}
```

While the following query
```
  places(options: { sort: { displayName: ASC } }) {
    displayName
    activity {
      id
    }
  }
```
yields
```
{
  "data": {
    "places": [
      { "displayName": "786 – Burger & brasserie","activity": [] },
      { "displayName": "8 à Huit", "activity": [] },
      { "displayName": "A One Shop", "activity": [] },
      { "displayName": "A.P.C. Surplus", "activity": [] },
      { "displayName": "A2Pas", "activity": [] },
      **{ "displayName": "9ème Sauvage", "activity": [ { "id": "some_id" } ] }**
    ]
  }
}
```

**Expected behavior**
It is expected that the sorting order is preserved regardless of what fields are queried.

**System (please complete the following information):**

* OS: macOS
* Version: occurs on both v3.10.1 and the latest commit (01ad3cb) 
* Node.js version: 16.17.1

**Additional context**
In case this might be relevant: "activity" is an interface.

neo4j-team-graphql:
We've been able to confirm this bug using the steps to reproduce that you provided - many thanks @vpctorr! :pray: We will now prioritise the bug and address it appropriately.

angrykoala:
Hi @vpctorr 
Thanks for this report, this is fixed in PR #2274 and will be merged and published shortly

vpctorr:
> Hi @vpctorr Thanks for this report, this is fixed in PR #2274 and will be merged and published shortly

Hey, thanks for fixing that real quick! 🙌

PR #2476, opened 2022-11-28
Remove additional SKIP and LIMIT from generated Cypher

### LINKED ISSUE #2396: `offset` / `limit` returning invalid results with `where` filtering
**Describe the bug**
Offset / limit don't seem to work properly, when used in accordance with filters, returning 0 results past a certain point when I know I should be getting some.

**Type definitions**
Isolated type definitions that cause the issue for me:

```graphql
type PostalCode @exclude(operations: [DELETE]) {
  archivedAt: DateTime
  number: String! @id(autogenerate: false)
  
  address: [Address!]! @relationship(type: "HAS_POSTAL_CODE", direction: IN)
}
extend type PostalCode
  @auth(
    rules: [
      { where: { archivedAt: null } }
    ]
  )

union AddressNode = Estate | Customer | Demand | Deed

type Address @exclude(operations: [DELETE]) {  
  archivedAt: DateTime
  uuid: ID! @id
  createdAt: DateTime! @timestamp(operations: [CREATE])
  updatedAt: DateTime! @timestamp(operations: [CREATE, UPDATE])

  address: String! @unique
  streetNumber: String
  route: String! @coalesce(value: "")
  postalCode: PostalCode! @relationship(type: "HAS_POSTAL_CODE", direction: OUT)
  locality: String! @coalesce(value: "")
  administrativeAreaLevel1: String! @coalesce(value: "")
  administrativeAreaLevel2: String
  country: String! @coalesce(value: "")
  location: Point!
  
  node: [AddressNode!]! @relationship(type: "HAS_ADDRESS", direction: IN)
}
extend type Address
  @auth(
    rules: [
      { where: { archivedAt: null } }
    ]
  )

type Mandate @exclude(operations: [DELETE]) {
  archivedAt: DateTime
  number: String! @unique # numéro
  createdAt: DateTime! @timestamp(operations: [CREATE])
  updatedAt: DateTime! @timestamp(operations: [CREATE, UPDATE])

  price: Float!
  
  valuation: Valuation! @relationship(type: "HAS_VALUATION", direction: OUT)
}
extend type Mandate
  @auth(
    rules: [
      { where: { archivedAt: null } }
    ]
  )

type Valuation @exclude(operations: [DELETE]) {
  archivedAt: DateTime
  uuid: ID! @id
  createdAt: DateTime! @timestamp(operations: [CREATE])
  updatedAt: DateTime! @timestamp(operations: [CREATE, UPDATE])

  estate: Estate @relationship(type: "VALUATION_FOR", direction: OUT)
}
extend type Valuation
  @auth(
    rules: [
      { where: { archivedAt: null } }
    ]
  )
  
enum EstateType {
  APARTMENT
  HOUSE_VILLA
  TOWNHOUSE
  LAND
  PARKING
  BOX
  BUILDING
  COMMERCIAL_PREMISE
  CHALET
  CASTLE
  OFFICE
  BUSINESS_FUND
}
type Estate @exclude(operations: [DELETE]) {
  archivedAt: DateTime
  uuid: ID! @id
  createdAt: DateTime! @timestamp(operations: [CREATE])
  updatedAt: DateTime! @timestamp(operations: [CREATE, UPDATE])

  estateType: EstateType!  
  area: Float!
  floor: Int
  
  address: Address @relationship(type: "HAS_ADDRESS", direction: OUT)
}
extend type Estate
  @auth(
    rules: [
      { where: { archivedAt: null } }
    ]
  )
```

**To Reproduce**

1. Import [this JSON data](https://github.com/neo4j/graphql/files/9971436/response.txt), matching the types described above
2. Run this Query:

```graphql
query Mandates($where: MandateWhere, $options: MandateOptions) {
  mandates(options: $options, where: $where) {
    valuation { 
      estate { 
        uuid 
      } 
    }
  }
}
```

with these variables:

```json
{
  "options": {},
  "where": {
    "price_GTE": 0,
    "valuation": {
      "estate": {
        "address": {
          "postalCode": {
            "number_IN": [
              "13001"
            ]
          }
        },
        "area_GTE": 0,
        "estateType_IN": [
          "APARTMENT"
        ],
        "floor_GTE": 0
      }
    }
  }
}
```

You should get about 26 results there.

If you try adding `"offset": 0` and `"limit": 20` to the options, you should get 20 results, accordingly.

However, if you set `offset` to 20 and `limit` to 40, you won't get any returned results.

From `DEBUG=@neo4j/graphql:*`:

```
Cypher:
MATCH (this:`Mandate`)
WHERE ((this.price >= $param0 AND EXISTS {
    MATCH (this)-[:HAS_VALUATION]->(this0:`Valuation`)
    WHERE EXISTS {
        MATCH (this0)-[:VALUATION_FOR]->(this1:`Estate`)
        WHERE (this1.area >= $param1 AND this1.floor >= $param2 AND this1.estateType IN $param3 AND EXISTS {
            MATCH (this1)-[:HAS_ADDRESS]->(this2:`Address`)
            WHERE EXISTS {
                MATCH (this2)-[:HAS_POSTAL_CODE]->(this3:`PostalCode`)
                WHERE this3.number IN $param4
            }
        })
    }
}) AND this.archivedAt IS NULL)

WITH *

SKIP $param5
LIMIT $param6
CALL {
    WITH this
    MATCH (this)-[this4:HAS_VALUATION]->(this_valuation:`Valuation`)
    WHERE this_valuation.archivedAt IS NULL
    CALL {
        WITH this_valuation
        MATCH (this_valuation)-[this5:VALUATION_FOR]->(this_valuation_estate:`Estate`)
        WHERE this_valuation_estate.archivedAt IS NULL
        WITH this_valuation_estate { .uuid } AS this_valuation_estate
        RETURN head(collect(this_valuation_estate)) AS this_valuation_estate
    }
    WITH this_valuation { estate: this_valuation_estate } AS this_valuation
    RETURN head(collect(this_valuation)) AS this_valuation
}
RETURN this { valuation: this_valuation } AS this

SKIP $param7
LIMIT $param8
Params:
{
  "param0": 0,
  "param1": 0,
  "param2": {
    "low": 0,
    "high": 0
  },
  "param3": [
    "APARTMENT"
  ],
  "param4": [
    "13001"
  ],
  "param5": {
    "low": 20,
    "high": 0
  },
  "param6": {
    "low": 40,
    "high": 0
  },
  "param7": {
    "low": 20,
    "high": 0
  },
  "param8": {
    "low": 40,
    "high": 0
  }
}
```

**Expected behavior**
Out of 26 total results, I would expect 6 results with `"offset": 20` and `"limit": 40`.

* OS: Manjaro Linux 22.0.0 Sikaris
* Version: 3.12.0
* Node.js version: v16.16.0

darrellwarde:
# Description

> **Note**
> 

Removes an additional `SKIP` and `LIMIT` from generated Cypher which was causing pagination to behave incorrectly.

Note: this essentially directly reverts a fix introduced in #2274 which causes the issue fixed in this PR, yet the integration introduced in that PR still passes. This is obviously unexpected, so should be investigated to ensure this doesn't introduce another regression. @angrykoala, I imagine you may be able to provide some more context here.

# Issue

> **Note**
>
>  

Closes #2396

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2476.surge.sh

mjfwebb:
@darrellwarde Am I correct in thinking that these tests generally get updated to match whatever the newly created cypher is and so aren't really reliable for regression testing?

darrellwarde:
> @darrellwarde Am I correct in thinking that these tests generally get updated to match whatever the newly created cypher is and so aren't really reliable for regression testing?

Those are the TCK tests - but we should still be eyeballing them to make sure the Cypher is valid.

But this is an actual integration test testing functionality, working before and after this reversion!

packages/graphql/src/translate/translate-read.ts
```diff
@@ -128,16 +128,6 @@ export function translateRead(
         );
     }
 
-    if (!projectionSubqueries.empty && hasOrdering) {
```
MacondoExpress, line 131: My understanding of the reason for this conditional block is very limited, the output produced by the TCK tests by removing it, looks good to me, but it may be just of the lack of tests.
Looking at the history seems that this part was touched by the Fulltext improvements and the CypherBuilder migration, does @angrykoala  or @Liam-Doodson  have more clue about it?

Liam-Doodson: My understanding of the logic here is also fairly limited. The change I made was to go from a function that added the sorting/pagination using raw cypher, to one that does the same thing using CypherBuilder.

As far as I can tell, there wasn't any logic in the old function that would've avoided this bug though.

packages/graphql/tests/integration/issues/2267.int.test.ts
```diff
@@ -71,9 +71,9 @@ describe("https://github.com/neo4j/graphql/issues/2267", () => {
         await session.run(`
         CREATE(:${Place} {displayName: "786 aa"})
         CREATE(:${Place} {displayName: "8 à Huita"})
-        CREATE(:${Place} {displayName: "9ème Sauvagea"})<-[:ACTIVITY]-(:${Story} {name: "A story"})
-        CREATE(:${Place} {displayName: "A One Shopa"})
         CREATE(:${Place} {displayName: "zaza"})
+        CREATE(:${Place} {displayName: "A One Shopa"})
+        CREATE(:${Place} {displayName: "9ème Sauvagea"})<-[:ACTIVITY]-(:${Story} {name: "A story"})
```
angrykoala, line 76: Why this change?

This test is quite delicate, as it is testing ordering, by changing the order the elements were created you may be changing the outcome of the test if the limit is lacking here

Could this be the reason this test is now passing despite reverting the changes?

.github/workflows/issue-labeled.yml
```diff
@@ -5,6 +5,34 @@ on:
     types: [labeled]
 
 jobs:
+  bug_report_label:
```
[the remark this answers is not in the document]
angrykoala: +1

packages/graphql/src/translate/translate-read.ts
angrykoala: My main worry is that this may break again the fix for #2267, but as you mention, our lack of tests may have a problem here

packages/graphql/tests/integration/issues/2396.int.test.ts
```diff
@@ -0,0 +1,859 @@
...
+                                            create: { node: { postalCode: { create: { node: { number: "13001" } } } } },
+                                        },
+                                        area: 48,
+                                        estateType: "APARTMENT",
+                                        floor: 3,
+                                    },
+                                },
+                            },
+                        },
+                    },
```
angrykoala, line 694: I know we usually try to replicate the user example as close as possible, but couldn't we replicate the same problem with a smaller dataset?

packages/graphql/tests/tck/issues/2267.test.ts
```diff
@@ -0,0 +1,94 @@
...
+                MATCH (this)<-[this0:ACTIVITY]-(this_Post:\`Post\`)
+                RETURN { __resolveType: \\"Post\\", name: this_Post.name } AS this_activity
+                UNION
+                WITH this
+                MATCH (this)<-[this1:ACTIVITY]-(this_Story:\`Story\`)
+                RETURN { __resolveType: \\"Story\\", name: this_Story.name } AS this_activity
+            }
+            RETURN collect(this_activity) AS this_activity
+            }
+            RETURN this { .displayName, activity: this_activity } AS this"
```
angrykoala, line 89: I'm afraid this may be incorrect.

I think the subquery with union will change the order of the rows and we need to order after the collect.
I may be wrong though

packages/graphql/tests/integration/issues/2267.int.test.ts
MacondoExpress: It may worth include the 2267 TCK tests to highlight the initial reason for the changes

darrellwarde: The test was always passing - I reordered things here to see if that was the reason the test didn't start failing following the change. We were creating the nodes in alphabetical order, and then querying them ordering them by this same order - so I am just removing this uncertainty.

angrykoala: Fair enough :+1:

.github/workflows/issue-labeled.yml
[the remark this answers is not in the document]
darrellwarde: Noted - I will leave this here for this PR, it was added in reflection of the issue which it is linked to, and will help us spot issues like this in future.

darrellwarde:
There is no regression in tests for this fix, and also no regression in performance.

packages/graphql/tests/integration/issues/2396.int.test.ts
darrellwarde: Hahahaha, perhaps, is there any major issue given that it's an exact replica? It is also cleaned up following execution.

angrykoala (approved):

PR #2489, opened 2022-11-30
Update cypher rfc with experimental option

angrykoala:
# Description
Updates RFC to cover the new `@cypher` behaviour to be opt-in


darrellwarde (approved):
Looks good - are we going to throw a helpful error if the user _hasn't_ aliased their return?

angrykoala:
I'm not sure that should be within the scope, as we may not be able to do it reliably at the moment

PR #2515, opened 2022-12-05
Replace `all` predicate with `any` for auth bind checks

darrellwarde:
# Description

> **Note**
> 

This replaces the `all` predicate with `any` when performing auth bind checks - this will mean that when validating the modified data, that only one path must match the specified rule, instead of all of them. This makes more sense in the context of auth, and probably should be the default. Note, this is technically a breaking change, so we need to have a discussion around this.

## Complexity

> **Note**
>

Complexity: Low

# Issue

> **Note**
>
>  

Closes


danstarns:
Hi, @darrellwarde, I am using this functionality and this raises a concern for me. It could be me not understanding the difference between `all` and `any` so feel free to call me out on that if I am wrong. 

If you have this schema: 

**TL;DL**
Users could post on behalf of others. 

```gql
type Post {
  id: ID
  creator: User! @relationship(type: "HAS_POST", direction: IN)
}

type User {
  id: ID
  name: String
}

extend type Post @auth(rules: [{ operations: [CREATE], bind: { creator: { id: "$jwt.sub" } } }])
```

You already have a user.

The user could then connect a post to themselves and that would satisfy the `any` check because there is only 1 post.

After the user has 1 post connected to themselves **another** user could then go and connect posts to that user because it no longer checks **all** `creator: { id: "$jwt.sub" } }` and is satisfied if just one matches that predicate.

darrellwarde:
> Hi, @darrellwarde, I am using this functionality and this raises a concern for me. It could be me not understanding the difference between `all` and `any` so feel free to call me out on that if I am wrong.
> 
> If you have this schema:
> 
> **TL;DR** Users could post on behalf of others.
> 
> ```graphql
> type Post {
>   id: ID
>   creator: User! @relationship(type: "HAS_POST", direction: IN)
> }
> 
> type User {
>   id: ID
>   name: String
>   posts: [Post!]! @relationship(type: "HAS_POST", direction: OUT)
> }
> 
> extend type Post @auth(rules: [{ operations: [CREATE], bind: { creator: { id: "$jwt.sub" } } }])
> ```
> 
> You already have a user.
> 
> The user could then connect a post to themselves and that would satisfy the `any` check because there is only 1 post.
> 
> After the user has 1 post connected to themselves **another** user could then go and connect posts to that user because it no longer checks **all** `creator: { id: "$jwt.sub" } }` and is satisfied if just one matches that predicate.

Interesting, I would hope some opinions would come out of the woodwork here!

My knowledge was that `any`, `all`, `single` and `none` apply only to situations where lists are involved. So taking a slightly different example:

```graphql
type System {
  name: String!
  administrators: [User!]! @relationship(type: "IS_ADMIN", direction: IN)
}

type User {
  id: ID!
}

extend type System @auth(rules: [{ bind: { administrators: { id: "$jwt.sub" } } }])
```

The current implementation says "after any operation, the ID of _all_ administrators must match the current JWT subject". This obviously doesn't make sense, that means that all administrators would have to be the same user! What we really want is "after any operation, the ID of _any_ administrators must match the current JWT subject". 

With your example:

```graphql
type Post {
  id: ID
  creator: User! @relationship(type: "HAS_POST", direction: IN)
}
```

Because `creator` is a non-nullable field, on creation, `creator` must be set, and because we've applied the `bind` rule, it will only allow the creator to have the same ID as the current JWT subject. The predicates described above don't apply here because this is a one-to-many relationship.

I would love to work through any examples of many-to-many that you or anybody has which might be impacted by this change, but if my understanding is correct, this should be the correct implementation! It should be noted that this has been raised by users who were having the issue of "but it's impossible for _all_ of them to match the JWT subject".

neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2515.surge.sh

packages/graphql/tests/integration/issues/bind-any.int.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    });
+
+    test("should allow the operation when predicate is any", async () => {
+        const neoSchema = new Neo4jGraphQL({
+            typeDefs,
+            driver,
+            plugins: {
+                auth: new Neo4jGraphQLAuthJWTPlugin({
+                    secret: "secret",
+                    bindPredicate: "any",
```
angrykoala, line 90: I don't see documentation for this parameter :thinking:

angrykoala (dismissed):

packages/graphql/tests/integration/issues/bind-any.int.test.ts
darrellwarde: Added something if you would like to check!

MacondoExpress:
I believe that there is some case where the predicate wanted is `all` and not `any`.
Consider this Scenario:
```graphql
 type Post {
  id: ID
  moderator: [User!]! @relationship(type: "MODERATES_POST", direction: IN)
}

type User {
  id: ID
  name: String
  isModerator: Boolean
  moderates: [Post!]! @relationship(type: "MODERATES_POST", direction: OUT)
}

extend type User @auth(rules: [{ operations: [CREATE, UPDATE], bind: { id: "$jwt.sub", isModerator: "$jwt.moderator" } } ])

extend type Post @auth(rules: [{ operations: [CREATE, UPDATE, CONNECT], bind: { moderator: { isModerator: true } } }])
```

In my opinion, the above should behave as:
 - A user could create a `User` only if the property `id` is matching with the JWT claim `sub` and if the property `isModerator` is matching the JWT claim `moderator`.
- A user could create or update a `Post` and connect it to moderators only if the `User` property `isModerator` is true.
With the current predicate behavior (`all`), no changes are needed to implement the above, although with the new predicate (`any`), if a `Post` is already connected with a moderator then it can be connected to a User with the property `isModerator` set to false.

Anyway as this change is opt-in, I will not see any issue to include this predicate as an option.

darrellwarde:
> I believe that there is some case where the predicate wanted is `all` and not `any`. Consider this Scenario:
> 
> ```graphql
>  type Post {
>   id: ID
>   moderator: [User!]! @relationship(type: "MODERATES_POST", direction: IN)
> }
> 
> type User {
>   id: ID
>   name: String
>   isModerator: Boolean
>   moderates: [Post!]! @relationship(type: "MODERATES_POST", direction: OUT)
> }
> 
> extend type User @auth(rules: [{ operations: [CREATE, UPDATE], bind: { id: "$jwt.sub", isModerator: "$jwt.moderator" } } ])
> 
> extend type Post @auth(rules: [{ operations: [CREATE, UPDATE, CONNECT], bind: { moderator: { isModerator: true } } }])
> ```
> 
> In my opinion, the above should behave as:
> 
> * A user could create a `User` only if the property `id` is matching with the JWT claim `sub` and if the property `isModerator` is matching the JWT claim `moderator`.
> * A user could create or update a `Post` and connect it to moderators only if the `User` property `isModerator` is true.
> 
> With the current predicate behavior (`all`), no changes are needed to implement the above, although with the new predicate (`any`), if a `Post` is already connected with a moderator then it can be connected to a User with the property `isModerator` set to false.
> 
> Anyway as this change is opt-in, I will not see any issue to include this predicate as an option.

I'm not sure if the above is accurate, I think those authorization rules aren't as you would ever use - for instance, why would you even need an `isModerator` flag for a field which is named `moderator`? Surely that's implied? Frankly, any authorization rule which doesn't compare to some unique value in a JWT isn't really a rule at all.

But agreed that as this is opt-in, it does not cause a huge issue. I will merge today.

PR #2455, opened 2022-11-22
Cypher run first column

angrykoala:
# Description

Removes runFirstColumn from top level cypher.

To do this, the field `Details` from the `EXPLAIN` result of the given query is used to collect the results.

Note that this solution seems to work, but may be a bit flaky

This PR depends on #2450 

This PR includes the new parameter `experimental` in `@cypher` directives

PR #2545, opened 2022-12-12
Add columnName for field cypher fields

angrykoala:
# Description
As a follow up for #2455 
Adds new behaviour with `columnName` argument to `@cypher` fields.

### DIFF
--- a/.changeset/clever-experts-return.md
+++ b/.changeset/clever-experts-return.md
@@ -1,0 +1,5 @@
+---
+"@neo4j/cypher-builder": patch
+---
+
+Support for UNWIND statement after CALL { ... }
--- a/packages/cypher-builder/src/clauses/Call.test.ts
+++ b/packages/cypher-builder/src/clauses/Call.test.ts
@@ -122,4 +122,31 @@ describe("CypherBuilder Call", () => {
             }
         `);
     });
+
+    it("CALL with unwind", () => {
+        const node = new Cypher.Node({ labels: ["Movie"] });
+
+        const matchClause = new Cypher.Match(node)
+            .where(Cypher.eq(new Cypher.Param("aa"), new Cypher.Param("bb")))
+            .return([node.property("title"), "movie"]);
+
+        const clause = new Cypher.Call(matchClause).unwind("*");
+        console.log(clause);
+        const queryResult = clause.build();
+        expect(queryResult.cypher).toMatchInlineSnapshot(`
+            "CALL {
+                MATCH (this0:\`Movie\`)
+                WHERE $param0 = $param1
+                RETURN this0.title AS movie
+            }
+            UNWIND *"
+        `);
+
+        expect(queryResult.params).toMatchInlineSnapshot(`
+            Object {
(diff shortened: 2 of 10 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2545.surge.sh

darrellwarde:
In `packages/graphql/src/graphql/directives/cypher.ts`, we need to add the new argument. I'm very surprised that this even passed type definition validation! Also, any documentation for this new argument?

.changeset/clever-experts-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neo4j/cypher-builder": patch
+---
+
+Support for UNWIND statement after CALL { ... }
```
darrellwarde, line 5: No changeset for the library needed?

packages/cypher-builder/src/clauses/Call.test.ts
```diff
@@ -122,4 +122,31 @@ describe("CypherBuilder Call", () => {
...
+        const clause = new Cypher.Call(matchClause).unwind("*");
+        console.log(clause);
+        const queryResult = clause.build();
+        expect(queryResult.cypher).toMatchInlineSnapshot(`
+            "CALL {
+                MATCH (this0:\`Movie\`)
+                WHERE $param0 = $param1
+                RETURN this0.title AS movie
+            }
+            UNWIND *"
```
darrellwarde, line 142: Do you not always need to alias an `UNWIND`?

.changeset/clever-experts-return.md
angrykoala: The changeset at #2455 already covers cypher changes, so no sense in adding a second one if we are releasing both top level and field in one go

packages/cypher-builder/src/clauses/Call.test.ts
darrellwarde: I think it was more of a general observation of, should it be possible to generate invalid Cypher like this through the Cypher Builder?

packages/graphql/src/translate/projection/subquery/translate-cypher-directive-projection.ts
```diff
@@ -230,6 +243,37 @@ function createCypherDirectiveApocProcedure({
     return apocClause;
 }
 
+function createExperimentalCypherStatement({
```
darrellwarde, line 246: I know we're marking this as experimental for our users, but is it really experimental for us? Not worth naming this function with a more long-term name?

darrellwarde (approved):
I realise you've primarily copied the existing tests, but some of this custom Cypher stuff is really poorly tested (again, not your doing). Field-level Cypher only tests for returning a string an a union... No list values? No interfaces? No "normal" relationship fields? Scalars? Enums? The list goes on! 😱 Feels like we're perhaps carrying a lot of risk here, and we should sure this up before we go "stable" with `columnName` next year.

darrellwarde (approved):

Issue #2577, opened 2022-12-15
Sorting by multiple fields fails if any field is `@cypher`


angrykoala:
**Describe the bug**
When sorting by multiple fields, in some cases the generated cypher is invalid

**Type definitions**
```graphql
    type Movie{
        id: ID!
        title: String!
        oneActorName: String @cypher(statement: "MATCH (this)<-[:ACTED_IN]-(a:Person) RETURN a.name")
    }
```

**To Reproduce**
The following query: 
```graphql
query SortWithTopLevelCypher_only {
    movies(options: { limit: 5, sort: { title: ASC, oneActorName: DESC } }) {
        title
        oneActorName
    }
}
```
Fails with the error:
```
Neo4jError: Variable `this_oneActorName` not defined (line 7, column 42 (offset: 122))
"ORDER BY this.title ASC, this_oneActorName DESC"
```

The generated Cypher: 
```cypher
MATCH (this:`Movie`)

WITH *
ORDER BY this.title ASC, this_oneActorName DESC
LIMIT $param0
CALL {
    WITH this
    UNWIND apoc.cypher.runFirstColumnSingle("MATCH (this)<-[:ACTED_IN]-(a:Person) RETURN a.name", { this: this, auth: $auth }) AS this_oneActorName
    RETURN head(collect(this_oneActorName)) AS this_oneActorName
}
RETURN this { .title, oneActorName: this_oneActorName } AS this
```

**System (please complete the following information):**

* Version: @neo4j/graphql@3.14.0
* Neo4j 5.1.0

neo4j-team-graphql:
Many thanks for raising this bug report @angrykoala. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

neo4j-team-graphql:
Many thanks for raising this bug report @angrykoala. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

[...]

PR #2573, opened 2022-12-15
Add tests on top-level connections with sort

angrykoala:
# Description
Add some tests related to connection cursor-based sorting.

* Removed unused code in translate-read
* Moved some tests from `sort.int.ts` to `connections/sort.int.test` 

## Complexity

Complexity: Low

PR #2578, opened 2022-12-15
Multiple sort bug

angrykoala:
# Description
Fixes the case of multiple sort fields in the same object (#2577). 

This PR depends on #2573 

## Complexity

Complexity:Low


darrellwarde (dismissed):
LGTM, good spot!

darrellwarde:
Actually, needs a changeset!

PR #2580, opened 2022-12-15
Performance sort connection

angrykoala:
# Description
Improves sort performance by executing it before subqueries if possible

This PR depends on #2573

# Complexity
Medium


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2580.surge.sh

packages/graphql/src/translate/translate-read.ts
```diff
@@ -136,6 +136,26 @@ export function translateRead(
     let connectionPreClauses: Cypher.Clause | undefined;
 
     if (isRootConnectionField) {
+        const hasConnectionOrdering = resolveTree.args.first || resolveTree.args.after || resolveTree.args.sort;
+        if (hasConnectionOrdering) {
+            const afterInput = resolveTree.args.after as string | undefined;
```
a-alle, line 141: can `resolveTree.args.after` be sth other than string if defined? I see the old code is explicitly checking its type before passing it to `cursorToOffset`

packages/graphql/tests/tck/issues/1364.test.ts
```diff
@@ -147,8 +148,9 @@ describe("https://github.com/neo4j/graphql/issues/1364", () => {
                 RETURN count(DISTINCT genre)\\", { this: this, auth: $auth }) AS this_totalGenres
                 RETURN head(collect(this_totalGenres)) AS this_totalGenres
             }
+            WITH *
+            ORDER BY this_totalGenres ASC
```
a-alle, line 152: Does it make sense to add a test when 2 cypher fields are sorted? And the combination of cypher and normal fields?

angrykoala: the second case you mention is already covered In #2578 

The case of 2 cyphers is not covered and I agree it'll be a good idea

packages/graphql/src/translate/translate-read.ts
darrellwarde: Not as far as I'm aware - it is a type `String` in the schema, so that should give it enough type safety.

darrellwarde (approved):
It looks good! Just to check, do we have a TCK test which also uses `after`? Just concerned that we will introduce `SKIP` into Cypher twice!

PR #2621, opened 2022-12-22
fix: set neo4j-version in GHA to older version for 4.x

tbwiss:
Our pull request tests which rely on a 4.x Neo4j DB fail due to an inability to connect to the Neo4j DB.
Trying to reset to an older version for 4.3 and 4.4 to see if the problem persists.


MacondoExpress (approved):
LGTM, Just to provide a little bit of context for future reference or for a benefit to others, the current latest Neo4j docker images released under the tag 4.3-enterprise and 4.4-enterprise have some connectivity issues with APOC. Will be fixed by future builds, but until then, is rolled back to the latest working images.

PR #2624, opened 2022-12-22
feat: GHA matrix with node versions for nightly build

tbwiss:
With this PR we test if the library passes integration tests with Node.js version 19.
At the moment we only test it on GHA against the latest node.js lts (v18.X)
For now, we only check it for the nightly build.

### DIFF
--- a/.github/workflows/reusable-integration-tests-on-prem-nightly.yml
+++ b/.github/workflows/reusable-integration-tests-on-prem-nightly.yml
@@ -39,7 +39,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [ "lts/*", 19 ]
+        node: [ "lts/*", "latest"]
         packages:
           [
             { package: "ogm", shard: 1/1 },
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-integration-tests-on-prem.yml
```diff
@@ -14,6 +14,7 @@ jobs:
   integration-tests:
     strategy:
       matrix:
+        node: [ "lts/*", 19 ]
```
tbwiss, line 17: The changes in this file are only to test if it works!
will be removed when successfully tested

tbwiss: Removed

.github/workflows/reusable-integration-tests-on-prem-nightly.yml
```diff
@@ -39,6 +39,7 @@ jobs:
 
     strategy:
       matrix:
+        node: [ "lts/*", 19 ]
```
MacondoExpress, line 42: I think it will be more convenient to test against the latest Node.JS build so that we could detect compatibility issues earliest possible.
```suggestion
        node: [ "lts/*", "latest"]
```

MacondoExpress (approved):

.github/workflows/reusable-integration-tests-on-prem-nightly.yml
tbwiss: Yes good one!

PR #2664, opened 2023-01-03
fix broken schema generation when point type with subscriptions

### LINKED ISSUE #2653: Subscriptions with Point type throw error during schema augmentation
**Describe the bug**
If you use a point field, and then try to generate subscriptions from your schema, an error is thrown regarding an incorrect type.

**Type definitions**
This is a simple type def that would throw this error:

```gql
type Location {
    id: ID!
    name: String!
    name_lc: String!
    description: String!
    point: Point!
}
```

**To Reproduce**

1. Create simple schema with a point field within the type definition
2. Add in subscriptions via Neo4jGraphQLSubscriptionsSingleInstancePlugin (The same problem occurs with Neo4jGraphQLSubscriptionsAMQPPlugin)
3. Try to run the program, and it will throw an error similar to:
```Error: TypeError[LocationSubscriptionWhere.point_IN]: Provided incorrect InputType: "[Point]"```

**Expected behavior**
It should not throw error

**Screenshots**

* OS: M1 mac
* Version: @neo4j/graphql@3.14.1
* Node.js version: 16.13.1

**Additional context**

### PR DESCRIPTION
**Pull request by a-alle**
# Description

Schema generation failed when Point or CartesianPoint typed fields existed in a type if subscriptions enabled. This PR fixes the crash and adds tests.

## Complexity

Complexity: Low

# Issue

Closes #2653

# Checklist

### CODE REVIEW

**Comment by neo4j-team-graphql** [comment]

Check out your changes at https://neo4j-graphql-toolbox-stage-2664.surge.sh

File: packages/graphql/tests/integration/subscriptions/spatial-types.int.test.ts [hunk 1]
```diff
@@ -0,0 +1,729 @@
...
+        session = await neo4j.getSession();
+    });
+
+    afterEach(async () => {
+        await cleanNodes(session, [typeMovie]);
+        await session.close();
+        await driver.close();
+    });
+
+    test("create type with Point field", async () => {
```
angrykoala, line 68: I see these are very similar tests, just done to Point and cartesian points. Would it be possible to reuse the same tests and just change the input with `test.each`?
 
 I assume is not that simple, so I'm only mentioning it in case it is relatively easy to do

angrykoala (approved):

packages/graphql/tests/integration/subscriptions/spatial-types.int.test.ts
a-alle: Not much else is similar, they are different fields in the graphql type and even if they weren't, the input types (PointInput, CartesianPointInput) have different fields (long/lat/height vs x/y/z). It can be done but so much has to be specified that I don't see the benefit in uniting them, unless you have a strong opinion on this :D

PR #2679, opened 2023-01-05
Fix invalid cypher when aggregating on connection where

### LINKED ISSUE #2670: Invalid cypher when aggregating on connection where
given:

```graphql
type Movie {
    title: String
    genres: [Genre!]! @relationship(type: "IN_GENRE", direction: OUT)
}
type Genre {
    name: String
    movies: [Movie!]! @relationship(type: "IN_GENRE", direction: IN)
}
```

and query

```graphql
{
    movies(where: { genresConnection: { node: { moviesAggregate: { count_GT: 10 } } } }) {
        title
    }
}
```

produces the following cypher:

```cypher
MATCH (this:`Movie`)
WHERE size([(this)-[this0:IN_GENRE]->(this1:`Genre`) WHERE var2 = true | 1]) > 0
RETURN this { .actorCount } AS this
```

where `var2` is not defined.

Liam-Doodson:
# Description

Fixes invalid cypher when filtering using aggregations within a connection filter. To achieve this the connection checks have been moved out of a `size` operation into a subquery that returns a `count` that can be compared. For example:

```
WHERE size([(this)-[this0:IN_GENRE]->(this1:\`Genre\`) WHERE this1.name = $param0 | 1]) = 0
```

Becomes:

```
CALL {
    WITH this
    MATCH (this)-[this0:IN_GENRE]->(this1:\`Genre\`)
    WHERE this1.name = $param0
    RETURN count(this0) AS var2
}
WITH *
WHERE var2 = 0
```

Note that these changes result in more db hits for the same query. However, the subqueries can be executed in parallel which means that query executions times seem to be fairly consistent between the old and new cypher (with the new cypher possibly slightly faster) across several performance tests locally and using the pipeline.

## Complexity

Complexity: Medium

# Issue

Closes #2670 

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2679.surge.sh

Andy2003:
This PR is only solving the issue for connections, but not for relations:

```graphql
type Movie {
  title: String!
  actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
}

type Actor {
  name: String!
}
```

with 

```graphql
{
  movies(where: {actors: { moviesAggregate: { count_GT: 10}  } }) {
    title
  }
}
```

generates

```cypher
"MATCH (this:`Movie`)
WHERE EXISTS {
    MATCH (this0:`Actor`)-[:ACTED_IN]->(this)
    WHERE var1 = true
}
RETURN this { .title } AS this"
```

packages/graphql/src/translate/where/property-operations/create-connection-operation.ts
```diff
@@ -92,18 +91,51 @@ export function createConnectionOperation({
...
+        let whereClause: Cypher.Match | Cypher.With = matchClause;
+        let innerSubqueriesAndWhereClause: Cypher.CompositeClause | undefined;
+
+        if (preComputedSubqueries && !preComputedSubqueries.empty) {
+            whereClause = new Cypher.With("*");
+            innerSubqueriesAndWhereClause = Cypher.concat(preComputedSubqueries, whereClause);
+        }
+
+        if (whereOperator) {
+            const newWhereOperator = listPredicateStr === "all" ? Cypher.not(whereOperator) : whereOperator;
```
Andy2003, line 107: If there is no where condition an `ALL` filter will result in no elements being returned:

```graphql
{
    movies(where: { genresConnection_ALL: {} }) {
        title
    }
}
```

results in

```
MATCH (this:`Movie`)
CALL {
    WITH this
    MATCH (this)-[this0:IN_GENRE]->(this1:`Genre`)
    RETURN count(this0) AS var2
}
WITH *
WHERE var2 = 0
RETURN this { .title } AS this
```

is this intended?

packages/graphql/tests/tck/advanced-filtering.test.ts
```diff
@@ -920,7 +934,14 @@ describe("Cypher Advanced Filtering", () => {
 
                 expect(formatCypher(result.cypher)).toMatchInlineSnapshot(`
                     "MATCH (this:\`Movie\`)
-                    WHERE size([(this)-[this0:IN_GENRE]->(this1:\`Genre\`) WHERE NOT this1.name = $param0 | 1]) = 0
+                    CALL {
```
Andy2003, line 937: do we need to check for at least one occurrence as well?

```cypher
MATCH (this:`Movie`)
CALL {
    WITH this
    MATCH (this)-[this0:IN_GENRE]->(this1:`Genre`)
    WHERE NOT (this1.name = $param0)
    RETURN count(this0) AS var2
}
CALL {
    WITH this
    MATCH (this)-[this0:IN_GENRE]->(this1:`Genre`)
    WHERE this1.name = $param0
    RETURN count(this0) AS var3
}
WITH *
WHERE var2 = 0 and var3  > 0
RETURN this { .actorCount } AS this"
```

Liam-Doodson: Hi @Andy2003! I'm not sure I follow. Why would we also need to check for more than one occurrence?

Liam-Doodson:
> This PR is only solving the issue for connections, but not for relations:
> 
> ```graphql
> type Movie {
>   title: String!
>   actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
> }
> 
> type Actor {
>   name: String!
>   movies: [Movie!]! @relationship(type: "ACTED_IN", direction: OUT)
> }
> ```
> 
> with
> 
> ```graphql
> {
>   movies(where: {actors: { moviesAggregate: { count_GT: 10}  } }) {
>     title
>   }
> }
> ```
> 
> generates
> 
> ```cypher
> "MATCH (this:`Movie`)
> WHERE EXISTS {
>     MATCH (this0:`Actor`)-[:ACTED_IN]->(this)
>     WHERE var1 = true
> }
> RETURN this { .title } AS this"
> ```

Hi Andy! After a short look this seems to be a similar but different issue so I have raised the following issue for it: #2708

packages/graphql/tests/tck/advanced-filtering.test.ts
Andy2003: Just b/c it is also done for retaltions:

https://github.com/neo4j/graphql/blob/d6837b84275f7bd69c59e7671cf5e44ffa302aed/packages/graphql/tests/tck/advanced-filtering.test.ts#L772-L782

packages/graphql/src/translate/where/property-operations/create-connection-operation.ts
Liam-Doodson: This was not previously supported and before this PR resulted in invalid cypher. I've created a feature request so the team can consider if we want to support this

packages/graphql/tests/tck/advanced-filtering.test.ts
Liam-Doodson: This has always been the behaviour for connections. However, I agree we may want to keep consistent with relationships so I have created the following issue to capture this #2713

packages/graphql/src/translate/create-aggregate-where-and-params.ts
```diff
@@ -77,7 +77,7 @@ export function aggregatePreComputedWhereFields(
     const subquery = new Cypher.Call(matchQuery).innerWith(matchNode);
     return {
         predicate: Cypher.and(...predicates),
-        preComputedSubquery: subquery,
+        preComputedSubquery: Cypher.concat(subquery),
```
angrykoala, line 80: I don't understand why do you need to concatenate a single query here :thinking:

angrykoala (dismissed):
Approved, just a question regarding the cypher builder

packages/graphql/src/translate/create-aggregate-where-and-params.ts
Liam-Doodson: I agree I thought this was a bit weird too. I did this because the value is passed straight up to `createWherePredicate` which expects a `Cypher.CompositeClause` and uses the `.empty` method. Because it uses that method I couldn't see a nice way around doing this. Do you have any thoughts?

angrykoala: mmm, that's fair, tbh I don't like the whole using the CompositeClause and the `.empty` but that would be a bigger refactor. For now, a comment should be enough

Issue #2681, opened 2023-01-05
Support of fulltext index for aggregation 


Andy2003:
With the deprecation of the old index handling, the following field is generated:

```graphql
moviesAggregate(
    fulltext: MovieFulltext @deprecated(reason : "This argument has been deprecated and will be removed in version 4.0.0 of the library. Please use the top-level query that corresponds to the index you wish to query instead. More information about the changes to @fulltext can be found here: https://neo4j.com/docs/graphql-manual/current/guides/v4-migration/#_fulltext_changes."),
    where: MovieWhere): MovieAggregateSelection!
```

Currently there is no replacement for this query field, so removing the deprecated argument results in loss of functionality.

AccsoSG:
It looks like the fulltext parameter doesn't even exist in the aggregate functions anymore. 
An essential function has been lost here. For a full-text search, it is very important to be able to return the count of a results set.

AccsoSG:
I noticed that the deprecated parameter is not readable via the graphql introspection. Is there a solution for this?

The problem is that our client applications no longer work with the new fulltext directive because they use codegen and rely on the old aggregate function.

darrellwarde:
Fear not, we are aware of this oversight and will have something more comprehensive in place for 4.0.0. 🙂

FloydanTheBeast:
Are there any updates on this? Seems like a very important feature, would love to see it released

PR #2687, opened 2023-01-06
RFC: Relay first GraphQL API

angrykoala:
# Description
RFC on uniform API based on Relay


docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+### Nodes
+
+The Entity (e.g. movies) properties for the projection will be defined in a `*Node` type:
+
+```graphql
+type MovieNode {
+    title: String!
+    released: Int
+    alternativeTitles: [String!]!
+    actors(
```
Liam-Doodson, line 148: I may be miss-remembering but did we not say this would be `actorsConnection` to keep consistent and because there some confusion over whether or not this was a connection or basic type

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+            node {
+                title
+            }
+        }
+    }
+}
+```
+
+> Note that the plain `fulltext` option without `where` may still filter the results.
+
```
Liam-Doodson, line 371: It would also be interesting to see an example of `fulltext` along with aggregations as we've had a number of requests for that e.g. #2681

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+-   `Connection`: Types containing `edges` and following Relay spec.
+-   `Edge`: Types containing `node` and, optionally, fields
+-   `Where`: Filtering Types
+-   `Aggregation`: Aggregation Types
+
+*   Types in a relationship will follow the convention `[RootType][fieldName]` at the beginning.
+*   Where possible, input types will mirror normal types. E.g. `PersonNode` will be filtered by `PersonNodeWhere`
+
+### Further discussion points
```
Liam-Doodson, line 456: Are these the points that were out of scope for this initial work? If so it might be worth making that clear

Liam-Doodson (commented):
This is looking really good! I'm excited to see this get implemented!

It might also be intersting to see how you propose to break up this work as it's got some quite big changes that seem fairly dependant on each other.

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+
+Query will be done with the following operations:
+
+```graphql
+type Query {
+    moviesConnection(
+        first: Int
+        after: String
+        last: Int
+        before: String
```
a-alle, line 61: Is there a reason for having both `first` and `last`? If (correct me if I'm wrong) they are both referring to the nr of elements to return then what happens if the user sets both of them with different values - which one has precedence? I propose to condense them down into a single field, like `take` from the wording commonly used in offset-based pagination.
Also, what happens if the user sets both `after` and `before`?

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+            startCursor
+            hasPreviousPage
+            hasNextPage
+            endCursor
+        }
+    }
+}
+```
+
+Following Relay spec, every `*Connection` type has the `pageInfo` property, and each `edge` has a cursor. Note that if limit-based pagination is enabled instead of cursor based, the API will not be Relay compliant.
```
Liam-Doodson, line 252: Is it possible to see an example of limit-based pagination with the new API? It'd also be good to see how you propose to allow users to configure the pagination type

 Also, this is a bit nit picky but is it not more commonly called offset-based?

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+
+```graphql
+type MovieActorsEdge {
+    cursor: String! # Required by Relay
+    node: PersonNode! # Required by Relay
+    fields: ActedIn # Fields of the relationship
+}
+```
+
+There will be an Edge type per entity, as well as a different Edge type per relationship, relationship edges may contain the property `fields` that will hold the relationship properties.
```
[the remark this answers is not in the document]
a-alle: I also don't like the name `fields` here. I propose we stay consistent with the wording we have in place for specifying them (the `@relationshipProperties` directive) and thus using sth in the lines of `properties` or `relationshipProperties`.

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+Note that this type does **not** contain any operations in itself, it has the same properties as the user-defined entity, with the only difference being the result of the nested `actors` and `director` types, that support the top-level input and return a `*Connection`.
+
+### Filtering
+
+Filtering follows the same structure as projections, following nodes and edges. The operations `AND`, `OR` and `NOT` are available at any level. Comparison operators are present inside each field.
+
+```graphql
+query MatrixMoviesFrom1999 {
+    moviesConnection(
+        where: { edges: { node: { AND: [{ title: { contains: "Matrix" } }, { released: { eq: 1999 } }] } } }
```
a-alle, line 170: Could we consider making all operators uppercased? I find it a bit confusing to keep just `AND`, `OR` and `NOT` uppercased, and the other operators "blending" with user-defined properties

Liam-Doodson: Yeah I agree, it would be much clearer what is a field and what is an operator!

docs/rfcs/rfc-040-relay-first-api.md
darrellwarde: I agree, the cursor connections specification gives freedom of choice about whether to implement `first` and `after` or `last` and `before`... Or both. I think implementing both in the same API is a mistake, and if anything there should be a toggle between the two, bearing in mind that we've never been asked for backwards pagination.

docs/rfcs/rfc-040-relay-first-api.md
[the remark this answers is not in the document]
darrellwarde: I actually disagree with the above here, see the following:

- Users of our library who are building the API - these users are aware of Neo4j as the datastore, so `@relationshipProperties` informs them how this data will be stored in backend. They need to be aware of this link.
- Users of the built APIs - the users who will see the `fields` property. To name this `properties` or `relationshipProperties` would be leaking information about the datastore to API consumers, which could be thousands of users for a public API. The database details should be abstracted from this users.

We always need to be thinking about these two user groups when implementing a new feature - nomenclature does not necessarily have to be uniform across type definitions and the generated schema.

docs/rfcs/rfc-040-relay-first-api.md
MacondoExpress: Nice suggestion!

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+-   Fulltext is not supported
+
+## Naming conventions
+
+These are some loose conventions on type and input naming:
+
+-   `Connection`: Types containing `edges` and following Relay spec.
+-   `Edge`: Types containing `node` and, optionally, fields
+-   `Where`: Filtering Types
+-   `Aggregation`: Aggregation Types
```
darrellwarde, line 451: Worth clarifying that types ending `Connection` are strictly reserved.

```suggestion
The following are strongly enforced naming conventions:

-   `*Connection`: Types containing `edges` and following Relay spec - these are reserved types according to the spec

These are some loose conventions on type and input naming:

-   `Edge`: Types containing `node` and, optionally, fields
-   `Where`: Filtering Types
-   `Aggregation`: Aggregation Types
```

darrellwarde (commented):
I've not tried anything out yet, but here are my initial observations from flicking through the RFC!

docs/rfcs/rfc-040-relay-first-api.md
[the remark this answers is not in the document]
angrykoala: I don't have that many opinions regarding fields vs properties. But to give some context on why we are not using `edge` here.

Fields are basically the edge fields, and in every instance we provide them inside an edges fields:

```graphql
actors {
    edges {
        fields { }
    }
}
```

The main reason of moving the edge fields to its own object (unlike now) is to separate between navigation/meta fields (`node`, `aggregation`, `pageInfo`...) from "data fields" such as `name` or `title`

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: I think there is already an example of this case:
```graphql
query TopLevelFulltextAndAggregation {
    moviesConnection(
        fulltext: { phrase: "Matrix", index: MovieTitle }
        where: { edges: { node: { released: { eq: 2001 } }, fulltext: { score: { gt: 6 } } } }
        sort: { edges: { fulltext: { score: DESC } } }
        first: 10
    ) {
        edges {
            node {
                title
            }
            fulltext {
                score
            }
        }
        aggregation {
            node {
                count
            }
        }
    }
}
```

If you are asking about aggregation as a filter, it works the same as any filter for nested aggregations, in this example, filtering on movies with >= 10 actors

```graphql
query TopLevelFulltextAndAggregation {
  moviesConnection(
    fulltext: { phrase: "Matrix", index: MovieTitle }
    where: {
      edges: {
        node: { actors: { aggregation: { count: { gte: 5 } } } }
        fulltext: { score: { gt: 6 } }
      }
    }
    sort: { edges: { fulltext: { score: DESC } } }
    first: 10
  ) {
    edges {
      node {
        title
      }
      fulltext {
        score
      }
    }
    aggregation {
      node {
        count
      }
    }
  }
}
```

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: Well, how this work is going to be split is still not decided, this is mostly non-blockers decisions that we could make at some point in the future

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: This is an example of the API following Relay spec allowing for both navigations. The toggle idea is good, and this does not have to be implemented if not required, I'll make a note in the RFC to clear that up

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: I'll add an example and description in its own section.

I will not add how the configuration would work, as that would fall into how we plan to do server configuration, which is out of scope for this API design

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: TBH I find quite ugly uppercased operators

the reason for AND OR NOT being uppercased is mainly that they exist within the data objects and need to be differentiated, but it is an aesthetic decision so we can change it if most people agree is better to have all of them being uppercase

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+    actors(
+        where: PersonConnectionWhere
+        first: Int
+        after: String
+        last: Int
+        before: String
+        directed: Boolean = true
+        sort: [MovieActorsConnectionSort!]
+    ): MovieActorsConnection!
+    director(where: MovieDirectorConnectionWhere, directed: Boolean = true): MovieDirectorConnection!
```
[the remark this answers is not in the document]
angrykoala: I think this is an error that we missed, indeed it should be its own type so it can be filtered by edges

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+        }
+    }
+}
+```
+
+Following Relay spec, every `*Connection` type has the `pageInfo` property, and each `edge` has a cursor. Note that if limit-based pagination is enabled instead of cursor based, the API will not be Relay compliant.
+
+#### Back pagination
+
+Following Relay spec, the fields `last` and `before` can be used instead of `first` and `after` to do backwards pagination.
```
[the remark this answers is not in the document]
angrykoala: This will be cleared in a section

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+query Query {
+    moviesConnection(
+        fulltext: { phrase: "Test", index: MovieTitle }
+        where: { edges: { fulltext: { score: { gt: 5 } } } }
+    ) {
+        totalCount
+        edges {
+            fulltext {
+                score
+            }
```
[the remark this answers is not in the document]
angrykoala: there are 2 reasons for taking this approach (our first choice was `fulltextScore too`

1. Consistency with aggregations and other "meta" fields
2. Redability, even if more verbose, fulltext -> score is much nicer than fulltextScore

docs/rfcs/rfc-040-relay-first-api.md
MacondoExpress: I agree with both the opinion honestly.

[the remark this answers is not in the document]
MacondoExpress: The verbosity of this API is real! There are a lot of places where we traverse through edges just for consistency without other gains.
The mutations are even worse. 
You would still need the type that contains the node because the `FullTextWhere`is placed outside the node.

```graphql
where: {  node: { AND: [{ title: { contains: "Matrix" } }, { released: { eq: 1999 } }]  } }
```

```graphql
where: { fulltext: { score: { gt: 3 } } }
```

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1347 @@
...
+    AND: [PersonMoviesAggregationEdgeWhere!]
+    OR: [PersonMoviesAggregationEdgeWhere!]
+    NOT: PersonMoviesAggregationEdgeWhere
+    nodes: MoviesAggregationWhere
+    edges: ActedInAggregationEdgeWhere
+}
+
+input ActedInAggregationEdgeWhere {
+    fields: ActedInAggregationWhere
+    count: IntWhere
```
MacondoExpress, line 902: I guess that if we are not addressing the support of multiple edges between two nodes in this phase/RFC, it may be better removes all the related references, like counting the edges. Thought?

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: yeah, but after some discussion we saw that where filtering would get extremely verbose so we didn't change it **yet**

We should get some examples and extra discussion, but it shouldn't be a blocker for the overall design anyway

docs/rfcs/rfc-040-relay-first-api.md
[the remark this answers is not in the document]
angrykoala: This is a fair point, I'll add it to a list of further discussions

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+                    aggregation {
+                        node {
+                            count
+                        }
+                        fields {
+                            year {
+                                min
+                                max
+                            }
+                        }
```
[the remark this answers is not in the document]
angrykoala: You are right, RFC updated

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: I think is worth adding optional capabilities for the future, I'll add a comment however to make sure it is clear

angrykoala: Yeah, you are probably right here

docs/rfcs/rfc-040-relay-first-api.md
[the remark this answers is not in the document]
angrykoala: I'll add this to further discussion points

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1282 @@
...
+            node {
+                title
+                actors {
+                    edges {
+                        node {
+                            name
+                        }
+                        fields {
+                            year
+                        }
```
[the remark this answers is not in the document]
angrykoala: So, the idea is to have a nice place where adding "metadata" and "navigation" fields, separate from data fields. However, it is true it makes edges more verbose, I'll add this to further discussion points

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: I'll add this to further discussion points

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1351 @@
...
+
+### Fulltext
+
+If fulltext is enable for an entity using the `@fulltext` directive, the `fulltext` input will be available, as well as the `score` fields and filters as a `Float`:
+
+```graphql
+query Query {
+    moviesConnection(
+        fulltext: { phrase: "Test", index: MovieTitle }
+        where: { edges: { fulltext: { score: { gt: 5 } } } }
```
Liam-Doodson, line 332: Is there a reason that the score filter cannot only be included in the top-level fulltext argument? It feels weird to be to be able to filter by a score that may not exist and I can see this leading to confusion/bugs if we are not careful.

I was thinking the API for this should be something like this:

```graphql
query Query {
    moviesConnection(
        fulltext: { phrase: "Test", index: MovieTitle, where: { edges: { fulltext: { score: { gt: 5 } } } } }
    ) {
        totalCount
        edges {
            fulltext {
                score
            }
        }
    }
}
```

We could then only allow the `score` filter to be used within this `where` input. To me that feels similar to how we do it for `create`/`connect`/`update` but does come with the down side of some extra type definitions.

docs/rfcs/rfc-040-relay-first-api.md
Liam-Doodson: Yeah I agree this shouldn't be a blocker. I didn't really have a strong opinion on it either way, just wanted to make sure it wasn't forgotten!

Liam-Doodson: As you say, it might be interesting to see an example in an "alternatives considered" section

docs/rfcs/rfc-040-relay-first-api.md
MacondoExpress: The issue here is that you will lose the ability to combine the FullText predicate with the common one.

Liam-Doodson: I was suggesting having access to all the filters at this level e.g. 

```graphql
query Query {
    moviesConnection(
        fulltext: { phrase: "Test", index: MovieTitle, where: { edges: { node: { title: { equals: "foo" } }, fulltext: { score: { gt: 5 } } } } }
    ) {
        totalCount
        edges {
            fulltext {
                score
            }
        }
    }
}
```

MacondoExpress: You would have still to maintain the other `where` input, imagine that you want to select a Movie filtered by the FullText score combined with a very long and complex predicate. Then, after some development, you decide that you don't want any more filtering by the FullText score. Now, the only options will be:
 - Keep the predicates inside the fulltext input without using any fulltext(assuming that we support it).
- Migrate the whole predicate to the normal where input (the very likely scenario). 
I see your concerns where are originated from, but in my opinion, the alternatives are worse!

docs/rfcs/rfc-040-relay-first-api.md
```diff
@@ -0,0 +1,1351 @@
...
+
+-   `Fields` in edges vs having the fields directly on top-level
+-   Uppercase operators (e.g. `AND: [{title: CONTAINS: "Matrix"}]`)
+-   Remove top level edge from `where` and mutations
+-   `fulltextScore` vs `fulltext { score }`
+-   Connections may have a `nodes` fields as a shortcut without passing through `edges`
+-   Sort by aggregations and 1-\* relationship
+-   Union / interfaces
+-   Nested [field]connection
+
```
MacondoExpress, line 486: I just noticed, that one thing that we don't analyze at all during the process (for good reason, as the scope is already large like it is now) is the Subscription related types.
Could it be worth adding it as a discussion point?

docs/rfcs/rfc-040-relay-first-api.md
angrykoala: That's a good point, I'll try to compile examples for when we have further discussions on these topics

docs/rfcs/rfc-040-relay-first-api.md
darrellwarde: I think the proposal here is more confusing; there should be a single `where` which allows you to do all combinations of filtering. However, I totally agree on the confusion of splitting the "fulltext logic", but this is a trade-off of having it all accessible in one field.

PR #2715, opened 2023-01-11
fix: issue on generated Cypher with interface type (#2709)

### LINKED ISSUE #2709: Possible issue on generated Cypher with interface type
**Describe the bug**
Reported in Trello.
Creating this bug report to track the bug and make the handling a bit smoother.

**Type definitions**
If applicable, add your type definitions to help us reproduce the bug.

```graphql
interface Production {
  title: String!
  actors: [Actor!]!
  distribution: [DistributionHouse!]!
}

type Movie implements Production @node(label: "Film") {
  title: String!
  actors: [Actor!]!
    @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
  runtime: Int!
  distribution: [DistributionHouse!]!
    @relationship(type: "DISTRIBUTED_BY", direction: IN)
}

type Series implements Production {
  title: String!
  actors: [Actor!]!
    @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
  episodes: Int!
  distribution: [DistributionHouse!]!
    @relationship(type: "DISTRIBUTED_BY", direction: IN)
}

interface ActedIn @relationshipProperties {
  role: String!
}

interface Actor {
  name: String!
  actedIn: [Production!]!
    @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
}

type MaleActor implements Actor {
  name: String!
  actedIn: [Production!]!
    @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
  rating: Int!
}
type FemaleActor implements Actor {
  name: String!
  actedIn: [Production!]!
    @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
  age: Int!
}

interface DistributionHouse {
  name: String!
}

type Dishney implements DistributionHouse {
  name: String!
  review: String!
}

type Prime implements DistributionHouse {
  name: String!
  review: String!
}

type Netflix implements DistributionHouse {
  name: String!
  review: String!
}
```
**To Reproduce**

With the type defs from above, execute the following GraphQL query:

```graphql
query {
  movies(
    where: {
      OR: [
        {
          distributionConnection_SOME: {
            node: { _on: { Netflix: {} }, name: "" }
          }
        }
      ]
    }
  ) {
    title
  }
}
```
This results in the following generated Cypher query:

```cypher
MATCH (this:`Film`)
WHERE size([(this1:`Dishney`)-[this0:DISTRIBUTED_BY]->(this) WHERE this1.name = $param0 | 1]) > 0
RETURN this { .title } AS this
```

The problem appears to be on the label type attached to the `DISTRIBUTED_BY` relationship, as far as i can see we're trying to only query movies that are distributed by "Netflix" but the generated Cypher always try to use "Dishney", ignoring the type specified within the GraphQL query.

* Neo4j Version: 4.4.x
* @neo4j/graphql Version: 3.12

tbwiss:
# Description

> **Note**
> 

Prior to this change the `refNode` was picked, in this particular situation, by which type/node that implements the `DistributionHouse` interface and which was ordered first. This led to the bug reported.
Now first it's evaluated if any of the node names match the `node._on` value (this info comes from the GraphQL query) or the entry type (in this particular case `DistributionHouse`). Only secondly the interface implementations are taken into consideration with the same logic as before.

## Complexity

> **Note**
>

Complexity: Low

# Issue

> **Note**
>
>  

Closes #2709 

# Checklist

### DIFF
--- a/packages/graphql/src/translate/where/property-operations/create-connection-operation.ts
+++ b/packages/graphql/src/translate/where/property-operations/create-connection-operation.ts
@@ -56,29 +56,18 @@ export function createConnectionOperation({
     const operations: (Cypher.BooleanOp | Cypher.RawCypher | undefined)[] = [];
 
     Object.entries(nodeEntries).forEach((entry) => {
-        console.log("entry", entry[1]);
-
-        let test;
-        const t = entry[1]?.node?._on;
-        if (t) {
-            test = Object.keys(t)[0] || undefined;
+        let nodeOnValue: string | undefined = undefined;
+        const nodeOnObj = entry[1]?.node?._on;
+        if (nodeOnObj) {
+            nodeOnValue = Object.keys(nodeOnObj)[0];
         }
-        console.log("test", test);
 
-        // the refNode is wrong!
-        // const refNode = context.nodes.find((x) => {
-
-        //     if (x.name === test) return;
-        //     return x.interfaces.some((i) => i.name.value === entry[0]);
-        // }) as Node;
-        let refNode = context.nodes.find((x) => x.name === test || x.name === entry[0]) as Node;
+        let refNode = context.nodes.find((x) => x.name === nodeOnValue || x.name === entry[0]) as Node;
         if (!refNode) {
             refNode = context.nodes.find((x) => x.interfaces.some((i) => i.name.value === entry[0])) as Node;
         }
 
         const relationField = connectionField.relationship;
-        // console.log("relationField", relationField);
-        console.log("refNode", refNode.name);
 
         const childNode = new Cypher.Node({ labels: refNode.getLabels(context) });
         const relationship = new Cypher.Relationship({
--- a/packages/graphql/src/translate/where/property-operations/create-connection-operation.ts
+++ b/packages/graphql/src/translate/where/property-operations/create-connection-operation.ts
@@ -56,11 +56,29 @@ export function createConnectionOperation({
     const operations: (Cypher.BooleanOp | Cypher.RawCypher | undefined)[] = [];
 
     Object.entries(nodeEntries).forEach((entry) => {
-        const refNode = context.nodes.find(
-            (x) => x.name === entry[0] || x.interfaces.some((i) => i.name.value === entry[0])
-        ) as Node;
+        console.log("entry", entry[1]);
+
+        let test;
+        const t = entry[1]?.node?._on;
+        if (t) {
+            test = Object.keys(t)[0] || undefined;
+        }
+        console.log("test", test);
+
+        // the refNode is wrong!
+        // const refNode = context.nodes.find((x) => {
+
+        //     if (x.name === test) return;
+        //     return x.interfaces.some((i) => i.name.value === entry[0]);
+        // }) as Node;
+        let refNode = context.nodes.find((x) => x.name === test || x.name === entry[0]) as Node;
+        if (!refNode) {
+            refNode = context.nodes.find((x) => x.interfaces.some((i) => i.name.value === entry[0])) as Node;
+        }
 
         const relationField = connectionField.relationship;
+        // console.log("relationField", relationField);
+        console.log("refNode", refNode.name);
 
         const childNode = new Cypher.Node({ labels: refNode.getLabels(context) });
         const relationship = new Cypher.Relationship({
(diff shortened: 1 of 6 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2715.surge.sh

Liam-Doodson (commented):
I'm not sure if this actually fixes the underlying issue of #2709 as I understood it. For the following query:

```gql
query {
  movies(
    where: {
      distributionConnection_SOME: {
        node: { name: "" }
      }
    }
  ) {
    title
  }
}
```

The following cypher is still produced:

```
MATCH (this:`Film`)
WHERE size([(this1:`Dishney`)-[this0:DISTRIBUTED_BY]->(this) WHERE this1.name = $param0 | 1]) > 0
RETURN this { .title } AS this
```

To me that's not the expected behaviour because only the Dishney types are checked in the predicate, not other implementations of DistributionHouse.

tbwiss:
You're right @Liam-Doodson !
What is the expected cypher query in this case? Something like this?
```cypher
MATCH (this:`Film`)
WHERE size([(this1:`Dishney`|`Netflix`|`Prime`)-[this0:DISTRIBUTED_BY]->(this) WHERE this1.name = $param0 | 1]) > 0
RETURN this { .title } AS this
```

Liam-Doodson:
> You're right @Liam-Doodson ! What is the expected cypher query in this case? Something like this?
> 
> ```cypher
> MATCH (this:`Film`)
> WHERE size([(this1:`Dishney`|`Netflix`|`Prime`)-[this0:DISTRIBUTED_BY]->(this) WHERE this1.name = $param0 | 1]) > 0
> RETURN this { .title } AS this
> ```

Yeah that looks right to me!

packages/graphql/src/translate/where/property-operations/create-connection-operation.ts
```diff
@@ -56,13 +56,31 @@ export function createConnectionOperation({
...
-        const childNode = new Cypher.Node({ labels: refNode.getLabels(context) });
+        let labels = refNode.getLabels(context);
+        const hasOnlyNodeObjectFilter = entry[1]?.node && !nodeOnObj;
+        if (hasOnlyNodeObjectFilter) {
+            const nodesImplementingInterface = context.nodes.filter((x) =>
+                x.interfaces.some((i) => i.name.value === entry[0])
+            );
+            if (nodesImplementingInterface.length) {
+                // remove the labels so we check all possibilities and not just the one label of the refNode
+                labels = [];
```
Liam-Doodson, line 80: I'm not sure this is right. Do we not need to have an OR of all the types of the interface as you suggested?

e.g.

```
MATCH (this:`Film`)
WHERE size([(this1:`Dishney`|`Netflix`|`Prime`)-[this0:DISTRIBUTED_BY]->(this) WHERE this1.name = $param0 | 1]) > 0
RETURN this { .title } AS this
```

It probably is a bit of an edge case but it would prevent matching nodes of a different type that share the same relationship. For example, with this type def it could end up also matching a `Publisher` if there are no labels as they also use the `DISTRIBUTED_BY` relationship type:

```gql
type Movie implements Production @node(label: "Film") {
    title: String!
    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
    runtime: Int!
    distribution: [DistributionHouse!]! @relationship(type: "DISTRIBUTED_BY", direction: IN)
    publisher: Publisher! @relationship(type: "DISTRIBUTED_BY", direction: IN)
}
```

tbwiss: It's good that you know or find these edge cases!
This approach with using an OR expression, `(this1:Dishney|Netflix|Prime)`, is sadly not Neo4j 4.x safe but there is another way to solve it. It's however a bit more verbose.

Liam-Doodson: Sorry to keep adding comments to this! Is it possible to add an integration test for this case as it seems like something that call be missed by changes in the future?

Liam-Doodson (approved):
I think it could be worth adding an integration test for the edge case that I mentioned to make sure we don't get any regressions in the future.

Other than that this looks good to go!

Liam-Doodson (approved):
LGTM!

PR #2716, opened 2023-01-11
boyscouting: disable CodeCov upload for dev nightly build

tbwiss:
Disable upload to CodeCov for nightly builds of the `dev` branch due to an increasing amount of `Too many uploads to the commit` error messages of `dev` nightly builds. It's disabled only for the scheduled nightly builds.


mjfwebb (dismissed):
Let's give this a try!

PR #2722, opened 2023-01-12
RFC - use selection set string for customResolvers requires

Liam-Doodson:
# Description

RFC for changing `@customResolvers` require from an array of fields to a selection set to allow for requiring related fields

## Complexity

Complexity: Medium


darrellwarde (approved):
Let's merge this as it's essentially implemented at this point!

Issue #2241, opened 2022-10-17
Enable constraints/filter conditions used to calculate Aggregate filters


imkleats:
**Is your feature request related to a problem? Please describe.**
Address disparity between ability to apply constraints (i.e. filter conditions) on `*Aggregate` fields but not being able to apply similar constraints/filter conditions *within* `*Aggregate` filters on the parent type.

Example: Given there are Groups that have Posts authored by Users, and some Users have a special characteristic of being an Administrator, suppose that you want to apply a filter that will return all Groups where there are fewer than 10 posts (or any other arbitrary number) written by Users who are Administrators.

##### Current Limitations

- We can query by the number of Posts in a Group.

```graphql
query {
  groups(where: {
    postsAggregate: {
      count_LT: 10
    }
  }) {
    ...groupSelectionSet
  }
}
```

- We can query for the Aggregate count of Posts that meet a certain condition in a Group:

```graphql
query {
  groups {
    ...groupSelectionSet
    adminPostAggregate: postsAggregate( where: {
        author: {
          role: "Administrator"
        }
    }) {
       count
    }
  }
}
```
- But we can't apply the conditions of the second aggregate to the filter being applied to the first aggregate.

##### Desired Experience

- We can apply constraints used in calculation of the Aggregate value used for our filtering:

```graphql
query {
  groups(where: {
    postsAggregate: {
      count_LT: 10
      where: {
        author: {
          role: "Administrator"
        }
      }
    }
  }) {
    ...groupSelectionSet
  }
}
```

- A more common use-case might be any time you're trying to use an Aggregate filter in the context of related nodes that have a logical delete instead of a hard delete.

**Describe the solution you'd like**
- Add a `where` input field to the generated `<Type 1><Type 2>AggregateInput` type that is of type `<Type 2>Where` to add constraints to the aggregate calculation that `<Type 1>` is being filtered against.
  - This would be consistent with the API established in the existing `<Type 1><Type 2>AggregationSelection`'s `where` argument, which is also of type `<Type 2>Where`.

**Describe alternatives you've considered**
We're using middleware to apply the constraints ourselves in an intermediate step that reduces an extended `<Type1><Type2>Aggregate` input (with the `where` argument added) into a list of IDs that can replace the original Aggregate filter field with a `<field>: { id_IN: [<list of IDs>] }` condition, which adds a round-trip to the database and isn't scalable when the number of matching IDs is large.

**Additional context**
Happy to address any additional necessary clarification or context in replies

[...]

PR #2723, opened 2023-01-12
RFC: Introduce Aggregation Query Filters

MacondoExpress:
# RFC Aggregation Query Filters

This RFC, introduces filters in the aggregation query generated.

## Complexity

Complexity: Low/Med

Closes: 
- #2241

### DIFF
--- a/docs/rfcs/rfc-034-aggregation-query-filters.md
+++ b/docs/rfcs/rfc-034-aggregation-query-filters.md
@@ -246,7 +246,8 @@ RETURN this { .name } AS this
 ## Risks
 
 - Schema size increase.
-- Even if there is a reason for this filters, the usage of these filters could be confusing.
+- Even if there is a reason for this filters, having two type of filters in aggregation could be confusing.
+- This is a breaking change, potentially could be revisited with the new knowledge obtained by the API uniformity work to avoid unnecessary breaking changes.
 
 ### Security consideration
 
--- a/docs/rfcs/rfc-034-aggregation-query-filters.md
+++ b/docs/rfcs/rfc-034-aggregation-query-filters.md
@@ -217,7 +217,9 @@ query {
                 count_LT: 10
             },
             {
-                count_LT: 10
+                node: {
+                  title_LONGEST_GT: 10
+                }
             },
         ]
         
--- a/docs/rfcs/rfc-034-aggregation-query-filters.md
+++ b/docs/rfcs/rfc-034-aggregation-query-filters.md
@@ -1,0 +1,257 @@
+# Aggregation Query Filters
+
+## Problem
+
+Given the following type definition:
+
+```graphql
+type Group {
+    name: String!
+    posts: [Post!]! @relationship(type: "HAS_POST", direction: OUT)
+}
+
+type Post {
+    title: String!
+    released: Int
+    author: Person @relationship(type: "WROTE_POST", direction: IN)
+}
+
+type Person {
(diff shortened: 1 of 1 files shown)


docs/rfcs/rfc-034-aggregation-query-filters.md
```diff
@@ -0,0 +1,257 @@
...
+  released_NOT: Int
+  released_IN: [Int]
+  released_NOT_IN: [Int]
+  released_LT: Int
+  released_LTE: Int
+  released_GT: Int
+  released_GTE: Int
+}
+
+input GroupPostsAggregateFilter {
```
MacondoExpress, line 101: In case we choose to implement this, it will be good to find a better name rather than `queryWhere` and `aggregationWhere`!

docs/rfcs/rfc-034-aggregation-query-filters.md
```diff
@@ -0,0 +1,258 @@
...
+        ]
+      }
+      aggregationWhere: {
+        OR: 
+        [
+            {
+                count_LT: 10
+            },
+            {
+                count_LT: 10
```
Liam-Doodson, line 220: Is one of these conditions a typo. They don't seem to line up with the title/cypher so I'm not too sure what this example is trying to show?

docs/rfcs/rfc-034-aggregation-query-filters.md
Liam-Doodson: Yeah this was my main thought reading this.

I was just looking at how the problem has been solved elsewhere and it looks like mongo uses `$match` for the equivalent of `queryWhere` and `$group` for `aggregationWhere` which I think makes the distinction a bit clearer.

If I was to apply similar thinking to our situation I'd change `queryWhere` to `aggregateWhere` and  `aggregationWhere` to `group`.

What do you think?

Liam-Doodson (changes requested):
Sounds like a good addition to me! I just think it's worth clarifying the example and having another think about the input names as you suggested

docs/rfcs/rfc-034-aggregation-query-filters.md
MacondoExpress: Good catch!

docs/rfcs/rfc-034-aggregation-query-filters.md
MacondoExpress: I think that in mongo you filter the Result set with `$match` and then you have the aggregation operation with `$group`, in our case what actually is not an aggregation, but a filter based on aggregation.
I'm happy with something like `aggregateWhere` and `groupWhere`, but, in my opinion, I believe that is better to wait for the API uniformity piece before taking a decision on this!

Liam-Doodson: I do think the naming of this is going to be tricky to get right because it's a weird concept to have a filter on a filter. To me having `where` in both the names makes it very difficult to distinguish between the behaviours.

I know the `group` input would be used as a filter but I don't think that means it needs to have `where` in the name in the same way we don't call other filters `titleWhere`, `releasedWhere`, `postsAggregateWhere` etc.

I'm happy to wait for the API uniformity work if you think it'll make the naming easier as you know more about what is happening there!

darrellwarde:
It's highly likely that this will be implemented after the proposed major API changes, and this RFC will need completely rewriting following this.

PR #2730, opened 2023-01-13
RFC: Generic Operators

MacondoExpress:
# RFC Generic Operators

Describes a more reusable API design for operators.

## Complexity

Complexity: Low


docs/rfcs/rfc-034-generic-operators.md
```diff
@@ -0,0 +1,332 @@
...
+**Posts with more than 10 likes**
+
+```graphql
+query Posts {
+    posts(where: { likesAggregate: { count_GT: 10 } }) {
+        content
+    }
+}
+```
+
```
Liam-Doodson, line 178: Is it worth putting the problem the current type defs that you are trying to solve?

I assume it's to reduce the size of our type defs & to have inputs that are more similar to other tools/APIs so it is easier for new users to get to grips with the library but I think it'd be good to have it spelled out

Liam-Doodson (dismissed):
I really like this suggestion! One of the first thoughts I had when starting was that it'd be nicer if our inputs where like this!

(I've approved but as this is quite a big change I think it'd be good if other people took a look before this is merged)

docs/rfcs/rfc-034-generic-operators.md
MacondoExpress: I think it's more the improve the readability and to be fair is because is nicer, but from a pragmatic point of view, unfortunately until GraphQL does not support Generics, it will result in a larger number of types!
I still believe that consistency and readability should have priority on this!

MacondoExpress:
Thank you @Liam-Doodson! I hope others agree with us!

docs/rfcs/rfc-034-generic-operators.md
```diff
@@ -0,0 +1,332 @@
...
+    NOT: MoviesLikesWhere
+    aggregation: PostLikesAggregateInput
+    all: PostLikesConnectionWhere
+    none: PostLikesConnectionWhere
+    single: PostLikesConnectionWhere
+    some: PostLikesConnectionWhere
+    # For simplicity the standard filters are substituted by the connections filters as discussed in the API uniformity piece.
+}
+
+input IntArrayWhere {
```
darrellwarde, line 215: I would personally use "List" instead of "Array" - this terminology used in the GraphQL spec.

```suggestion
input IntListWhere {
```

docs/rfcs/rfc-034-generic-operators.md
```diff
@@ -0,0 +1,332 @@
...
+    OR: [IntArrayWhere!]
+    AND: [IntArrayWhere!]
+    NOT: IntArrayWhere
+    equal: [Int!]
+    includes: Int
```
darrellwarde, line 220: This possibly feels like the correct RFC to begin including #2740, might we want:

```suggestion
    all: [Int!]
    some: Int
```

docs/rfcs/rfc-034-generic-operators.md
```diff
@@ -0,0 +1,332 @@
...
+    single: PostLikesConnectionWhere
+    some: PostLikesConnectionWhere
+    # For simplicity the standard filters are substituted by the connections filters as discussed in the API uniformity piece.
+}
+
+input IntArrayWhere {
+    OR: [IntArrayWhere!]
+    AND: [IntArrayWhere!]
+    NOT: IntArrayWhere
+    equal: [Int!]
```
darrellwarde, line 219: I believe `equal` should always be `equals` - there are a few instances of this throughout your example. All of the operators end with an "s".

```suggestion
    equals: [Int!]
```

darrellwarde (commented):
You have confused `Post` and `Movie` throughout your example, best go and fix that! 🙃 

My biggest concern as I have expressed elsewhere, is that this will block any future endeavours of having filters configurable on a per-field basis. However, I think the benefits far outweigh the drawbacks here, and we could still enable configuration on a per-type basis.

MacondoExpress:
I agreed with you @darrellwarde , although, to be fair if I have to be honest, I'm not even a fan of having the granularity defined at the field level.
Please @Liam-Doodson , and @darrellwarde , take a look at the new section `Technical challenge`, personally, I'm not against having `set`/`assign`/`to`/`update` as an operator in the update input, but I have the gut feeling that could be controversial!

Liam-Doodson:
> I agreed with you @darrellwarde , although, to be fair if I have to be honest, I'm not even a fan of having the granularity defined at the field level. Please @Liam-Doodson , and @darrellwarde , take a look at the new section `Technical challenge`, personally, I'm not against having `set`/`assign`/`to`/`update` as an operator in the update input, but I have the gut feeling that could be controversial!

I think I'd just have `set` on the `FloatUpdate` type. It keeps it more consistent with the rest of the API which I think is more important here

docs/rfcs/rfc-034-generic-operators.md
```diff
@@ -0,0 +1,506 @@
...
+    some: PostLikesConnectionWhere
+    # For simplicity the standard filters are substituted by the connections filters as discussed in the API uniformity piece.
+}
+
+input IntListWhere {
+    OR: [IntListWhere!]
+    AND: [IntListWhere!]
+    NOT: IntListWhere
+    all: [Int!]
+    some: Int
```
darrellwarde, line 220: Let's change these back - I definitely don't think these make sense.

`all` is just `equals` and `some` just doesn't make sense at all here. I would have the following operations:

- `equals`
- `includes`
- `intersects` (future, let's leave this out for now)

darrellwarde (changes requested):
Before we merge this, let's fix the list filters - I don't like the changes we proposed.

PR #2761, opened 2023-01-19
Granular configuration of startup validation for type definitions and resolvers

### LINKED ISSUE #2394: @customResolvers directive breaks integration with NestJS and other GraphQL Frameworks
**Describe the regression**

The `@customResolvers` directive which replaces the previous `@computed` directive critically breaks our usage of this library in conjunction with `NestJS` GraphQL. 

Previously, if no custom resolver was provided to the Neo4jGraphQL (or OGM) constructor, the schema generation would not fail. This is important to us because we first generate the schema via `@neo4j/graphql` before passing the generated schema to NestJS. 

Then custom resolvers for mutations, queries, and field-level resolvers are implemented in the NestJS-style of Dependency-Injection containers. This is mission critical for a complex service like ours. 

Previously, we built out a `@neo4j/graphql`-only service, but this presented limitations in complexity due to the way that resolvers must be passed at schema-construction time. Very happy with our workflow of `neo4j -> generated schema -> nestjs + custom resolvers implementation` 

**Last working version**
Still works, but will be broken in 4.0 as the `@computed` directive is deprecated

**To Reproduce**

1. Define custom field-level resolvers in the schema-definition `.graphql` files
2. Do *not* pass the resolver at Neo4jGraphQL construction time
3. Schema generation fails

**Expected behavior**

**Requesting a `config.x` flag that can be passed in the `Neo4jGraphQLConstructor` to disable the enforcement of the presenec of custom resolvers**. 

This would enable this library to operate alongside NestJS and other graphql frameworks.

Liam-Doodson:
# Description

Adds a new config option `startupValidation` that allows users to turn off validation checks run against either `typeDefs` or `customResolvers` or both.

The existing `skipValidateTypeDefs` config option has been deprecated.

# Examples

## Creating a schema that will avoid all checks:

```js
    typeDefs,
    driver,
    config: {
        startupValidation: false,
    },
});

const generatedSchema = neoSchema.getSchema();
```

## Creating a schema that just performs typeDef validation:

```js
    typeDefs,
    driver,
    config: {
        startupValidation: {
            resolvers: false,
        },
    },
});

const generatedSchema = neoSchema.getSchema();
```

OR

```js
    typeDefs,
    driver,
    config: {
        startupValidation: {
            typeDefs: true,
            resolvers: false,
        },
    },
});

const generatedSchema = neoSchema.getSchema();
```

## Complexity

Complexity: Low

# Issue

Closes #2394


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2761.surge.sh

packages/graphql/src/types.ts
```diff
@@ -357,6 +357,18 @@ export interface CypherQueryOptions {
     replan?: CypherReplanning;
 }
 
+/** The startup validation checks to run */
+export interface StartupValidationOptions {
+    typeDefs?: boolean;
+    customResolver?: boolean;
```
darrellwarde, line 363: I think this should be `customResolvers` (plural), or maybe even just `resolvers`! The latter is attractive because it mirrors the fields which our users pass into the constructor (`typeDefs` and `resolvers`).

darrellwarde (changes requested):
I think we need a little tweak to this API - we can discuss the best choice!

packages/graphql/src/types.ts
Liam-Doodson: I think `resolvers` is the best shout because as you say it mirrors the constructor. I'll update it to that if you're happy!

darrellwarde: Let's do it!

darrellwarde (approved):
LGTM!

PR #2776, opened 2023-01-24
fix: remove playwright workaround

tbwiss:
Remove the workaround for the "playwright install connectivity issue" added in https://github.com/neo4j/graphql/pull/2729


Liam-Doodson (approved):
LGMT! Thanks for looking at this 😄

PR #2785, opened 2023-01-26
Update aggregation operations to properly check durations

### LINKED ISSUE #2697: Aggregation: wrong cypher to compare durations
Direct comparison of duration is not supported and so, the cypher queries generated by these tests are wrong:

* https://github.com/neo4j/graphql/blob/0d571437d2bf24790fa718ffdd81071d20cf49c5/packages/graphql/tests/tck/aggregations/where/edge/duration.test.ts
* https://github.com/neo4j/graphql/blob/0d571437d2bf24790fa718ffdd81071d20cf49c5/packages/graphql/tests/tck/aggregations/where/node/duration.test.ts

Example:

```cypher
MATCH (this:`Post`)
            CALL {
                WITH this
                MATCH (this1:`User`)-[this0:LIKES]->(this:`Post`)
                RETURN any(var2 IN collect(this0.someDuration) WHERE var2 >= $param0) AS var3
            }
            WITH *
            WHERE var3 = true
            RETURN this { .content } AS this
```

should be

```cypher
MATCH (this:`Post`)
            CALL {
                WITH this
                MATCH (this1:`User`)-[this0:LIKES]->(this:`Post`)
                RETURN any(var2 IN collect(this0.someDuration) WHERE datetime() + var2 >= datetime() + $param0) AS var3
            }
            WITH *
            WHERE var3 = true
            RETURN this { .content } AS this
```

Background:

> The behavior you describe is intended and documented:
> https://neo4j.com/docs/cypher-manual/current/syntax/operators/#cypher-ordering
> 
> > [Duration](https://neo4j.com/docs/cypher-manual/current/syntax/temporal/#cypher-temporal-durations) values cannot be compared, since the length of a day, month or year is not known without knowing which day, month or year it is. Since Duration values are not comparable, the result of applying a comparison operator between two Duration values is null.
> 
> If you want to compare duration values, you first need to add them to a common date/datetime like this:
> 
> ```
> RETURN date() + duration("P1D") <=  date() + duration("P1Y")
> ```
> 
> Alternatively, if you know that your durations only use the same [component groups](https://neo4j.com/docs/cypher-manual/current/syntax/temporal/#cypher-temporal-accessing-components-durations), you can compare those:
> 
> ```
> RETURN duration("P1D").days <=  duration("P365D").days
> ```
 _Originally posted by @sherfert in https://github.com/neo4j/neo4j/issues/13019#issuecomment-1375340256_

angrykoala:
# Description

Fixes the cypher generating comparison of duration fields in aggregation

## Complexity

Complexity: Low

# Issue

Closes #2697

### DIFF
--- a/packages/graphql/src/translate/create-aggregate-where-and-params.ts
+++ b/packages/graphql/src/translate/create-aggregate-where-and-params.ts
@@ -21,7 +21,7 @@ import Cypher from "@neo4j/cypher-builder";
 import type { Node, Relationship } from "../classes";
 import type { RelationField, Context, GraphQLWhereArg, PredicateReturn } from "../types";
 import { aggregationFieldRegEx, AggregationFieldRegexGroups, ListPredicate, whereRegEx } from "./where/utils";
-import { createBaseOperation } from "./where/property-operations/create-comparison-operation";
+import { createBaseOperation, createDurationOperation } from "./where/property-operations/create-comparison-operation";
 import { NODE_OR_EDGE_KEYS, LOGICAL_OPERATORS, AGGREGATION_AGGREGATE_COUNT_OPERATORS } from "../constants";
 import mapToDbProperty from "../utils/map-to-db-property";
 
@@ -123,6 +123,7 @@ export function aggregateWhere(
             const target = key === "edge" ? cypherRelation : aggregationTarget;
             const refNodeOrRelation = key === "edge" ? relationship : refNode;
             if (!refNodeOrRelation) throw new Error(`Edge filter ${key} on undefined relationship`);
+
             const {
                 returnProjections: innerReturnProjections,
                 predicates: innerPredicates,
@@ -253,12 +254,21 @@ function createEntityOperation(
         });
     } else {
         const innerVar = new Cypher.Variable();
-        const innerOperation = createBaseOperation({
-            operator: logicalOperator || "EQ",
-            property: innerVar,
-            param: paramName,
-        });
 
+        let innerOperation: Cypher.Operation;
+        if (fieldType === "Duration") {
+            innerOperation = createDurationOperation({
+                operator: logicalOperator || "EQ",
+                property: innerVar,
+                param: paramName,
+            });
+        } else {
+            innerOperation = createBaseOperation({
+                operator: logicalOperator || "EQ",
+                property: innerVar,
+                param: paramName,
+            });
+        }
         const dbFieldName = mapToDbProperty(refNodeOrRelation, fieldName);
         const collectedProperty =
             fieldType === "String" && logicalOperator !== "EQUAL"
--- a/packages/graphql/tests/integration/aggregations/where/edge/duration.int.test.ts
+++ b/packages/graphql/tests/integration/aggregations/where/edge/duration.int.test.ts
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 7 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2785.surge.sh

packages/graphql/src/translate/create-aggregate-where-and-params.ts
```diff
@@ -253,12 +254,21 @@ function createEntityOperation(
...
         const innerVar = new Cypher.Variable();
-        const innerOperation = createBaseOperation({
-            operator: logicalOperator || "EQ",
-            property: innerVar,
-            param: paramName,
-        });
 
+        let innerOperation: Cypher.Operation;
+        if (fieldType === "Duration") {
+            innerOperation = createDurationOperation({
```
Liam-Doodson, line 260: I've got a feeling the mistake we made here was using `createBaseOperation` instead of `createComparisonOperation`.

That also seems to do something different for `Point` fields which I'm guessing is something else that doesn't work as it should with aggregations.

It looks like it might be a bit annoying to pass the `context.neo4jDatabaseInfo` here but I think it's worth it to keep a single source of truth for handling these operations

packages/graphql/tests/integration/aggregations/where/edge/duration.int.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    let neoSchema: Neo4jGraphQL;
+
+    beforeAll(async () => {
+        neo4j = new Neo4j();
+        driver = await neo4j.getDriver();
+    });
+
+    beforeEach(async () => {
+        const typeDefs = `
+        type User {
```
Liam-Doodson, line 44: It looks like you're not using the unique types here and this could also do with another layer of indentation

Liam-Doodson (changes requested):
This fix looks good for fixing the `Duration` bug!. However, looking at the changes you've made I've got a feeling there is a similar bug for `Point` fields which we also missed and both could be fixed by using existing functionality.

I think it's probably worth checking that out before merging this

packages/graphql/src/translate/create-aggregate-where-and-params.ts
angrykoala: I agree that this is probably the root cause of this problem, however it is not a quick and simple refactor I think, as not only the context but the format of the provided fields need to be different

I agree with the `Point` and I was planning on addressing that in a separate PR

angrykoala: Oh, we need to fix it properly, but this is not going to increase our technical debt towards that so no sense in having that broken while we fix it properly.

The points on separate PR is simply to avoid large PRs, as point tests need to be added to guarantee it is a problem that need fixing

Liam-Doodson (approved):
LGTM!

packages/graphql/tests/tck/aggregations/where/edge/duration.test.ts
```diff
@@ -69,7 +69,7 @@ describe("Cypher Aggregations where edge with Duration", () => {
             CALL {
                 WITH this
                 MATCH (this1:\`User\`)-[this0:LIKES]->(this)
-                RETURN any(var2 IN collect(this0.someDuration) WHERE var2 = $param0) AS var3
+                RETURN any(var2 IN collect(this0.someDuration) WHERE datetime() + var2 = datetime() + $param0) AS var3
```
darrellwarde, line 72: Is there a risk that the two separate calls of `datetime()` yield slightly different results, which could compromise the comparison? Is it worth having something like the below instead:

```suggestion
                WITH *, datetime() AS base
                RETURN any(var2 IN collect(this0.someDuration) WHERE base + var2 = base + $param0) AS var3
```

darrellwarde (approved):
LGTM, just a small observation!

packages/graphql/tests/tck/aggregations/where/edge/duration.test.ts
angrykoala: There shouldn't be a risk here, the whole operation runs in a transaction and datetime uses the same value for the whole transaction.

As a test of that, the following cypher:

```cypher
WITH datetime() as d1
CALL apoc.util.sleep(1000)
RETURN d1, datetime()
```

RETURN the exact same time for d1 and the last datetime

PR #2790, opened 2023-01-26
Docs on @cypher columnName

angrykoala:
# Description
Updated documentation for the `@cypher` directive in 4.0.0

implementation of these changes in #2769


docs/modules/ROOT/pages/type-definitions/cypher.adoc
```diff
@@ -3,6 +3,8 @@
 
 The `@cypher` directive binds a GraphQL field to the result(s) of a Cypher query.
 
+This directive can be use both for properties in a type or as top level queries:
```
Liam-Doodson, line 6:
```suggestion
This directive can be used both for properties in a type or as top level queries:
```

Liam-Doodson (changes requested):
These changes look good! I think the migration guide also needs an update for v4 because it makes `columnName` sound optional which will not be the case when this is released

PR #2862, opened 2023-02-10
XOR operation

angrykoala:
# Description

Adds `Cypher.xor` boolean operation

```typescript
const predicate1 = Cypher.coalesce(new Cypher.Variable());
const predicate2 = Cypher.max(new Cypher.Variable());

const xor = Cypher.xor(predicate1, predicate2);
```

Would result on:
```cypher
(coalesce(var0) XOR max(var1))
```


darrellwarde (approved):
Beautiful, amazing how easy that was!

PR #2865, opened 2023-02-10
Add `.vscode/settings.json` to ensure uniform formatting

darrellwarde:
# Description

> **Note**
> 

We still have a problem of PRs with completely wrong formatting.

As I was setting up my machine, I started completely from scratch, and have come up with the barest project configuration to ensure that formatting happens as it should.

This depends on the `rvest.vs-code-prettier-eslint` extension, which I have add as a recommended extension. It takes into account both ESLint and Prettier rules, unlike when using those two extensions separately. The ESLint can still be used for syntax highlighting linting errors.

This can of course be overridden by user settings.

## Complexity

> **Note**
>

Complexity: Low

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ tmp/
 .vscode/*
 !.vscode/extensions.json
 !.vscode/graphql.code-snippets
+!.vscode/settings.json
 
 # OSX
 .DS_Store
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,12 +2,7 @@
     // See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
     // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
     // List of extensions which should be recommended for users of this workspace.
-    "recommendations": [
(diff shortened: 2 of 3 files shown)


angrykoala:
I'm not sure about this, the new extension seems to be waaay less famous than the previous one. Also, I'm not sure what our problems on formatting are, but in theory prettier+eslint should work if correctly configured no?

darrellwarde:
> I'm not sure about this, the new extension seems to be waaay less famous than the previous one. Also, I'm not sure what our problems on formatting are, but in theory prettier+eslint should work if correctly configured no?

The issue is, the ESLint extension in VSCode doesn't deal with formatting (as in, it doesn't apply Prettier formatting), so the fact that it is configured with Prettier rules does not mean that the Prettier rules get applied. The Prettier extension just ignores ESLint configuration.

This extension takes both configurations into consideration when applying formatting. I was actually looking more at ratings over downloads, the Prettier extension only has 3.5 stars whilst this one has around 4.5 stars. Trying it out, it's rock solid and it performs well.

Not naming names, but we have at least 2 people on the team who raise PRs with incorrect formatting!! 😆 This basic configuration will ensure that at least basic formatting is applied.

PR #2943, opened 2023-02-28
Add comma-dangle to .eslintrc

darrellwarde:
# Description

> **Note**
> 

The current combined Prettier+ESLint combination does not add trailing commas for multi-line statements as Prettier would like. This adds the comma-dangle rule to our .eslintrc which will fix this behaviour.

## Complexity

> **Note**
>

Complexity: Low

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
         es2021: true,
     },
     rules: {
-        "comma-dangle": ["error", "only-multiline"],
+        "comma-dangle": ["error", "always-multiline"],
         "eslint-comments/no-unused-disable": "error",
         // Expensive rules disabled below
         "import/default": "off",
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
         es2021: true,
     },
     rules: {
+        "comma-dangle": ["error", "only-multiline"],
```
Liam-Doodson, line 9: Should this not be `always-multiline`?

> "always-multiline" requires trailing commas when the last element or property is in a different line than the closing ] or } and disallows trailing commas when the last element or property is on the same line as the closing ] or }

darrellwarde: With reflection, this is probably the right move. I've changed to `always-multiline` - let's merge, and then I'll raise a separate PR fixing all of the issues on both `dev` and `4.0.0` (I think the merge from `dev` to `4.0.0` will be messy)? Seems to be a slightly bigger problem than we think, there is a lot of disparity.

Liam-Doodson (approved):
LGTM!

PR #2956, opened 2023-03-01
Surface LICENSE information for npm/github & other tools

cliffordfajardo:
# Description
Currently this repo does not surface its LICENSE in manner that make the information surfaceable to tools like: Github, NPM etc

Example 1: This repo on the left side, a repo with license on the right side
CleanShot 2023-03-01 at 14 42 05@2x

## What are the changes?
- change `addlicense` to `LICENSE`, which is a[ common license file naming scheme](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository#including-an-open-source-license-in-your-repository)
- add `license` field to package.json, following `package.json` [doc recommendations](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#license)

## Motivation for changes
I wanted to use this library at the company I work for but was unable to since our custom internal tool which checks for a license couldn't find it in the `package.json`  via a `license` key or a file called `LICENSE` in the repo


tbwiss:
Thank you very much for pointing this out @cliffordfajardo!
As you may have noticed we have license information for each package in this monorepo but missed adding the license info to the root of the project.
This will be addressed in https://github.com/neo4j/graphql/pull/2968 instead therefore I'll close this PR.

PR #2970, opened 2023-03-03
RFC: Authorization

darrellwarde:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=Rr8NmsQ)

# Description

> **Note**
> 

This PR proposes a simplified RFC covering _only_ the new authorization functionality.

It switches to using the filtering input types we generate at the moment.

Additionally, with the philosophy that RFCs are meant to be somewhat living documents, this excludes some contentious topics such as authorization for Subscriptions, in order to allow for concise reviews and to unblock development.

## Complexity

> **Note**
>

Complexity: Medium

## Issues

- #2749


docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+
+The following is a type proposal for auth configuration:
+
+```ts
+type JWKS = {
+  uri: string;
+}
+
+type AuthorizationConfig = {
+  secret: string | JWKS
```
Liam-Doodson, line 15: Would this not lose functionality as the current type for secret looks like this for JWKS:

```ts
jwksEndpoint: string | ((req: RequestLike) => string); // Probably the equivalent of the `uri` here
```

And it looks like this for jwts:

```ts
secret: jsonwebtoken.Secret | ((req: RequestLike) => jsonwebtoken.Secret);
```

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+
+enum AuthorizationValidateAfter {
+  CREATE
+  UPDATE
+  DELETE
+  CREATE_RELATIONSHIP
+  DELETE_RELATIONSHIP
+}
+
+input UserAuthorizationFilterRule {
```
Liam-Doodson, line 331:
```suggestion
input UserAuthorizationValidationRule {
```

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+  NOT: UserAuthorizationWhere
+  jwtPayload: JWTPayloadWhere
+  node: UserWhere
+}
+
+enum AuthorizationFilterOperation {
+  READ
+  UPDATE
+  DELETE
+  CREATE_RELATIONSHIP
```
Liam-Doodson, line 304: For the same reason we're leaning towards `CONNECT`/`DISCONNECT` for the condition schema generation, should it also not be `CONNECT`/`DISCONNECT` here? This is what users currently understand our operations as. 

I do agree we should move in this direction in the future but I don't know if we're there yet

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+
+enum AuthorizationValidateAfter {
+  CREATE
+  UPDATE
+  DELETE
+  CREATE_RELATIONSHIP
+  DELETE_RELATIONSHIP
+}
+
+input PostAuthorizationFilterRule {
```
Liam-Doodson, line 386:
```suggestion
input PostAuthorizationValidationRule {
```

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+  before: [AuthorizationValidateBefore!]! = [READ, CREATE, UPDATE, DELETE, CREATE_RELATIONSHIP, DELETE_RELATIONSHIP]
+  after: [AuthorizationValidateAfter!]! = [CREATE, UPDATE, DELETE, CREATE_RELATIONSHIP, DELETE_RELATIONSHIP]
+  requireAuthentication: Boolean! = true
+  where: PostAuthorizationWhere!
+}
+
+directive @authorization(
+  filter: [PostAuthorizationFilterRule!]
+  validate: [PostAuthorizationValidateRule!]!
+) on OBJECT | FIELD_DEFINITION | INTERFACE
```
Liam-Doodson, line 396: Is there anyway we could provide this or a similar directive for `SCHEMA` level rules? Obviously, we'd only be able to check the jwt at that level but I know this is something customers have asked for

Liam-Doodson: Even if this is out of scope here, it's probably worth noting this in an out of scope section

docs/rfcs/rfc-035-authorization.md
darrellwarde: I was actually about to rewrite the read/write controls RFC with `CREATE_RELATIONSHIP` and `DELETE_RELATIONSHIP`, as this will align the terminology across that, authorization, and Subscriptions, and future-proof us so we don't have to make more breaking changes in the future.

Liam-Doodson: As long as we're going to be aligned in multiple places then I think this is fine then

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+    name
+  }
+}
+```
+
+This will generate Cypher and parameters along the lines of the following:
+
+```cypher
+MATCH (this:User)
+WHERE this.name = $name
```
MacondoExpress, line 693: I believe this syntax is dangerous, you will be able to obtain all the DB information by playing with the API.
In this example, `Unauthorized` is raised only if `this.name` is different from "Bob". This means that I can use the "Unauthorized" error as a source of truth.
If I want to know if Bob is a friend of Darrell I can write:
```graphql
{
  users(where: { name: "Bob",  friends_SOME: {
        name: "Darrell"
    }  }) {
    id
    name
  }
}
```
 And again as I know that Bob exists if not I should have received an error before, I can continue leaking information.

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+  node: ${typename}Where
+}
+
+input ${typename}AuthorizationFilterRule {
+  operations: [AuthorizationFilterOperation!]! = [READ, UPDATE, DELETE, CREATE_RELATIONSHIP, DELETE_RELATIONSHIP]
+  requireAuthentication: Boolean! = true
+  where: ${typename}AuthorizationWhere!
+}
+
+input ${typename}AuthorizationFilterRule {
```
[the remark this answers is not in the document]
darrellwarde:
```suggestion
input ${typename}AuthorizationValidateRule {
```

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+  node: PostWhere
+}
+
+input PostAuthorizationFilterRule {
+  operations: [AuthorizationFilterOperation!]! = [READ, UPDATE, DELETE, CREATE_RELATIONSHIP, DELETE_RELATIONSHIP]
+  requireAuthentication: Boolean! = true
+  where: PostAuthorizationWhere!
+}
+
+input PostAuthorizationFilterRule {
```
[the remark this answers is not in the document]
darrellwarde:
```suggestion
input PostAuthorizationValidateRule {
```

docs/rfcs/rfc-035-authorization.md
darrellwarde: I'm not sure I follow... This is taking the user's filtering arguments, and once they have been applied, we perform the authorization check. This is exactly as it performs at the moment, literally identical to `allow`.

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+
+The generated types for the directive will depends on `jwtPayload` when instantiating the feature. Given the following:
+
+```ts
+new Neo4jGraphQL({
+  typeDefs,
+  features: {
+    authorization: {
+      secret: "secret",
+      jwtPayload: {
```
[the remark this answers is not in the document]
darrellwarde: Liam also mentioned this on the last RFC... I don't think I like enforcing users into using the type name `JWTPayload`, however, I would be open to introducing a labelling directive:

```gql
type SomeOtherType @jwtPayload {
  sub: String!
  roles: [String!]!
}
```

Thoughts?

docs/rfcs/rfc-035-authorization.md
darrellwarde: My intention is to explore GraphQL Shield for this kind of thing... 👀

docs/rfcs/rfc-035-authorization.md
a-alle: I think the idea is that you have a jwt with sub "A". You try to get user Bob:
`MATCH (u:User) WHERE u.name="Bob" AND ..apoc.validate.. `
Your jwt sub is checked against Bob node's ID and they will obviously not match. Therefore, validation error is thrown, indicating that Bob does indeed exist (bc otherwise apoc would not have thrown an error). 
Therefore whatever your condition is before `apoc.validate`, you will know it's true if you get an error. I think @MacondoExpress is trying to point out that if you work it out step by step you can infer a lot of information.

a-alle: However, if I remember correctly we have already discussed this is part of existing functionality so we should support it, but **strongly** suggest to use the `filter` argument instead to avoid this kind of data leaking. I might be wrong however so please correct me.

docs/rfcs/rfc-035-authorization.md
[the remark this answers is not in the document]
a-alle: yees!

docs/rfcs/rfc-035-authorization.md
darrellwarde: That's exactly it, we're only doing as the user has instructed, and why `filter` is going to become such a pivotal rule!! It all comes down to good documentation communicating best practices.

docs/rfcs/rfc-035-authorization.md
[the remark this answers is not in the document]
darrellwarde: We would check for only one instance of `@jwtPayload` in the entire schema. I don't think something on schema would work, let's take advantage of the type safety of GraphQL!

docs/rfcs/rfc-035-authorization.md
MacondoExpress: The problem is that the `authorization check` happens only if the user's where predicate returns True if it returns False then I receive an empty response. This difference in the response enables a malicious user to leak information.
Let's make an example:
```graphql
type User @authorization(
	validate: [{ after: [], where: { node: { id: "$jwt.sub" } } }]
) {
	id: String!
	name: String!
        verySecretNumber: Int
}
```
Let's consider that you Darrell have an account on this social platform with an id = 20 and your very secret number is 105.
If I with the id = 35 try to query the `users` with my id, as follow: 
```graphql
{
    users(where: { id:  20 }) {
	id
	name
        verySecretNumber
   }
}
```
I receive an error "Unauthorized" coming from the `apoc.util.validatePredicate(NOT (this.id = $jwt.sub), "Unauthorized", [])`.
But then I notice that if I make this query
```graphql
{
    users(where: { AND: [{ id:  20 }, NOT: { id: 20 }] }) {
	id
	name
        verySecretNumber
   }
}
```
I just receive an empty response.
Now I have the plan to leak your very secret number.
What I know is that if this query:
```
{
    users(where: { { id:  20, verySecretNumber: 105 } }) {
	id
	name
        verySecretNumber
   }
}
```
Returns an "Unauthorized" error then it means that your secret number is 105, if instead returns an empty response it means that I have to keep searching for it. Of course, I don't know initially your secret number but using a binary search I can find it in a couple of iterations.

MacondoExpress: That sounds good to me it can be solvable by good documentation indeed!

docs/rfcs/rfc-035-authorization.md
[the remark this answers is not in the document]
MacondoExpress: I like it! potentially could be beneficial for some users to have not a single instance of `@jwtPayload`, if we apply the directive on an interface and we allow only one interface to be defined with the `@jwtPayload`, then it will be easier to support multiple `jwtPayload` implementations if requested, as we could reuse the interface filtering syntax.

docs/rfcs/rfc-035-authorization.md
Liam-Doodson: Yeah this is something I raised a few times in this `@auth` redesign. It's because we only perform the validation on data from the database. For now it should be fine as long as we document this behaviour well!

At some point in the future, I think we should consider some mechanism to throw these requests out even earlier (e.g. the user is called "Bob" but hasn't included a `where: { name: "Bob" }` filter, then throw a forbidden error before ever hitting the db).

However, this is probably out of scope for this work but it might be worth capturing in an out of scope section

docs/rfcs/rfc-035-authorization.md
```diff
@@ -0,0 +1,759 @@
...
+If, following the user's filter application, the data to be returned does not satisfy the rule, the error "Unauthorized" will be thrown.
+
+#### After operations
+
+Like disabling `after` operations, you will need to do the same with `before` to run only after an operation:
+
+```gql
+type User
+  @authorization(
+    validate: [{ before: [], where: { node: { id: "$jwt.sub" } } }]
```
angrykoala, line 716: I really don't like this approach of using empty `before` and `after` to define when an operation happens...
Not only seems like an ugly way of setting a flag, it is unintuitive as we are setting a before field to run the operations before.

Maybe I am missing something, but feels like this could use an enum or boolean instead :thinking:

darrellwarde: Yeah this is a sticking point, the issue is that the operations are different for `before` and `after`, so it's hard to unify them under a single boolean or something similar. I think this is an improvement to two completely different objects (`pre` and `post`), because I think most of the time people will want the same rule applied before and after operations.

angrykoala: I think I don't follow on what the problem is, do you mind having a quick chat whenever you are free to check this?

docs/rfcs/rfc-035-authorization.md
darrellwarde: Indeed, I've updated this section, thanks!

docs/rfcs/rfc-035-authorization.md
darrellwarde: I think what I don't understand about the above proposal; if we're going to force users to use a particular filter, why would we not just apply it for them using the `filter` argument? This seems to be anecdotally the same logic as HMRC and self-assessment - "we _know_ what tax you owe, but we're going to make you calculate it yourself or we will punish you"?

PR #2971, opened 2023-03-03
change Math operator to support unary operations, move outside string…

MacondoExpress:
# Add Math operators, separate string concat

Add other operators to the Math operators and separate the string.concat implementation from the math.plus.
This is a breaking change as the plus and minus are now treated as unary plus and unary minus and while the previous behavior is moved to `add/subtract` respectively. 

## Complexity

Complexity: Low

# Checklist

### DIFF
--- a/packages/cypher-builder/src/clauses/With.test.ts
+++ b/packages/cypher-builder/src/clauses/With.test.ts
@@ -71,7 +71,7 @@ describe("CypherBuilder With", () => {
         });
 
         test("With expression aliased", () => {
-            const expr = Cypher.plus(new Cypher.Param("The "), new Cypher.Param("Matrix"));
+            const expr = Cypher.append(new Cypher.Param("The "), new Cypher.Param("Matrix"));
             const alias = new Cypher.Variable();
             const withQuery = new Cypher.With([expr, alias]);
 
--- a/packages/cypher-builder/src/Cypher.ts
+++ b/packages/cypher-builder/src/Cypher.ts
@@ -77,7 +77,9 @@ export {
     endsWith,
     matches,
 } from "./expressions/operations/comparison";
-export { plus, minus } from "./expressions/operations/math";
+export { plus, minus, add, divide, multiply, subtract, remainder, pow } from "./expressions/operations/math";
+// Concat exported as append as concat is colliding with the Composite Clause concat
+export { concat as append } from "./expressions/operations/string";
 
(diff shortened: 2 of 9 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-2971.surge.sh

.changeset/shaggy-dingos-melt.md
```diff
@@ -0,0 +1,7 @@
+---
+"@neo4j/cypher-builder": patch
```
darrellwarde, line 2: If this is a breaking change, should this not be `minor`?

.changeset/shaggy-dingos-melt.md
```diff
@@ -0,0 +1,7 @@
...
+"@neo4j/graphql": patch
```
darrellwarde, line 3: The change doesn't reflect the library at all - should this even have a changesets entry?

packages/cypher-builder/src/clauses/With.test.ts
```diff
@@ -71,7 +71,7 @@ describe("CypherBuilder With", () => {
         });
 
         test("With expression aliased", () => {
-            const expr = Cypher.plus(new Cypher.Param("The "), new Cypher.Param("Matrix"));
+            const expr = Cypher.append(new Cypher.Param("The "), new Cypher.Param("Matrix"));
```
darrellwarde, line 74: Based on the comment above the export, do we want to consider exporting these functions under their namespace?

```suggestion
            const expr = Cypher.string.concat(new Cypher.Param("The "), new Cypher.Param("Matrix"));
```

This would also apply to math operators:

```ts
Cypher.math.multiply(literal1, literal2);
```

darrellwarde (commented):
The changes look good, I just wonder whether or not this functionality _should_ be available in the root!

packages/cypher-builder/src/clauses/With.test.ts
MacondoExpress: I was thinking something the same, but I was not sure what should be exported under a different namespace and what to export from the root namespace!

darrellwarde: @angrykoala, any thoughts on the above?

angrykoala:
Is this still relevant @MacondoExpress ?

PR #2972, opened 2023-03-03
Improve docs instructions on @node labels and constraints

a-alle:
# Description

Updates docs to instruct the users that the first entry in the `@node(labels:[])` list is the one considered when creating constraints.

## Complexity

Complexity: Low


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-2972.surge.sh

docs/modules/ROOT/pages/guides/v4-migration/index.adoc
```diff
@@ -154,6 +154,7 @@ type Tech @node(label: "TechDB") @plural(value: "Techs") {
 
 There is no concept of a "main label" in the Neo4j database. As such, keeping these two separate arguments causes a disconnect between the database and the GraphQL library.
 As a result, the `label` and `additionalLabels` arguments have been condensed into a single argument `labels` which will accept a list of string labels that used when a node of the given GraphQL type is created.
+Please note that defining `labels` means you take control of the database labels of the node. Because indexes and constraints in the Neo4j database can only be assigned one label, for the moment the first label in the `labels` list will be used.
```
mjfwebb, line 157: Do we want to use the phrase "for the moment", here? I think if we write that then we should provide some context about why this is only temporary.

darrellwarde: I would remove "for the moment", this isn't going to change in the database AFAIK. A slightly nitpicky suggestion also not to start the second sentence with "because":

```suggestion
Please note that defining `labels` means you take control of the database labels of the node. Indexes and constraints in Neo4j only support a single label, for which the first element of the `labels` argument will be used.
```

a-alle: Hm the reason for the "for the moment" was the discussion about potentially adding an argument on @unique to reference the label for the constraint to be created on. Not sure how the general feeling is on this now, I believe it was due a discussion in the team.

darrellwarde: It would still be the case that we take the first element though - that wouldn't go away. This argument would just provide a way to override this behaviour.

docs/modules/ROOT/pages/type-definitions/database-mapping.adoc
```diff
@@ -157,6 +157,16 @@ MATCH (this:Dog:K9)
 RETURN this { .name } as this
 ----
 
+Please note that defining `labels` means you take control of the database labels of the node. Because indexes and constraints in the Neo4j database can only be assigned one label, for the moment the first label in the `labels` list will be used.
```
mjfwebb, line 160:
```suggestion
Please note that defining `labels` means you take control of the database labels of the node. Indexes and constraints in Neo4j only support a single label, for which the first element of the `labels` argument will be used.
```

mjfwebb (changes requested):
Just need the same edit here :)

mjfwebb (approved):

docs/modules/ROOT/pages/type-definitions/indexes-and-constraints.adoc
```diff
@@ -41,21 +41,21 @@ type Colour {
 }
 ----
 
-The `@node` directive is used to change the database label mapping in this next example, so a unique constraint will be asserted for the label `Color` and the property `hexadecimal`:
+The `@node` directive is used to change the database label mapping in this next example, so a unique constraint will be asserted for the first label in the list, ie `Color`, and the property `hexadecimal`:
```
darrellwarde, line 44: Get rid of the "ie" here?

```suggestion
The `@node` directive is used to change the database label mapping in this next example, so a unique constraint will be asserted for the first label in the list, `Color`, and the property `hexadecimal`:
```

darrellwarde (dismissed):
LGTM, just one suggestion!

darrellwarde (approved):

PR #3008, opened 2023-03-10
Add support for DISTINCT in aggregation functions in Cypher Builder

angrykoala:
# Description
This PR adds support for DISTINCT in aggregation functions in the cypher builder
https://neo4j.com/docs/cypher-manual/current/functions/aggregating/#_counting_with_and_without_duplicates

Example:
```js
Cypher.count(param).distinct()
```

Results in: `count(DISTINCT $param)`


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-3008.surge.sh

angrykoala:
@darrellwarde 
The problem here is that distinct is only valid in aggregation functions, so it is actually sort of attached to the function, not to the expression in the argument so something like `count(expr.distinct())` would be a no-go

Another option could be to treat it a bit like a function on itself: `count(distinct(expr))` in a way that an aggregation function can accept either an expression or a `Distinct` statement.

This could also be extended to RETURN statements (that work the same way) but that can be a bit tricky as multiple columns can be passed to a RETURN, but distinct only exists at the beggining, independently of columns:
```RETURN DISTINCT 1, 5`

PR #3028, opened 2023-03-13
Introduce authorization to the schema model

MacondoExpress:
# Introduce authorization to the schema model

This PR introduces the authorization design to the schema model.
This is to be considered an initial implementation, Composite Entitities are not considered at this phase and more functionality may be added in the future to the Authorization classes during the Operation Tree scope.

## Complexity

Complexity: Low


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-3028.surge.sh

packages/graphql/src/schema-model/annotation/Annotation.ts
```diff
@@ -17,6 +17,6 @@
  * limitations under the License.
  */
 
-import type { CypherAnnotation } from "./CypherAnnotation";
-
-export type Annotation = CypherAnnotation;
+export interface Annotation {
+    name: string;
+}
```
darrellwarde, line 22: @angrykoala, I remember that your intention was for this to be:

```ts
export type Annotation = AuthorizationAnnotation | CypherAnnotation
```

Was there a reason for this which might be scuppered by this?

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
```diff
@@ -0,0 +1,129 @@
...
+        this.requireAuthentication = requireAuthentication === undefined ? true : requireAuthentication;
+        this.where = new AuthorizationFilterWhere(where);
+        this.when = when ?? ["BEFORE", "AFTER"];
+    }
+}
+
+export const AuthorizationFilterRuleWhereArguments = {
+    jwtPayload: "jwtPayload",
+    node: "node",
+};
```
darrellwarde, line 120: This seems to be unused?

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
```diff
@@ -0,0 +1,129 @@
...
+export class AuthorizationFilterWhere {
```
darrellwarde, line 121: I would change the naming of this as to not to conflict with filter rules.

```suggestion
export class AuthorizationWhere {
```

packages/graphql/src/schema-model/generate-model.ts
```diff
@@ -16,79 +16,117 @@
...
+
+    // hydrate interface to typennames map
+    definitionNodes.objectTypes.forEach((el) => {
+        if (!el.interfaces) {
+            return;
+        }
+        const objectTypeName = el.name.value;
+        el.interfaces?.forEach((i) => {
+            const interfaceTypeName = i.name.value;
+            const before = interfaceToImplementingTypeNamesMap.get(interfaceTypeName);
```
darrellwarde, line 52: `before` doesn't seem to be a descriptive name for this variable - I can't tell what it actually means.

darrellwarde (changes requested):
Generally looks good, I have a few comments? 🙂

packages/graphql/src/schema-model/annotation/Annotation.ts
angrykoala: Good eye. Actually yes, this may be a problem, because typescript interfaces are stupid, this interface will be matched by anything that has a name property (regardless on wether or not the class implements that interface), that means that if we have a function such as
```ts
function doSomethingWithAnnotation(ann: Annotation)
```
The class `Node` for example, will be allowed to be passed into this, as it has a `name` property

The fix for this would either use an abstract class instead of an interface or the approach that @darrellwarde mentions, that explicitly says what things are annotations

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
```diff
@@ -0,0 +1,129 @@
...
+    "DELETE",
+    "CREATE_RELATIONSHIP",
+    "DELETE_RELATIONSHIP",
+] as const;
+
+export type AuthorizationFilterOperation =
+    (typeof AuthorizationFilterOperationRule)[keyof typeof AuthorizationFilterOperationRule];
+
+export type AuthorizationValidateOperation =
+    (typeof AuthorizationValidateOperationRule)[keyof typeof AuthorizationValidateOperationRule];
```
angrykoala, line 48: I don't completely understand what these types are supposed to be, may there be a  more concise or explicit way of defining these?

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+});
+
+describe("AuthorizationValidateRule", () => {
+    it("initialize class correctly", () => {
+        const rule = {
+            where: { node: { name: { equals: "Keanu" } } },
+            when: ["BEFORE"],
+        } as AuthorizationValidateRuleConstructor;
+        const authValidateRule = new AuthorizationValidateRule(rule);
```
angrykoala, line 97: Do we need these extra classes? I have the feeling that we could pass the validateRule into the annotation directly and parse them there instead of creating a rule class

packages/graphql/src/schema-model/generate-model.ts
```diff
@@ -122,36 +160,24 @@ function createFieldAnnotations(directives: readonly DirectiveNode[]): Annotatio
...
-
-function getArgumentValueByType(argumentValue: ValueNode): unknown {
-    // TODO: parse other kinds
-    if (argumentValue.kind === Kind.STRING) {
-        return argumentValue.value;
-    }
-    if (argumentValue.kind === Kind.LIST) {
-        return argumentValue.values.map((v) => getArgumentValueByType(v));
-    }
+function createEntityAnnotation(directives: readonly DirectiveNode[]): Annotation[] {
```
angrykoala, line 172:
```suggestion
function createEntityAnnotations(directives: readonly DirectiveNode[]): Annotation[] {
```

packages/graphql/src/schema-model/generate-model.ts
```diff
@@ -16,79 +16,117 @@
...
 
 export function generateModel(document: DocumentNode): Neo4jGraphQLSchemaModel {
     const definitionNodes = getDefinitionNodes(document);
+
+    // init interface to typennames map
+    const interfaceToImplementingTypeNamesMap = definitionNodes.interfaceTypes.reduce((acc, entity) => {
+        const interfaceTypeName = entity.name.value;
+        acc.set(interfaceTypeName, []);
+        return acc;
+    }, new Map<string, string[]>());
```
angrykoala, line 42: I would consider moving these to separate methods, that way the context could be in the name instead of the comment

```ts
initInterfacesToTypenamesMap(definitionNodes, interfaceTypeName)
```

packages/graphql/src/schema-model/generate-model.ts
```diff
@@ -16,79 +16,117 @@
...
+
+    // hydrate interface to typennames map
+    definitionNodes.objectTypes.forEach((el) => {
+        if (!el.interfaces) {
+            return;
+        }
```
angrykoala, line 48: Sames as above

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.test.ts
MacondoExpress: At this point no, we started the design with it and then never change back, but potentially in the future they could serve some purpose. I would say there is no harm to keep it as it!

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
MacondoExpress: Good catch!

packages/graphql/src/schema-model/annotation/Annotation.ts
MacondoExpress: We changed a bit the implementation and moved back to the union type!

packages/graphql/src/schema-model/parser/utils.ts
```diff
@@ -0,0 +1,48 @@
+/*
```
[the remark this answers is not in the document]
MacondoExpress: Big fan of it, I would not do it now as this PR is already a blocker, but absolutely!

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
```diff
@@ -0,0 +1,115 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const AuthorizationAnnotationArguments = {
+    filter: "filter",
+    validate: "validate",
+};
```
darrellwarde, line 23: I find this an odd export; it's only used in one place for logging the values, so this could well be an array?

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
```diff
@@ -0,0 +1,115 @@
...
+];
+
+export const AuthorizationValidateOperationRule: ReadonlyArray<AuthorizationValidateOperation> = [
+    "READ",
+    "CREATE",
+    "UPDATE",
+    "DELETE",
+    "CREATE_RELATIONSHIP",
+    "DELETE_RELATIONSHIP",
+];
```
darrellwarde, line 52: Why did we decide against inferring the types from these declared as a `const` array in the end? Seems to be that we now have duplicate specification of these strings now.

darrellwarde (approved):
This looks good, I just have a couple of small observations/questions! 🙂

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
MacondoExpress: You're right, changed, no reason, we changed this file over multiple iterations and now it could be better typed as a simple array!

packages/graphql/src/schema-model/annotation/AuthorizationAnnotation.ts
MacondoExpress: It could be inferred directly but as in another comment was pointed out, it was less readable, after the change I agree that even if a bit redundant is way easier to understand this type.

PR #3087, opened 2023-03-23
Merging content for restructuring of docs

lidiazuin:
Work in progress. Please do not merge.


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-docs-3087.surge.sh

darrellwarde:
As discussed, a few points that need changing:

- [ ] Remove the orphaned `type-definitions/index.adoc` page and children
- [ ] Create a new top-level section "Migration Guides", and move the "Migration from `neo4j-graphql-js`" page into it at the same level as the other migration guides
- [ ] I actually just noticed the "OGM" page under the "Reference" section - I would move that up to the top-level!
- [ ] Find the missing API reference page for the library and pop it under the "Reference" section
- [ ] The "Array methods" page seems to have gone missing. I would put this under the "Mutations" page, and also move the "Mathematical operators" page from in "Queries" to "Mutations"
- [ ] Finally, I would swap the order of the "Custom Resolvers" and "Subscriptions" pages in the contents

But overall, I think this is a much needed cleanup and looking really promising! 🙂

lidiazuin:
Hi @darrellwarde ! Thanks for the checklist. The Array Methods page was merged with the Update page under Mutations. Should I restore it to a different page instead?

darrellwarde:
> Hi @darrellwarde ! Thanks for the checklist. The Array Methods page was merged with the Update page under Mutations. Should I restore it to a different page instead?

I see! In that case, I would consider also merging Mathematical Operators into the Update page also, or perhaps as a sub-page of Update, whatever you think is best!

lidiazuin:
> > Hi @darrellwarde ! Thanks for the checklist. The Array Methods page was merged with the Update page under Mutations. Should I restore it to a different page instead?
> 
> I see! In that case, I would consider also merging Mathematical Operators into the Update page also, or perhaps as a sub-page of Update, whatever you think is best!

Cool, I've merged it. With that, I think everything from the checklist is covered and we are good to go, if there isn't anything else you might have spotted?

darrellwarde:
Under [https://neo4j-graphql-docs-3087.surge.sh/graphql-manual/3.0/api-reference/](https://neo4j-graphql-docs-3087.surge.sh/graphql-manual/3.0/api-reference/), `Neo4jGraphQL` is still a dead link,  I think that page still hasn't been mapped in!

darrellwarde:
Perhaps a bigger problem actually, it seems that all links to other pages in the documentation have become broken, perhaps because the targets have not been updated. This is particularly easy to spot on the Directives reference page. Maybe we were doing these links wrong, but as it stands, these will need updating!

lidiazuin:
> https://neo4j-graphql-docs-3087.surge.sh/graphql-manual/3.0/api-reference/

This was actually due to an accidental deletion of the page, but I fixed this already. About updating the targets, I'm working on it!

darrellwarde:
> @darrellwarde all links should be fixed now, but I think I broke the yarn.lock file again 🙃

With the `yarn.lock` file:

1. Merge the `4.0.0` branch into your branch.
2. If there are conflicts other than the `yarn.lock` file, resolve these manually.
3. If the `yarn.lock` file is conflicting, don't manually update! Just run `yarn` at the root of the repo and it will resolve them for you. Install yarn using `npm i -g yarn` if you don't have it.
4. Commit the changes, and push!

I should have explained this earlier, my bad! 🙃

darrellwarde:
@lidiazuin I think this is almost good to merge. One concern I've just realised, is the Filtering and Sorting under "Queries and Aggregations". Filtering and sorting are available on Mutations and Subscriptions also, so having them hidden under Queries could be detrimental. Perhaps we could merge them with pagination, "Filtering, Sorting and Pagination"? Too long?

lidiazuin:
@darrellwarde under the Subscription section, there's another page for Filtering specific to that - https://neo4j-graphql-docs-3087.surge.sh/graphql-manual/3.0/subscriptions/filtering/

Would that be the case to have a specific page for each case of filtering, sorting and pagination under mutations and subscription?

darrellwarde:
> @darrellwarde under the Subscription section, there's another page for Filtering specific to that - https://neo4j-graphql-docs-3087.surge.sh/graphql-manual/3.0/subscriptions/filtering/
> 
> Would that be the case to have a specific page for each case of filtering, sorting and pagination under mutations and subscription?

That's a great observation, and actually, I don't believe sorting is available for Mutations, so maybe we just need to add a filtering page in there at some point in future. 

With that being said, shall we merge this in? I think it's looking good, and we can start progressing from here!

PR #3100, opened 2023-03-27
custom schema validation

MacondoExpress:
# Description

This PR introduces a new mechanism to test the correct schema passed by the user for cases when a Neo4j GraphQL definition depends on the usage of the user schema.
This PR is introduced mainly to have a test for the new Authorization directive: https://github.com/neo4j/graphql/pull/2970 where the computed definition depends on the GraphQL definition where the directive is used.
This PR introduces some architectural components: 
- `makeValidationDocument`: Generates a new DocumentNode with the computed Definition, only used for validation purposes,
 - `validateUserDefinition`: Raise an error if a Definition is used incorrectly by the user,
 - `validateSDL`: Internal utility to visit the validation document node with the GraphQL standard validation and custom ones,
 - `Enricher`: This is a reducer callback that starting from the user Definition node, defines a new computed version of it, and adds it to the accumulator.
 - `EnricherContext`: Context used to store information between enrichment invocations.

## Complexity

Complexity: Medium

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-stage-3100.surge.sh

packages/graphql/src/schema/validation/schema-validation.ts
```diff
@@ -0,0 +1,96 @@
...
+                (definitions, enricher) => definitions.reduce(enricher, [] as DefinitionNode[]),
+                document.definitions
+            )
+            .concat(...additionalDefinitions),
+    };
+}
+
+function makeValidationDocument(userDocument: DocumentNode, augmentedDocument: DocumentNode): DocumentNode {
+    const enricherContext = new EnricherContext(userDocument, augmentedDocument);
+    const enrichers: Enricher[] = [];
```
a-alle, line 58: discussion: I'm unsure about the `Enricher` name here. To me an enricher should only add things to an original, not modify it like in this case. 
However as you pointed out, there are really 2 behaviours here: adding types (really enriching) and modifying definitions, so probably one single name will never accurately reflect this.
What are your thoughts about something like this? 
```js
const transformer = new Transformer(userDocument); // when to specify the referenceDocument and document that gets modified are up for debate
transformer.registerEnricher(authorizationDefinitionEnricher)
transformer.registerEditor(authorizationUsageEnricher)
transformer.registerStatic(getStaticAuthorizationDefinitions())
return transformer.transform(augmentedDocument);
```
Even though `registerEnricher` and `registerEditor` might have the same implementation, it's clearer that they have different side effects and acommodates for different behaviours in the future if needed. Also then we'll contain the document transforming logic and the context inside the Transformer and apply it how we want to the transformer functions.

MacondoExpress: The problem with Enricher and Editor being two separate entity is that even if now one enricher add new definitions and the other mutates the current definitions, this is only the current behavior of the authorizations enrichers while for others this can be done in a single pass. Considering that for every pass the whole Augmented Document is enumerated, having two separated entities (Enrichers and Editor) will lead to enumerating the Augmented Document twice, even when can be potentially achievable in a single pass. Of course, this is an initial design of this architecture component that can and very likely will be revisited during the refactoring of the augmented schema that will unlock different options!

MacondoExpress:
Thank you @a-alle  , addressed all the comments and it looks cleaner now!

packages/graphql/src/schema/validation/enrichers/authorization.ts
```diff
@@ -0,0 +1,162 @@
...
+function containsAuthorization(object: ObjectLikeDefinitionNode): boolean {
+    switch (object.kind) {
+        case Kind.INTERFACE_TYPE_EXTENSION:
+        case Kind.OBJECT_TYPE_EXTENSION:
+        case Kind.INTERFACE_TYPE_DEFINITION:
+        case Kind.OBJECT_TYPE_DEFINITION: {
+            const hasFields = !!object.fields;
+            return !!(
+                getAuthorizationDirectiveDefinition(object) ||
+                (hasFields && object.fields.some(getAuthorizationDirectiveDefinition))
```
a-alle, line 49: hasFields it not necessary, you could do optional chaining on object.fields

PR #3160, opened 2023-04-05
RFC: Updated method of specifying JWT payload

darrellwarde:
# Description

> **Note**
> 

Introduces an additional directive, `@jwtClaim`, for specifying JWT claims with more complex paths.

## Complexity

> **Note**
>

Complexity: Low


docs/rfcs/rfc-036-authorization.md
```diff
@@ -49,25 +50,43 @@ new Neo4jGraphQL({
...
+As part of this proposal, users will be able to define the structure of their JWT payload using two GraphQL directives:
 
 ```gql
 directive @jwtPayload on OBJECT
+directive @jwtClaim(
+  path: String!
+) on FIELD_DEFINITION
 ```
 
 This is a directive purely for flagging which type should be used as the representation of the JWT payload structure, and it can only be used once.
```
mjfwebb, line 62: This paragraph doesn't seem to match the new additions any more.

docs/rfcs/rfc-036-authorization.md
```diff
@@ -49,25 +50,43 @@ new Neo4jGraphQL({
...
 type JWTPayload @jwtPayload {
-  sub: String!
   roles: [String!]!
+  application1Groups: [String!]! @jwtClaim(path: "applications[0].groups")
 }
 ```
 
 This will be used for the generation of filters later down the line.
 
+We will automatically map the following JWT payload fields:
```
mjfwebb, line 81: It could be nice to add a note explaining why we are mapping these fields.

mjfwebb (commented):
A couple of comments only :)

docs/rfcs/rfc-036-authorization.md
```diff
@@ -20,27 +20,28 @@ type RemoteJWKS = {
 
 type Key = string | RemoteJWKS
 
-type AuthorizationConfig = {
+type AuthConfig = {
   key: Key | ((req: RequestLike) => Key);
   verify: boolean;
   // https://github.com/panva/jose/blob/main/docs/interfaces/jwt_verify.JWTVerifyOptions.md
   verifyOptions: JWTVerifyOptions;
+  globalAuthentication: boolean;
```
a-alle, line 28: I feel like we've discussed this before but I don't remember why it was not a good idea: if we support `@authentication` could this be replaced by a schema-level `@authentication` ?

darrellwarde: I don't actually know how this got here, I don't recall adding that line! 😅 We should really have an RFC for the `@authentication` directive!

mjfwebb (approved):
The only additional parts I would like to see are some usage examples of AuthConfig (for example, showing a different use of the AuthConfig.key), but that can also wait for documentation.

PR #3177, opened 2023-04-11
Fix apollo federation compatibility tests

Liam-Doodson:
# Description

- Renamed the docker service from `server` to `products` - this was causing the DNS errors
- Added missing test data that is required for >= federation 2.3
- Removed an necessary filter
- Added a missing `@relationship` directive

## Complexity

Complexity: Low


darrellwarde (approved):
Awesome.

PR #3345, opened 2023-05-23
Switch to bundled Apoc in GHA runs

darrellwarde:
# Description

> **Note**
> 

Due to 4.3 compatibility, we were using `NEO4JLABS_PLUGINS=\[\"apoc\"\]` to include Apoc, which fetches remotely published versions.

With 4.3 deprecation around the corner from the library, I have duplicated some test runs to use `NEO4J_PLUGINS=\[\"apoc-core\"\]` instead, which is compatible with 4.4 and 5.x, and uses the bundled Apoc.

The duplicated 4.3 jobs will need deleting in `4.0.0`.

## Complexity

> **Note**
>

Complexity: Low


mjfwebb (approved):
Great! :-1:

PR #3369, opened 2023-05-25
Add support for clauses as input of sub-clauses mixins where possible

angrykoala:
# Description

To make composability easier, this PR adds the option to pass already built clauses into the subclasses methods.

This way, the following:
```
const with= new Cypher.With();
new Cypher.Match().with(with)
```

is possible


mjfwebb (approved):
Yeah, I love this change. :)

PR #3606, opened 2023-07-06
Adds a test satisfying #2548

### LINKED ISSUE #2548: In `@auth`, `allowUnauthenticated` rule is ignored when other role-based rules are present.
**Describe the bug**
When using an object auth, the allowUnauthenticated rule is ignored if other role-based rules are present, e.g.
```
@auth(rules: [
    { operations: [READ], allow unauthenticated: true, where: {isPublic: true}}
    { operations: [READ], roles: ["ADMIN"]}
])
```
The first rule is ignored.

**Type definitions**
```
type User
@auth(rules: [
    { operations: [READ], allowUnauthenticated: true, where: {isPublic: true}}
    { operations: [READ], roles: ["ADMIN"]}
])
{
  userId: ID! @id
  isPublic: Boolean
}
```

**To Reproduce**
1. Create the above schema and create a few users, some with `isPublic=true` 
2. An unauthenticated select should return the users with `isPublic=true`, instead we get an `Unautheticated` error thrown

**Additional context**
This is due to an exception being raised when the second rule is checked, completely ignoring that the first rule allows it.

darrellwarde:
# Description

Adds a test satisfying #2548.

## Complexity

Complexity: Low

# Issue

Closes #2548

### DIFF
--- a/packages/graphql/tests/integration/issues/2548.int.test.ts
+++ b/packages/graphql/tests/integration/issues/2548.int.test.ts
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-import type { Driver, Session } from "neo4j-driver";
+import type { Driver } from "neo4j-driver";
 import type { GraphQLSchema } from "graphql";
 import { graphql } from "graphql";
 import Neo4j from "../neo4j";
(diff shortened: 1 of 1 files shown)


mjfwebb (approved):
:+1:

PR #3607, opened 2023-07-06
bug-fix 3591

### LINKED ISSUE #3591: OGM does not generate types when used within Schema Configuration directives.
**Describe the bug**

OGM does not generate types when used within Schema Configuration directives.

**To Reproduce**
Schema Configuration directives break OGM-generated types.
The following code:
```typescript
 const typeDefs = `
 type User @query(aggregate: false) {
    id: ID! @id
    company: [Company!]! @relationship(type: "WORKS_AT", direction: OUT)
  }

  type Company {
    id: ID! @id
  }

`;

const ogm = new OGM({
  typeDefs,
  // @ts-ignore
  driver: {},
});
generate({
  ogm,
  noWrite: true,
});

```
raise `TypeError: Cannot read properties of undefined (reading 'split')`.
Although this does not happen when aggregation is enabled by changing the argument aggregate to `true` `@query(aggregate:true)`

* Version: [@neo4j/graphql-ogm@3.22.0]

MacondoExpress:
# Description

Ignore the schema configuration directives for OGM purposes.

## Complexity

Complexity: low

Closes 
#3591

# Checklist


packages/ogm/src/utils/filter-document.ts
```diff
@@ -30,10 +30,20 @@ const excludedDirectives = [
...
+    "settable",
 ];
 
 function filterDocument(typeDefs: Neo4jGraphQLConstructor["typeDefs"]): DocumentNode {
-    const merged = mergeTypeDefs(Array.isArray(typeDefs) ? (typeDefs as string[]) : [typeDefs as string]);
+    // hack to keep aggregation enabled for OGM
+    const schemaExtension = "extend schema @query(aggregate: true)";
+    const merged = mergeTypeDefs(
+        Array.isArray(typeDefs) ? (typeDefs as string[]).concat(schemaExtension) : [typeDefs as string, schemaExtension]
+    );
```
darrellwarde, line 46: I see you add the schema extension in at this point - how does it not get removed by the reducer below? I would have thought it would have needed to be added at the end. Because if schema extensions aren't being filtered, then users will be able to add schema extensions also and potentially produce unexpected behaviour.

packages/ogm/src/utils/filter-document.ts
```diff
@@ -56,7 +66,28 @@ function filterDocument(typeDefs: Neo4jGraphQLConstructor["typeDefs"]): Document
...
+                                                        return {
+                                                            ...arg,
+                                                            value: {
+                                                                ...arg.value,
+                                                                value: true,
+                                                            } as BooleanValueNode, // expect a BooleanValue GraphQL validation already happened
+                                                        };
+                                                    }
+                                                    return arg;
+                                                }),
```
darrellwarde, line 86: Looks to me like this will only work if the `aggregate` argument has already been included in the directive, but aren't we going to be setting this to `false` by default in `4.0.0`? So this will have to add the `aggregate` argument regardless of whether it was in the original input.

darrellwarde (changes requested):
I think this might need a couple of tweaks!

packages/ogm/src/utils/filter-document.ts
MacondoExpress: Yeah, it's very hacky but if you see below only object types are transformed, so the extension definitions remain intact!

packages/ogm/src/utils/filter-document.ts
MacondoExpress: Ouch, I'm pretty sure you're right here :(

packages/ogm/src/utils/filter-document.ts
darrellwarde: Which is a problem right? Because it means people could pass in the `@mutation` directive for example and disable mutations, which would break the OGM. I think the reducer _does_ need to worry about schema extensions, and the extensions that we need should be added on following filtering.

darrellwarde (approved):

PR #3728, opened 2023-08-01
Fix use of unnormalized type name in generation

### LINKED ISSUE #3723: OGM TS Type Generations not working on uppercase types on 4.0.0 Beta
**Describe the bug**
In the new 4.0.0 Beta the type generation for uppercase typenames seems to be broken again.
Its the same Issue as in #3539 and was fixed by @darrellwarde in #3556 .

Seems like there are some new node.name instead of normalizedNodeName in the code.

image

darrellwarde:
# Description

Not sure what has gone on here, but this issue is also present on `dev` and not just on the 4.0.0 branch as reported.

## Complexity

Complexity: Low

# Issue

Closes #3723

### DIFF
--- a/packages/ogm/src/generate.ts
+++ b/packages/ogm/src/generate.ts
@@ -186,7 +186,7 @@ async function generate(options: IGenerateOptions): Promise<undefined | string> 
                     args?: any;
                     context?: any;
                     rootValue?: any;
-                }): Promise<${node.name}[]>
+                }): Promise<${normalizedNodeName}[]>
                 public create(args: {
                     input: ${normalizedNodeName}CreateInput[];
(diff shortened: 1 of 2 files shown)


packages/ogm/src/generate.ts
```diff
@@ -186,7 +186,7 @@ async function generate(options: IGenerateOptions): Promise<undefined | string>
                     args?: any;
                     context?: any;
                     rootValue?: any;
-                }): Promise<${node.name}[]>
+                }): Promise<${normalizedNodeName}[]>
```
tbwiss, line 189: question: Do we also need to use `normalizedNodeName` on line 218 in this file?

tbwiss (approved):

PR #3729, opened 2023-08-01
Improve the checking for non-existent parameters when evaluating authorization rules

darrellwarde:
# Description

Address the following issue reported on Discord:

> Hi, I started to play around with the new  `@authorization` directive using 3.23.0. Our jwt object that is passed to the executed query contains a list of stores where the user is admin.
> What we want to achieve is that the admin user is only able to query clients that belong to their stores. The authorization directive looks like this:
> `@authorization(validate: [{when: [BEFORE], where: {node: {config: {store_IN: "$jwt.storeAdmins"}}}}])`
> However this will throw an error when a graphql query is executed: `Type mismatch: expected List<T> but was Any, Map, Node or Relationship`
>
> The issue seems to be that the validatePredicate cypher that is generated is using COALESCE for the store_IN condition: `WHERE this0.store IN coalesce($jwt.storeAdmins, $jwtDefault)`
> $jwtDefault is an empty object so `this0.store IN coalesce($jwt.storeAdmins, $jwtDefault)` will throw an error when it is used with an _IN filter.
>
> This behavior can be easily checked in neo4j browser as well:
> ```
> :params {
>   jwt: {
>    adminStores: ["www.acme.com"]
>   },
>   jwtDefault: {}
> }
> ```
>
> This will execute: `return "www.acme.com" in coalesce($jwt.adminStores, [])`
> This will throw error: `return "www.acme.com" in coalesce($jwt.adminStores, $jwtDefault)`
>
> Do you have any recommendation how to handle this? 
>
> Great work btw, I really love the improvements I see so far around the new authorization directive!

## Complexity

Complexity: Medium

### DIFF
--- a/.changeset/young-ligers-poke.md
+++ b/.changeset/young-ligers-poke.md
@@ -1,0 +1,5 @@
+---
+"@neo4j/graphql": patch
+---
+
+Improve the checking for non-existent parameters when evaluating authorization rules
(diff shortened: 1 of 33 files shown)


a-alle (approved):
Looks good

PR #3770, opened 2023-08-10
use adapters to get the relationship direction

MacondoExpress:
Support query direction


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-3770.surge.sh

packages/graphql/src/schema-model/relationship/model-adapters/RelationshipAdapter.ts
```diff
@@ -97,6 +97,38 @@ export class RelationshipAdapter {
     public findAttribute(name: string): AttributeAdapter | undefined {
         return this.attributes.get(name);
     }
+    /**
+     * translation-only
+     *
+     * @param directed the direction asked during the query, for instance "friends(directed: true)"
+     * @returns the direction to use in the CypherBuilder
+     **/
+    public getCypherDirection(directed?: boolean): "left" | "right" | "undirected" {
```
angrykoala, line 106: A thing I don't like about this approach is that we are coupling our internal model (schema model) to the cypher-builder API

It is not a huge problem, as the API is still somewhat agnostic and we own both pieces, but it feels like this is the wrong place for this (at least while the adaptor are so coupled with the schema model)

angrykoala:
Is this worth it at the moment? I'm not so sure about coupling the cypher logic to the model tbh

Also, a couple of small comments on the logic to generate these values

packages/graphql/src/schema-model/relationship/model-adapters/RelationshipAdapter.ts
MacondoExpress: I agree, but the plan was to keep these adapters in the schema-model folder only during the refactoring, but then move them outside in a specialized fashion. Happy to delete these adapaters if the transition will not work out.

PR #3823, opened 2023-08-22
Breaking changes to the `@id` directive

darrellwarde:
# Description

The `@id` directive has had a number of breaking changes.

The `autogenerate` argument has been removed. With this value set to `false`, the `@id` directive was essentially an alias for the `@unique` directive. Use the `@unique` directive instead.

The `unique` argument has been removed. We consider it necessary for a globally unique identifier to be backed by a unique node property constraint, so we have removed the ability to disable this. If you wish to automatically generate an ID without it being backed by a constraint, we recommend using the `@populatedBy` directive with a callback to generate appropriate values.

The `global` argument has been removed. This quite key feature of specifying the globally unique identifier for Relay was hidden away inside the `@id` directive. This functionality has been moved into its own directive, `@relayId`, which is used with no arguments. The use of the `@relayId` directive also implies that the field will be backed by a unique node property constraint.

Note, if using the `@id` and `@relayId` directive together on the same field, this will be an autogenerated ID compatible with Relay, and be backed by a single unique node property constraint. If you wish to give this constaint a name, use the `@unique` directive also with the `constraintName` argument.

## Complexity

Complexity: Medium


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-3823.surge.sh

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -104,8 +130,14 @@ async function createIndexesAndConstraints({ nodes, session }: { nodes: Node[];
     }
 
     for (const constraintToCreate of constraintsToCreate) {
+        let createConstraintStr = "CREATE CONSTRAINT";
+
+        if (constraintToCreate.constraintName) {
+            createConstraintStr = createConstraintStr + ` ${constraintToCreate.constraintName}`;
```
mjfwebb, line 136:
```suggestion
            createConstraintStr = `${createConstraintStr} ${constraintToCreate.constraintName}`;
```

packages/graphql/src/schema/make-augmented-schema.test.ts
```diff
@@ -567,40 +567,31 @@ describe("makeAugmentedSchema", () => {
...
                 type User {
                     id: ID!
-                    email: ID! @id(global: true)
+                    email: ID! @id @relayId
                 }
             `;
 
             expect(() => makeAugmentedSchema(typeDefs)).toThrow(
-                `Type User already has a field "id." Either remove it, or if you need access to this property, consider using the "@alias" directive to access it via another field`
+                `Type User already has a field \`id\`, which is reserved for Relay global node identification.\nEither remove it, or if you need access to this property, consider using the \`@alias\` directive to access it via another field`
```
mjfwebb, line 588: I think it would be better to reduce the use of backticks in error messages. I suggest a style similar to Typescript: `"Error message on field 'id' has an error"`

packages/graphql/src/schema/parse/get-unique-meta.ts
```diff
@@ -40,29 +34,14 @@ function getUniqueMeta(
...
-            | undefined;
-        // If unique argument is absent from @id directive, default is to use unique constraint
-        uniqueId = idDirectiveUniqueArgument ? idDirectiveUniqueArgument.value : true;
-    }
-
-    if (uniqueId) {
-        return {
-            constraintName: `${type.name.value}_${fieldName}`,
-        };
+    if (directives.some((d) => ["id", "relayId"].includes(d.name.value))) {
```
mjfwebb, line 43: Opinionated code style suggestion:

```suggestion
    if (directives.some((directive) => ["id", "relayId"].includes(directive.name.value))) {
```

mjfwebb: The strings `id`and `relayId` should really be constants somewhere.

packages/graphql/src/graphql/directives/id.ts
```diff
@@ -17,26 +17,11 @@
  * limitations under the License.
  */
 
-import { DirectiveLocation, GraphQLBoolean, GraphQLDirective, GraphQLNonNull } from "graphql";
+import { DirectiveLocation, GraphQLDirective } from "graphql";
 
 export const idDirective = new GraphQLDirective({
     name: "id",
     description:
         "Indicates that the field is an identifier for the object type. By default; autogenerated, and has a unique node property constraint in the database.",
```
a-alle, line 25: this description should be updated

packages/graphql/src/schema/parse/get-unique-meta.ts
```diff
@@ -40,29 +34,14 @@ function getUniqueMeta(
...
+        return {};
```
a-alle, line 44: this is a bit weird, returning an empty object to signal the existence of a unique constraint.. Also, any reason why we discarded the fallback constraintName `${type.name.value}_${fieldName}`? It's nice to be able to identify which constraints were created by the library.

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -104,8 +130,14 @@ async function createIndexesAndConstraints({ nodes, session }: { nodes: Node[];
...
+        }
+
         const cypher = [
-            `CREATE CONSTRAINT ${constraintToCreate.constraintName}`,
+            createConstraintStr,
```
a-alle, line 140: also this implication of not having a default constraint name is not that nice

darrellwarde:
@a-alle, @Liam-Doodson, as per our discussion this morning, this now no longer configures the constraint.

PR #3927, opened 2023-09-11
Generic changesets workflow for all releases

darrellwarde:
# Description

I was writing some documentation for major releases and it struck me that setup could be a little easier. So I present a generic GHA workflow for all changesets releases, including all future major releases that we will do.

It might not work. 😆 

## Complexity

Complexity: Low


.github/workflows/changesets.yml
```diff
@@ -98,3 +83,52 @@ jobs:
...
+    if: ${{ github.ref_name == 'dev' && needs.release.outputs.published == 'true' }}
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: master
+          fetch-depth: 0
+          token: ${{ secrets.NEO4J_TEAM_GRAPHQL_PERSONAL_ACCESS_TOKEN }}
```
mjfwebb, line 101: doesn't this checkout step need to happen on both `merge-dev-into-master` and `merge-dev-into-dev`?

mjfwebb (commented):
I'm a little confused by the replacement to the changsets-dev workflow as there's a number of steps that don't seem to have been recreated (like installing deps, creating .npmrc etc)

Is there something I'm missing here?

.github/workflows/changesets.yml
darrellwarde: This is a good spot, thank you!

darrellwarde:
> I'm a little confused by the replacement to the changsets-dev workflow as there's a number of steps that don't seem to have been recreated (like installing deps, creating .npmrc etc)
> 
> Is there something I'm missing here?

They are all still there, they are just collapsed in the diff!

mjfwebb (approved):
Thank you!

Issue #1249, opened 2022-04-11
cypherParams are not passed on to nested query 


fnfbraga:
**Describe the bug**
When creating a nested query which depends on cypherParams these are not readable.

`error: Expected parameter(s): cypherParams
`
This is because the `translateRead` function does not accept the `context.cypherParams`

For now we patched your package but would like to request the changes:

replace `let cypherParams = {}`
with 
`let cypherParams = context.cypherParams || {}`

 and 
`cypherParams = { ...cypherParams, ...topLevelMatch[1] }`
with
`cypherParams = { cypherParams, ...topLevelMatch[1] }`

neo4j-team-graphql:
Many thanks for raising this bug report @fnfbraga. :bug: We will now attempt to reproduce the bug based on the steps you have provided.

Please ensure that you've provided the necessary information for a minimal reproduction, including but not limited to:
* Type definitions
* Resolvers
* Query and/or Mutation (or multiple) needed to reproduce

If you have a support agreement with Neo4j, please link this GitHub issue to a new or existing Zendesk ticket.

Thanks again! :pray:

darrellwarde:
Hey @fnfbraga, it definitely sounds like there is a problem here, but can you please provide some type definitions and Query/Mutation to run to reproduce please? That way we can write tests to prevent regressions in the future. Thanks!

fnfbraga:
Hi @darrellwarde 

This essentially happens when you use `apoc.cypher.runFirstColumn` which I am not sure if its 3 or 4 levels of nesting in a query.

SO once you have the types

` 
type Level1 {
  id: ID!
  props: Level2!
}

type Level2 {
  id: ID!
  props: Level3!
}

type Level3 {
  id: ID!
  props: Level4!
  data: @cypher(
		statement: """
		  MATCH (p:Person {company: cypherParams.company})
		  RETURN p		
                 """
	)
}
`

darrellwarde:
I think possibly your reply got cut off there @fnfbraga?

fnfbraga:
> I think possibly your reply got cut off there @fnfbraga?

thank you for the heads up, updated

fnfbraga:
@darrellwarde would you like me to create a PR for this? Thank you.

tbwiss:
@fnfbraga feel free to do so!
Just keep in mind to also add a test with the example described above.
Thank you in advance!

tbwiss:
Hi @fnfbraga!
[...]

PR #3928, opened 2023-09-11
Support for dynamic labels

MacondoExpress:
Support for dynamic labels

This PR fixes:
- 'Label in Node directive Select Movie with label Film',
- 'https://github.com/neo4j/graphql/issues/1249 should contain the cypherParams that are passed via the context'

It also removes parentNode from the transpile contract, and shares code between the old NodeDirective and the new Schema Model Concrete Entity.

### DIFF
--- a/packages/graphql/src/classes/NodeDirective.ts
+++ b/packages/graphql/src/classes/NodeDirective.ts
@@ -17,10 +17,11 @@
  * limitations under the License.
  */
 
-import dotProp from "dot-prop";
+
 import { Neo4jGraphQLError } from "./Error";
 import Cypher from "@neo4j/cypher-builder";
 import type { Neo4jGraphQLContext } from "../types/neo4j-graphql-context";
+import { mapLabelsWithContext } from "../schema-model/utils/map-labels-with-context";
 
 export interface NodeDirectiveConstructor {
     labels?: string[];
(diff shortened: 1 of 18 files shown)


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-3928.surge.sh

packages/graphql/src/schema-model/entity/model-adapters/ConcreteEntityAdapter.ts
```diff
@@ -28,6 +28,9 @@ import type { Annotations } from "../../annotation/Annotation";
 import { ConcreteEntityOperations } from "./ConcreteEntityOperations";
 import type { InterfaceEntityAdapter } from "./InterfaceEntityAdapter";
 import type { UnionEntityAdapter } from "./UnionEntityAdapter";
+import type { Neo4jGraphQLContext } from "../../../types/neo4j-graphql-context";
+import dotProp from "dot-prop";
+import Cypher from "@neo4j/cypher-builder";
```
angrykoala, line 33: Why are we importing these now?

packages/graphql/src/classes/NodeDirective.ts
```diff
@@ -37,42 +37,15 @@ export class NodeDirective {
         if (!typeName) {
             throw new Neo4jGraphQLError("Could not generate label string in @node directive due to empty typeName");
         }
-        const labels = this.getLabels(typeName, context).map((l) => this.escapeLabel(l));
+        const labels = this.getLabels(typeName, context).map((label) => Cypher.utils.escapeLabel(label));
```
angrykoala, line 40: I'm not 100% sure about this change, are we using this somewhere else? Why is this part of the new translation layer?

packages/graphql/src/translate/queryAST/ast/QueryAST.ts
```diff
@@ -38,8 +42,17 @@ export class QueryAST {
         //         target: new Cypher.NamedVariable("this"),
         //     })
         // );
-
-        const result = this.operation.transpile({ returnVariable: new Cypher.NamedVariable("this") });
+        const queryASTEnv = new QueryASTEnv(neo4jGraphQLContext);
```
angrykoala, line 45: Are we sure we want the context in the ENV? Maybe we should add the graphqlContext directly in the context

packages/graphql/src/translate/queryAST/ast/operations/interfaces/InterfaceConnectionPartial.ts
```diff
@@ -19,24 +19,23 @@
...
         const relationship = new Cypher.Relationship({ type: this.relationship.type });
         const relDirection = this.relationship.getCypherDirection(this.directed);
 
         const clause = new Cypher.Match(
-            new Cypher.Pattern(parentNode).withoutLabels().related(relationship).withDirection(relDirection).to(node)
+            new Cypher.Pattern(context.target).withoutLabels().related(relationship).withDirection(relDirection).to(node)
         );
 
-        const nestedContext = new QueryASTContext({ target: node, relationship, source: parentNode });
+        const nestedContext = context.push({ target: node, relationship});
```
angrykoala, line 38: :clap:

angrykoala (commented):
Good job on adding the QueryASTContext to the relevant functions.

My only concern is polluting the `Env` with the GraphQL context

packages/graphql/src/translate/queryAST/ast/QueryAST.ts
MacondoExpress: I added directly to the Env because it will be relevant across the whole tree. We can definitely add it to the Context but we will re-assign the reference every time we push a new Context. At least this was my reasoning on the Env rather than in the Context!

packages/graphql/src/classes/NodeDirective.ts
MacondoExpress: This is to unify the dynamic labels logic between the two implementations "Schema Model" and "Node" to avoid having duplicate code achieving the exact same result; `this.escapeLabel` was literally just invoking the `Cypher.utils.escapeLabels`!

PR #3957, opened 2023-09-15
Abstract edge case fixes

MacondoExpress:
Support concrete entitiy filtering and interface shared filters 
```
Tests improved in branch B:  [
  'Interface Relationships Interface Relationship Query through connection with where excluding type',
  'Interface Relationships Interface Relationship Query through connection with where overriding',
  'Interface Relationships Simple Interface Relationship Query with offset and limit',
  'Interface Relationships Type filtering using onType'
]
```


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-3957.surge.sh

packages/graphql/src/translate/queryAST/ast/operations/interfaces/InterfaceReadOperation.ts
```diff
@@ -103,4 +107,8 @@ export class InterfaceReadOperation extends Operation {
     public addSort(...sortElement: Sort[]): void {
         this.sortFields.push(...sortElement);
     }
+
+    protected getSortFields(context: QueryASTContext, target: Cypher.Variable): SortField[] {
+        return this.sortFields.flatMap((sf) => sf.getSortFields(context, target, false));
+    }
```
angrykoala, line 113: I think the name of this function this may lead to a problem in the future

`getSortFields` is the name used in the contract for sort, which is "public" in class `Sort`

In this case, we have a protected function with the same name. At the moment is not a problem, but in the future we will probably try to merge all the interfaces to have a common contract between all nodes

I think we should consider renaming this protected function, to avoid confusion/conflicts with the contract in the future

packages/graphql/src/translate/queryAST/factory/OperationFactory.ts
```diff
@@ -241,25 +217,52 @@ export class OperationsFactory {
...
 
                     return this.hydrateConnectionOperationAST({
                         relationship,
                         target: concreteEntity,
                         resolveTree,
                         context,
                         operation: connectionPartial,
+                        whereArgs: whereArgs,
                     });
-                }
```
angrykoala, line 261: These big blocks are getting complex to follow, please move them to separate functions if possible

packages/graphql/src/translate/queryAST/factory/OperationFactory.ts
```diff
@@ -320,14 +323,16 @@ export class OperationsFactory {
...
     }: {
         relationship: RelationshipAdapter;
         target: ConcreteEntityAdapter;
         resolveTree: ResolveTree;
         context: Neo4jGraphQLTranslationContext;
         operation: T;
+        whereArgs: Record<string, any>;
     }): T {
-        let whereArgs = (resolveTree.args.where || {}) as Record<string, any>;
+        //let whereArgs = (resolveTree.args.where || {}) as Record<string, any>;
```
angrykoala, line 335: :eyes:

packages/graphql/src/translate/queryAST/utils/filter-concrete-entities-with-on-filter.ts
```diff
@@ -0,0 +1,61 @@
+import type { ConcreteEntityAdapter } from "../../../schema-model/entity/model-adapters/ConcreteEntityAdapter";
+import type { InterfaceEntityAdapter } from "../../../schema-model/entity/model-adapters/InterfaceEntityAdapter";
+import type { UnionEntityAdapter } from "../../../schema-model/entity/model-adapters/UnionEntityAdapter";
+import { isUnionEntity } from "./is-union-entity";
+
+/**
+ * Returns the concrete entities presents in the where [_on] argument,
+ * if the where argument is not defined then returns all the concrete entities of the composite target.
+ **/
+export function filterConcreteEntities(
```
angrykoala, line 10: This name does not match the file name.

Could we try a more explicit name?

```
getConcreteEntitiesInOnArgumentOfWhere
```

That's a bit verbose, so maybe there is a better name

And update the file so the name is the same

packages/graphql/src/translate/queryAST/ast/operations/interfaces/InterfaceReadOperation.ts
MacondoExpress: Sure, I agree! Just a thing, this name and behavior mimic what we have in the class InterfaceConnectionReadOperation, I tried to keep everything as close as possible between these classes so that will be easier to unify later!

packages/graphql/src/translate/queryAST/utils/filter-concrete-entities-with-on-filter.ts
MacondoExpress: Good catch, I've played around with it and I've losed track of the name of it!

packages/graphql/src/translate/queryAST/ast/operations/interfaces/InterfaceReadOperation.ts
angrykoala: Oh, that's great, is just so we can distinguish that this is not yet the same interface :)

PR #3984, opened 2023-09-19
Fix labels filter on cypher directive with unions

angrykoala:
# Description

Fix tck tests `1848`:

```
Test Suites: 4 failed, 237 passed, 241 total
Tests:       9 failed, 1276 passed, 1285 total
```

### DIFF
--- a/packages/graphql/src/translate/queryAST/ast/fields/attribute-fields/CypherUnionAttributePartial.ts
+++ b/packages/graphql/src/translate/queryAST/ast/fields/attribute-fields/CypherUnionAttributePartial.ts
@@ -51,7 +51,17 @@ export class CypherUnionAttributePartial extends QueryASTNode {
     }
 
     public getFilterPredicate(variable: Cypher.Variable): Cypher.Predicate {
-        return (variable as Cypher.Node).hasLabels(...this.target.getLabels());
+        const labels = this.target.getLabels();
+
+        // TODO: Refactor when `.hasLabel` on variables is supported in CypherBuilder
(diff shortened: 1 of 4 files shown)


packages/graphql/src/translate/queryAST/ast/fields/attribute-fields/CypherUnionAttributePartial.ts
```diff
@@ -51,7 +51,17 @@ export class CypherUnionAttributePartial extends QueryASTNode {
...
+        const labels = this.target.getLabels();
+
+        // TODO: Refactor when `.hasLabel` on variables is supported in CypherBuilder
+        const predicates = labels.map((label) => {
+            return new Cypher.RawCypher((env) => {
+                const varName = env.compile(variable);
+                const labelStr = Cypher.utils.escapeLabel(label);
+                return `${varName}:${labelStr}`;
+            });
+        });
```
MacondoExpress, line 63: If I understood correctly the reason behind maybe the following works as well!

```suggestion
        const predicates = labels.map((label) => {
            return (variable as Cypher.Node).hasLabel(label);
        });
```

MacondoExpress (approved):

packages/graphql/src/translate/queryAST/ast/fields/attribute-fields/CypherUnionAttributePartial.ts
angrykoala: It works, but it is unsafe, as we do not guarantee that a variable is a Node at the moment

PR #4061, opened 2023-10-03
Top level interface query

a-alle:
# Description

This PR implements the simple top-level interface and union query fields. 
Also fixes a bug with the __id field which was pointing to the incorrect cypher variable.
Depends on #4067 to be merged into dev.

## Complexity

Complexity: Medium

# Issue

> **Note**
>
>  

Closes 

# Checklist


neo4j-team-graphql:

Check out your changes at https://neo4j-graphql-toolbox-4061.surge.sh

packages/graphql/src/translate/queryAST/utils/is-nested-context.ts
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+import type Cypher from "@neo4j/cypher-builder";
+import type { QueryASTContext } from "../ast/QueryASTContext";
+
+/** Checks if provided context has the target field defined (which would make it a nested translation context) */
+export function isNestedContext(context: QueryASTContext): context is QueryASTContext & { target: Cypher.Node } {
+    return context.target !== undefined;
+}
```
angrykoala, line 26: Are we sure this is correct?

At the moment, nestedContext happens when there is a source. A target is always defined (except on the new cases). Has this changed that much?

angrykoala (approved):
A couple of small comments. Looks good!

packages/graphql/src/translate/queryAST/utils/is-nested-context.ts
MacondoExpress: Sorry didn't have time to fully review yet, but I  wanted to raise this as well, if you look at the `propertyFilter` file changes, you'll find for instance:

```typescript
private getPropertyRef(queryASTContext: QueryASTContext): Cypher.Property {
    if (this.attachedTo === "node") {
        if (!isNestedContext(queryASTContext)) throw new Error("No parent node found!");
        return queryASTContext.target.property(this.attribute.databaseName);
    } else if (this.attachedTo === "relationship" && queryASTContext.relationship) {
        return queryASTContext.relationship.property(this.attribute.databaseName);
    } else {
        throw new Error("Transpilation error");
    }
}
```

The assertion line:
```typescript
if (!isNestedContext(queryASTContext)) throw new Error("No parent node found!");
```
Give the sense that a PropertyFilter can be translated only with "nested contexts" although this is not true, for normal operations like Read/Connection, in fact, the context passed to a top-level Read Operation is directly passed to the filters without creating a "nested context", even if the `isNestedContext` returns true because the target, in that case, is defined. I feel it's worth changing this name to a less opinionated name without inferring that if the target is defined then the context is "nested"!

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -501,20 +530,22 @@ function makeAugmentedSchema({
         ...(hasGlobalNodes ? { Node: { __resolveType: (root) => root.__resolveType } } : {}),
     };
 
-    unionTypes.forEach((union) => {
-        // It is possible to make union types "writeonly". In this case adding a resolver for them breaks schema generation.
-        const unionTypeInSchema = parsedDoc.definitions.find((def) => {
-            if (def.kind === Kind.UNION_TYPE_DEFINITION && def.name.value === union.name.value) return true;
+    schemaModel.compositeEntities.forEach((compositeEntity) => {
```
MacondoExpress, line 535: Here it's iterating over the whole compositeEntities and then again the whole definitions, this could have an impact on the performance for type definitions with a large number of composite entities!

packages/graphql/src/translate/queryAST/utils/is-nested-context.ts
a-alle: mmm yes sure, good points! I was mistaken thinking that target being undefined means you are in a nestedContext, it's more of the other way around sort of.
Would `hasTarget` work or do you have any suggestions?

MacondoExpress: Works for me!

packages/graphql/src/translate/queryAST/ast/QueryAST.ts
```diff
@@ -18,28 +18,40 @@
...
+import { ReadOperation } from "./operations/ReadOperation";
 import type { QueryASTNode } from "./QueryASTNode";
 import { QueryASTContext, QueryASTEnv } from "./QueryASTContext";
 import { createNodeFromEntity } from "../utils/create-node-from-entity";
 import type { Neo4jGraphQLContext } from "../../../types/neo4j-graphql-context";
+import type { CompositeReadOperation } from "./operations/composite/CompositeReadOperation";
 
 export class QueryAST {
-    private operation: ReadOperation;
+    private operation: ReadOperation | CompositeReadOperation;
```
MacondoExpress, line 29: Maybe it's just worth changing to
```suggestion
    private operation: Operation;
```

packages/graphql/src/translate/queryAST/ast/QueryAST.ts
```diff
@@ -18,28 +18,40 @@
...
 
-    constructor(operation: ReadOperation) {
+    constructor(operation: ReadOperation | CompositeReadOperation) {
```
MacondoExpress, line 31:
```suggestion
    constructor(operation: Operation) {
```

packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeReadOperation.ts
```diff
@@ -31,7 +31,7 @@ import type { QueryASTContext } from "../../QueryASTContext";
 
 export class CompositeReadOperation extends Operation {
     private children: CompositeReadPartial[];
-    private entity: InterfaceEntityAdapter | UnionEntityAdapter;
+    private _entity: InterfaceEntityAdapter | UnionEntityAdapter;
```
MacondoExpress, line 34: Is this attribute ever being used?

packages/graphql/tests/tck/union.test.ts
```diff
@@ -28,7 +28,7 @@ describe("Cypher Union", () => {
     let typeDefs: DocumentNode;
     let neoSchema: Neo4jGraphQL;
 
-    beforeAll(() => {
+    beforeEach(() => {
```
MacondoExpress, line 31: Any particular reason? 👀 
```suggestion
    beforeAll(() => {
```

MacondoExpress (commented):
It looks good! I left some comments! sorry for being late!

packages/graphql/tests/tck/union.test.ts
a-alle: because some tests use the experimental flag set to true while others don't

packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeReadOperation.ts
a-alle: hmm apparently not anymore

packages/graphql/tests/tck/union.test.ts
MacondoExpress: Ah I see, I didn't catch it!

packages/graphql/src/schema/make-augmented-schema.ts
a-alle: yeah it's not ideal, however this is pretty much how it was before (before it was looping through all the unions and all the interfaceRelationships, now it's looping through all composite entities so the only additional ones are interfaces that are not used as targets of relationships, which are usually a very little minority). 
However this is a bigger issue to do with the entire schema generation flow that shouldn't be in the scope of this PR.
I left a TODO comment to bring awareness to the issue.

MacondoExpress: Sounds good, didn't expected a resolution just wanted to raise awareness of the bottleneck!

MacondoExpress (approved):

PR #4136, opened 2023-10-16
Fix `@jwtClaims` not working in `@authentication` rules

### LINKED ISSUE #4112: @jwtClaim doesn't seem to add the targeted field to the roles array
# Describe the bug
It seems that `@jwtClaim` is not adding the targeted roles in the roles array, and thus `@authentication` fails to find them. I tested several different setup, but even the one in the doc is failing.

## To Reproduce
All tokens has been generated with jwt.io with the key 123456 , inclued the "admin" role, and are valid for a year.

The server: a classic apollo server with the following
```
    typeDefs,
    driver,
    features: {
      authorization: {
        key: "123456", // key used for the test to simplify token generation
      },
    }
  })
```
and lauched with
```
const { url } = await startStandaloneServer(server, {
    listen: { port: 4001 },
    context: async ({ req }) => ({
      token: req.headers.authorization,
    }),
  })
```

the typeDefs is the following:
```
const typeDefs = `#graphql
  type JWT @jwt {
    roles: [String!]!
  }

type Category @authentication(operations: [READ], jwt: { roles_INCLUDES: "admin" } ) {
    name: String! @unique
  }

```

I ran the following tests, every time in order to list the categories, and the expected result here is that it should be ok as the admin role is defined in the jwt:
### "roles" field at the root 
```
type JWT @jwt {
    roles: [String!]!
  }
```

```
@authentication(operations: [READ], jwt: { roles_INCLUDES: "admin" }
```

token used: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJPbmxpbmUgSldUIEJ1aWxkZXIiLCJpYXQiOjE2OTcxMTc2NzIsImV4cCI6MTcyODY1MzgyMSwiYXVkIjoid3d3LmV4YW1wbGUuY29tIiwic3ViIjoianJvY2tldEBleGFtcGxlLmNvbSIsInJvbGVzIjpbImFkbWluIl19.jv9vXPuPjsTBLkBcpsQL9wG--N_C898NcHJc0ReWpmI

**Result: OK**

**Map myApplication.roles to roles with jwtClaims (from the doc)**
```
type JWT @jwt {
    roles: [String!]! @jwtClaim(path: "myApplication.roles")
  }
```

```
@authentication(operations: [READ], jwt: { roles_INCLUDES: "admin" }
```

token used: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJPbmxpbmUgSldUIEJ1aWxkZXIiLCJpYXQiOjE2OTcxMTc2NzIsImV4cCI6MTcyODY1MzgyMSwiYXVkIjoid3d3LmV4YW1wbGUuY29tIiwic3ViIjoianJvY2tldEBleGFtcGxlLmNvbSIsIm15QXBwbGljYXRpb24iOnsicm9sZXMiOlsiYWRtaW4iXX19.gJsmG0M0BMYWQ9610r6vEVjYBWiAOTGEJnIro-ZhVFQ

**_Result: KO !_**

### Map roles to groups with jwtClaims
That is actually the one I really want to solve, because cognito provide a token with the field `cognito:groups` for the groups at the root. As I cannot define `cognito:groups` as a graphql field, I have to map it.
```
type JWT @jwt {
    roles: [String!]! @jwtClaim(path: "groups")
  }
```

```
@authentication(operations: [READ], jwt: { roles_INCLUDES: "admin" }
```

token used: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJPbmxpbmUgSldUIEJ1aWxkZXIiLCJpYXQiOjE2OTcxMTc2NzIsImV4cCI6MTcyODY1MzgyMSwiYXVkIjoid3d3LmV4YW1wbGUuY29tIiwic3ViIjoianJvY2tldEBleGFtcGxlLmNvbSIsImdyb3VwcyI6WyJhZG1pbiJdfQ.UcZG2NsGBbz0_xDDQcY1T4zufWEAQRpSeLPJqaKbG9U

**_Result: KO !_**

## Conclusion
I ran a lot of other tests, but basically I never managed to have @jwtClaim working as I was expected. So maybe I understand this feature wrong or I am not using it the right way.

## Versions
```
"@apollo/server": "^4.9.4",
"@neo4j/graphql": "^4.3.1",
"apollo-server": "^3.12.1",
"graphql": "^16.8.1",
"neo4j-driver": "^5.13.0"
```

a-alle:
# Description

This PR adds the logic to "expand" a jwt claim with the `@jwtClaims` alias when checking the `@authentication` rules.
Also does this for `@authentication` on top level Cypher fields.
Also does this for `@subscriptionAuthorization` (the claims map was not being provided to `wrapSubscriptions`).
Adds tests for all the above-mentioned cases.

Fixes #4112 

## Complexity

Complexity: Low

# Issue

> **Note**
>
>  

Closes 

# Checklist

### DIFF
--- a/.changeset/cold-buttons-glow.md
+++ b/.changeset/cold-buttons-glow.md
@@ -1,0 +1,5 @@
+---
+"@neo4j/graphql": patch
+---
+
+Fix `@jwtClaim` not working in `@authentication` rules
--- a/packages/graphql/src/translate/authorization/utils/filter-by-values.ts
+++ b/packages/graphql/src/translate/authorization/utils/filter-by-values.ts
@@ -69,11 +69,9 @@ function getReceivedValue<T>({
 }): T | undefined {
     const mappedJwtClaim = jwtClaims?.get(fieldName);
     if (mappedJwtClaim) {
-        return expandJwtClaim(mappedJwtClaim, receivedValues);
+        // expand the jwt claim
+        return dotProp.get<T>(receivedValues, mappedJwtClaim);
     } else {
         return receivedValues[fieldName];
     }
-}
-function expandJwtClaim<T>(mappedJwtClaim: string, jwtData: Record<string, T>): T | undefined {
-    return dotProp.get<T>(jwtData, mappedJwtClaim);
 }
--- a/packages/graphql/src/translate/authorization/utils/filter-by-values.ts
+++ b/packages/graphql/src/translate/authorization/utils/filter-by-values.ts
@@ -23,17 +23,18 @@ import { parseFilterProperty } from "../../../schema/resolvers/subscriptions/whe
 
 export function filterByValues<T>(
     whereInput: Record<string, T | Array<Record<string, T>> | Record<string, T>>,
-    receivedValues: Record<string, T>
+    receivedValues: Record<string, T>,
+    jwtClaims?: Map<string, string>
 ): boolean {
     for (const [k, v] of Object.entries(whereInput)) {
         if (Object.keys(multipleConditionsAggregationMap).includes(k)) {
             const comparisonResultsAggregationFn = multipleConditionsAggregationMap[k];
             let comparisonResults;
             if (k === "NOT") {
-                comparisonResults = filterByValues(v as Record<string, T>, receivedValues);
+                comparisonResults = filterByValues(v as Record<string, T>, receivedValues, jwtClaims);
             } else {
                 comparisonResults = (v as Array<Record<string, T>>).map((whereCl) => {
-                    return filterByValues(whereCl, receivedValues);
+                    return filterByValues(whereCl, receivedValues, jwtClaims);
                 });
             }
 
@@ -42,7 +43,16 @@ export function filterByValues<T>(
             }
         } else {
             const { fieldName, operator } = parseFilterProperty(k);
-            const receivedValue = receivedValues[fieldName];
+            let receivedValue: T | undefined;
+            console.log("here", fieldName, receivedValues);
+            const mappedJwtClaim = jwtClaims?.get(fieldName);
+            if (mappedJwtClaim) {
+                receivedValue = expandJwtClaim(mappedJwtClaim, receivedValues);
+                console.log("mapped", receivedValue);
+            } else {
+                receivedValue = receivedValues[fieldName];
+                console.log("NOT mapped", receivedValue);
+            }
             if (!receivedValue) {
                 return false;
             }
@@ -54,4 +64,16 @@ export function filterByValues<T>(
         }
     }
     return true;
+}
+
+function expandJwtClaim<T>(mappedJwtClaim: string, jwtData: Record<string, T>): T | undefined {
+    let receivedValue: T | undefined;
+    let paths = mappedJwtClaim.split(/(?<!\\)\./);
+
+    paths = paths.map((p) => p.replaceAll(/\\\./g, "."));
+
+    for (const p of paths) {
+        receivedValue = (receivedValue || jwtData)[p];
+    }
+    return receivedValue;
 }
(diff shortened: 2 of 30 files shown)


packages/graphql/src/translate/authorization/utils/filter-by-values.ts
```diff
@@ -42,7 +43,13 @@ export function filterByValues<T>(
...
         } else {
             const { fieldName, operator } = parseFilterProperty(k);
-            const receivedValue = receivedValues[fieldName];
+            let receivedValue: T | undefined;
+            const mappedJwtClaim = jwtClaims?.get(fieldName);
+            if (mappedJwtClaim) {
+                receivedValue = expandJwtClaim(mappedJwtClaim, receivedValues);
+            } else {
+                receivedValue = receivedValues[fieldName];
+            }
```
angrykoala, line 52: Nitpick

This could be a separate method to get the `receivedValue` from a `fieldName`

packages/graphql/src/translate/authorization/utils/filter-by-values.ts
```diff
@@ -55,3 +62,15 @@ export function filterByValues<T>(
...
+
+function expandJwtClaim<T>(mappedJwtClaim: string, jwtData: Record<string, T>): T | undefined {
+    let receivedValue: T | undefined;
+    let paths = mappedJwtClaim.split(/(?<!\\)\./);
+
+    paths = paths.map((p) => p.replaceAll(/\\\./g, "."));
+
+    for (const p of paths) {
+        receivedValue = (receivedValue || jwtData)[p];
+    }
```
angrykoala, line 74: I'm not sure anymore, but I think we used to have an util/lib for this kind of path access operations :thinking:

packages/graphql/src/translate/translate-top-level-cypher.ts
```diff
@@ -52,10 +52,12 @@ export function translateTopLevelCypher({
     if (annotation) {
         applyAuthentication({ context, annotation });
     }
+    // TODO: is this even possible now?
```
angrykoala, line 55: What is not possible now? :rofl:

angrykoala (dismissed):

.changeset/cold-buttons-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neo4j/graphql": patch
+---
+
+Fix `@jwtClaims` not working in `@authentication` rules
```
darrellwarde, line 5:
```suggestion
Fix `@jwtClaim` not working in `@authentication` rules
```

packages/graphql/src/translate/authorization/utils/filter-by-values.ts
darrellwarde: Yeah, why not using `dot-prop` here?

a-alle: oh, yes you're right 🤦‍♀️

packages/graphql/src/translate/translate-top-level-cypher.ts
a-alle: because we don't allow `@authorization` directive on top-level cypher fields anymore..

packages/graphql/src/translate/authorization/utils/filter-by-values.ts
```diff
@@ -55,3 +57,23 @@ export function filterByValues<T>(
...
+    const mappedJwtClaim = jwtClaims?.get(fieldName);
+    if (mappedJwtClaim) {
+        return expandJwtClaim(mappedJwtClaim, receivedValues);
+    } else {
+        return receivedValues[fieldName];
+    }
+}
+function expandJwtClaim<T>(mappedJwtClaim: string, jwtData: Record<string, T>): T | undefined {
+    return dotProp.get<T>(jwtData, mappedJwtClaim);
+}
```
darrellwarde, line 79: Is this function worthwhile having now it is just a single function call?

darrellwarde (dismissed):
Looks good, just a question about we need the `expandJwtClaim` function now `dot-prop` is used!

PR #4234, opened 2023-11-01
Fix pipeline (DBNAME collision)

MacondoExpress:
Two tests are using the same database name and this cause problem with the latest neo4j images.


a-alle (approved):
if this works let's maybe introduce a mechanism to do the database naming such that to avoid this problem in the future!

PR #4509, opened 2024-01-12
Fixes the missing takeover of custom directives for interfaces

Andy2003:
# Description

Fixes the missing takeover of custom directives for interfaces

## Complexity

Complexity: low

### Closes

closes #4566

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [ ] Documentation has been updated
- [x] TCK tests have been updated
- [ ] Integration tests have been updated
- [ ] Example applications have been updated
- [x] New files have copyright header
- [x] CLA (https://neo4j.com/developer/cla/) has been signed

PR #4516, opened 2024-01-16
chore: improve annotation handling

Andy2003:
# Description

Until now, you had to register each annotation in several places:

- in the union type `Annotation`
- in the enum `AnnotationsKey`
- in the type `Annotations`
- in the function `annotationToKey`

This was very error-prone.

This improvement:

- changes `Annotation` to be an interface with a read only property `name` being the name of the directive
- improves the way annotations of the library are stored:
  - created a new object (record) `annotationsParsers` containing all known annotations - it is enforced that the keys of this object matches the `Annotation::name` property
  - remove of the `AnnotationKey` enum since it can now be directly looked up via the `annotationsParsers` object
  - removed reverse lookup method `annotationToKey` since this name is now directly accessible via `Annotation::name`

## Complexity

Complexity: Low

# Checklist


neo4j-team-graphql:
Thanks for the Neo4j GraphQL Toolbox updates.

The Neo4j GraphQL Toolbox has now been torn down - reopening this PR will republish it.

packages/graphql/src/schema/validation/custom-rules/directives/valid-directive-field-location.ts
```diff
@@ -194,7 +194,7 @@ function validFieldOfInterfaceTypeLocation({
     directiveNode: DirectiveNode;
     parentDef: InterfaceTypeDefinitionNode | InterfaceTypeExtensionNode;
 }) {
-    if (SCHEMA_CONFIGURATION_FIELD_DIRECTIVES.includes(directiveNode.name.value)) {
+    if (SCHEMA_CONFIGURATION_FIELD_DIRECTIVES.includes(directiveNode.name.value as any)) {
```
angrykoala: to avoid `as any` here, please use the `isInArray` utility function

```suggestion
    if (isInArray(SCHEMA_CONFIGURATION_FIELD_DIRECTIVES,directiveNode.name.value)) {
```

packages/graphql/src/schema-model/library-directives.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const additionalDirectives = ["alias", "relationship", "relationshipProperties", "node", SHAREABLE] as const;
+
+export type LibraryDirectives =
+    | keyof typeof annotationsParsers
+    | (typeof additionalDirectives)[number]
+    | typeof DEPRECATED;
+const createDirectivesArray = <T extends readonly LibraryDirectives[]>(arr: T) => arr;
+
+export const LIBRARY_DIRECTIVES = createDirectivesArray([
```
angrykoala: I may be missing something, but I don't see  `LIBRARY_DIRECTIVES` being used anywhere

Could this variable be removed?

packages/graphql/src/schema-model/library-directives.ts
```diff
@@ -0,0 +1,95 @@
...
+import type { DEPRECATED } from "../constants";
+import { SHAREABLE } from "../constants";
+
+const additionalDirectives = ["alias", "relationship", "relationshipProperties", "node", SHAREABLE] as const;
+
+export type LibraryDirectives =
+    | keyof typeof annotationsParsers
+    | (typeof additionalDirectives)[number]
+    | typeof DEPRECATED;
+const createDirectivesArray = <T extends readonly LibraryDirectives[]>(arr: T) => arr;
```
angrykoala: I'm not sure I understand what this function is for. Seems like an identity function that doesn't change the value nor the typings

packages/graphql/src/schema-model/generate-model.ts
```diff
@@ -436,9 +435,9 @@ function generateConcreteEntity(
 
     // schema configuration directives are propagated onto concrete entities
     const schemaDirectives = definitionCollection.schemaExtension?.directives?.filter((x) =>
-        SCHEMA_CONFIGURATION_OBJECT_DIRECTIVES.includes(x.name.value)
+        SCHEMA_CONFIGURATION_OBJECT_DIRECTIVES.includes(x.name.value as any)
```
angrykoala: Please, use `isInArray` utility function to avoid casting to any

packages/graphql/src/schema-model/generate-model.ts
```diff
@@ -460,21 +459,6 @@ function getLabels(entityDefinition: ObjectTypeDefinitionNode): string[] {
...
-
-    // We only ever want to create one annotation even when an entity contains several key directives
-    const keyDirectives = directives.filter((directive) => directive.name.value === "key");
-    if (keyDirectives) {
-        entityAnnotations.push(parseKeyAnnotation(keyDirectives));
-    }
-    const annotations = parseAnnotations(directives);
-
-    return entityAnnotations.concat(annotations);
-}
```
angrykoala, line 476: Good thing you were able to get rid of this convoluted logic

packages/graphql/src/schema-model/annotation/Annotation.ts
```diff
@@ -17,140 +17,83 @@
...
+ Keep this for type checking!
+ We cannot directly assign the type `AnnotationsParsers` to the const `annotationsParsers`
+ because this will create a circular reference for type resolution.
+*/
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const typeCheckingForAnnotationsParsers: AnnotationsParsers = annotationsParsers;
+
+export type Annotations = {
+    [P in keyof typeof annotationsParsers]?: ReturnType<(typeof annotationsParsers)[P]>;
+};
```
angrykoala: Removing the long `if` chained statements would be a good shout, however I'm worried about the increased complexity of this approach.

These types are quite hard to read and understand, and I'm not sure it is worth the increase of complexity to reduce verbosity.

In another comment I made, we can see how these indirect types are being implicitly casted to `any` removing any form of static typing .


Another smell of these types being a bit too complex is the circular reference issue a few lines below

packages/graphql/src/schema-model/parser/parse-annotation.ts
```diff
@@ -52,66 +28,14 @@ export function parseAnnotations(directives: readonly DirectiveNode[]): Annotati
             // multiple interfaces can have this annotation - must constrain this flexibility by design
             return directivesMap;
         }
-        const annotation = parseDirective(directive);
+        const annotation = annotationsParsers[directive.name.value]?.(
+            directive,
+            directives.filter((other) => other.name.value === directive.name.value)
+        );
```
angrykoala: This bit here is quite hard to follow, it would be worth adding some intermediate variables to aid readability

However, more importantly:
* I've checked the typings, the result of `annotationsParsers[directive.name.value]` is `any`, removing any form of static type checking on this bit here, which includes the annotation that gets returned by it
* I've noticed that here we are passing 2 parameters to the parsing function, however, most parsing functions only accept one parameter (I think key is the exception). This bit here is relying on the lack of typings of `annotationsParsers[directive.name.value]` to handle the different interfaces implicitly. This is extremely brittle, as any annotationParser adding a different interface (which may happen and is the reasoning behind having different parsers) will silently fail

packages/graphql/src/schema-model/annotation/Annotation.ts
```diff
@@ -17,140 +17,83 @@
...
-    selectable = "selectable",
-    settable = "settable",
-    subscription = "subscription",
-    subscriptionsAuthorization = "subscriptionsAuthorization",
-    timestamp = "timestamp",
-    unique = "unique",
-    relayId = "relayId",
+export interface Annotation {
+    readonly name: string;
 }
```
angrykoala, line 70: While I wouldn't consider this a blocking problem. I want to mention that a Typescript interface is a bit brittle on this case due to duck-typing.

What I mean is that any object with a property `name` will fulfil this interface (and hence, passed as ok on any function using this type), including classes **not** extending from this interface

By comparison, while the union is, undoubtedly, more verbose, it will statically check for the Annotation being one of the classes defined in it. Making it more robust for type checking

angrykoala (changes requested):
Hi @Andy2003 

Thanks for your contribution, I've carefully reviewed this as it points out a flaw in the current annotations design. A good set of your changes are great and ready to be merged, some require a few small changes but are fine overall.

However, I'm a bit worried about a few aspects I regarding the `annotationsParsers` approach as well as the use of a `name` property to define annotations

My main concerns is that the types required to reduce the current verbosity are too complex. In the team we are trying to reduce the complexity of the Typescript types we use, as these often lead to less readable code and, down the line, brittle static typing.

In fact, this PR introduces a couple of brittle types (objects being implicitly casted to any, and hence, no static checks at all).

I would recommend splitting the changes that are good to go (like removing old code or using constants instead of plain strings) or almost ready into a separate PR (or updating this one)

The whole `annotationsParsers` would require bigger changes to fix these issues, and I think may require a whole different approach, unless there is a way to improve these typings with the current solution I cannot see.

packages/graphql/src/schema-model/library-directives.ts
Andy2003: You are right, I need it for the changes in PR #4509, but for now we can remove it from this PR

packages/graphql/src/schema-model/library-directives.ts
Andy2003: For example take these lines:

```typescript
const UNION_DIRECTIVES = createDirectivesArray(["query", "plural"] as const);
export type UnionDirective = (typeof UNION_DIRECTIVES)[number];
```

Here `UnionDirective` is `"query" | "plural"`

If I would define the `UNION_DIRECTIVES` like this:

```typescript
const UNION_DIRECTIVES: readonly LibraryDirectives[] = ["query", "plural"] as const;
```

the `UnionDirective` is `"private" | "plural" | "customResolver" | "jwt" | "selectable" | "populatedBy" | "subscription" | "authorization" | "default" | "settable" | "cypher" | ...`

angrykoala: I see, but then if you do:

```ts
const UNION_DIRECTIVES = ["query", "plural"] as const;
export type UnionDirective = (typeof UNION_DIRECTIVES)[number];
```

You would get the same type

If what worries you here is the type checking of the array being a `LibraryDirective`, the relatively new `satisfies` keyword would do the trick:

```ts
const UNION_DIRECTIVES = ["query", "plural"] as const satisfies readonly LibraryDirective[];
export type UnionDirective = (typeof UNION_DIRECTIVES)[number];
```

packages/graphql/src/schema-model/parser/parse-annotation.ts
```diff
@@ -18,100 +18,23 @@
...
-        }
-        return directivesMap;
-    }, new Map<string, Annotation>());
-    return Array.from(annotations.values());
-}
+export function parseAnnotations(directives: readonly DirectiveNode[]): Partial<Annotations> {
+    const groupedDirectives = new Map<string, DirectiveNode[]>();
+    directives.forEach((directive) => {
+        const directivesOfName = groupedDirectives.get(directive.name.value) ?? [];
+        groupedDirectives.set(directive.name.value, [...directivesOfName, directive]);
```
angrykoala: Please, change this this `forEach` for a `for ... of`. This is purely for consistency

packages/graphql/src/schema-model/parser/annotations-parser/authentication-annotation.ts
```diff
@@ -33,7 +32,7 @@ const authenticationDefaultOperations: AuthenticationOperation[] = [
     "DELETE_RELATIONSHIP",
     "SUBSCRIBE",
 ];
-export function parseAuthenticationAnnotation(directive: DirectiveNode): AuthenticationAnnotation {
+export const parseAuthenticationAnnotation: AnnotationFactory<AuthenticationAnnotation> = (directive) => {
```
angrykoala: For consistency (and to avoid some gotcha) I'd recommend keeping this as a `function` for top level named functions.

I guess you made this change to add the `AnnotationFactory` types, but because the `annotationsParsers` are now statically set in the array with the type set, this shouldn't be needed (the type will fail if the function does not match)

packages/graphql/src/schema-model/annotation/Annotation.ts
```diff
@@ -17,140 +17,115 @@
...
-    if (ann instanceof SettableAnnotation) return AnnotationsKey.settable;
-    if (ann instanceof TimestampAnnotation) return AnnotationsKey.timestamp;
-    if (ann instanceof UniqueAnnotation) return AnnotationsKey.unique;
-    if (ann instanceof SubscriptionAnnotation) return AnnotationsKey.subscription;
-    if (ann instanceof JWTClaimAnnotation) return AnnotationsKey.jwtClaim;
-    if (ann instanceof JWTPayloadAnnotation) return AnnotationsKey.jwtPayload;
-    if (ann instanceof RelayIDAnnotation) return AnnotationsKey.relayId;
-    throw new Error("annotation not known");
-}
+export type AnnotationFactory<T extends Annotation> = (
```
angrykoala: Maybe `AnnotationParser` is a better type for this? Just to keep the same naming, no other reason

packages/graphql/src/schema-model/parser/parse-annotation.ts
```diff
@@ -18,100 +18,23 @@
...
-            return undefined;
-    }
+    const result: Partial<Annotations> = {};
+    Object.entries(annotationsParsers).forEach(([name, parser]) => {
+        const relevantDirectives = groupedDirectives.get(name);
+        if (relevantDirectives?.length) {
+            const [first] = relevantDirectives;
+            result[name] = parser(first!, relevantDirectives);
+        }
+    });
```
angrykoala: As before change it to `for...of` (alternatively, maybe a `.reduce` could be a bit more succinct here)

There are also a couple of things I'd like to discuss about this logic. Overall I think this is a good change compared to the previous version, and some of them are minor points that may not require changes, but would like to know your opinion on these:

* This code assumes that the directive name (`directive.name.value`) is going to match the annotation name. I think that is true at the moment, and we can live with this, but it would be a bit better to make that either explicit through types or to decouple that need (if it is easy enough to do so, as I said, this is good enough)

A small type improvement that I could propose here:

```ts
const relevantDirectives = groupedDirectives.get(name) ?? [];
const firstDirective = relevantDirectives[0];
if (firstDirective) {
  result[name] = parser(first, relevantDirectives);
}
```

Functionally, it is the same as before, but by checking `undefined` of the first element instead of `.length` we can avoid the non-null assertion `first!`

packages/graphql/src/schema-model/library-directives.ts
```diff
@@ -0,0 +1,82 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import type { Annotations } from "./annotation/Annotation";
+import type { DEPRECATED } from "../constants";
+import { SHAREABLE } from "../constants";
+
+const additionalDirectives = ["alias", "relationship", "relationshipProperties", "node", SHAREABLE] as const;
+
+export type LibraryDirectives = keyof Annotations | (typeof additionalDirectives)[number] | typeof DEPRECATED;
```
angrykoala: I see that `additionalDirectives` is not really used other than for typings, so maybe this whole bit can be simplified?

```suggestion
type AdditionalDirectives = "alias" | "relationship" | "relationshipProperties" | "node" | typeof SHAREABLE | typeof DEPRECATED 

export type LibraryDirectives = keyof Annotations | AdditionalDirectives;
```

packages/graphql/src/schema-model/library-directives.ts
```diff
@@ -0,0 +1,82 @@
...
+    "plural",
+    "limit",
+    "fulltext",
+    "node",
+    "jwt",
+    SHAREABLE,
+    "deprecated",
+    ...SCHEMA_CONFIGURATION_OBJECT_DIRECTIVES,
+] as const satisfies readonly LibraryDirectives[];
+export type ObjectDirective = (typeof OBJECT_DIRECTIVES)[number];
```
angrykoala: This is more of a personal grudge maybe, but I don't think `type[number]` is an obvious way of describing the intention of this type.

As it is kind of unavoidable if the array is the source of truth, what do you think about adding a type helper (maybe a new file in utils)?

```ts
// Type matching the elements in object/array
export type ValueOf<T> = T[keyof T];
```

It is the same exact type, but I think helps convey the meaning when applied to these:

```suggestion
export type ObjectDirective = ValueOf<typeof OBJECT_DIRECTIVES>;
```


Thoughts on this?

angrykoala (changes requested):

packages/graphql/src/schema-model/library-directives.ts
Andy2003: For the PR #4509 i will need this array, so I want to keep it as is.

packages/graphql/src/schema-model/annotation/Annotation.ts
```diff
@@ -17,140 +17,119 @@
...
-    [AnnotationsKey.settable]: SettableAnnotation;
-    [AnnotationsKey.timestamp]: TimestampAnnotation;
-    [AnnotationsKey.unique]: UniqueAnnotation;
-    [AnnotationsKey.subscription]: SubscriptionAnnotation;
-    [AnnotationsKey.jwtClaim]: JWTClaimAnnotation;
-    [AnnotationsKey.jwtPayload]: JWTPayloadAnnotation;
-    [AnnotationsKey.relayId]: RelayIDAnnotation;
+type CheckAnnotationName<T> = {
+    [P in keyof T]: T[P] & { name: P };
 };
```
angrykoala, line 75: While I'm still not 100% happy with complex typings, I can see the benefit of adding this one, maybe adding a quick comment just so anyone reading this have the context of why this type exists?

Maybe something on the lines?
```
// Ensures that annotation name property matches its Annotations object key
```

packages/graphql/src/schema-model/library-directives.ts
```diff
@@ -0,0 +1,83 @@
...
+ * limitations under the License.
+ */
+import type { Annotations } from "./annotation/Annotation";
+import type { DEPRECATED } from "../constants";
+import { SHAREABLE } from "../constants";
+import type { ValueOf } from "../utils/value-of";
+
+const additionalDirectives = ["alias", "relationship", "relationshipProperties", "node", SHAREABLE] as const;
+
+export type LibraryDirectives = keyof Annotations | ValueOf<typeof additionalDirectives> | typeof DEPRECATED;
```
angrykoala: Could we remove the export of this? I think it is not being used outside of this file :thinking: (Ignore this comment if it is being used and I just missed it)

```suggestion
type LibraryDirectives = keyof Annotations | ValueOf<typeof additionalDirectives> | typeof DEPRECATED;
```

angrykoala (commented):
Looking good, nice work here. Just a couple of minor comments.

I know some of my comments are just stylistic but it is mostly for consistency

angrykoala (approved):

PR #4555, opened 2024-01-22
Add explicit check for a skip on license-header-check

mjfwebb:
# Description

Prior to this change, a skipping license-header-check would cause this quality-gate job to skip.

This change makes it so that if the license-header-check skips or succeeds then quality-gate will still succeed/fail depending on the other needed jobs.

## Complexity

Low


Liam-Doodson (approved):
LGTM! Let's give this a go!

PR #4567, opened 2024-01-23
build: deliberately fail quality-gate is any needed job doesn't succeed

mjfwebb:
# Description

Currently, a failing test will cause the quality-gate to skip. Instead we want the quality-gate to be reported as failed if one of its needed jobs fails.

## Complexity

Low


Liam-Doodson (approved):
LGTM!

PR #4623, opened 2024-01-30
4615 aggregations with no result returns null but the graphql type is nonnull

angrykoala:
Makes aggregation types nullable, even if the original property is non-nullable.

This is because, in case of no nodes existing in the database, a null value will be returned by the aggregation

Schema tests had to be updated. New integration and schema tests added

### DIFF
--- a/.changeset/fair-meals-reflect.md
+++ b/.changeset/fair-meals-reflect.md
@@ -1,0 +1,7 @@
+---
+"@neo4j/graphql": major
+---
+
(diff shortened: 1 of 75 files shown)


neo4j-team-graphql:
Thanks for the Neo4j GraphQL Toolbox updates.

The Neo4j GraphQL Toolbox has now been torn down - reopening this PR will republish it.

darrellwarde (approved):

PR #4710, opened 2024-02-15
Remove code-cov while preserving test coverage archival

mjfwebb:
# Description

This PR removes code-cov test coverage uploads but preserves the archival of the test coverage artifacts.

## Complexity

Low


darrellwarde:
Do we need the `disable-coverage-archival` or can we just always archive? Does it really matter?

Looks like something is a bit broken on the job which merged the artifacts.

angrykoala (approved):

PR #4801, opened 2024-02-29
Update README.md's discord linkout

inf3rnus:
Linkout info is useful for disc noobs like myself :)


angrykoala:
Hi @inf3rnus Thanks for that PR, however, this readme isn't really a place for explanation on how discord works as that is just a set of useful links

PR #4820, opened 2024-03-05
Migrate all the integration/issues tests to use the UniqueType utility.

MacondoExpress:
As title.


packages/graphql/tests/integration/issues/1150.int.test.ts
```diff
@@ -31,45 +32,55 @@ describe("https://github.com/neo4j/graphql/issues/1150", () => {
     let driver: Driver;
     let neo4j: Neo4jHelper;
 
+    let Battery: UniqueType;
+    let CombustionEngine: UniqueType;
+    let Drive: UniqueType;
+    let DriveComposition: UniqueType;
```
angrykoala, line 38: Nitpick, no need to change it, but in the cases where we are using `beforeAll` these could be replaced by a const:

```
const Battery = new UniqueType("Battery")
```

angrykoala (approved):

PR #4828, opened 2024-03-06
Migrate all the integration/aggregations tests to use UniqueType

MacondoExpress:
A chunk of migration on the migration to use UniqueTypes, in this case for all the integration/aggregations/ files


packages/graphql/tests/integration/aggregations/top-level/datetime.int.test.ts
```diff
@@ -225,7 +211,7 @@ describe("aggregations-top_level-datetime", () => {
 
             expect(gqlResult.errors).toBeUndefined();
 
-            expect((gqlResult.data as any).moviesAggregate).toEqual({
+            expect((gqlResult.data as any)[Movie.operations.aggregate]).toEqual({
```
angrykoala, line 214: **nitpick** I find it a bit nicer if we put the path in the expectation, this way we avoid the `any`:

```
expect(gqlResult.data).toEqual({
  [Movie.operations.aggregate]: {
   // ...
  }
})
```

angrykoala (approved):

PR #4849, opened 2024-03-11
Update flaky tests in issues

angrykoala:
# Description

This PR adds a TestHelpers class to standardize integration tests and fix flaky tests and updates issue tests to use this helper.

The following tests has not been updated: 349, 464, 505, 915, 4056
The test helper is still a draft and may need some changes to fix race conditions in cloud tests


packages/graphql/tests/integration/utils/tests-helper.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        return graphqlRuntime({
+            schema,
+            source: query,
+            contextValue: await this.getContextValue(),
+            ...args,
+        });
+    }
+
+    public async runGraphQLWithToken(
```
MacondoExpress, line 91: Is it not better to have it as an optional parameter of the method `runGraphQL`?

packages/graphql/tests/integration/utils/tests-helper.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    public createBearerToken(secret: string, extraData?: Record<string, any>) {
+        return createBearerToken(secret, extraData);
+    }
+
+    public async initNeo4jGraphQL(options: Omit<Neo4jGraphQLConstructor, "driver">): Promise<Neo4jGraphQL> {
+        if (helperLock) {
+            throw new Error("TestHelper still open. Did you forget calling .close()?");
+        }
+        helperLock = true;
```
MacondoExpress, line 48: I would set this variable after `await this.getDriver()`, because `getDriver` could throw and IMO it would be better to have the lock as false in case of error!

packages/graphql/tests/integration/utils/tests-helper.ts
```diff
@@ -0,0 +1,198 @@
...
+        }
+        const driver = await this.getDriver();
+        if (preClose) {
+            try {
+                await preClose();
+            } catch (err) {
+                // Ignore error
+            }
+        }
+        await this.cleanNodes(driver, this.uniqueTypes); // TODO: take the database into account
```
MacondoExpress: Is the TODO message still valid? I think `cleanNodes` in the last revision takes into account the database!

packages/graphql/tests/integration/issues/1050.int.test.ts
```diff
@@ -103,36 +97,25 @@ describe("https://github.com/neo4j/graphql/issues/1050", () => {
...
-        await session.run(`
+        await testHelper.runCypher(`
           CREATE (c:${testUser.name} {id: 'abc'})
             -[:OWNS]->(i:${testInbox.name} {ownerId: 'abc'})
             -[:CONTAINS]->(m:${testMessage.name} {ownerId: 'abc', subject: 'Hello', body: 'World'})
             <-[:ATTACHED_TO]-(a:${testAttachment.name} {ownerId: 'abc', contents: 'something interesting'})
         `);
 
-        const query = gql`
+        const query = /* */ `
```
MacondoExpress:
```suggestion
        const query = /* GraphQL */ `
```

packages/graphql/tests/integration/issues/2100.int.test.ts
```diff
@@ -17,117 +17,101 @@
...
+                id: ID
+                records: [Record!]! @relationship(type: "HAS_BUSSING", direction: OUT)
+            }
+            type ${BussingRecordType} implements Record {
+                id: ID!
+                attendance: Int
+                markedAttendance: Boolean!
+                    @cypher(
+                        statement: """
+                        MATCH (this)<-[:PRESENT_AT_SERVICE|ABSENT_FROM_SERVICE]-(member:Member)
```
MacondoExpress: No relevant to your PR but I noticed that we're missing the UniqueTypes inside the `@cypher` here!

packages/graphql/tests/integration/issues/2100.int.test.ts
```diff
@@ -17,117 +17,101 @@
...
+            }
+    
+            type ${BacentaType} implements Church @authentication {
+                id: ID @id @unique
+                name: String!
+                serviceLogs: [${ServiceLogType}!]! @relationship(type: "HAS_HISTORY", direction: OUT)
+                bussing(limit: Int!): [${BussingRecordType}!]!
+                    @cypher(
+                        statement: """
+                        MATCH (this)-[:HAS_HISTORY]->(:ServiceLog)-[:HAS_BUSSING]->(records:BussingRecord)-[:BUSSED_ON]->(date:TimeGraph)
```
MacondoExpress: Also here

packages/graphql/tests/integration/utils/tests-helper.ts
```diff
@@ -0,0 +1,198 @@
...
+import { createBearerToken } from "../../utils/create-bearer-token";
+import { UniqueType } from "../../utils/graphql-types";
+
+const INT_TEST_DB_NAME = "neo4jgraphqlinttestdatabase";
+const DEFAULT_DB = "neo4j";
+
+let helperLock = false;
+
+export class TestHelper {
+    private database: string = DEFAULT_DB;
```
MacondoExpress, line 36: I see below that we verify the connectivity using the variable `USE_DEFAULT_DB` IMO it would make sense that `checkConnectivity` and other methods target the same database, thoughts?

packages/graphql/tests/integration/utils/tests-helper.ts
```diff
@@ -0,0 +1,198 @@
...
+    public async initNeo4jGraphQL(options: Omit<Neo4jGraphQLConstructor, "driver">): Promise<Neo4jGraphQL> {
+        if (helperLock) {
+            throw new Error("TestHelper still open. Did you forget calling .close()?");
+        }
+        helperLock = true;
+        if (this.neo4jGraphQL) {
+            throw new Error("Neo4jGraphQL already initialized");
+        }
+        const driver = await this.getDriver();
+        this.neo4jGraphQL = new Neo4jGraphQL({
```
MacondoExpress, line 53: Nitpick: Maybe we set the validate as false by default to reduce the load and improve the tests speed

packages/graphql/tests/integration/issues/4532.int.test.ts
```diff
@@ -17,25 +17,22 @@
...
-import { UniqueType } from "../../utils/graphql-types";
-import { default as Neo4jHelper } from "../neo4j";
+import type { UniqueType } from "../../utils/graphql-types";
+import { TestHelper } from "../utils/tests-helper";
 
 describe("https://github.com/neo4j/graphql/issues/4532", () => {
-    let driver: Driver;
-    let neo4j: Neo4jHelper;
-    let neo4jGraphql: Neo4jGraphQL;
+    let testHelper: TestHelper;
```
MacondoExpress, line 24: in this case I believe that having the testHelper defined here may cause some flakyness as two describe below have different typeDefs

MacondoExpress (commented):
Left some comments but overall seems an improvement in the ergonomic and for the flakyness

packages/graphql/tests/integration/utils/tests-helper.ts
angrykoala: I find this a bit simpler, as the parameter could be confusing because we have other parameters, but we can discuss with the team, it is just what we prefer at the end of the day

packages/graphql/tests/integration/utils/tests-helper.ts
angrykoala: tbh, if the driver throws the tests are going to fail anyway

In any case, the lock must be before the driver (or we would have a race condition as the driver takes time to generate), I can add a try ... catch to release the lock  in case of an error, but then the close will fail so it is not as simple ...

packages/graphql/tests/integration/utils/tests-helper.ts
angrykoala: yep, good catch

packages/graphql/tests/integration/issues/2100.int.test.ts
angrykoala: quite relevant as it may be causing race conditions, I'll fix that!

packages/graphql/tests/integration/issues/1050.int.test.ts
```diff
@@ -134,12 +134,12 @@ describe("https://github.com/neo4j/graphql/issues/1050", () => {
         `;
 
         const result = await testHelper.runGraphQL(query, {
-            contextValue: await testHelper.getContextValue({
+            contextValue: {
```
MacondoExpress, line 137: Way better!

packages/graphql/tests/integration/utils/tests-helper.ts
angrykoala: mmm, in theory, checkConnectivity will update this value if needed, I'm not sure what you mean

packages/graphql/tests/integration/utils/tests-helper.ts
angrykoala: mmm, we could discuss that, however a quick test show that the performance improvements is less than 1% which is well under the margin of error so I think the impact is negligible at the moment

packages/graphql/tests/integration/issues/4532.int.test.ts
angrykoala: mm, it should be fine, as the typedefs are set in initNeo4jGraphQL and there is a hook throwing an error if you try to call this method twice on the same  testHelper. The different describes in the same file are executed in series

packages/graphql/tests/integration/utils/tests-helper.ts
MacondoExpress: I see what you mean, my point was mainly due to the fact that if a driver fails on a test, then all the other tests will fail because the lock is never being released. Not important anyway!

packages/graphql/tests/integration/utils/tests-helper.ts
MacondoExpress: I think that the `env.USE_DEFAULT_DB` is used only to verify the connectivity but is ignored for other purposes like `runCypher` or `runGraphQL`, I don't have tested tho, though it's just my impression reading the code!

packages/graphql/tests/integration/utils/tests-helper.ts
MacondoExpress: Not important anyway, as could be added with ease afterward, but I remember that when you have some type-dependant directives you will have a larger performance impact, although this could have changed between releases!

packages/graphql/tests/integration/utils/tests-helper.ts
```diff
@@ -189,7 +189,12 @@ export class TestHelper {
     }
 
     /** Removes all nodes with the given labels from the database */
-    private async cleanNodes(driver: neo4j.Driver, labels: Array<string | UniqueType>): Promise<neo4j.Result> {
+    private async cleanNodes(driver: neo4j.Driver, labels: Array<string | UniqueType>): Promise<void> {
+        if (labels.length === 0) {
```
MacondoExpress, line 196: Nice! Perhaps maybe is worth adding this extra check to the deprecated functions here: `packages/graphql/tests/utils/clean-nodes.ts` while waiting for the full migration!

MacondoExpress (dismissed):

packages/graphql/tests/integration/utils/tests-helper.ts
angrykoala: I don't think that will be a problem as the types in typescript ensure that the array is not empty

MacondoExpress (approved):

PR #4877, opened 2024-03-14
Fix root connection integration test

angrykoala:
# Description
Use test helper in `root-connection.int.test.ts`, this should improve this test flakyness


MacondoExpress (approved):
Finally!

PR #4878, opened 2024-03-14
update integration tests under path relationship-properties/ to new test methodologies

MacondoExpress:
as title


packages/graphql/tests/integration/relationship-properties/connect-union.int.test.ts
```diff
@@ -0,0 +1,171 @@
...
+import { TestHelper } from "../utils/tests-helper";
+
+describe("Relationship properties - connect on union", () => {
+    let Movie: UniqueType;
+    let Actor: UniqueType;
+    let Show: UniqueType;
+    let testHelper: TestHelper;
+
+    beforeAll(async () => {
+        testHelper = new TestHelper();
```
angrykoala, line 31: testHelper can be created as a const in the describe

angrykoala (approved):

PR #4880, opened 2024-03-14
Fix spatial types integration test

angrykoala:


MacondoExpress (dismissed):

MacondoExpress:
The test is still failing, perhaps there could be another reason why these tests is failing!

angrykoala:
I found that the expectations where assuming an order, which is not true for the request made, I updated them to use `SameMembers` instead

MacondoExpress (approved):

MacondoExpress:
> I found that the expectations where assuming an order, which is not true for the request made, I updated them to use `SameMembers` instead

Nice!

PR #4888, opened 2024-03-15
Update interface relationship tests with testHelpers

angrykoala:


packages/graphql/tests/integration/interface-relationships/declare-relationship/interface-filters.int.test.ts
```diff
@@ -376,18 +340,14 @@ describe("interface filters of declared interface relationships", () => {
     let seriesScreenTime;
     let episodeNr;
 
-    beforeAll(async () => {
-        neo4j = new Neo4jHelper();
-        driver = await neo4j.getDriver();
-    });
+    beforeAll(async () => {});
```
mjfwebb: Looks like this could be removed

angrykoala:
```suggestion
```

mjfwebb (approved):

PR #4909, opened 2024-03-18
Update connect or create tests

angrykoala:
# Description
Update connect or create tests to use TestHelper


packages/graphql/tests/integration/connect-or-create/connect-or-create-authorization.int.test.ts
```diff
@@ -60,7 +48,7 @@ describe("connectOrCreate", () => {
             }
             `;
 
-            queryUpdate = gql`
```
mjfwebb, line 63: we could also use `/* GraphQL */` on these potentially

mjfwebb, line 63: we could also use `/* GraphQL */` on these potentially

mjfwebb (approved):

PR #4938, opened 2024-03-22
Refactor assert indexes and constraints to use Schema Model

mjfwebb:
# Description

This PR updates the existing asserts indexes and constraints functions to use the schema model as the data source instead of the Node class.

Along with this refactor a small change was made to unique annotation to make the `constraintName` argument optional

## Complexity

Low


packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -46,21 +47,22 @@ export async function assertIndexesAndConstraints({
 
     try {
         if (options?.create) {
-            await createIndexesAndConstraints({ nodes, session });
+            await createIndexesAndConstraints({ schemaModel, session });
         } else {
-            await checkIndexesAndConstraints({ nodes, session });
+            await checkIndexesAndConstraints({ schemaModel, session });
         }
```
a-alle, line 53: this code seems to do the same thing regardless of the if condition

angrykoala (approved):

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -83,47 +85,69 @@ async function createIndexesAndConstraints({ nodes, session }: { nodes: Node[];
...
+            entity.annotations.fulltext.indexes.forEach((index) => {
+                const indexName = index.indexName || index.name; // TODO remove indexName assignment and undefined check once the name argument has been removed.
                 if (indexName === undefined) {
                     throw new Error("The name of the fulltext index should be defined using the indexName argument.");
                 }
                 const existingIndex = existingIndexes[indexName];
                 if (!existingIndex) {
                     const properties = index.fields.map((field) => {
-                        const stringField = node.primitiveFields.find((f) => f.fieldName === field);
+                        const attributeAdapter = entity.findAttribute(field);
```
a-alle, line 114: not too important but this is not an AttributeAdapter type, so the naming is a bit confusing

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -83,47 +85,69 @@ async function createIndexesAndConstraints({ nodes, session }: { nodes: Node[];
...
                         indexName: indexName,
-                        label: node.getMainLabel(),
+                        label: entityAdapter.getMainLabel(),
                         properties,
                     });
                 } else {
                     index.fields.forEach((field) => {
-                        const stringField = node.primitiveFields.find((f) => f.fieldName === field);
-                        const fieldName = stringField?.dbPropertyNameUnescaped || field;
+                        const attributeAdapter = entity.findAttribute(field);
```
a-alle, line 131: also not an adapter

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -172,63 +202,43 @@ async function checkIndexesAndConstraints({ nodes, session }: { nodes: Node[]; s
...
-                    indexErrors.push(`Missing @fulltext index '${indexName}' on Node '${node.name}'`);
+                    indexErrors.push(`Missing @fulltext index '${indexName}' on Node '${entity.name}'`);
 
                     return;
                 }
 
                 index.fields.forEach((field) => {
-                    const stringField = node.primitiveFields.find((f) => f.fieldName === field);
-                    const fieldName = stringField?.dbPropertyNameUnescaped || field;
+                    const attributeAdapter = entity.findAttribute(field);
```
a-alle, line 223: not an adapter

a-alle: also it feels like this code is extremely similar to the else above in createIndexesAndConstraints. Is there a benefit to extracting this into a utility function?

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -272,18 +281,33 @@ async function getMissingConstraints({
...
+            if (anyLabelHasConstraint === false) {
+                const constraintName =
+                    uniqueField.annotations.unique.constraintName || `${entity.name}_${uniqueField.databaseName}`;
+
                 missingConstraints.push({
-                    constraintName: field.unique!.constraintName,
-                    label: node.getMainLabel(),
-                    property,
+                    constraintName,
+                    label: entityAdapter.getMainLabel(),
```
a-alle, line 305: should we then also use the main label in the constraint name instead of `entity.name`?

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -272,18 +281,33 @@ async function getMissingConstraints({
...
+                    property: uniqueField.databaseName,
```
a-alle, line 306: above in this file we do `attribute.databaseName || fieldName` but here there is no fallback, however the uniqueField here is also just an Attribute. Should this maybe be a getter in the Attribute/AttributeAdapter class so we can make sure we have the same behaviour everywhere?

PR #4941, opened 2024-03-22
Move install of webpack-cli to reusable federation tests workflow

mjfwebb:
This PR moves the webpack-cli dependency install to `.github/workflows/reusable-federation-tests.yml` so it is more clear what is is used for.


.github/workflows/reusable-federation-tests.yml
```diff
@@ -16,6 +16,8 @@ jobs:
           cache: yarn
       - name: Install dependencies
         run: yarn
+      - name: Install webpack-cli
+        run: yarn add -D webpack-cli
```
angrykoala, line 20: doesn't this also add webpack to the package json?

not important, as it is going to be removed anyway, but maybe a simple `npm i webpack-cli` (or the equivalent yarn) is a bit more accurate to what we try to do here

angrykoala (approved):

PR #4947, opened 2024-03-25
Fix flakyness in array-push.int.test.ts

angrykoala:
The test array-push.int.test.ts suffers from flakyness caused by random datetimes, when it ends in .000 ms, then gets rounded by the GraphQL response (so, basically, a test that randomly fails 1/1000 times)

Check this issue in this run: https://github.com/neo4j/graphql/actions/runs/8418889021/job/23050344223

This PR changes the random times for a explicit time, and adds a test with a explicit case for rounding with .000


mjfwebb (approved):
Nice catch!

PR #4956, opened 2024-03-27
Fix e2e tests

angrykoala:
# Description
Applies the same changes as in integration tests on the E2E tests, using the testHelper to handle setup and teardown

Websockets and TestServers remain unchanged


packages/graphql/src/schema/resolvers/composition/wrap-query-and-mutation.ts
```diff
@@ -123,11 +123,15 @@ export const wrapQueryAndMutation =
             features,
             subscriptionsEnabled,
             executor,
-            neo4jDatabaseInfo,
             authorization: authorizationContext,
-            // Consider anything in here overrides
+        };
+
+        const finalContext = {
```
MacondoExpress: Does this change reduce the flakiness of e2e tests?  If not I would remove this from this PR as there could be some side effects produced by this change unrelated to our tests that are worth seeing in isolation!

MacondoExpress (dismissed):

PR #4963, opened 2024-03-28
Move test-helper to common test utils

angrykoala:
# Description

This PR moves the test helper to the general test utils, to be reused outside of the integration tests


angrykoala:
eslint checks may fail due to this PR changing more than 300 files

MacondoExpress (approved):

PR #5085, opened 2024-05-07
Hide validation warning under env variable #4843

### LINKED ISSUE #4843: Objects fully defined by custom queries should not throw warnings about unresolved fields.
### Type definitions

```GraphQL
extend schema @mutation(operations: []) @subscription(events: [])

type Movie @query(read:false) {
    actors: [Actor!]!
}

type Actor {
   name: String
}

type Query {
   getMovies():[Movie!]!
}
```

### Test data

_No response_

### Steps to reproduce

1. Generate schema or use OGM for type generation
2. Warning "Object types need a way to be resolved for field:  actors" is received 

### What happened

Consider we have a type of object that is fully returned by a custom query. Let's say that object is the Movie we have above. Our getMovies is implemented by a custom query that fully returns the object:

```ts
module.exports = {
  Query: {
      getMovies: getMoviesImplementationFunction,
   }
}
```

When upgrading from 4.4.5 to 5.1.0 for neo4j/graphql and neo4j/graphql-ogm, we get a warning when building schema or generating types using OGM that the actors field does not have a way to be resolved.  

We are using `@query(read:false)` to avoid having the query autogenerated and thus have full control over it's implementation.  We're also not generating any mutations. Thus,  we would expect not to have that warning thrown in this scenario. 

### Expected behaviour

No warning thrown, as it behaved in 4.4.5

### Version

5.1.0

### Database version

4.4

### Relevant log output

_No response_

angrykoala:
To avoid verbose output on valid use cases when using the OGM, the warning on object fields without a resolver is now shown under the DEBUG env variable

### DIFF
--- a/packages/graphql/src/schema/validation/custom-rules/warnings/object-fields-without-resolver.test.ts
+++ b/packages/graphql/src/schema/validation/custom-rules/warnings/object-fields-without-resolver.test.ts
@@ -25,7 +25,6 @@ import type {
 } from "graphql";
 import { gql } from "graphql-tag";
 import validateDocument from "../../validate-document";
-import { VALIDATE_OBJECT_FIELD_WARN_MSG } from "./object-fields-without-resolver";
 
 const additionalDefinitions = {
     enums: [] as EnumTypeDefinitionNode[],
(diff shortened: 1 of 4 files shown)


packages/graphql/src/schema/validation/custom-rules/warnings/object-fields-without-resolver.test.ts
```diff
@@ -44,48 +43,45 @@ describe("WarnObjectFieldsWithoutResolver", () => {
     afterEach(() => {
         warn.mockReset();
     });
+    describe("Does not show warning", () => {
+        test("Error on object field array  without resolver throw warning in debug", () => {
```
mjfwebb, line 47:
```suggestion
        test("Error on object field array without resolver throw warning in debug", () => {
```

mjfwebb (approved):

PR #5132, opened 2024-05-15
Add type checking for existing `@populatedBy` functionality

darrellwarde:
# Description

`@populatedBy` lacked any form of type checking. This PR adds this type checking ahead of adding support for remaining scalars.

This additionally fixes the following issues:

* Strings containing non-numeric values would evaluate to 0 as a `BigInt`, and not throw an error
* List values were not supported using `@populatedBy`
* `@populatedBy` could be applied to any field type

## Complexity

Complexity: Low


packages/graphql/src/classes/CallbackBucket.ts
```diff
@@ -62,12 +65,41 @@ export class CallbackBucket {
...
                 }
-
-                params[cb.paramName] = param;
             })
         );
 
         return { cypher, params };
     }
+
+    private parseCallbackResult(result: any, type: TypeMeta): any {
```
mjfwebb: We can use `CallbackReturnValue` from `packages/graphql/src/types/index.ts` as the return type here, as long as we add `Integer` to it. If that's not possible, then we can make a new type:

`type CallbackResult = Integer | number | string | boolean | null | CallbackResult[];`

packages/graphql/src/schema/validation/custom-rules/directives/populatedBy.ts
```diff
@@ -16,18 +16,30 @@
...
+        directiveNode: DirectiveNode;
+        traversedDef: ObjectOrInterfaceWithExtensions | FieldDefinitionNode;
+    }) {
         const callbackArg = directiveNode.arguments?.find((x) => x.name.value === "callback");
         if (!callbackArg) {
             // delegate to DirectiveArgumentOfCorrectType rule
             return;
         }
+        if (traversedDef.kind !== Kind.FIELD_DEFINITION) {
+            // delegate
```
mjfwebb: It could be nice to complete this comment, like the above so we know where we delegate to.

mjfwebb (approved):

PR #5136, opened 2024-05-16
Allow `@populatedBy` to be used on temporal fields

### LINKED ISSUE #4886: `@populatedBy` does not work in conjunction with `DateTime` data type: callback function is not called
**Describe the bug**
The `@populatedBy` directive does not work if the field the directive is declared for has the `DateTime` datatype. The callback function never gets called.

**Type definitions**
````graphql
type Request {
  status: Status!
  approvedAt: DateTime @populatedBy(callback: "setApprovedAtTimestampCallback", operations: [UPDATE])
}

enum Status {
  REQUESTED,
  APPROVED
}
````

**To Reproduce**

1. Declare the callback function:
````javascript
const setApprovedAtTimestampCallback: async (root, _parent, _args, context) => {
        return root.status === 'APPROVED' ? new Date().toISOString() : null;
},
````
2. Execute the following Mutation...
````graphql
mutation {
updateRequests(update: { status: APPROVED }){ info { nodesCreated } }
}
````
3. See error: the callback function is never called

**Expected behavior**
The callback function should be called and invoked as described in the docs.

**System:**
* Version: @neo4j/graphql@5.2.0

### PR DESCRIPTION
**Pull request by darrellwarde**
# Description

Allow `@populatedBy` to be used on temporal fields.

## Complexity

Complexity: Medium

# Issue

Closes #4886

### CODE REVIEW

File: packages/graphql/tests/integration/directives/populatedBy.int.test.ts [hunk 1]
```diff
@@ -1465,6 +1465,972 @@ describe("@populatedBy directive", () => {
...
+                                callback
+                            }
+                        }
+                    }
+                `;
+
+                const result = await testHelper.executeGraphQL(mutation);
+
+                expect(result.errors).toHaveLength(1);
+                expect(result.errors?.[0]?.message).toBe("DateTime cannot represent non temporal value: banana");
```
angrykoala: I would consider doing a single expect check on errors

```suggestion
                expect(result.errors).toEqual(["DateTime cannot represent non temporal value: banana"]);
```

packages/graphql/tests/integration/directives/populatedBy.int.test.ts
```diff
@@ -1465,6 +1465,972 @@ describe("@populatedBy directive", () => {
...
+                            callbacks: {
+                                callback,
+                            },
+                        },
+                    },
+                });
+
+                const movieId = generate({
+                    charset: "alphabetic",
+                });
```
angrykoala: Maybe we can avoid randomly generated test data? This makes debugging harder and may cause flaky tests

packages/graphql/tests/integration/directives/populatedBy.int.test.ts
```diff
@@ -1465,6 +1465,972 @@ describe("@populatedBy directive", () => {
             });
         });
 
+        describe("@populatedBy - DateTime", () => {
```
angrykoala: Maybe these tests could be on a separate file?

packages/graphql/tests/integration/directives/populatedBy.int.test.ts
mjfwebb: We can't just use `toEqual` here because the response is an object containing a `message` property. We could do this instead, though it is probably even more verbose than the original:

```ts
expect(result.errors).toEqual(
    expect.arrayContaining([
        expect.objectContaining({
            message: "DateTime cannot represent non temporal value: banana",
        }),
    ])
);
```

packages/graphql/tests/integration/directives/populatedBy.int.test.ts
mjfwebb: I think this should probably be addressed in a holistic future change

mjfwebb (approved):

PR #5153, opened 2024-05-22
New API nested filters.

MacondoExpress:
# Description

Add schema generation step for nested filters on the new API


packages/graphql/src/api-v6/schema-generation/schema-types/filter-schema-types/TopLevelFilterSchemaTypes.ts
```diff
@@ -49,16 +56,32 @@ export class TopLevelFilterSchemaTypes extends FilterSchemaTypes<ConcreteEntity>
...
                     OR: itc.NonNull.List,
                     NOT: itc,
                     ...this.convertAttributesToFilters([...this.entity.attributes.values()]),
+                    ...this.convertRelationshipsToFilters([...this.entity.relationships.values()]),
                 },
             };
         });
     }
+
+    private convertRelationshipsToFilters(relationships: Relationship[]): Record<string, InputTypeComposer> {
```
angrykoala: Maybe `createRelationshipFilterTypes` is a better name?

packages/graphql/src/api-v6/schema-generation/schema-types/filter-schema-types/TopLevelFilterSchemaTypes.ts
```diff
@@ -49,16 +56,32 @@ export class TopLevelFilterSchemaTypes extends FilterSchemaTypes<ConcreteEntity>
...
+        const fields: ([string, InputTypeComposer | GraphQLScalarType] | [])[] = filterTruthy(
```
angrykoala: Maybe it is more readable with Array?
```suggestion
        const fields: Array<[string, InputTypeComposer | GraphQLScalarType] | []> = filterTruthy(
```

angrykoala: Also, why filterTruthy?

it looks like this map always return a value

angrykoala (approved):

packages/graphql/src/api-v6/schema-generation/schema-types/filter-schema-types/TopLevelFilterSchemaTypes.ts
MacondoExpress: Sounds good! I'll be changing `convertAttributeToFilters` to `createPropertyFilters` as well in this case!

packages/graphql/src/api-v6/schema-generation/schema-types/filter-schema-types/TopLevelFilterSchemaTypes.ts
MacondoExpress: No problem!

MacondoExpress: Good catch I  changed it at the last minute but I forgot to remove `filterTruthy`

PR #5163, opened 2024-05-23
Add temporal types

MacondoExpress:


packages/graphql/tests/api-v6/integration/types/datetime.int.test.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    beforeEach(() => {
+        Movie = testHelper.createUniqueType("Movie");
+    });
+
+    afterEach(async () => {
+        await testHelper.close();
+    });
+
+    describe("find", () => {
```
angrykoala: I think that describe in describe is a bit messy, I'd rather have separate files here

angrykoala (approved):

PR #5179, opened 2024-05-30
Generic changesets workflow

darrellwarde:


.github/workflows/changesets.yml
```diff
@@ -55,7 +56,7 @@ jobs:
         uses: changesets/action@aba318e9165b45b7948c60273e0b72fce0a64eb9 # v1.4.7
         with:
           title: changesets for branch `${{ github.ref_name }}`
-          version: yarn changeset-version-${{ github.ref_name }}
+          version: yarn changeset version --since ${{ github.ref_name }} && yarn
```
darrellwarde, line 59: The `&&` here is the only thing I have my doubts about, but we won't know until it tries it!

PR #5184, opened 2024-05-30
Migrate 4667 tests to use old tooling

MacondoExpress:
as title.


packages/graphql/tests/tck/issues/4667.test.ts
```diff
@@ -36,7 +37,7 @@ describe("https://github.com/neo4j/graphql/issues/4667", () => {
 
         const neoSchema = new Neo4jGraphQL({ typeDefs });
 
-        const query = /* GraphQL */ `
+        const query = /* GraphQL */ gql`
```
darrellwarde:
```suggestion
        const query = gql`
```

darrellwarde (approved):
Looks good, just get rid of the additional code block "decorator"!

PR #5188, opened 2024-05-31
V6 - Pagination

angrykoala:
Adds cursor pagination, pageInfo and after to the v6 api


packages/graphql/src/api-v6/queryIRFactory/ReadOperationFactory.ts
```diff
@@ -88,8 +89,10 @@ export class ReadOperationFactory {
         const nodeResolveTree = connectionTree.fields.edges?.fields.node;
         const sortArgument = connectionTree.args.sort;
         const firstArgument = connectionTree.args.first;
+        const afterArgument = connectionTree.args.after ? cursorToOffset(connectionTree.args.after) : undefined;
 
-        const pagination = firstArgument ? new Pagination({ limit: firstArgument }) : undefined;
+        const hasPagination = firstArgument || afterArgument;
```
MacondoExpress, line 94:
```suggestion
        const hasPagination = Boolean(firstArgument ?? afterArgument);
```

packages/graphql/src/api-v6/queryIRFactory/ReadOperationFactory.ts
```diff
@@ -143,8 +146,10 @@ export class ReadOperationFactory {
         const sortInputFields = this.getSortInputFields({ entity: relTarget, relationship, sortArgument });
 
         const firstArgument = connectionTree.args.first;
+        const afterArgument = connectionTree.args.after ? cursorToOffset(connectionTree.args.after) : undefined;
```
MacondoExpress, line 149: As this will be duplicated between top-level and nested, we could handle it in a separate method `getPagination`

packages/graphql/tests/api-v6/integration/pagination/first-after.int.test.ts
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+        `;
+
+        const gqlResult = await testHelper.executeGraphQL(query);
+        expect(gqlResult.errors).toBeFalsy();
+        expect(gqlResult.data).toEqual({
+            [Movie.plural]: {
+                connection: {
+                    edges: expect.toBeArrayOfSize(2),
```
MacondoExpress, line 73: Expectations do not match the test name. Argument `after` behavior is not tested, as it will be valid for any queries with `first/limit` as `2`.

MacondoExpress (changes requested):

Issue #5270, opened 2024-06-25
Authorization filter causing undefined error


0xbmart:
**Describe the bug**

When trying to query the user associated to the passed JWT using a cypher directive while there are authorization filters to filter out blocking user, i am receiving an error stating a variable is undefined in the resulting cypher query 

**Type definitions**

```
  type User @node(labels: ["User"]) @authorization(
		filter: [
			{ where: { node: { NOT: { blockedUsers_SOME: { to: { id: "$jwt.sub" } } } } } },
			#{ where: { node: { blockedUsers_NONE: { to: { id: "$jwt.sub" } } } } },
		]
	) {
		id: ID! @unique @id

    blockedUsers: [UserBlockedUser!]! @relationship(type: "HAS_BLOCKED", direction: OUT)
  }

  type UserBlockedUser @node(labels: ["UserBlockedUser"]) @authorization(
		filter: [
			{ where: { node: { from: { id: "$jwt.sub" } } } }
		]
	) {
		id: ID! @id @unique

		from: User! @relationship(type: "HAS_BLOCKED", direction: IN) @settable(onCreate: true, onUpdate: false)
		to: User! @relationship(type: "IS_BLOCKING", direction: OUT) @settable(onCreate: true, onUpdate: false)
	}

  type Query {
    getMe: User @cypher(statement: "OPTIONAL MATCH (u:User {id: $jwt.sub}) RETURN u", columnName: "u")
  }
```

**To Reproduce**

1. Pull [this repo](https://github.com/bmart7/neo4j-auth-issue)
2. `npm run up-build` to start the server
3. Then run the following Query with the following `Authorization` header:
- Authorization: "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI4YzFiMDE4NC1kODU3LTRkOWQtOGUxMy1mYmFiNGJkYWVkMmUiLCJuYW1lIjoiSm9obiBEb2UiLCJpYXQiOjE1MTYyMzkwMjJ9.Vf_ydofF9cKOJmMmtxXVozpt6Lux9HErskAzi6izNuw"

```
query GetMe {
  getMe {
    id
    __typename
  }
}
```

5. See error
```
WITH u AS this0
MATCH (this0)-[:HAS_BLOCKED]->(this1:UserBlockedUser)
OPTIONAL MATCH (this1)-[:IS_BLOCKING]->(this2:User)
WITH *, count(this2) AS toCount
[...]

PR #5306, opened 2024-07-03
Add tests for cypher directive with auth

angrykoala:
# Description
Follow up tests for another case similar to #5270

### DIFF
--- a/packages/graphql/tests/integration/directives/cypher/cypher-auth.int.test.ts
+++ b/packages/graphql/tests/integration/directives/cypher/cypher-auth.int.test.ts
@@ -90,6 +90,7 @@ describe("https://github.com/neo4j/graphql/issues/5270", () => {
 
         await testHelper.executeCypher(`
                 CREATE (:${User} {id: "${userId}"})
+                CREATE (:${User} {id: "1234"})
                 CREATE (:${Thing})
             `);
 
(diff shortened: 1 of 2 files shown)


packages/graphql/tests/integration/directives/cypher/cypher-auth.int.test.ts
```diff
@@ -0,0 +1,106 @@
...
+                        __typename
+                    }
+                }
+            }
+        `;
+
+        const userId = "my-user-id";
+
+        await testHelper.executeCypher(`
+                CREATE (:${User} {id: "${userId}"})
```
MacondoExpress, line 92: To validate the assertion below, we have to have at least another User, if not, the assertion will be good with or without the filter from the authorization directive.

MacondoExpress (changes requested):
Left a comment but overall LGTM, thank you!

packages/graphql/tests/integration/directives/cypher/cypher-auth.int.test.ts
angrykoala: :thinking: the assertion is not 100% correct, as the bug fixed here is actually invalid cypher that doesn't even run, I'll take a look to improve the test

PR #5352, opened 2024-07-12
Graphql tree cleanup

angrykoala:
# Description
This PR change the types in the graphql tree to make it easier to read and update as well as removing some intermediate types


packages/graphql/src/api-v6/queryIRFactory/resolve-tree-parser/graphql-tree/where.ts
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** Args for `where` in nested connections (with edge -> node) */
+export type GraphQLWhereArgs = LogicalOperation<{
```
MacondoExpress: Shall we remove the `args` suffix? I don't think it adds any value
```suggestion
export type GraphQLWhere= LogicalOperation<{
```

MacondoExpress (approved):
Nice improvements! I'm thinking of removing the args suffix everywhere as it doesn't seem to add a lot of value!

PR #5361, opened 2024-07-15
Add `@unique` directive to v6

MacondoExpress:
Add `@unique` support in v6 branch.

Note: 
The `constraintName` is still optional and the `getMissingConstraints` still returns a default constraint name.
Removed the `assertIndexesAndConstraints` test related to the `@id` directive as the `@id` directive is no longer relevant after the `unique` argument was moved outside the directive


packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -262,55 +262,55 @@ async function getMissingConstraints({
...
-    debug(`About to execute Cypher: ${constraintsCypher}`);
-    const constraintsResult = await session.run<{ labelsOrTypes: [string]; properties: [string] }>(constraintsCypher);
-
-    constraintsResult.records
+    const CYPHER_SHOW_UNIQUE_CONSTRAINTS = "SHOW UNIQUE CONSTRAINTS";
+    debug(`About to execute Cypher: ${CYPHER_SHOW_UNIQUE_CONSTRAINTS}`);
+    const constraintsQueryResult = await session.run<{ labelsOrTypes: [string]; properties: [string]; name: string }>(
+        CYPHER_SHOW_UNIQUE_CONSTRAINTS
+    );
+    // Map that holds as key the label name and as value an object with properties -> constraint name
```
angrykoala: This is not accurate, as the value of the map is another map, not an object

As a note, maybe it is a bit easier to reason about if it where an object instead? nested maps can get confusing in the code

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -262,55 +262,55 @@ async function getMissingConstraints({
...
-
-            if (anyLabelHasConstraint === false) {
-                // TODO: The fallback value of `${entity.name}_${uniqueField.databaseName}` should be changed to use the main label of the entity
-                // But this can only be done once the translation layer has been updated to use the schema model instead of the Node class
-                const constraintName =
-                    uniqueField.annotations.unique.constraintName || `${entity.name}_${uniqueField.databaseName}`;
+                // TODO: remove default value once the constraintName argument is required
+                return (
+                    constraintsForLabel.get(uniqueField.databaseName) ===
+                    (constraintName ?? `${entity.name}_${uniqueField.databaseName}`)
```
angrykoala: Could this be a separate variable for better readability?

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
```diff
@@ -262,55 +262,55 @@ async function getMissingConstraints({
...
-    const constraintsResult = await session.run<{ labelsOrTypes: [string]; properties: [string] }>(constraintsCypher);
-
-    constraintsResult.records
+    const CYPHER_SHOW_UNIQUE_CONSTRAINTS = "SHOW UNIQUE CONSTRAINTS";
+    debug(`About to execute Cypher: ${CYPHER_SHOW_UNIQUE_CONSTRAINTS}`);
+    const constraintsQueryResult = await session.run<{ labelsOrTypes: [string]; properties: [string]; name: string }>(
+        CYPHER_SHOW_UNIQUE_CONSTRAINTS
+    );
+    // Map that holds as key the label name and as value an object with properties -> constraint name
+    const existingConstraints: Map<string, Map<string, string>> = constraintsQueryResult.records
```
angrykoala: This function is getting quite big, any change of moving the existing constraints generation to a new function? `getExistingConstraints`?

packages/graphql/tests/api-v6/integration/combinations/alias-unique/alias-unique.int.test.ts
```diff
@@ -0,0 +1,166 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { generate } from "randomstring";
+import type { UniqueType } from "../../../../utils/graphql-types";
+import { isMultiDbUnsupportedError } from "../../../../utils/is-multi-db-unsupported-error";
+import { TestHelper } from "../../../../utils/tests-helper";
+
+describe("assertIndexesAndConstraints/alias-unique", () => {
```
angrykoala: Maybe we can change this name to be more consistent with other tests?

angrykoala (approved):

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
MacondoExpress: Sounds good, let me separate this method, I tried to limited the refactor here but there is no reason why we cannot improve it a little bit more!

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
MacondoExpress: Right, no preference TBF!

packages/graphql/src/classes/utils/asserts-indexes-and-constraints.ts
MacondoExpress: Sure!

PR #5365, opened 2024-07-16
WIP: Improve FilterFactory

angrykoala:
# Description

Improve filter factory

### DIFF
--- a/packages/graphql/src/api-v6/queryIRFactory/FilterFactory.ts
+++ b/packages/graphql/src/api-v6/queryIRFactory/FilterFactory.ts
@@ -31,6 +31,7 @@ import { DurationFilter } from "../../translate/queryAST/ast/filters/property-fi
 import { PropertyFilter } from "../../translate/queryAST/ast/filters/property-filters/PropertyFilter";
 import { SpatialFilter } from "../../translate/queryAST/ast/filters/property-filters/SpatialFilter";
 import { fromGlobalId } from "../../utils/global-ids";
+import { asArray, filterTruthy } from "../../utils/utils";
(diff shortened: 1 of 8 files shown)


packages/graphql/src/api-v6/queryIRFactory/FilterFactory.ts
```diff
@@ -56,176 +57,85 @@ export class FilterFactory {
...
+                    return this.createNodeFilter({ entity, where: value });
+                }
+
+                if (key === "properties" && relationship) {
+                    const edgePropertiesFilters = this.createEdgePropertiesFilters({
+                        where: value,
+                        relationship,
+                    });
+
+                    return new LogicalFilter({
```
MacondoExpress: IMO this logical operator logic should be defined inside `createEdgePropertiesFilters` as is done for `createNodeFilters`. 
I would personally avoid having methods that return `Filter[]` as I would wrap these filters in a `LogicalFilter`.
For instance, here you would have the `LogicalFilter` broken for relationship properties.
Try this:
```gql
query {
    movies {
        connection {
            edges {
                node {
                    title
                    actors(
                        where: {
                            edges: {
                                properties: { NOT: { year: { equals: 1999 } } }
                            }
                        }
                    ) {
                        connection {
                            edges {
                                node {
                                    name
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

MacondoExpress (changes requested):
It looks way better than before, nice! 
I left a comment with a bug, but it could be easily solved by applying the same strategy to all the methods that may return a predicate!

MacondoExpress (approved):

PR #5459, opened 2024-08-12
Top-level Delete v6

mjfwebb:
# Description

Adds top-level delete resolver, translation, factory

### DIFF
--- a/packages/graphql/src/api-v6/queryIR/DeleteOperation.ts
+++ b/packages/graphql/src/api-v6/queryIR/DeleteOperation.ts
@@ -1,0 +1,189 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


packages/graphql/src/api-v6/queryIR/DeleteOperation.ts
```diff
@@ -0,0 +1,189 @@
...
+    private transpileTopLevel(
+        selection: SelectionClause,
+        context: QueryASTContext<Cypher.Node>
+    ): OperationTranspileResult {
+        this.validateSelection(selection);
+        const filterSubqueries = wrapSubqueriesInCypherCalls(context, this.filters, [context.target]);
+        const predicate = this.getPredicate(context);
+        const extraSelections = this.getExtraSelections(context);
+
+        const nestedOperations: (Cypher.Call | Cypher.With)[] = this.getNestedDeleteSubQueries(context);
```
angrykoala: nitpick/question. Isn't the type `Array<Cypher.Call | Cypher.With>` a little bit easier to read?

angrykoala: Also, I think these nested operations are not tested, AFAIK the current scope of the top level delete does not need this code. The code looks fine but I'm a bit wary of adding code that is not run. Maybe add this code in the nested delete PR if that is something you are already working on?

angrykoala (approved):

PR #5596, opened 2024-09-26
Migrate all eq filters

MacondoExpress:
# Description
Migrated all the tests to use `_EQ` rather than explicit equality. Fix `_EQ` with subscriptions.


packages/graphql/src/schema/resolvers/subscriptions/where/utils/parse-filter-property.ts
```diff
@@ -17,17 +17,30 @@
...
 export function parseFilterProperty(key: string): { fieldName: string; operator: string | undefined } {
-    const match = whereRegEx.exec(key);
-    if (!match) {
-        throw new Error(`Failed to match key in filter: ${key}`);
+    // eslint-disable-next-line prefer-const
+    let { fieldName: stuff, operator: operatorStuff, isNot } = parseWhereField(key);
+
+    // These conversions are only temporary necessary until the the _NOT operator exists, after that we can just return the output of parseWhereField
+    if (operatorStuff === "EQ") {
+        operatorStuff = undefined;
```
angrykoala: what is `operatorStuff`?

packages/graphql/src/schema/resolvers/subscriptions/where/utils/parse-filter-property.ts
```diff
@@ -17,17 +17,30 @@
...
+            operatorStuff = `NOT_${operatorStuff}`;
+        } else {
+            operatorStuff = "NOT";
+        }
     }
-    return { fieldName, operator };
+    return { fieldName: stuff, operator: operatorStuff };
+}
+
+function isAPossibleNotOperator(
```
angrykoala: I see that operators are prepended with `NOT_` if they are in the list defined in this function. What is special about these operators? Maybe we could reflect that in the name of the function?

packages/graphql/src/translate/queryAST/ast/filters/utils/create-point-operation.ts
```diff
@@ -19,15 +19,15 @@
...
+import type { FilterOperator } from "../Filter";
 
 export function createPointOperation({
     operator,
     property,
     param,
     attribute,
 }: {
-    operator: WhereOperator | "EQ";
+    operator: FilterOperator;
```
angrykoala, line 30: Nice

packages/graphql/src/translate/queryAST/factory/OperationFactory.ts
```diff
@@ -403,6 +403,7 @@ export class OperationsFactory {
 
         if (resolveTree.name === "_entities" && reference) {
             const { __typename, ...referenceWhere } = reference;
+            console.log(`reference where: ${JSON.stringify(referenceWhere, null, 2)}`);
```
angrykoala: :eyes:

packages/graphql/tests/integration/deprecations/implicit-eq-filtering.test.ts
```diff
@@ -0,0 +1,193 @@
+/*
```
angrykoala: Is this file running on the tests? it is missing the `int.test.ts`

packages/graphql/tests/integration/deprecations/implicit-eq-filtering.test.ts
```diff
@@ -0,0 +1,193 @@
...
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
```
angrykoala: Is it worth adding a case for nested? just for completeness?

angrykoala (changes requested):

packages/graphql/src/schema/resolvers/subscriptions/where/utils/parse-filter-property.ts
MacondoExpress: 👀  Ops, my bad!

angrykoala (approved):

PR #5608, opened 2024-10-01
Deprecate options argument in favour of pagination 

MacondoExpress:
This PR deprecates the `options` argument in favour of `offset` / `limit` / `sort` arguments.

**Deprecated syntax**
```graphql
{
  movies(options: { offset: 1, limit: 10, sort: [{ id: DESC }] }) {
    id
    title
  }
}
```

**Current syntax**

```graphql
{
  movies(offset: 1, limit: 10, sort: [{ id: DESC }]) {
    id
    title
  }
}
```

This PR fixes also a bug in the schema where the sorting arguments for Interfaces and Union were a list of nullable elements instead of non nullable elements.


packages/graphql/src/schema/generation/sort-and-options-input.ts
```diff
@@ -41,7 +41,8 @@ export function withOptionsInputType({
     }
     // TODO: Concrete vs Abstract discrepancy
     // is this intended? For ConcreteEntity is NonNull, for InterfaceEntity is nullable
-    const sortFieldType = entityAdapter instanceof ConcreteEntityAdapter ? sortInput.NonNull.List : sortInput.List;
+    //const sortFieldType = entityAdapter instanceof ConcreteEntityAdapter ? sortInput.NonNull.List : sortInput.List;
```
angrykoala: Commented code

angrykoala (approved):

PR #5630, opened 2024-10-08
Remove deprecated update input

angrykoala:
# Description

Remove deprecated top level arguments for nested operations in mutations:

-   create
-   delete
-   connect
-   disconnect
-   connectOrCreate

For example, the following is no longer valid:

_invalid_

```graphql
mutation UpdatePeople {
    updatePeople(create: { movies: { node: { title: "The Good" } } }) {
        people {
            name
        }
    }
}
```

_valid_

```graphql
mutation UpdatePeople {
    updatePeople(update: { movies: { create: { node: { title: "The Good" } } } }) {
        people {
            name
        }
    }
}
```


packages/graphql/tests/e2e/subscriptions/delete-relationship.e2e.test.ts
```diff
@@ -474,301 +482,6 @@ subscription SubscriptionPerson {
         ]);
     });
 
-    test("disconnect via update - disconnect subscription sends events one way: interface type, by common field", async () => {
```
darrellwarde, line 477: Why the deletion over changing this test here?

darrellwarde (commented):
This looks good! I won't comment on every one, but any reason you opted to remove some of these tests instead of updating them?

packages/graphql/tests/e2e/subscriptions/delete-relationship.e2e.test.ts
angrykoala: this was already tested through the other path

angrykoala:
Lots of those are redundant, either because we had duplicate tests, or because now that we have a single path, the other was already tested.

PR #5634, opened 2024-10-08
Remove deprecated `bookmark` field

darrellwarde:
# Description

Removes the deprecated `bookmark` field.

## Complexity

Complexity: Medium (due to number of files)


packages/graphql/tests/integration/info.int.test.ts
```diff
@@ -80,44 +79,13 @@ describe("info", () => {
...
 
-        expect(typeof (gqlResult?.data as any)?.[Movie.operations.create].info.bookmark).toBe("string");
         expect((gqlResult?.data as any)?.[Movie.operations.create].info.nodesCreated).toBe(2);
         expect((gqlResult?.data as any)?.[Movie.operations.create].info.relationshipsCreated).toBe(1);
         expect((gqlResult?.data as any)?.[Movie.operations.create][Movie.plural]).toEqual([
             { title, actors: [{ name }] },
         ]);
     });
 
-    test("should return info from a delete mutation", async () => {
```
MacondoExpress: In this case, I would not remove the test itself I would return the remaining info fields.

darrellwarde: Good spot!

MacondoExpress (approved):

PR #5638, opened 2024-10-08
Bump minimum versions - Neo4j 5.s and Node.js 20.x

darrellwarde:
# Description

The Neo4j GraphQL Library now requires Neo4j 5.x and Node.js >=20.

## Complexity

Complexity: Low

### DIFF
--- a/.changeset/wild-knives-yawn.md
+++ b/.changeset/wild-knives-yawn.md
@@ -1,0 +1,6 @@
+---
+"@neo4j/graphql": major
+"@neo4j/graphql-ogm": major
+---
+
+The Neo4j GraphQL Library now requires Node.js 20 or greater.
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -22,7 +22,7 @@
         "dist/**/*.js.map"
     ],
     "engines": {
-        "node": ">=16.0.0"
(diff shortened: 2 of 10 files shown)


packages/graphql/package.json
```diff
@@ -22,7 +22,7 @@
         "dist/**/*.js.map"
     ],
     "engines": {
-        "node": ">=16.0.0"
+        "node": ">=20.0.0"
```
MacondoExpress, line 25: Shall we bump also the introspector/ogm packages?

darrellwarde: I have only adjusted what is here... 🫠 We don't currently have this in the `package.json` of the other packages! Presumably because the GraphQL Library is a dependency of them.

mjfwebb:
@darrellwarde can you please update the title/description to include mention of this PR also removing support for Neo4j 4.4?

.changeset/wild-knives-yawn.md
```diff
@@ -0,0 +1,6 @@
+---
+"@neo4j/graphql": major
+"@neo4j/graphql-ogm": major
```
MacondoExpress, line 3: It's also breaking for the introspector now!

MacondoExpress (approved):

PR #5639, opened 2024-10-08
Remove aggregation filters that are not aggregation filters

darrellwarde:
# Description

Remove aggregation filters that are not aggregation filters. These were deprecated in 5.x.

## Complexity

Complexity: Medium


packages/graphql/tests/integration/aggregations/where/AND-OR-operations.int.test.ts
```diff
@@ -174,274 +174,4 @@ describe("Nested within AND/OR", () => {
...
-                            {
-                                AND: [
-                                    {
-                                        node: {
-                                            testString_SHORTEST_LT: 4
-                                        }
-                                    }
-                                    {
-                                        node: {
-                                            testString_EQUAL: "${testString5}"
```
MacondoExpress, line 194: In this case, I would change the operator instead of deleting the whole test. This test assertion is not about this specific filter but about testing the logical operator with aggregation.

packages/graphql/tests/integration/aggregations/where/AND-OR-operations.int.test.ts
```diff
@@ -174,274 +174,4 @@ describe("Nested within AND/OR", () => {
...
-        expect(result.data).toEqual({
-            [postType.plural]: expect.toIncludeSameMembers([
-                {
-                    content: content5,
-                },
-            ]),
-        });
-    });
-
-    test("AND within an AND with NOT", async () => {
```
MacondoExpress: Same as before, I know it could sound superfluous, but these tests were very helpful during the refactor.

MacondoExpress (commented):
the setting `aggregationFilters` in `excludeDeprecatedFields` is still present in the file `src/types/index.ts`

MacondoExpress (approved):

darrellwarde:
> the setting `aggregationFilters` in `excludeDeprecatedFields` is still present in the file `src/types/index.ts`

Correct me if I'm wrong, but this one is still used for the aggregation filters without `LENGTH` that need removing. 🙂

MacondoExpress:
> > the setting `aggregationFilters` in `excludeDeprecatedFields` is still present in the file `src/types/index.ts`
> 
> Correct me if I'm wrong, but this one is still used for the aggregation filters without `LENGTH` that need removing. 🙂

You're absolutely right, forgot about these!

PR #5649, opened 2024-10-09
Remove `options` of `assertIndexesAndConstraints`

darrellwarde:
# Description

The deprecated `options` argument of `assertIndexesAndConstraints` has been removed. Database migrations are outside of the scope of the Neo4j GraphQL Library, and all indexes and constraints will have to be managed manually.

## Complexity

Complexity: Low


.changeset/lovely-toys-tan.md
```diff
@@ -0,0 +1,6 @@
+---
+"@neo4j/graphql": major
+"@neo4j/graphql-ogm": major
```
MacondoExpress: This is no longer needed right? as we don't have anymore OGM on v6, actually we should edit the previous changeset to reflect that!

darrellwarde: I already edited the changesets which were already merged, it'll just be the ones in PRs remaining!

MacondoExpress (approved):

PR #5662, opened 2024-10-14
Remove deprecated subscriptions

angrykoala:
# Description
Removes support for non-cdc subscriptions. This means the only available engine for subscriptions is `Neo4jGraphQLSubscriptionsCDCEngine`:

```ts
new Neo4jGraphQL({
    typeDefs,
    driver,
    features: {
        subscriptions: new Neo4jGraphQLSubscriptionsCDCEngine({
            driver,
        }),
    },
});
```

The default behaviour of subscriptions has also been updated to use CDC, so now passing `true` will use the CDC engine with the default parameters and driver:

```ts
new Neo4jGraphQL({
    typeDefs,
    driver,
    features: {
        subscriptions: true,
    },
});
```

----

Remove support for relationship subscriptions:

-   `*RelationshipCreated`
-   `*RelatiosnhipDeleted`

> A follow up PR will remove the deprecated code and update cypher


.changeset/curly-taxis-sleep.md
```diff
@@ -0,0 +1,8 @@
+---
+"@neo4j/graphql": major
+---
+
+Remove support for relationship subscriptions:
+
+-   `*RelationshipCreated`
+-   `*RelatiosnhipDeleted`
```
darrellwarde:
```suggestion
-   `*RelationshipDeleted`
```

.changeset/great-beans-sell.md
```diff
@@ -0,0 +1,29 @@
...
+Removes support for non-cdc subscriptions. This means the only available engine for subscriptions is `Neo4jGraphQLSubscriptionsCDCEngine`:
+
+```ts
+new Neo4jGraphQL({
+    typeDefs,
+    driver,
+    features: {
+        subscriptions: new Neo4jGraphQLSubscriptionsCDCEngine({
+            driver,
+        }),
```
darrellwarde, line 14: I think let's leave this like this like now for this PR, but after seeing this and the fact that literally the only thing that can be passed in is the CDC engine, I'm definitely camp "this should be turned into an object for configuration" - maybe a follow-up PR?

darrellwarde (approved):
The PR looks good, but I do think we should maybe re-discuss the API here?

.changeset/great-beans-sell.md
angrykoala: The one advantage of this is that it leaves the door open for capturing the events by extending this class, I know it is not the primary use case, but some users have been using the CDC subscriptions to capture these events for some external processing.

PR #5676, opened 2024-10-18
Reduce api library tests matrix

angrykoala:
# Description

This change reduce the api library tests workflow runs from 8 to 5


mjfwebb (approved):
Good stuff!

PR #5693, opened 2024-10-23
Fix #5671

### LINKED ISSUE #5671: Combining PUSH and POP in update mutations
While reading the docs about combining PUSH and POP in update mutations (https://neo4j.com/docs/graphql/5/mutations/update/#_combined_methods), the question arose if it's possible to combine a POP and a PUSH on the same array in a single mutation, for example to update an array ['a', 'b', 'c'] to ['a', 'b', 'd'] (tags array, see below) in a single mutation:

```
mutation {
    updateMovies (update: { tags_POP: 1, tags_PUSH: "d" }) {
        movies {
            title
            tags
        }
    }
}
```

the order is essential, since a PUSH followed by a POP would leave the array unchanged.
however, apparently no matter the order of the POP and PUSH operations inside "update", that is exactly what happens - the array doesn't change.

does PUSH take precedence? are the operations in "update" unordered?

is this a potential bug?

angrykoala:
# Description

Throws an error when the same field is updated multiple times on same update operation.

For example:

```graphql
mutation {
    updateMovies(update: { tags_POP: 1, tags_PUSH: "d" }) {
        movies {
            title
            tags
        }
    }
}
```

### DIFF
--- a/packages/graphql/src/translate/queryAST/factory/parsers/parse-mutation-field.ts
+++ b/packages/graphql/src/translate/queryAST/factory/parsers/parse-mutation-field.ts
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+type MutationOperation = "PUSH" | "POP" | "ADD" | "SUBTRACT" | "MULTIPLY" | "DIVIDE" | "INCREMENT" | "DECREMENT";
+
+export type MutationRegexGroups = {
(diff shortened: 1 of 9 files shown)


packages/graphql/src/translate/queryAST/factory/parsers/parse-mutation-field.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+const mutationRegEx =
+    /(?<fieldName>[_A-Za-z]\w*?)(?:_(?<operator>PUSH|POP|ADD|SUBTRACT|MULTIPLY|DIVIDE|INCREMENT|DECREMENT))?$/;
+
+export function parseMutationField(field: string): MutationRegexGroups {
+    const match = mutationRegEx.exec(field);
+
+    return match?.groups as MutationRegexGroups;
+}
```
darrellwarde, line 34: I know we don't do an awful lot of unit testing... But this is a prime example of where I think we should be, just absolutely rinsing a large number of field names through this function and ensuring that they parse as expected.

packages/graphql/tests/integration/array-methods/array-pop-errors.int.test.ts
```diff
@@ -253,12 +253,9 @@ describe("array-pop-errors", () => {
 
         const gqlResult = await testHelper.executeGraphQL(update);
 
-        expect(gqlResult.errors).toBeDefined();
-        expect(
-            (gqlResult.errors as GraphQLError[]).some((el) =>
-                el.message.includes("Cannot mutate the same field multiple times in one Mutation")
-            )
-        ).toBeTruthy();
```
darrellwarde, line 261: So this was already a thing? Or no?

darrellwarde (approved):
LGTM, just a small comment about our general sentiment around unit testing... 👀

PR #5694, opened 2024-10-23
Deprecate query direction options

MacondoExpress:
This PR deprecates `DEFAULT_DIRECTED` / `DEFAULT_UNDIRECTED` `DIRECTED_ONLY` / `UNDIRECTED_ONLY` as `@relationship.queryDirection` argument values. The options with the prefix `DEFAULT` are deprecated following the deprecation of the generated `directed` argument. The options with the suffix `_ONLY` have been changed to `DIRECTED` / `UNDIRECTED` as the suffix `_ONLY`.

This PR also adds warnings in case a deprecated value is used.

Less important changes:
- The algorithm to define which direction to use was modified (`getCypherDirection`, `getRelationshipDirection`).
- Redundant types are removed.


MacondoExpress:
I'm adding some integration test coverage, right now!

packages/graphql/tests/tck/undirected-relationships/query-direction-aggregations.test.ts
```diff
@@ -95,7 +95,7 @@ describe("QueryDirection in relationships aggregations", () => {
         expect(formatParams(result.params)).toMatchInlineSnapshot(`"{}"`);
     });
 
-    test("query with a UNDIRECTED_ONLY relationship", async () => {
+    test("query with a UNDIRECTED_ONLY (deprecated) relationship", async () => {
```
angrykoala: could it be worth moving these tests to a `deprecated` folder? so we remember to remove them later

angrykoala (approved):

packages/graphql/tests/tck/undirected-relationships/query-direction-aggregations.test.ts
MacondoExpress: Done!

PR #5703, opened 2024-10-24
Fix cypher sort be included 

### LINKED ISSUE #5631: Sorting schema should not be generated for cypher fields with NonNullable arguments
Given the following type definitions, schema is generated for sorting on the field `custom_string_with_param`. 

This should not occur as that field takes a NonNullable argument.

```graphql
type Movie {
    custom_string_with_param(param: String!): String!
        @cypher(statement: "RETURN $param as c", columnName: "c")
}
```

Excerpt from the schema generated:

```graphql
            \\"\\"\\"
            Fields to sort Movies by. The order in which sorts are applied is not guaranteed when specifying many fields in one MovieSort object.
            \\"\\"\\"
            input MovieSort {
              custom_string_with_param: SortDirection
            }
```

We should ensure that custom cypher fields with NonNullable arguments do not have sort schema generated.

MacondoExpress:
Fixed a bug that caused `@cypher` fields to be included as sortable fields in the generated even when this is not supported.

Fixes: #5631

### DIFF
--- a/.changeset/five-seals-itch.md
+++ b/.changeset/five-seals-itch.md
@@ -1,0 +1,5 @@
+---
+"@neo4j/graphql": major
+---
+
+Fixed a bug that caused `@cypher` field to be added as sortable fields in the schema even for not supported cases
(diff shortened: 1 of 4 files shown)


mjfwebb (approved):
Great!

PR #5733, opened 2024-10-30
Fix cdc tests on prem

angrykoala:
# Description

Tests for subscriptions are incorrectly skipped on prem. This fixes that

### DIFF
--- a/packages/graphql/tests/utils/tests-helper.ts
+++ b/packages/graphql/tests/utils/tests-helper.ts
@@ -98,6 +98,12 @@ export class TestHelper {
                 "CDC is note enable in test helper. Did you forget to set cdc:true or used isCDCEnabled by mistake?"
             );
         }
+
+        const dbInfo = await this.getDatabaseInfo();
+        if (!dbInfo.isAura()) {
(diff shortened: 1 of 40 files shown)


packages/graphql/tests/utils/tests-helper.ts
```diff
@@ -92,15 +91,21 @@ export class TestHelper {
         return this.neo4jGraphQL;
     }
 
-    public async isCDCEnabled(): Promise<boolean> {
+    public async assertCDCEnabled(): Promise<boolean> {
```
MacondoExpress, line 94: Not important for now: But I would not run an `ALTER DATABASE` inside a method called `isCDCEnabled` 

edit: I notice now is changed to `assertCDCEnabled` however this method could return false for Aura instances.

Not a blocker for the release IMO anyway!

MacondoExpress (approved):

packages/graphql/tests/utils/tests-helper.ts
angrykoala: we need to improve the ergonomics, I called assert for the reason uou mention, I just want to unblock release atm, later we can chat on improving this

PR #5755, opened 2024-11-04
Remove unique

angrykoala:
# Description
Remove `@unique` directive and update tests

Some code is still leftover from `connectOrCreate`, which is implicitly removed by this PR


packages/graphql/src/schema-model/entity/model-adapters/ConcreteEntityAdapter.ts
```diff
@@ -80,9 +79,6 @@ export class ConcreteEntityAdapter {
 
             if (attributeAdapter.isConstrainable()) {
                 this.constrainableFieldsKeys.push(attribute.name);
-                if (attributeAdapter.isUnique()) {
```
MacondoExpress, line 83: I believe the whole logic could `isConstrainable` can be removed from both here and Interfaces,  could you confirm this?

packages/graphql/src/schema-model/entity/model-adapters/InterfaceEntityAdapter.ts
```diff
@@ -112,10 +111,6 @@ export class InterfaceEntityAdapter {
      * used to generate different types for the Entity that contains these Attributes
      */
 
-    public get uniqueFields(): AttributeAdapter[] {
-        return this.uniqueFieldsKeys.map((key) => getFromMap(this.attributes, key));
```
MacondoExpress, line 116: `uniqueFieldsKeys` attribute has to be removed as well

MacondoExpress (commented):
a couple of leftover related to the the `@unique` fields but apart from that LGTM!

MacondoExpress (approved):

PR #5777, opened 2024-11-11
Remove deprecated `options` argument

darrellwarde:
# Description

The deprecated `options` argument has been removed.

Consider the following type definitions:

```graphql
type Movie {
  title: String!
}
```

The migration is as below:

```graphql
# Old syntax
{
  movies(options: { first: 10, offset: 10, sort: [{ title: ASC }] }) {
    title
  }
}

# New syntax
{
  movies(first: 10, offset: 10, sort: [{ title: ASC }]) {
    title
  }
}
```

## Complexity

Complexity: Low


packages/graphql/src/translate/queryAST/factory/Operations/ReadFactory.ts
```diff
@@ -189,17 +180,10 @@ export class ReadFactory {
...
-        raiseOnMixedPagination({
-            optionsArg,
-            sort: resolveTree.args.sort,
-            limit: resolveTree.args.limit,
-            offset: resolveTree.args.offset,
-        });
+        const sortArg = resolveTree.args.sort as Record<string, any>[];
+        const limitArg = resolveTree.args.limit as number;
+        const offsetArg = resolveTree.args.offset as number;
+
```
MacondoExpress, line 186: the block below is not longer necessary

packages/graphql/src/types/index.ts
```diff
@@ -231,7 +231,7 @@ export interface ConnectionQueryArgs {
  * Representation of the options arg
  * passed to resolvers.
  */
-export interface GraphQLOptionsArg {
+export interface GraphQLSortingAndPaginationArgs {
```
MacondoExpress, line 234: 😍

MacondoExpress (commented):
There is an unreachable code block, apart from that it looks good!

packages/graphql/src/translate/queryAST/factory/Operations/ReadFactory.ts
darrellwarde: I think it may be! Despite being called `getOptions`, I still think it deals with applying sorting and pagination. Otherwise this function is literally just assigning values to three variables and then returning nothing! 😆

MacondoExpress: ah right! 🤦‍♂️

MacondoExpress (approved):
LGTM!

PR #5781, opened 2024-11-12
Configure jest to use prettier 2

MacondoExpress:
Configure Jest to use Prettier 2 as Prettier 3 is not yet supported.

### DIFF
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -65,6 +65,7 @@
         "koa-router": "13.0.1",
         "mock-jwks": "1.0.10",
         "nock": "13.5.6",
+        "prettier-2": "npm:prettier@^2",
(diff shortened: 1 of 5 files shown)


packages/graphql/package.json
```diff
@@ -65,6 +65,7 @@
         "koa-router": "13.0.1",
         "mock-jwks": "1.0.10",
         "nock": "13.5.6",
+        "prettier-2": "npm:prettier@^2",
```
mjfwebb: We tend to specify versions here instead of `^2`, I think?

MacondoExpress: right, moved also to the global repository settings, as we may need the same fix for other packages!

mjfwebb (approved):

PR #5782, opened 2024-11-12
feat(renovate): add groups to separate major and minor PRs

mjfwebb:
This PR updates renovate.json so that Renovate will create a single PR for minor and patch dependency updates. Major dependency updates will still be created with individual PRs.


renovate.json
```diff
@@ -18,11 +21,23 @@
         "**/__fixtures__/**"
     ],
     "ignoreDeps": ["neo4j"],
+    "separateMajorMinor": true,
+    "groupName": "all",
```
darrellwarde: What's the deal with `groupName` out here as well as inside `packageRules`?

mjfwebb: yeah that's probably not necessary! I looked at this with Liam and we both thought that it was needed but I can't find it written in their docs anywhere today. I'll remove the top level one.

renovate.json
```diff
@@ -18,7 +19,18 @@
         "**/__fixtures__/**"
     ],
     "ignoreDeps": ["neo4j"],
+    "separateMajorMinor": true,
```
darrellwarde, line 22: This is true by default I believe! https://docs.renovatebot.com/configuration-options/#separatemajorminor

Not that we need to remove it, but just a heads up.

darrellwarde (approved):

PR #5788, opened 2024-11-13
deprecate single element relationships

angrykoala:
# Description

Deprecates single item relationships

### DIFF
--- a/.changeset/chilled-zebras-help.md
+++ b/.changeset/chilled-zebras-help.md
@@ -17,3 +17,5 @@ type Movie {
     director: [Person!]! @relationship(type: "DIRECTED", direction: "IN")
 }
 ```
+
+1-1 relationships cannot be reliably enforced, leading to a data inconsistent between the schema and the database. For this reason, these have been removed in favor of the more accurate list relationships.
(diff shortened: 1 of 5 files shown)


.changeset/chilled-zebras-help.md
```diff
@@ -0,0 +1,19 @@
+---
+"@neo4j/graphql": patch
+---
+
+Deprecate single element relationships:
```
darrellwarde, line 5: I think this changeset is by far not long enough - this needs a lot of explanation as to the rationale IMO!

darrellwarde (dismissed):
I think looks good, but consider my comment?

.changeset/chilled-zebras-help.md
angrykoala: Isn't that explanation better left for the migration guide?

darrellwarde: I dunno - I know this isn't the breaking change itself but this is what changesets outputs:

```
🦋  warn This Changeset includes a major change and we STRONGLY recommend adding more information to the changeset:
🦋  warn WHAT the breaking change is
🦋  warn WHY the change was made
🦋  warn HOW a consumer should update their code
```

PR #5819, opened 2024-11-19
Remove single relationships

angrykoala:
# Description

Removes 1-1 relationships


packages/graphql/src/schema/create-relationship-fields/create-relationship-fields.ts
```diff
@@ -176,6 +176,7 @@ export function createRelationshipFields({
     userDefinedFieldDirectivesForNode: Map<string, Map<string, DirectiveNode[]>>;
     features?: Neo4jFeaturesSettings;
 }): void {
+    // HERE?
```
MacondoExpress:
```suggestion
```

packages/graphql/src/schema/validation/custom-rules/warnings/single-relationship.ts
```diff
@@ -1,60 +0,0 @@
...
-                for (const directive of field.directives ?? []) {
-                    if (directive.name.value === relationshipDirective.name) {
-                        isRelationship = true;
-                    }
-                }
-
-                const isList = Boolean(getListTypeNode(field));
-
-                if (isRelationship && !isList) {
-                    console.warn(
```
MacondoExpress, line 40: I would change this to be a validation rule, so that an error is raised before reaching make-augmented-schema

packages/graphql/tests/integration/directives/authorization/roles.int.test.ts
```diff
@@ -538,93 +538,6 @@ describe("auth/roles", () => {
...
-                type ${typeComment} @node {
-                    id: String
-                    content: String
-                    post: ${typePost}! @relationship(type: "HAS_COMMENT", direction: IN)
-                }
-
-                type ${typePost} @node {
-                    id: String
-                    content: String
-                    creator: ${typeUser}! @relationship(type: "HAS_POST", direction: IN)
```
MacondoExpress, line 557: Is this test no longer necessary?

packages/graphql/tests/integration/directives/authorization/roles.int.test.ts
```diff
@@ -698,92 +611,6 @@ describe("auth/roles", () => {
...
```
MacondoExpress, line 717: Same as above, is there is any reason to delete this test rather than change the `creator` type

MacondoExpress (approved):
Lovely! Just left some comments

PR #5820, opened 2024-11-19
Make `@node` mandatory to generate Queries and Mutations.

MacondoExpress:
Change the way how `@node` behaves, `@node` is now required, and GraphQL Object types without the directive `@node` will no longer considered as a Neo4j Nodes representation. Queries and Mutations will be generated only for types with the `@node` directive.


angrykoala (approved):

packages/graphql/src/schema-model/parser/parse-attribute.ts
```diff
@@ -115,7 +115,8 @@ function parseTypeNode(
                 return new InterfaceType(typeNode.name.value, isRequired);
             } else if (isInput(definitionCollection, typeNode.name.value)) {
                 return new InputType(typeNode.name.value, isRequired);
-            } else {
+            }
+            else {
```
mjfwebb:
```suggestion
            } else {
```

packages/apollo-federation-subgraph-compatibility/src/type-defs.ts
```diff
@@ -42,7 +42,7 @@ export const typeDefs = gql`
 
     directive @custom on OBJECT
 
-    type Product @custom @key(fields: "id") @key(fields: "sku package") @key(fields: "sku variation { id }") {
+    type Product @node @custom @key(fields: "id") @key(fields: "sku package") @key(fields: "sku variation { id }") {
```
MacondoExpress, line 45: @darrellwarde can you verify the changes made here?

darrellwarde: No? 😆 It looks fine to me! It's a node in our database right?!

packages/graphql/tests/schema/issues/1575.test.ts
```diff
@@ -78,15 +78,6 @@ describe("https://github.com/neo4j/graphql/issues/1575", () => {
               node: Foo!
             }
 
-            input FooOptions {
```
darrellwarde: I don't know how this ended up in here?

darrellwarde (approved):
LGTM!!

packages/apollo-federation-subgraph-compatibility/src/type-defs.ts
MacondoExpress: It is!

PR #5834, opened 2024-11-22
Filter cdc events skipping nodes not present in the schema

### LINKED ISSUE #5787: `Neo4jGraphQLSubscriptionsCDCEngine` with enabled CDC@Aura kills the neo4j-graphql server whenever massive data changes happen within a short period of time
**Describe the bug**
Since API v6 the `Neo4jGraphQLSubscriptionsDefaultEngine` is deprecated, so we replaced it with the only current available one and enabled CDC with mode=`DIFF` at our Aura instance.
This works totally fine as long as there are no massive changes within the database.
But: we have scheduled tasks that import data into our aura instance every day. The amount of data varies, but it is not uncommon that millions of nodes are (re-)inserted into the database.
This causes a **LOT** of CDC events which propagate back to our neo4j-graphQL server - **which dies and shuts down**. This even happens at my local machine where a lot of CPU and RAM is available, so this seems not to be a bottleneck here.
I tried different configurations for `pollTime` reaching from 100ms to 5000ms but this seems to have no effect on the problem here - the server dies either way.

The thing is: we **do not need** the CDC events from the database. We only need subscriptions for change events happening within the neo4j-graphQL server - exactly what `Neo4jGraphQLSubscriptionsDefaultEngine` was doing. Since that was dropped, the only current working solution for us now is to disable subscriptions in total.

On the other hand we have some applications which rely on the subscription-functionality - these apps do not work anymore.

We definitely do NOT want to stay on v5 since we aim to be up-to-date with the API at all time.

So, what are our options?
Do you see a way to re-enable something like `Neo4jGraphQLSubscriptionsDefaultEngine` again in API v7?
Can we somehow mimic the behaviour of this engine by ourselves and pass it to `Neo4jGraphQL`?

By the way: to enable CDC in our situation feels really bad since it massively blows up the transaction log with CDC events no one needs.

angrykoala:
# Description

CDC subscription optimization. Only node events with labels present in the GraphQL schema will be queried. This will reduce the number of subscription events queried by skipping events to nodes that cannot be subscribed through GraphQL

### DIFF
--- a/.changeset/tasty-onions-laugh.md
+++ b/.changeset/tasty-onions-laugh.md
@@ -1,0 +1,5 @@
+---
+"@neo4j/graphql": patch
+---
+
+CDC subscription optimization. Only node events with labels present in the GraphQL schema will be queried. This will reduce the number of subscription events queried by skipping events to nodes that cannot be subscribed through GraphQL
--- a/packages/graphql/src/classes/subscription/Neo4jGraphQLSubscriptionsCDCEngine.ts
+++ b/packages/graphql/src/classes/subscription/Neo4jGraphQLSubscriptionsCDCEngine.ts
@@ -20,6 +20,7 @@
 import { EventEmitter } from "events";
 import type { Driver, QueryConfig } from "neo4j-driver";
 import { Memoize } from "typescript-memoize";
+import type { Neo4jGraphQLSchemaModel } from "../../schema-model/Neo4jGraphQLSchemaModel";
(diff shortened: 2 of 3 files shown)


darrellwarde (commented):
I think we should have some tests for this perhaps?

angrykoala:
I'm not sure if this is really testable without lots of mocks, as it has to be tested e2e (using a real database) but the external behaviour of a service running this hasn't changed

PR #5835, opened 2024-11-22
Add option`onlyGraphQLEvents` for subscriptions

### LINKED ISSUE #5787: `Neo4jGraphQLSubscriptionsCDCEngine` with enabled CDC@Aura kills the neo4j-graphql server whenever massive data changes happen within a short period of time

angrykoala:
# Description

Adds the option `filterGraphQLEvents` to only trigger CDC on GraphQL events

### DIFF
--- a/packages/graphql/src/classes/subscription/Neo4jGraphQLSubscriptionsCDCEngine.ts
+++ b/packages/graphql/src/classes/subscription/Neo4jGraphQLSubscriptionsCDCEngine.ts
@@ -21,7 +21,7 @@ import Cypher from "@neo4j/cypher-builder";
 import { EventEmitter } from "events";
 import type { Driver, QueryConfig } from "neo4j-driver";
 import { Memoize } from "typescript-memoize";
-import environment from "../../environment";
+import { APP_ID } from "../../constants";
 import type { Neo4jGraphQLSchemaModel } from "../../schema-model/Neo4jGraphQLSchemaModel";
 import type { Neo4jGraphQLSubscriptionsEngine, SubscriptionEngineContext, SubscriptionsEvent } from "../../types";
 import { CDCApi } from "./cdc/cdc-api";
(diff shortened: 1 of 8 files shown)


packages/graphql/src/classes/subscription/Neo4jGraphQLSubscriptionsCDCEngine.ts
```diff
@@ -97,7 +103,16 @@ export class Neo4jGraphQLSubscriptionsCDCEngine implements Neo4jGraphQLSubscript
     }
 
     private async pollEvents(): Promise<void> {
-        const cdcEvents = await this.cdcApi.queryEvents(this.subscribeToLabels);
+        let txFilter: Cypher.Map | undefined;
+        if (this.filterGraphQLEvents) {
+            const app = `${environment.NPM_PACKAGE_NAME}@${environment.NPM_PACKAGE_VERSION}`;
```
darrellwarde: This is a value used elsewhere, so we should have it as a constant somewhere to ensure these don't fall out of sync.

packages/graphql/src/classes/subscription/Neo4jGraphQLSubscriptionsCDCEngine.ts
```diff
@@ -35,18 +37,22 @@ export class Neo4jGraphQLSubscriptionsCDCEngine implements Neo4jGraphQLSubscript
     private closed = false;
 
     private subscribeToLabels: string[] | undefined;
+    private filterGraphQLEvents: boolean;
 
     constructor({
         driver,
         pollTime = 1000,
         queryConfig,
+        filterGraphQLEvents = false,
```
darrellwarde: I think this nomenclature is awkward - when this is true it... Filters _out_ GraphQL events?

`onlyGraphQLEvents`?

darrellwarde (dismissed):

PR #5837, opened 2024-11-22
Add validation rules to disable nullable elements in a list, when used inside `@node` or `@relationshipProperties`.

MacondoExpress:
As the title, this PR adds a validation rule to avoid a list of nullable elements present in the `@node` or `@relationshipProperties` types as they are not natively supported in Neo4j.

Test: 3905.test.ts is skipped as`@cypher` fields with non-`@node` target are currently not implemented in the 7.x branch


packages/graphql/src/schema/validation/custom-rules/valid-types/valid-list-in-node-type.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Validates that list types used in type annotated with the node directive are supported by Neo4j
+ **/
+export function ValidListInNodeType(context: SDLValidationContext): ASTVisitor {
+    return {
+        ObjectTypeDefinition(objectTypeDefinitionNode: ObjectTypeDefinitionNode, _key, _parent) {
+            const { directives } = objectTypeDefinitionNode;
+
+            const nodeUsage = directives?.find((directive) => directive.name.value === nodeDirective.name);
```
angrykoala, line 40: `nodeUsage` is confusing. Maybe `isNode`?

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-list-in-node-type.ts
```diff
@@ -0,0 +1,116 @@
...
+            const relationshipPropertiesUsage = directives?.find(
```
angrykoala, line 41: Same here

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-list-in-node-type.ts
```diff
@@ -0,0 +1,116 @@
...
+                (directive) => directive.name.value === relationshipPropertiesDirective.name
+            );
+            if (!directives || (!nodeUsage && !relationshipPropertiesUsage)) {
+                return; // Skip non-node types or relationshipProperties types
+            }
```
angrykoala: I think this is a bit more readable as 2 separate ifs. i.e. two cases we want to return

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-list-in-node-type.ts
```diff
@@ -0,0 +1,116 @@
...
+                            errorMsg,
+                        })
+                    );
+                }
+            });
+        },
+    };
+}
+
+type TypePath = (string | Kind.LIST_TYPE | Kind.NON_NULL_TYPE)[];
```
angrykoala: Nitpick (no need to change it) I find the array syntax less confusing when dealing with nested types:

```suggestion
type TypePath = Array<string | Kind.LIST_TYPE | Kind.NON_NULL_TYPE>;
```

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-list-in-node-type.ts
```diff
@@ -0,0 +1,116 @@
...
+
+function typeNodeToString(typeNode: TypeNode): string {
+    if (typeNode.kind === Kind.NON_NULL_TYPE) {
+        return `${typeNodeToString(typeNode.type)}!`;
+    } else if (typeNode.kind === Kind.LIST_TYPE) {
+        return `[${typeNodeToString(typeNode.type)}]`;
+    }
+    return typeNode.name.value;
+}
```
angrykoala: Doesn't this already exists somewhere? Could we reuse that utility or it is too much of a different context?

angrykoala (approved):

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-list-in-node-type.ts
MacondoExpress: The reason why I appended `Usage` is to hint that is not used as a boolean, as later it is used to collect the name of the directive.
I'll accept any suggestion here!

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-list-in-node-type.ts
MacondoExpress: You're right we have an utility for that!

PR #5850, opened 2024-11-27
Add point and time filters on schema

angrykoala:
# Description

Point and Time filters

### DIFF
--- a/packages/graphql/src/graphql/input-objects/generic-operators/BooleanScalarFilters.ts
+++ b/packages/graphql/src/graphql/input-objects/generic-operators/BooleanScalarFilters.ts
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-import { GraphQLBoolean, GraphQLInputObjectType } from "graphql";
+import { GraphQLBoolean, GraphQLInputObjectType, GraphQLList, GraphQLNonNull } from "graphql";
 
 export const BooleanScalarFilters = new GraphQLInputObjectType({
     name: "BooleanScalarFilters",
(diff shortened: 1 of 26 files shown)


packages/graphql/src/graphql/input-objects/generic-operators/BooleanScalarFilters.ts
```diff
@@ -28,3 +28,12 @@ export const BooleanScalarFilters = new GraphQLInputObjectType({
         },
     },
 });
+
+export const BooleanListFilters = new GraphQLInputObjectType({
```
MacondoExpress, line 32: Do you know if this is something we currently support? if not we can keep it out of the scope for now!

packages/graphql/src/graphql/input-objects/generic-operators/DateScalarFilters.ts
```diff
@@ -34,3 +34,12 @@ export const DateScalarFilters = new GraphQLInputObjectType({
         lessThanEquals: { type: GraphQLDate },
     },
 });
+
+export const DateListFilters = new GraphQLInputObjectType({
```
MacondoExpress, line 38: As the difference between these ListFilters is only the type, maybe we can generate them rather than write them manually. But, I'm not suggesting, just a thought!

MacondoExpress (approved):
LGTM!

packages/graphql/src/graphql/input-objects/generic-operators/DateScalarFilters.ts
angrykoala: I though about that, but realistically, we would still need to call the generators with the relevant information, so while it makes it a bit less verbose it won't be for much without adding complexity

packages/graphql/src/graphql/input-objects/generic-operators/BooleanScalarFilters.ts
angrykoala: Good point

angrykoala: Actually, we do support these on Cypher fields, but only equals, so I'll update this

PR #5856, opened 2024-11-28
(6.x) Cypher filtering many to many relationships

mjfwebb:
# Description

This PR adds the ability to filter on many to many (m:m) relationship cypher fields. An example:

Type definitions:
```gql
type Movie @node {
    title: String
    actors: [Actor!]!
        @cypher(
            statement: """
            MATCH (this)<-[:ACTED_IN]-(actor:Actor)
            RETURN actor
            """
            columnName: "actor"
        )
}

type Actor @node {
    name: String
    movies: [Movie!]!
        @cypher(
            statement: """
            MATCH (this)-[:ACTED_IN]->(movie:Movie)
            RETURN movie
            """
            columnName: "movie"
        )
}
```

You can now query with a filter based on the `actors` field:
```gql
query {
  movies(where: { actors: { name: "Jada Pinkett Smith" } }) {
      title
  }
}
```

## Complexity

Medium

# Issue

Closes #554.

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [ ] Documentation has been updated
- [x] TCK tests have been updated
- [x] Integration tests have been updated
- [ ] Example applications have been updated
- [x] New files have copyright header
- [x] CLA (https://neo4j.com/developer/cla/) has been signed

PR #5858, opened 2024-11-29
(7.x) Cypher filtering many to many relationships

mjfwebb:
As https://github.com/neo4j/graphql/pull/5856 but for 7.x


packages/graphql/tests/schema/interfaces.test.ts
```diff
@@ -331,6 +331,11 @@ describe("Interfaces", () => {
               AND: [MovieWhere!]
               NOT: MovieWhere
               OR: [MovieWhere!]
+              customQuery: MovieWhere
```
MacondoExpress: This seems wrong to me

MacondoExpress (approved):

PR #5865, opened 2024-12-03
Fix `@default` so it works for temporal types and `BigInt` fields

### LINKED ISSUE #5436: Invalid `@default` validation rules
**Describe the bug**

Current `@default` validation does not provide the correct error message for some typeDefs.

For instance:
`Time` issue:
```graphql
type typeNode {
  time: Time @default(value: "10:00:00")
}
```
Raise the following error: "@default.value is not a valid Time" with a correct `Time` value. 
The error in this case should be that `Time` is not supported.

DateTime case:
`DateTime` is a supported type, however, no validation rules happen for a List of `DateTime` values.
```graphql
type typeNode {
  dateTimes: [DateTime] @default(value: ["dummy"])
}
```
The above, in fact, does not raise any error while `dateTime: DateTime @default(value: "dummy")` raise `@default.value is not a valid DateTime`.

Float case:
The `Float` type is supported, however, when not specifying the fraction part it raises an error:
```graphql
type typeNode {
  float: Float @default(value: 1)
}
```
Raise the error: "@default.value on Float fields must be of type Float".
However, this is different from the GraphQL input specification:
https://spec.graphql.org/October2021/#sel-GAHXTHDCAACEB68G

* OS: MacOS
* Version: @neo4j/graphql@5.5.2
* Node.js version: 20

darrellwarde:
# Description

Fix `@default` so it works for temporal types and `BigInt` fields

## Complexity

Complexity: Low

# Issue

Closes #5436

### DIFF
--- a/packages/graphql/src/schema/get-obj-field-meta.ts
+++ b/packages/graphql/src/schema/get-obj-field-meta.ts
@@ -439,7 +439,7 @@ function getObjFieldMeta({
                                 primitiveField.defaultValue = parseInt(value.value, 10);
                                 break;
                             case "Float":
-                                if (value?.kind !== Kind.FLOAT) {
+                                if (value?.kind !== Kind.FLOAT && value?.kind !== Kind.INT) {
                                     throw new Error(typeError);
                                 }
                                 primitiveField.defaultValue = parseFloat(value.value);
--- a/packages/graphql/src/schema/get-obj-field-meta.ts
+++ b/packages/graphql/src/schema/get-obj-field-meta.ts
@@ -432,6 +432,12 @@ function getObjFieldMeta({
                                 }
                                 primitiveField.defaultValue = parseInt(value.value, 10);
                                 break;
+                            case "BigInt":
+                                if (value?.kind !== Kind.INT && value?.kind !== Kind.STRING) {
+                                    throw new Error(typeError);
+                                }
+                                primitiveField.defaultValue = parseInt(value.value, 10);
+                                break;
                             case "Float":
                                 if (value?.kind !== Kind.FLOAT) {
                                     throw new Error(typeError);
@@ -440,7 +446,7 @@ function getObjFieldMeta({
                                 break;
                             default:
                                 throw new Error(
-                                    "@default directive can only be used on types: Int | Float | String | Boolean | ID | DateTime | Enum"
+                                    "@default directive can only be used on fields of type Int, Float, String, Boolean, ID, BigInt, DateTime, Date, Time, LocalDateTime or LocalTime."
                                 );
                         }
                     }
--- a/packages/graphql/src/schema/validation/custom-rules/directives/default.ts
+++ b/packages/graphql/src/schema/validation/custom-rules/directives/default.ts
@@ -16,13 +16,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { DirectiveNode, FieldDefinitionNode, EnumTypeDefinitionNode, StringValueNode } from "graphql";
+import type { DirectiveNode, EnumTypeDefinitionNode, FieldDefinitionNode, StringValueNode } from "graphql";
 import { Kind } from "graphql";
+import { GRAPHQL_BUILTIN_SCALAR_TYPES } from "../../../../constants";
+import { parseLocalTime } from "../../../../graphql/scalars/LocalTime";
+import { parseTime } from "../../../../graphql/scalars/Time";
+import { DocumentValidationError } from "../utils/document-validation-error";
+import type { ObjectOrInterfaceWithExtensions } from "../utils/path-parser";
 import { assertArgumentHasSameTypeAsField } from "../utils/same-type-argument-as-field";
 import { getInnerTypeName, isArrayType } from "../utils/utils";
-import { DocumentValidationError } from "../utils/document-validation-error";
-import { GRAPHQL_BUILTIN_SCALAR_TYPES, isSpatial, isTemporal } from "../../../../constants";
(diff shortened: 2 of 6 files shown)


packages/graphql/src/schema/validation/custom-rules/directives/default.ts
```diff
@@ -48,21 +50,38 @@ export function verifyDefault(enums: EnumTypeDefinitionNode[]) {
         }
 
         if (!isArrayType(traversedDef)) {
-            if (isSpatial(expectedType)) {
-                throw new DocumentValidationError(`@default is not supported by Spatial types.`, ["value"]);
-            } else if (isTemporal(expectedType)) {
+            if (["DateTime", "LocalDateTime", "Date"].includes(expectedType)) {
```
MacondoExpress: Nitpick: IMO `[GraphQLDateTime.name, GraphQLLocalDateTime.name, GraphQLDate.name]`  is better to avoid potential typos.

packages/graphql/src/schema/validation/custom-rules/directives/default.ts
```diff
@@ -48,21 +50,38 @@ export function verifyDefault(enums: EnumTypeDefinitionNode[]) {
...
                 if (Number.isNaN(Date.parse((defaultArg?.value as StringValueNode).value))) {
                     throw new DocumentValidationError(
                         `@default.${defaultArg.name.value} is not a valid ${expectedType}`,
                         ["value"]
                     );
                 }
+            } else if (expectedType == "Time") {
```
MacondoExpress: Nitpick:
```suggestion
            } else if (expectedType === "Time") {
```

packages/graphql/src/schema/validation/custom-rules/directives/default.ts
```diff
@@ -48,21 +50,38 @@ export function verifyDefault(enums: EnumTypeDefinitionNode[]) {
...
+                try {
+                    parseTime((defaultArg?.value as StringValueNode).value);
+                } catch {
+                    throw new DocumentValidationError(
+                        `@default.${defaultArg.name.value} is not a valid ${expectedType}`,
+                        ["value"]
+                    );
+                }
+            } else if (expectedType == "LocalTime") {
```
MacondoExpress: Nitpick:
```suggestion
            } else if (expectedType === "LocalTime") {
```

packages/graphql/src/schema/validation/custom-rules/directives/default.ts
```diff
@@ -48,21 +50,38 @@ export function verifyDefault(enums: EnumTypeDefinitionNode[]) {
...
+                    throw new DocumentValidationError(
+                        `@default.${defaultArg.name.value} is not a valid ${expectedType}`,
+                        ["value"]
+                    );
+                }
             } else if (
                 !GRAPHQL_BUILTIN_SCALAR_TYPES.includes(expectedType) &&
-                !enums.some((x) => x.name.value === expectedType)
+                !enums.some((x) => x.name.value === expectedType) &&
+                expectedType != "BigInt"
```
MacondoExpress: Nitpick: 
```suggestion
                expectedType !== "BigInt"
```

packages/graphql/src/schema/validation/validate-document.test.ts
```diff
@@ -1215,6 +1215,74 @@ describe("validation 2.0", () => {
...
+                    validateDocument({
+                        document: doc,
+                        additionalDefinitions,
+                        features: {},
+                    });
+
+                expect(executeValidate).not.toThrow();
+            });
+
+            test("@default on Time must be valid datetime correct", () => {
```
MacondoExpress: I would change the test titles as they all ending with `must be valid datetime correct`

packages/graphql/src/schema/validation/validate-document.test.ts
```diff
@@ -1215,6 +1215,74 @@ describe("validation 2.0", () => {
...
+                const executeValidate = () =>
+                    validateDocument({
+                        document: doc,
+                        additionalDefinitions,
+                        features: {},
+                    });
+
+                expect(executeValidate).not.toThrow();
+            });
+
```
MacondoExpress, line 1285: Nitpick: It may worth to add a test with `BigInt` as the other types are tested!

packages/graphql/src/schema/get-obj-field-meta.ts
```diff
@@ -432,6 +432,12 @@ function getObjFieldMeta({
...
                                 primitiveField.defaultValue = parseInt(value.value, 10);
                                 break;
+                            case "BigInt":
+                                if (value?.kind !== Kind.INT && value?.kind !== Kind.STRING) {
+                                    throw new Error(typeError);
+                                }
+                                primitiveField.defaultValue = parseInt(value.value, 10);
+                                break;
                             case "Float":
                                 if (value?.kind !== Kind.FLOAT) {
```
MacondoExpress: Unrelated to this PR, but I believe here we should accept also `INT`
```suggestion
                                if (value?.kind !== Kind.FLOAT || value?.kind !== Kind.INT) {
```

MacondoExpress (commented):
LGTM, left some comments!

packages/graphql/src/schema/validation/custom-rules/directives/default.ts
darrellwarde: Too much go 🥲

packages/graphql/src/schema/get-obj-field-meta.ts
darrellwarde: Unrelated, but just done and integration test added!

PR #5868, opened 2024-12-05
Generic mutation operators

angrykoala:
# Description

Add suport for generic update operators

### DIFF
--- a/.changeset/green-jobs-jam.md
+++ b/.changeset/green-jobs-jam.md
@@ -2,4 +2,15 @@
 "@neo4j/graphql": minor
 ---
 
-Add suport for generic update operators
+Add suport for generic update operators:
+
+```graphql
+mutation {
+    updateMovies(update: { name: { set: "The Matrix" } }) {
+        movies {
+            id
+            name
+        }
+    }
+}
+```
--- a/.changeset/green-jobs-jam.md
+++ b/.changeset/green-jobs-jam.md
@@ -1,0 +1,5 @@
+---
+"@neo4j/graphql": minor
+---
+
+Add suport for generic update operators
--- a/packages/graphql/src/schema/generation/get-mutation-input-from-attribute-type.ts
+++ b/packages/graphql/src/schema/generation/get-mutation-input-from-attribute-type.ts
@@ -122,13 +122,13 @@ export function getMutationInputFromAttributeType(attribute: AttributeAdapter): 
         return DateScalarMutations;
     }
 
-    // if (attribute.typeHelper.isEnum()) {
-    //     return `${attribute.getTypeName()}EnumScalarFilters`;
-    // }
+    if (attribute.typeHelper.isEnum()) {
+        return `${attribute.getTypeName()}EnumScalarMutations`;
+    }
 
-    // if (attribute.typeHelper.isUserScalar()) {
-    //     return `${attribute.getTypeName()}ScalarFilters`;
-    // }
+    if (attribute.typeHelper.isUserScalar()) {
+        return `${attribute.getTypeName()}ScalarMutations`;
+    }
 
     throw new Error(`No scalar mutation found for attribute ${attribute.type.name}`);
 }
--- a/packages/graphql/src/schema/generation/get-mutation-input-from-attribute-type.ts
+++ b/packages/graphql/src/schema/generation/get-mutation-input-from-attribute-type.ts
@@ -18,107 +18,149 @@
  */
 
 import type { GraphQLInputType } from "graphql";
-import { BigIntScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/BigIntScalarMutations";
-import { BooleanScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/BooleanScalarMutations";
-import { CartesianPointMutations } from "../../graphql/input-objects/generic-mutation-operations/CartesianPointMutations";
-import { DateScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/DateScalarMutations";
-import { DateTimeScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/DateTimeScalarMutations";
-import { DurationScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/DurationScalarMutations";
-import { FloatScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/FloatScalarMutations";
-import { IDScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/IDScalarMutations";
-import { IntScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/IntScalarMutations";
-import { LocalDateTimeScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/LocalDateTimeScalarMutations";
-import { LocalTimeScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/LocalTimeScalarMutations";
-import { PointMutations } from "../../graphql/input-objects/generic-mutation-operations/PointMutations";
-import { StringScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/StringScalarMutations";
-import { TimeScalarMutations } from "../../graphql/input-objects/generic-mutation-operations/TimeScalarMutations";
+import {
+    BigIntListMutations,
+    BigIntScalarMutations,
+} from "../../graphql/input-objects/generic-mutation-operations/BigIntScalarMutations";
+import {
+    BooleanListMutations,
+    BooleanScalarMutations,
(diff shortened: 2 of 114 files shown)


.changeset/green-jobs-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neo4j/graphql": minor
+---
+
+Add suport for generic update operators
```
MacondoExpress: I would expand this changeset as it could be used to understand what to expect from the new syntax with an example like:
```
 mutation {
            updateMovies(update: { name: { set: "The Matrix" } }) {
                movies {
                    id
                    name
		}
       }
}
```

packages/graphql/src/schema/generation/get-mutation-input-from-attribute-type.ts
```diff
@@ -0,0 +1,182 @@
...
+    StringScalarMutations,
+} from "../../graphql/input-objects/generic-mutation-operations/StringScalarMutations";
+import {
+    TimeListMutations,
+    TimeScalarMutations,
+} from "../../graphql/input-objects/generic-mutation-operations/TimeScalarMutations";
+import type { AttributeAdapter } from "../../schema-model/attribute/model-adapters/AttributeAdapter";
+
+export function getMutationInputFromAttributeType(attribute: AttributeAdapter): GraphQLInputType | string {
+    // TODO: Keep working here
```
MacondoExpress: is there something else to be added here?

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -155,6 +155,27 @@ function makeAugmentedSchema({
...
+            description: `Mutations for a list for ${enumType.name.value}`,
+            fields: {
+                set: { type: `[${enumType.name.value}!]!` },
+                push: { type: `[${enumType.name.value}!]!` },
+                pop: { type: enumType.name.value },
+            },
+        });
+    });
+
     // Generates the filters for scalar, which are reused
```
MacondoExpress: This comment is no longer valid as here we generate also the mutation inputs

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -155,6 +155,27 @@ function makeAugmentedSchema({
             },
         });
     });
+
+    // Generates the mutations for enums, which are reused
+    Array.from(enumTypes.values()).forEach((enumType) => {
```
MacondoExpress, line 160: I would start moving these outside this large `makeAugmentedSchema` function as a separate step!

packages/graphql/src/translate/utils/get-mutation-field-statements.ts
```diff
@@ -100,3 +114,66 @@ export function getMutationFieldStatements({
...
+    withVars,
+}: {
+    nodeOrRel: Node | Relationship;
+    param: string;
+    key: string;
+    varName: string;
+    operations: any;
+    withVars: string[];
+}): string {
+    if (Object.entries(operations).length > 1) {
```
MacondoExpress, line 134: I would use `assertNonAmbiguousUpdate` that took care of `@alias` fields as well

packages/graphql/tests/integration/array-methods/array-pop.int.test.ts
```diff
@@ -576,7 +576,7 @@ describe("array-pop", () => {
 
         const update = `
             mutation {
-                ${typeMovie.operations.update} (update: { tags_POP: 1, moreTags_POP: 2 }) {
+                ${typeMovie.operations.update} (update: { tags_POP: 1, moreTags: {pop: 2} }) {
```
MacondoExpress:
```suggestion
                ${typeMovie.operations.update} (update: { tags: { pop:  1} , moreTags: {pop: 2} }) {
```

packages/graphql/tests/integration/math.int.test.ts
```diff
@@ -135,7 +135,7 @@ describe("Mathematical operations tests", () => {
 
             const query = /* GraphQL */ `
             mutation($id: ID, $value: ${type}) {
-                ${movie.operations.update}(where: { id_EQ: $id }, update: {viewers_${operation}: $value}) {
+                ${movie.operations.update}(where: { id_EQ: $id }, update: {viewers: {${operation}: $value }}) {
```
MacondoExpress, line 138: Unrelated change but it will avoid us to pass again to this test!
```suggestion
                ${movie.operations.update}(where: { id: { equals: $id } }, update: {viewers: {${operation}: $value }}) {
```

packages/graphql/tests/integration/types/date.int.test.ts
```diff
@@ -189,7 +189,7 @@ describe("Date", () => {
                 mutation {
                     ${
                         Movie.operations.update
-                    }(where: { id_EQ: "${id}"}, update: { date_SET: "${date.toISOString()}" }) {
+                    }(where: { id_EQ: "${id}"}, update: { date: {set: "${date.toISOString()}" } }) {
```
MacondoExpress, line 192: same reason as above:
```suggestion
                    }(where: { id: { equals: "${id}" } }, update: { date: {set: "${date.toISOString()}" } }) {
```

packages/graphql/tests/integration/types/duration.int.test.ts
```diff
@@ -189,7 +189,7 @@ describe("Duration", () => {
 
             const mutation = /* GraphQL */ `
                     mutation ($id: ID!, $duration: Duration) {
-                        ${Movie.operations.update}(where: { id_EQ: $id }, update: { duration_SET: $duration }) {
+                        ${Movie.operations.update}(where: { id_EQ: $id }, update: { duration: { set: $duration } }) {
```
MacondoExpress, line 192: same reason as above:
```suggestion
                        ${Movie.operations.update}(where: { id: { equals: "${id}" } }, update: { duration: { set: $duration } }) {
```

packages/graphql/tests/integration/types/localtime.int.test.ts
```diff
@@ -166,7 +166,7 @@ describe("LocalTime", () => {
 
             const mutation = /* GraphQL */ `
                     mutation ($id: ID!, $time: LocalTime) {
-                        ${Movie.operations.update}(where: { id_EQ: $id }, update: { time_SET: $time }) {
+                        ${Movie.operations.update}(where: { id_EQ: $id }, update: { time: {set: $time} }) {
```
MacondoExpress, line 169: same reason as above:
```suggestion
                        ${Movie.operations.update}(where: { id: { equals: $id } }, update: { time: {set: $time} }) {
```

packages/graphql/tests/integration/types/point.int.test.ts
```diff
@@ -194,7 +194,7 @@ describe("Point", () => {
             mutation UpdatePhotographs($id: String!, $longitude: Float!, $latitude: Float!) {
                 ${Photograph.operations.update}(
                     where: { id_EQ: $id }
```
MacondoExpress, line 196:
```suggestion
                    where: { id: { equals: $id } }
```

packages/graphql/tests/integration/types/point.int.test.ts
```diff
@@ -263,7 +263,7 @@ describe("Point", () => {
             mutation UpdatePhotographs($id: String!, $longitude: Float!, $latitude: Float!, $height: Float!) {
                 ${Photograph.operations.update}(
                     where: { id_EQ: $id }
```
MacondoExpress, line 265:
```suggestion
                    where: { id: { equals: $id  } }
```

MacondoExpress (commented):
LGTM, left some comments!

packages/graphql/src/schema/generation/get-mutation-input-from-attribute-type.ts
angrykoala: No, I left that comment by mistake

packages/graphql/src/translate/utils/get-mutation-field-statements.ts
angrykoala: it may be worth adding a test but this is an extra level check that 2 mutations cannot be inside the field, but assertNonAmbiguous is still being called to ensure that title + aliases are not being called twice

packages/graphql/tests/integration/types/date.int.test.ts
angrykoala: I'd prefer to keep this PR without those changs to avoid conflicts, particularly after the discussion on changing the names for the filters (`equals` vs `eq`)

MacondoExpress: Cool, no problem, we can change it later on when we migrate all the tests!

PR #5879, opened 2024-12-12
Aggregation filters

angrykoala:
# Description

Add generic filters for aggregations:

```graphql
{
    posts(where: { likesAggregate: { node: { rating: { average: { eq: 3.2 } } } } }) {
        title
    }
}
```


packages/graphql/src/translate/queryAST/factory/FilterFactory.ts
```diff
@@ -608,6 +613,29 @@ export class FilterFactory {
                 throw new Error(`Invalid operator ${key}`);
         }
     }
+    private parseGenericOperatorForAggregation(key: string): AggregationLogicalOperator {
+        // we convert them to the previous format to keep the same translation logic
+        switch (key) {
+            case "equals":
+            case "eq":
+                return "EQUAL";
+            case "lessThan": // TODO: remove long syntax once tests have been updated
```
MacondoExpress: This has already been removed!

packages/graphql/src/translate/queryAST/factory/FilterFactory.ts
```diff
@@ -608,6 +613,29 @@ export class FilterFactory {
                 throw new Error(`Invalid operator ${key}`);
         }
     }
+    private parseGenericOperatorForAggregation(key: string): AggregationLogicalOperator {
```
MacondoExpress, line 612: I believe we can reuse `parseGenericOperator`!

packages/graphql/src/translate/queryAST/factory/FilterFactory.ts
```diff
@@ -919,4 +987,27 @@ export class FilterFactory {
         }
         return targetPoint;
     }
+
+    private parsedGenericAggregationOperator(key: string): AggregationOperator {
```
MacondoExpress:
```suggestion
    private parseGenericAggregationOperator(key: string): AggregationOperator {
```

MacondoExpress (approved):
Left some comments but apart of that it looks great!

packages/graphql/src/translate/queryAST/factory/FilterFactory.ts
angrykoala: These are not the same filters, even if they have the same name, the typings are defined to use one or the other, and I rather keep the nice typings

PR #5880, opened 2024-12-12
Implement query complexity estimation

a-alle:
# Description

Adds query complexity estimation using a dedicated lib by implementing a FieldExtensionsEstimator for relationship fields, all other fields have a default value of 1 by employing the SimpleEstimator with a defaultValue of 1 (this value is not configurable at this point in time).
A method is exposed to retrieve the list of estimators, which can then be extended if the client wishes to do so.

## Complexity

Complexity: Medium

# Issue

> **Note**
>
>  

Closes 

# Checklist

### DIFF
--- a/packages/graphql/src/classes/ComplexityEstimatorHelper.ts
+++ b/packages/graphql/src/classes/ComplexityEstimatorHelper.ts
@@ -60,8 +60,8 @@ export class ComplexityEstimatorHelper {
                     complexity: ({childComplexity, args}) => {
                         const fieldDefaultComplexity = 1
                         const defaultLimitIfNotProvided = 1
-                        if(args.limit || args.first) {
-                            return childComplexity + (args.limit || args.first) * fieldDefaultComplexity
+                        if(args.limit ?? args.first) {
+                            return childComplexity + (args.limit ?? args.first) * fieldDefaultComplexity
                         }
                         return childComplexity + defaultLimitIfNotProvided
                         
--- a/packages/graphql/src/classes/ComplexityEstimatorHelper.ts
+++ b/packages/graphql/src/classes/ComplexityEstimatorHelper.ts
@@ -31,10 +31,8 @@ export class ComplexityEstimatorHelper {
 
     public registerField(parentObjectTypeNameOrInterfaceTypeName: string, fieldName: string): void {
         if(this.useComplexityEstimators) {
-            const existingFieldsForTypeName = this.objectTypeNameToFieldNamesMapForComplexityExtensions.get(parentObjectTypeNameOrInterfaceTypeName) || []
+            const existingFieldsForTypeName = this.objectTypeNameToFieldNamesMapForComplexityExtensions.get(parentObjectTypeNameOrInterfaceTypeName) ?? []
             this.objectTypeNameToFieldNamesMapForComplexityExtensions.set(parentObjectTypeNameOrInterfaceTypeName, existingFieldsForTypeName.concat(fieldName))
-
-            console.log(this.objectTypeNameToFieldNamesMapForComplexityExtensions)
         }
     }
 
@@ -94,7 +92,7 @@ export class ComplexityEstimatorHelper {
 
     public getComplexityEstimators(): ComplexityEstimator[] {
         if (!this.useComplexityEstimators) {
-            return [simpleEstimator({ defaultComplexity: 1 })]
+            return []
         }
         return [
             fieldExtensionsEstimator(),
--- a/packages/graphql/src/classes/ComplexityEstimatorHelper.ts
+++ b/packages/graphql/src/classes/ComplexityEstimatorHelper.ts
@@ -1,0 +1,108 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { type DefinitionNode,type GraphQLFieldExtensions, type GraphQLSchema, GraphQLInterfaceType, GraphQLObjectType, Kind } from "graphql";
+import type { ComplexityEstimator} from "graphql-query-complexity";
+import { fieldExtensionsEstimator, simpleEstimator } from "graphql-query-complexity";
+
+export class ComplexityEstimatorHelper {
+    private objectTypeNameToFieldNamesMapForComplexityExtensions: Map<string, string[]>;
+    private useComplexityEstimators: boolean;
+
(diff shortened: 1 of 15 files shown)


packages/graphql/src/classes/ComplexityEstimatorHelper.ts
```diff
@@ -0,0 +1,108 @@
...
+        this.useComplexityEstimators = useComplexityEstimators
+        this.objectTypeNameToFieldNamesMapForComplexityExtensions = new Map<string, string[]>();
+    }
+
+    public registerField(parentObjectTypeNameOrInterfaceTypeName: string, fieldName: string): void {
+        if(this.useComplexityEstimators) {
+            const existingFieldsForTypeName = this.objectTypeNameToFieldNamesMapForComplexityExtensions.get(parentObjectTypeNameOrInterfaceTypeName) || []
+            this.objectTypeNameToFieldNamesMapForComplexityExtensions.set(parentObjectTypeNameOrInterfaceTypeName, existingFieldsForTypeName.concat(fieldName))
+
+            console.log(this.objectTypeNameToFieldNamesMapForComplexityExtensions)
```
angrykoala: Is this a leftover `console.log`?

packages/graphql/src/classes/ComplexityEstimatorHelper.ts
```diff
@@ -0,0 +1,108 @@
...
+    private useComplexityEstimators: boolean;
+
+    constructor(useComplexityEstimators: boolean) {
+        this.useComplexityEstimators = useComplexityEstimators
+        this.objectTypeNameToFieldNamesMapForComplexityExtensions = new Map<string, string[]>();
+    }
+
+    public registerField(parentObjectTypeNameOrInterfaceTypeName: string, fieldName: string): void {
+        if(this.useComplexityEstimators) {
+            const existingFieldsForTypeName = this.objectTypeNameToFieldNamesMapForComplexityExtensions.get(parentObjectTypeNameOrInterfaceTypeName) || []
```
angrykoala: **nitpick** I think we are moving to use `??` for these checks which is slightly nicer than `||`
```suggestion
            const existingFieldsForTypeName = this.objectTypeNameToFieldNamesMapForComplexityExtensions.get(parentObjectTypeNameOrInterfaceTypeName) ?? []
```

packages/graphql/src/classes/ComplexityEstimatorHelper.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+import { type DefinitionNode,type GraphQLFieldExtensions, type GraphQLSchema, GraphQLInterfaceType, GraphQLObjectType, Kind } from "graphql";
+import type { ComplexityEstimator} from "graphql-query-complexity";
+import { fieldExtensionsEstimator, simpleEstimator } from "graphql-query-complexity";
+
+export class ComplexityEstimatorHelper {
+    private objectTypeNameToFieldNamesMapForComplexityExtensions: Map<string, string[]>;
+    private useComplexityEstimators: boolean;
+
+    constructor(useComplexityEstimators: boolean) {
```
angrykoala, line 27: Could it make sense to use an enum or list of estimators instead of boolean? (something like `estimator: "simple" | "complexity"` or `"default"`) Mostly so we can add more options in the future without having a breaking change if we decide different estimators

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -482,6 +442,60 @@ function makeAugmentedSchema({
     };
 }
 
+function getFinalDefinitionNodes(schemaExtensions: SchemaExtensionNode | undefined, definitions: readonly DefinitionNode[], complexityEstimatorHelper: ComplexityEstimatorHelper): DefinitionNode[] {
```
angrykoala: Nice on moving this to a function. What about naming it `getDefinitionNodesForAugmentedSchema`? I'm not too fond of `final` but it is fine as is too

angrykoala (commented):
Looks good to me :+1:

packages/graphql/src/classes/ComplexityEstimatorHelper.ts
MacondoExpress: +1

packages/graphql/src/classes/ComplexityEstimatorHelper.ts
MacondoExpress: I agree, using the name "useComplexityEstimators" seems to imply that if this is false, then it does not use any ComplexityEstimators, while later on the method "getComplexityEstimators" is used to return the "fieldExtensionsEstimator" or not. I would suggest  passing an object with the fields `{  simpleEstimator: boolean, fieldExtensionsEstimator: boolean } ` maybe with `simpleEstimator` defaulted to true

packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -108,6 +111,8 @@ class Neo4jGraphQL {
 
             this.authorization = new Neo4jGraphQLAuthorization(authorizationSettings);
         }
+
+        this.complexityEstimatorHelper = new ComplexityEstimatorHelper(!!this.features.complexityEstimators);
```
MacondoExpress, line 114: I would personally set the complexityEstimatorHelper as optional and define it only when `features.complexityEstimators` is true, same with the `make-augmented-schema`, I would make the complexityEstimatorHelper dependency optional!

MacondoExpress (approved):

packages/graphql/tests/e2e/setup/apollo-server.ts
```diff
@@ -86,6 +103,30 @@ export class ApolloTestServer implements TestGraphQLServer {
         const server = new ApolloServer({
             schema,
             plugins: [
+                {
+                    requestDidStart() {
+                        return  Promise.resolve({
+                             didResolveOperation({ request, document }) {
+                                const estimators = neo4jGraphql.getComplexityEstimators()
+                                if(estimators.length) {
```
darrellwarde: I think we may have found the culprit of our formatting issues! 👀 Formatting is a bit all over the place (lines not ending in semicolons, missing spaces after `if` - can you look into this?

darrellwarde (approved):
Other than my comment about formatting and what's already been pointed out, LGTM!

packages/graphql/src/classes/ComplexityEstimatorHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+                extensions: {
+                    // COMPLEXITY FORMULA
+                    // c = c_child + lvl_limit * c_field, where
+                    // c_field = 1
+                    // lvl_limit defaults to 1
+                    // c_child comes from simpleEstimator 
+                    complexity: ({childComplexity, args}) => {
+                        const fieldDefaultComplexity = 1
+                        const defaultLimitIfNotProvided = 1
+                        if(args.limit || args.first) {
```
MacondoExpress:
```suggestion
                        if(args.limit ?? args.first) {
```

packages/graphql/src/classes/ComplexityEstimatorHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+                            return childComplexity + (args.limit || args.first) * fieldDefaultComplexity
```
MacondoExpress:
```suggestion
                            return childComplexity + (args.limit ?? args.first) * fieldDefaultComplexity
```

packages/graphql/tests/e2e/setup/apollo-server.ts
a-alle: I think this is fine now?

darrellwarde (approved):

PR #5883, opened 2024-12-12
Update generic filters for custom scalars and enums

angrykoala:
# Description


MacondoExpress (commented):
It looks great, can you refactor an enum tests to use the generic style?

MacondoExpress (commented):
LGTM, can you add a test for enum as well?

PR #5885, opened 2024-12-12
fix authentication tests

MacondoExpress:
Fix the authentication tests, during a merge resolution I removed the alias for `includes` , `startsWith` ,`endsWith`.

### DIFF
--- a/packages/graphql/src/schema/resolvers/subscriptions/where/utils/get-filtering-fn.ts
+++ b/packages/graphql/src/schema/resolvers/subscriptions/where/utils/get-filtering-fn.ts
@@ -78,9 +78,9 @@ export function getFilteringFn<T>(
 
     const operators = { ...operatorCheckMap, ...overrides };
 
-    const op = operators[operator];
-    if (!op) {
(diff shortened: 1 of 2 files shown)


packages/graphql/src/schema/resolvers/subscriptions/where/utils/get-filtering-fn.ts
```diff
@@ -74,5 +78,9 @@ export function getFilteringFn<T>(
 
     const operators = { ...operatorCheckMap, ...overrides };
 
-    return operators[operator];
+    const op = operators[operator];
```
angrykoala: Maybe a more detailed name for `op`?

angrykoala (approved):

packages/graphql/src/schema/resolvers/subscriptions/where/utils/get-filtering-fn.ts
MacondoExpress: Good call!

PR #5892, opened 2024-12-16
Connect always creates new relationships

mjfwebb:
# Description

This PR changes connect functionality so that when we connect, it always adds a new relationship.

## Complexity

Low


packages/graphql/tests/integration/interfaces/relationships/create/connect.int.test.ts
```diff
@@ -163,4 +163,97 @@ describe("interface relationships", () => {
             },
         });
     });
+
+    test("should nested create connect a new relationship using interface relationship fields", async () => {
```
angrykoala: I'm not sure I understand the title of this test

angrykoala (approved):

PR #5897, opened 2024-12-16
Deprecate non generic filtering

MacondoExpress:
# Add deprecations to non generic filtering 

This PR adds deprecation to attribute relationship filters using the suffix syntax such as `_EQ` / `_SOME`


packages/graphql/src/schema/generation/augment-where-input.ts
```diff
@@ -250,9 +269,15 @@ function fieldConfigsToFieldConfigMap({
...
+        let directives;
+        if (deprecatedDirectives.length) {
+            directives = deprecatedDirectives;
+        } else if (field.deprecationReason) {
+            directives = [{ name: DEPRECATED, args: { reason: field.deprecationReason } }];
+        }
         fieldsConfigMap[field.name] = {
             type: field.typeName,
-            directives: deprecatedDirectives,
+            directives,
```
mjfwebb, line 280: Probably fine, but what happens if directives is undefined?

.changeset/nine-games-clap.md
```diff
@@ -0,0 +1,7 @@
+---
+"@neo4j/graphql": patch
+---
+
+
```
mjfwebb:
```suggestion

```

.changeset/nine-games-clap.md
```diff
@@ -0,0 +1,7 @@
...
+Deprecate relationship filtering using the non-generic version such as `actors_SOME:{ title_EQ: "The Matrix" }` in favor of the dedicated input `actors: { some: { title: { eq: "The Matrix" } } }`.
```
mjfwebb:
```suggestion
Deprecate relationship filtering using the non-generic version such as `actors_SOME: { title_EQ: "The Matrix" }` in favor of the dedicated input `actors: { some: { title: { eq: "The Matrix" } } }`.
```

mjfwebb: I see the terms "non-generic" and "dedicated" both used here. Is that intentional?

mjfwebb (commented):
Nice! Just a couple of comments.

.changeset/nine-games-clap.md
MacondoExpress: Good point, not there was no specific reasoning behind!

packages/graphql/src/schema/generation/augment-where-input.ts
MacondoExpress: From the typings it should not be possible!

mjfwebb (approved):

PR #5898, opened 2024-12-17
Remove aggregations for ID fields, remove comparators if not specified for String fields

MacondoExpress:
This PR removes aggregations and aggregation filters for ID fields.
Apart from that, the filters: `lt`, `lte`, `gte`, `gt`, `matches` are no longer generated in the `StringScalarFilters` and `IDScalarFilters` if not specified in the feature settings.


packages/graphql/src/schema-model/attribute/model-adapters/AttributeAdapter.ts
```diff
@@ -177,17 +177,24 @@ export class AttributeAdapter {
...
-            this.typeHelper.isGraphQLBuiltInScalar() && !this.typeHelper.isBoolean();
-        const isTemporalWithoutDate = this.typeHelper.isTemporal() && !this.typeHelper.isDate();
+        if (
+            this.typeHelper.isList() ||
+            this.typeHelper.isID() ||
+            this.typeHelper.isBoolean() ||
+            this.typeHelper.isDate()
+        ) {
+            return false;
+        }
```
mjfwebb, line 195: This is so much easier to read. We should change the others to be like this! :+1:

packages/graphql/tests/integration/interfaces/relationships/declare-relationship/interface-simple.int.test.ts
```diff
@@ -1333,7 +1333,7 @@ describe("interface with declared relationships", () => {
             ]
         );
     });
-
+    // FLAKY_TEST
```
mjfwebb, line 1336: Is this noted with the intention to work on it in this PR? Otherwise we should probably make a note to work on this separately.

mjfwebb (approved):
Lovely schema size reduction!

packages/graphql/src/schema-model/attribute/model-adapters/AttributeAdapter.ts
MacondoExpress: Thank you!

packages/graphql/tests/integration/interfaces/relationships/declare-relationship/interface-simple.int.test.ts
MacondoExpress: I just noticed that this test is flaky and I thought that by adding this comment could help us on keeping track of them!

PR #5902, opened 2024-12-18
Fix flaky test

darrellwarde:
# Description

Fix flaky test missing `expect.toIncludeSameMembers`.


MacondoExpress (approved):
even if the test is still failing, I think this is an improvement good to be merged

PR #5936, opened 2025-01-10
De-duplicate relationship projection when not using connection API

mjfwebb:
# Description

This PR introduces a change in the result projection where there are multiple relationships between two nodes.

In the case of using the connection API then multiple relationships will still be represented, as there is the ability to select the relationship properties. In the non-connection API case, the duplicate results will only return distinct results.

For example:

Data set:
```cypher
CREATE (m:Movie} {title: "Movie One"})
CREATE (a:Actor} {name: "Actor One"})
CREATE (a)-[:ACTED_IN {role: "Role One"}]->(m)
CREATE (a)-[:ACTED_IN {role: "Role Two"}]->(m)
```

## Connection API (actorsConnection):
```gql
query {
    movies {
        title
        actorsConnection {
            edges {
                node {
                    name
                }
                properties {
                    role
                }
            }
        }
    }
}
```

With the results:

```gql
movies: [
    {
        title: "Movie One",
        actorsConnection: {
            edges: expect.toIncludeSameMembers([
                {
                    node: {
                        name: "Actor One",
                    },
                    properties: {
                        role: "Role One",
                    },
                },
                {
                    node: {
                        name: "Actor One",
                    },
                    properties: {
                        role: "Role Two",
                    },
                },
            ]),
        },
    },
],
```

## Non-connection API (actors):
```gql
query {
    movies {
        title
        actors {
            name
        }
    }
}
```

With the results:

```gql
movies: [
    {
        title: "Movie One",
        actors: [
            {
                name: "Actor One",
            },
        ],
    },
],
```

## Complexity

Low

### DIFF
--- a/packages/graphql/tests/integration/directives/relationship/duplicate-relationship/duplicate-relationship-read-interface.int.test.ts
+++ b/packages/graphql/tests/integration/directives/relationship/duplicate-relationship/duplicate-relationship-read-interface.int.test.ts
@@ -55,7 +55,7 @@ describe("Interface: Multiple relationships results difference between Connectio
         `;
         await testHelper.initNeo4jGraphQL({ typeDefs });
 
-        // Deliberately create duplicate relationships
+        //  Create duplicate relationships
         await testHelper.executeCypher(`
             CREATE (m:${Movie} {title: "Movie One"})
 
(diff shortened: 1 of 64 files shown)


packages/graphql/tests/integration/directives/relationship/duplicate-relationship/duplicate-relationship-read-interface.int.test.ts
```diff
@@ -0,0 +1,220 @@
...
+                productions: [${Production}!]! @relationship(type: "ACTED_IN", properties: "ActedIn", direction: OUT)
+            }
+
+            type ActedIn @relationshipProperties {
+                role: String!
+            }
+        `;
+        await testHelper.initNeo4jGraphQL({ typeDefs });
+
+        // Deliberately create duplicate relationships
```
MacondoExpress:
```suggestion
        //  Create duplicate relationships
```
Nitpick: `Deliberately` to me seems imply a case not expected

MacondoExpress (approved):
Neat!

PR #5944, opened 2025-01-21
Top Level aggregations in connections

angrykoala:
# Description
Add top level aggregations inside connection operations:

```graphql
query {
    moviesConnection {
        aggregate {
            node {
                count
                id {
                    longest 
                }
            }
        }
    }
}
```

### DIFF
--- a/packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeConnectionReadOperation.ts
+++ b/packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeConnectionReadOperation.ts
@@ -35,7 +35,7 @@ export class CompositeConnectionReadOperation extends Operation {
     protected sortFields: Array<{ node: Sort[]; edge: Sort[] }> = [];
     private pagination: Pagination | undefined;
 
-    private aggregationField: ConnectionAggregationField | undefined; // TODO: multiple aggregations
+    private aggregationField: ConnectionAggregationField | undefined;
 
     constructor(children: CompositeConnectionPartial[]) {
         super();
--- a/packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeConnectionReadOperation.ts
+++ b/packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeConnectionReadOperation.ts
@@ -25,6 +25,7 @@ import { filterTruthy } from "../../../../../utils/utils";
 import { hasTarget } from "../../../utils/context-has-target";
 import { QueryASTContext } from "../../QueryASTContext";
 import type { QueryASTNode } from "../../QueryASTNode";
+import type { ConnectionAggregationField } from "../../fields/ConnectionAggregationField";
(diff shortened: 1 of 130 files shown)


packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeConnectionReadOperation.ts
```diff
@@ -138,4 +159,38 @@ export class CompositeConnectionReadOperation extends Operation {
             return [...nodeFields, ...edgeFields];
         });
     }
+
+    // NOTE: duplicate from ConnectionReadOperation
+    private transpileAggregation(context: QueryASTContext): {
```
MacondoExpress, line 164: Not sure if it will be right approach the proposal I'm saying but:
Can we expose this method from the `aggregationField` and avoid duplicity between the `CompositeConnectionReadOperation` and the `ConnectionReadOperation`

packages/graphql/tests/integration/aggregations/top-level/alias.int.test.ts
```diff
@@ -67,24 +67,119 @@ describe("aggregations-top_level-alias", () => {
...
+
+        const maxDate = new Date();
+        maxDate.setDate(maxDate.getDate() + 1);
+
+        await testHelper.initNeo4jGraphQL({ typeDefs });
+
+        await testHelper.executeCypher(
+            `
+                    CREATE (:${typeMovie} {testString: "${testString}", id: "1", title: "1", imdbRating: 1, createdAt: datetime("${minDate.toISOString()}")})
+                    CREATE (:${typeMovie} {testString: "${testString}", id: "22", title: "22", imdbRating: 2, createdAt: datetime()})
```
MacondoExpress, line 153: Even if these tests were already made, in my opinion, it will cover more cases if some entries are filtered out by the `testString` filter!

MacondoExpress (approved):
LGTM!

packages/graphql/tests/integration/aggregations/top-level/alias.int.test.ts
angrykoala: As you said, those cases are already covered. And these tests are based on the existing ones, Improving tests on aggregations is a bit out of scope for this, already quite big, PR. But I'll keep this in mind as we do need to improve these tests once the aggregations work is complete

packages/graphql/src/translate/queryAST/ast/operations/composite/CompositeConnectionReadOperation.ts
angrykoala: TBH, I don't think that is the best option. Sure, the code is the same, but it is more coincidental than anything, as transpilation for compositeConnection doesn't necessarily have to be the same, and I'm hesitant to couple the nested element of the tree. I'd rather have this duplicate code than break the encapsulation on those

PR #5953, opened 2025-01-24
2025 version

angrykoala:
# Description

Add support for CalVer editions of neo4j

### DIFF
--- a/packages/graphql/src/classes/Neo4jDatabaseInfo.test.ts
+++ b/packages/graphql/src/classes/Neo4jDatabaseInfo.test.ts
@@ -29,4 +29,9 @@ describe("Neo4jDatabaseInfo", () => {
             return new Neo4jDatabaseInfo("this_seems_not_valid", "enterprise");
         }).toThrow();
     });
+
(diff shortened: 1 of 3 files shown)


MacondoExpress (approved):

packages/graphql/src/classes/Neo4jDatabaseInfo.test.ts
```diff
@@ -29,4 +29,9 @@ describe("Neo4jDatabaseInfo", () => {
             return new Neo4jDatabaseInfo("this_seems_not_valid", "enterprise");
         }).toThrow();
     });
+
+    test("should accept CalVar", () => {
+        const dbInfo = new Neo4jDatabaseInfo("2025.01.0-aura", "enterprise");
```
darrellwarde, line 34: How about the case of "2025.01.0-19853" which @LackOfMorals produced?

angrykoala: That would be changed to "2025.1.0-19853" which is semver valid

PR #5963, opened 2025-01-27
Validation rules refactor and included new errors for when the directive node is expected but missing

MacondoExpress:
As title, this PR refactor the way how the validation are done.
The previous validation functions associated with each directive are now moved to their own Rule per directive.
It also includes a list of new rules for when the `@node` directive is expected but missing, for instance, when a directive expected to be used within a `@node` type is found but the `@node` directive is not.

Nice follow-ups (out of the scope of PR).
- The list of directives tested was increased but is still not comprehensive; for instance, no validation rules were present for the Vector directive.

- The new errors include invalid usage when the `@node` directive is not found. However, it does not cover cases when the type of a field is expected to reference an Object type that contains the `@node` directive; This could be particularly useful for identifying mistakes when the `@relationship` and `@cypher` target is not annotated as a `@node`.


packages/graphql/src/schema/validation/validate-sdl.ts
```diff
@@ -18,22 +18,29 @@
...
 import { mapError } from "./utils/map-error";
 
+type Neo4jValidationRule = <T extends SDLValidationContext>(context: T) => ASTVisitor;
 export function validateSDL(
     documentAST: DocumentNode,
-    rules: ReadonlyArray<SDLValidationRule>,
-    schemaToExtend?: Maybe<GraphQLSchema>
+    rules: ReadonlyArray<Neo4jValidationRule>,
+    schemaToExtend?: Maybe<GraphQLSchema>,
+    callbacks?: any
```
angrykoala: any better type than `any`? Maybe `unknown` if we are just checking callbacks exist

packages/graphql/src/schema/validation/Neo4jValidationContext.ts
```diff
@@ -0,0 +1,89 @@
...
+                        acc[typeName].extensions.push(def);
+                    } else {
+                        acc[typeName].extensions = [def];
+                    }
+                } else {
+                    acc[typeName].definition = def;
+                }
+            }
+            return acc;
+        }, {});
```
angrykoala: I would move this to a separate function, just to add a bit on context, and to avoid the constructor having all this logic

packages/graphql/src/schema/validation/Neo4jValidationContext.ts
```diff
@@ -0,0 +1,89 @@
...
+        this.callbacks = callbacks;
+        this.typeMapWithExtensions = ast.definitions.reduce((acc, def): TypeMapWithExtensions => {
+            if (
+                def.kind === Kind.OBJECT_TYPE_DEFINITION ||
+                def.kind === Kind.INTERFACE_TYPE_DEFINITION ||
+                def.kind === Kind.UNION_TYPE_DEFINITION ||
+                def.kind === Kind.ENUM_TYPE_DEFINITION ||
+                def.kind === Kind.OBJECT_TYPE_EXTENSION ||
+                def.kind === Kind.INTERFACE_TYPE_EXTENSION ||
+                def.kind === Kind.UNION_TYPE_EXTENSION
```
angrykoala: **nitpick**: Could using `includes` or `isInArray` make this a bit more succinct?

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-field-types.ts
```diff
@@ -63,10 +61,10 @@ function verifyRelationshipFieldType({
...
 function getValidationFunction(directiveName: string): ValidationFunction | undefined {
     switch (directiveName) {
-        case "id":
-            return verifyId;
-        case "timestamp":
-            return verifyTimestamp;
+        //case "id":
+        //return verifyId;
+        // case "timestamp":
+        //     return verifyTimestamp;
```
angrykoala:
```suggestion
```

packages/graphql/src/schema/validation/custom-rules/utils/same-type-argument-as-field.ts
```diff
@@ -17,8 +17,18 @@
  * limitations under the License.
  */
 import type { ArgumentNode, EnumTypeDefinitionNode, FieldDefinitionNode, ValueNode } from "graphql";
```
angrykoala, line 19: **nitpick**: Not really of this PR, so no need to change it, but maybe it is not worth having a `utils` folder along with all the custom rules for a single file. Maybe we can just move this to the `utils` folder in `schema/validation`

packages/graphql/src/schema/validation/node-is-missing.test.ts
```diff
@@ -0,0 +1,391 @@
...
+import gql from "graphql-tag";
+import validateDocument from "./validate-document";
+
+const additionalDefinitions = {
+    enums: [] as EnumTypeDefinitionNode[],
+    interfaces: [] as InterfaceTypeDefinitionNode[],
+    unions: [] as UnionTypeDefinitionNode[],
+    objects: [] as ObjectTypeDefinitionNode[],
+};
+describe("directive is required to be used withing @node directive", () => {
```
angrykoala:
```suggestion
describe("directive is required to be in a type with @node", () => {
```

packages/graphql/src/schema/validation/node-is-missing.test.ts
```diff
@@ -0,0 +1,391 @@
...
+        const executeValidate = () =>
+            validateDocument({
+                document: doc,
+                additionalDefinitions,
+                features: {},
+            });
+        expect(executeValidate).not.toThrow(
+            'Directive "relationship" requires to be used within the "@node" directive'
+        );
+    });
```
angrykoala: Do we need happy path tests here? This should be covered extensively in the `@relationship` tests themselves right?

I'm happy keeping this, not harm in having extra tests, but maybe is a good opportunity for reducing redundant tests?

packages/graphql/src/schema/validation/custom-rules/utils/same-type-argument-as-field.ts
```diff
@@ -50,37 +60,87 @@ export function assertArgumentHasSameTypeAsField({
...
         }
     }
 }
 
 function doTypesMatch(expectedType: string, argumentValueType: ValueNode, enums: EnumTypeDefinitionNode[]): boolean {
-    const isSpatialOrTemporal = isSpatial(expectedType) || isTemporal(expectedType);
-    if (isSpatialOrTemporal) {
-        return true;
+    if (expectedType === GraphQLID.name) {
+        return !!(fromValueKind(argumentValueType, enums, expectedType) === GraphQLString.name.toLowerCase());
```
angrykoala: *nitpick** I find using `Boolean()` better conveys what we are trying to do here. But no need to change it

packages/graphql/src/schema/validation/custom-rules/location-helpers/is-node-type.ts
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { type ObjectTypeDefinitionNode } from "graphql";
+import { nodeDirective } from "../../../../graphql/directives";
+import { asArray } from "../../../../utils/utils";
+import type { TypeMapWithExtensions } from "../../Neo4jValidationContext";
+
+export function typeIsANodeType({
```
angrykoala: Same as with utils. Maybe it is worth moving these to the `schema/validation/utils/location`? Being inside `custom-rules` indicate this is a rule, which is a bit confusing

packages/graphql/src/schema/validation/custom-rules/location-helpers/is-in-root-type.ts
```diff
@@ -0,0 +1,50 @@
...
+        throw new Error(
+            `Internal validation error: field with path: ${pathToNode.join(", ")} is in a type that does not exist in the typeMapWithExtensions`
+        );
+    }
+    const parentTypeAndExtensions = typeMapWithExtensions[parentOfTraversedDef.name.value];
+    if (!parentTypeAndExtensions) {
+        throw new Error(
+            `Internal validation error: field with path: ${pathToNode.join(", ")} is in a type that does not exist in the typeMapWithExtensions`
+        );
+    }
```
angrykoala: This bit of logic seems very similar to other helper methods like `is-in-subscription-type`. Maybe it is worth having a common function or grouping these helpers to share the same logic?

packages/graphql/src/schema/validation/custom-rules/directives/authentication.ts
```diff
@@ -0,0 +1,106 @@
...
+    return {
+        FieldDefinition(fieldDefinitionNode: FieldDefinitionNode, _key, _parent, path, ancestors) {
+            if (
+                !fieldDefinitionNode.directives?.find(
+                    (directive) => directive.name.value === authenticationDirectiveScaffold.name
+                )
+            ) {
+                return;
+            }
+
```
angrykoala, line 46: This looks like it could be a helper, as most directive rules seem to need this same check

packages/graphql/src/schema/validation/custom-rules/directives/authentication.ts
```diff
@@ -0,0 +1,106 @@
...
+            const isValidLocation =
+                (fieldIsInNodeType({ path, ancestors, typeMapWithExtensions }) ||
+                    fieldIsInRootType({ path, ancestors, typeMapWithExtensions })) &&
+                !fieldIsInSubscriptionType({ path, ancestors, typeMapWithExtensions });
+
+            const { isValid, errorMsg } = assertValid(() => {
```
angrykoala, line 52: Maybe this assignment could be inside `assertValid` just to keep the scope of `isValidLocation` close to where it is used

packages/graphql/src/schema/validation/custom-rules/directives/authentication.ts
```diff
@@ -0,0 +1,106 @@
...
+                if (!isValidLocation) {
+                    throw new DocumentValidationError(
+                        `Directive "${authenticationDirectiveScaffold.name}" requires to be used within the "@node" directive or in root types: Query, and Mutation`,
+                        []
+                    );
+                }
+            });
+            const pathToHere = getPathToNode(path, ancestors);
```
angrykoala: maybe `pathToNode` is a betterName? :sweat_smile: 

```suggestion
            const pathToNode = getPathToNode(path, ancestors);
```

packages/graphql/src/schema/validation/custom-rules/directives/coalesce.ts
```diff
@@ -16,51 +16,103 @@
...
+                assertArgumentHasSameTypeAsField({
+                    directiveName: coalesceDirective.name,
+                    traversedDef: fieldDefinitionNode,
+                    argument: valueArg,
+                    enums: enumsTypes,
+                });
+            });
+            const pathToHere = getPathToNode(path, ancestors);
+
+            if (!isValid) {
```
angrykoala: is pathToHere used outside the `!isValid`? Maybe move it inside if it isn't

packages/graphql/src/schema/validation/custom-rules/directives/relationship.ts
```diff
@@ -16,181 +16,136 @@
...
-                    `@relationship.properties invalid. Properties type ${propertiesValue} must use directive \`@relationshipProperties\`.`,
-                    ["properties"]
+            const { isValid, errorMsg, errorPath } = assertValid(() => {
+                if (!isValidLocation) {
+                    if (fieldIsInInterfaceType({ path, ancestors, typeMapWithExtensions })) {
+                        // throw more specific error for interface types as in the past it was possible to have relationships on interfaces
+                        throw new DocumentValidationError(
+                            `Invalid directive usage: Directive @${relationshipDirective.name} is not supported on fields of interface types (${parentOfTraversedDef?.name.value}). Since version 5.0.0, interface fields can only have @${declareRelationshipDirective.name}. Please add the @relationship directive to the fields in all types which implement it.`,
+                            []
+                        );
```
angrykoala, line 110: Do we still want to keep the `Since version 5.0.0` message? It is a bit outdated by now

packages/graphql/src/schema/validation/validate-document.test.ts
```diff
@@ -5082,9 +5075,9 @@ describe("validation 2.0", () => {
                     expect(errors[0]).not.toBeInstanceOf(NoErrorThrownError);
                     expect(errors[0]).toHaveProperty(
                         "message",
-                        "Invalid @relationshipProperties field: Cannot use the @authorization directive on relationship properties."
+                        'Directive "authorization" requires to be used within the "@node" directive'
```
angrykoala: Shouldn't this have `@`? `Directive "@authorization"`?

angrykoala (changes requested):

packages/graphql/src/schema/validation/Neo4jValidationContext.ts
MacondoExpress: I agree but I'm not sure, how easy and clean it will be as I have to reduce the type DefinitionNode to a specific subtype that contains only `ObjectTypeExtensionNode | InterfaceTypeExtensionNode | UnionTypeExtensionNode | ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode | UnionTypeDefinitionNode | EnumTypeDefinitionNode`, an easy way will be to manually set the type rather to let it be inferred.

packages/graphql/src/schema/validation/node-is-missing.test.ts
MacondoExpress: I would say yes, in my opinion, there is a chance that a bug appear in the validation and an error is raised when it shouldn't, this will be the worse case to me!

packages/graphql/src/schema/validation/custom-rules/directives/relationship.ts
MacondoExpress: maybe not, but I would change it separately from this PR to isolate the scope as much as possible!

PR #6000, opened 2025-02-11
merge aggregation-6-into-aggregation-7

MacondoExpress:
As title PR to sync aggregation work done in the 6.x branch with the aggregation work done in the 7.x branhc


MacondoExpress (commented):
@angrykoala  I took a look and it seems what we expected, LGTM

PR #6003, opened 2025-02-14
Count aggregation

angrykoala:
# Description

Add count fields in aggregations with support for nodes and edges count:

```graphql
query {
    moviesConnection {
        aggregate {
            count {
                nodes
            }
        }
    }
}
```

```graphql
query {
    movies {
        actorsConnection {
            aggregate {
                count {
                    nodes
                    edges
                }
            }
        }
    }
}
```

### DIFF
--- a/packages/graphql/src/schema/generation/aggregate-types.ts
+++ b/packages/graphql/src/schema/generation/aggregate-types.ts
@@ -74,6 +74,27 @@ export function withAggregateSelectionType({
     return aggregateSelection;
 }
 
+/** Top level count */
+export function getCountType(composer: SchemaComposer): ObjectTypeComposer {
+    const CountFieldName = "Count";
+    return composer.getOrCreateOTC(CountFieldName, (countField) => {
+        countField.addFields({
+            nodes: new GraphQLNonNull(GraphQLInt),
+        });
+    });
+}
+
+/** Nested count */
+export function getCountConnectionType(composer: SchemaComposer): ObjectTypeComposer {
+    const CountFieldName = "CountConnection";
+    return composer.getOrCreateOTC(CountFieldName, (countField) => {
+        countField.addFields({
+            nodes: new GraphQLNonNull(GraphQLInt),
+            edges: new GraphQLNonNull(GraphQLInt),
+        });
+    });
+}
+
 /** Create aggregate field inside connections */
 function createConnectionAggregate({
     entityAdapter,
@@ -88,26 +109,33 @@ function createConnectionAggregate({
     composer: SchemaComposer;
     features: Neo4jFeaturesSettings | undefined;
 }): ObjectTypeComposer {
-    const aggregateNode = composer.createObjectTC({
-        name: entityAdapter.operations.aggregateTypeNames.node,
-        fields: {
-            count: {
-                type: new GraphQLNonNull(GraphQLInt),
-                resolve: numericalResolver,
-                args: {},
-            },
-        },
-        directives: graphqlDirectivesToCompose(propagatedDirectives),
-    });
-    aggregateNode.addFields(makeAggregableFields({ entityAdapter, aggregationTypesMapper, features }));
+    const aggregableFields = makeAggregableFields({ entityAdapter, aggregationTypesMapper, features });
+    let aggregateNode: ObjectTypeComposer | undefined;
+    const hasNodeAggregateFields = Object.keys(aggregableFields).length > 0;
+    if (hasNodeAggregateFields) {
+        aggregateNode = composer.createObjectTC({
+            name: entityAdapter.operations.aggregateTypeNames.node,
+            fields: {},
+            directives: graphqlDirectivesToCompose(propagatedDirectives),
+        });
+        aggregateNode.addFields(aggregableFields);
+    }
 
-    return composer.createObjectTC({
+    const connectionAggregate = composer.createObjectTC({
         name: entityAdapter.operations.aggregateTypeNames.connection,
         fields: {
-            node: aggregateNode.NonNull,
+            count: getCountType(composer),
         },
         directives: graphqlDirectivesToCompose(propagatedDirectives),
     });
+
+    if (aggregateNode) {
+        connectionAggregate.addFields({
+            node: aggregateNode.NonNull,
+        });
+    }
+
+    return connectionAggregate;
 }
 
 function makeAggregableFields({
--- a/packages/graphql/src/schema/generation/aggregate-types.ts
+++ b/packages/graphql/src/schema/generation/aggregate-types.ts
@@ -81,7 +81,7 @@ export function getCountType(composer: SchemaComposer): ObjectTypeComposer {
         countField.addFields({
             nodes: {
                 type: new GraphQLNonNull(GraphQLInt),
-                resolve: (source, args, context, info) => numericalResolver(source, args, context, info),
+                resolve: numericalResolver,
             },
         });
     });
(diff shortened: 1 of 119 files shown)


packages/graphql/src/schema/generation/aggregate-types.ts
```diff
@@ -74,6 +74,36 @@ export function withAggregateSelectionType({
     return aggregateSelection;
 }
 
+/** Top level count */
+export function getCountType(composer: SchemaComposer): ObjectTypeComposer {
+    const CountFieldName = "Count";
```
MacondoExpress: I would name it `countFieldTypeName` the field name is `count` lower case, same below!

packages/graphql/src/schema/generation/aggregate-types.ts
```diff
@@ -74,6 +74,36 @@ export function withAggregateSelectionType({
...
+    return composer.getOrCreateOTC(CountFieldName, (countField) => {
+        countField.addFields({
+            nodes: {
+                type: new GraphQLNonNull(GraphQLInt),
+                resolve: (source, args, context, info) => numericalResolver(source, args, context, info),
```
MacondoExpress: I believe that the extra arrow function is not needed here.
```suggestion
                resolve: numericalResolver(source, args, context, info),
```

packages/graphql/src/schema/generation/aggregate-types.ts
```diff
@@ -74,6 +74,36 @@ export function withAggregateSelectionType({
...
+}
+
+/** Nested count */
+export function getCountConnectionType(composer: SchemaComposer): ObjectTypeComposer {
+    const CountFieldName = "CountConnection";
+    return composer.getOrCreateOTC(CountFieldName, (countField) => {
+        countField.addFields({
+            nodes: {
+                type: new GraphQLNonNull(GraphQLInt),
+                resolve: (source, args, context, info) => numericalResolver(source, args, context, info),
```
MacondoExpress: same as above.

packages/graphql/src/translate/queryAST/factory/Operations/ConnectionFactory.ts
```diff
@@ -268,22 +268,20 @@ export class ConnectionFactory {
...
+                const aggregationOperation = this.aggregateFactory.createAggregationOperation({
+                    entityOrRel: relationship ?? target,
+                    resolveTree: resolveTreeAggregate,
+                    context,
+                });
+                // NOTE: This will always be true on 7.x and this attribute should be removed
+                aggregationOperation.isInConnectionField = true;
+                const aggregationField = new ConnectionAggregationField({
+                    alias: resolveTreeAggregate.name, // Alias is hanlded by graphql on top level
+                    nodeAlias: nodeField?.alias ?? "node",
```
MacondoExpress: This `nodeAlias` seems no longer in use, this makes the above `nodeField` and `edgeField` unused variables.

packages/graphql/src/translate/queryAST/factory/Operations/ConnectionFactory.ts
```diff
@@ -294,28 +292,26 @@ export class ConnectionFactory {
...
```
MacondoExpress: same as above

packages/graphql/src/translate/queryAST/ast/fields/aggregation-fields/CountField.ts
```diff
@@ -24,10 +24,22 @@ import { AggregationField } from "./AggregationField";
 
 export class CountField extends AggregationField {
     private entity: Entity;
+    public edgeVar: Cypher.Variable | undefined;
```
MacondoExpress, line 27: I would investigate a different approach here, instead of constructing a single node for both `nodes` and `edges` I would construct multiple CountField as we do for AttributeFields. Then use `attachedTo` to chose which variable count.

packages/graphql/src/translate/queryAST/ast/fields/aggregation-fields/CountField.ts
```diff
@@ -43,6 +55,20 @@ export class CountField extends AggregationField {
...
+        if (this.countFields.nodes) {
+            resultMap.set("nodes", this.getAggregationExpr(target));
+        }
+        if (this.countFields.edges) {
+            if (!this.edgeVar) {
+                throw new Error(
+                    "Edge variable not defined in Count field. This is likely a bug with the GraphQL library."
+                );
+            }
+            resultMap.set("edges", this.getAggregationExpr(this.edgeVar));
```
MacondoExpress, line 69: Independently from the comment above, this `edgeVar` variable seems not needed. I tested the below, and the whole aggregation test suite works fine. This is because the client already invokes the `getAggregationProjection` with the correct variable depending on the context.
```suggestion
        if (this.countFields.edges) {
            resultMap.set("edges", this.getAggregationExpr(target));
```

MacondoExpress (commented):
Left some comments

packages/graphql/src/schema/generation/aggregate-types.ts
angrykoala: To be honest, I'm not 100% sure if we need this arrow function, but this pattern `resolve: (source, args, context, info) => numericalResolver(source, args, context, info),` is what we use elsewhere in the codebase for numericalResolvers, if we prefer removing the arrow functions I'd rather do it and test it in a separate PR in all the resolvers

packages/graphql/src/translate/queryAST/ast/fields/aggregation-fields/CountField.ts
angrykoala: I think that approach would add unnecessary complexity. The `attachedTo` solution involves an arbitrary number of attributes, in this case it is a single condition, I think it is more complex than this solution. Happy to discuss and change in the future, but it is a non-trivial change for this PR

packages/graphql/src/schema/generation/aggregate-types.ts
MacondoExpress: I think is safe to remove it, is already used without the arrow function:
`packages/graphql/src/schema/aggregations/field-aggregation-composer.ts:99`
`packages/graphql/src/schema/to-compose.ts:88`
!

packages/graphql/src/translate/queryAST/ast/fields/aggregation-fields/CountField.ts
angrykoala: I think this may actually be a gap in the tests. As we don't yet support multiple relationships between 2 nodes, all the tests are covering a case where the count of nodes is the same as relationships, so these pass. 

With change you are proposing, the aggregation count cypher looks like this:

```cypher
MATCH (this)<-[this0:ACTED_IN]-(this1:cBfniRfaActor)
RETURN { nodes: count(this1), edges: count(this1) } AS var2
```

MacondoExpress: You're right, I didn't notice it, for the count fields the target is always passed as "node" in the "AggregationOperation" while in `nodeFields` and `edgeFields` the target depends on the context!

packages/graphql/src/translate/queryAST/ast/fields/aggregation-fields/CountField.ts
MacondoExpress: Fair enough!

MacondoExpress (approved):

PR #6004, opened 2025-02-14
Add count.nodes and count.edges implementation

MacondoExpress:
# Description

As title, this PR add the `count: { nodes: { eq: 1 } }` and `count: { edges: { eq: 1 } } ` to the connection aggregation input.

I left the PR in draft, while waiting to see what will be the plan for the `DISTINCT` in the count.


packages/graphql/tests/integration/aggregations/where/count-edges.int.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                    title
+                    likes {
+                        name
+                    }
+                }
+            }
+        `;
+
+        const gqlResult = await testHelper.executeGraphQL(query);
+        expect(gqlResult.data).toBeDefined();
```
angrykoala, line 116: We don't need this assertion. The expect at the bottom will already check that is defined

packages/graphql/tests/integration/aggregations/where/count-edges.int.test.ts
```diff
@@ -0,0 +1,156 @@
...
+        expect(gqlResult.errors).toBeUndefined();
+        expect(gqlResult.data![Post.plural]).toHaveLength(2);
```
angrykoala, line 118: We don't really need this assertion either. The includeSameMembers will already check the length

packages/graphql/tests/integration/aggregations/where/count-edges.int.test.ts
```diff
@@ -0,0 +1,156 @@
...
+        expect(gqlResult.data![Post.plural]).toIncludeSameMembers([
+            {
+                title: "Post B",
+                likes: expect.toIncludeSameMembers([{ name: "Alice" }, { name: "Bob" }]),
+            },
+            {
+                title: "Post C",
+                likes: expect.toIncludeSameMembers([{ name: "Alice" }, { name: "Bob" }, { name: "Frank" }]),
+            },
+        ]);
```
angrykoala, line 128: **nitpick** I find a bit more readable adding the `Post.plural` into the equals:

```js
expect(gqlResult.data).toEquals({
  [Post.plural]: expect.toIncludeSameMemers([
   // Elements in array
  ]);
})
```

I find this a bit nicer for two reasons:

* Avoid using `!` after data
* It has a closer meaning to what we are actually testing: "we want data to be an object with posts", rather than "we want the posts (if any) to be this array"


In general, I've been refactoring tests written in the previous style to the new one

packages/graphql/src/schema/generation/aggregate-types.ts
```diff
@@ -132,6 +133,63 @@ function makeAggregableFields({
...
+        features,
+    });
+    if (edgeWhereInputType) {
+        aggregateWhereInput.addFields({ edge: edgeWhereInputType });
+    }
+    return aggregateWhereInput;
+}
+
+/**
+ * Deprecated Aggregation filters, for the Aggregation inside Connection input see withConnectionAggregateInputType
```
angrykoala, line 191: It may be worth adding the `@deprecated` tag to this comment to ensure this is slowly being replaced

packages/graphql/src/translate/queryAST/ast/filters/aggregation/CountFilter.ts
```diff
@@ -27,25 +26,44 @@ import { Filter } from "../Filter";
...
         super();
         this.comparisonValue = comparisonValue;
         this.operator = operator;
+        this.attachedTo = attachedTo;
+    }
+
+    protected getTarget(queryASTContext: QueryASTContext<Cypher.Node>): Cypher.Node | Cypher.Relationship {
+        const target = this.attachedTo === "node" ? queryASTContext.target : queryASTContext.relationship;
+        if (!target) {
+            throw new Error("No target found");
```
angrykoala, line 49: May be worth pointing out that this is an internal error that is caused by a `@neo4j/graphql` bug, as this is not a path that should be executed under normal circumstances

angrykoala (approved):

packages/graphql/tests/integration/aggregations/where/count-edges.int.test.ts
MacondoExpress: 100% agree, in this case I was started from the count test template and I didn't wanted to refactor the whole assertions!

PR #6010, opened 2025-02-18
Fix tests aggregation 6

angrykoala:
# Description
Some unit tests and e2e tests were failing after changes in aggregations, this PR fixes the tests. 

It also fixes the `@shareable` directive in new common fields generated for count


packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -125,6 +125,7 @@ class Neo4jGraphQL {
     }
 
     public async getSubgraphSchema(): Promise<GraphQLSchema> {
+        console.log("Get subgraph schema");
```
MacondoExpress:
```suggestion
```

MacondoExpress (approved):

Issue #6005, opened 2025-02-14
Aggregation results vary based on duplicate relationships between nodes


MacondoExpress:
**Describe the bug**
The library currently aggregates related node values in a way that depends on how many relationships exist between the same two nodes. For instance, if a node is connected via multiple relationships, its aggregated values (such as the sum of ages) will be counted multiple times. This behavior causes inconsistent aggregation results when duplicate relationships exist between nodes. The expected outcome is that each unique related node is counted only once, regardless of the number of relationships linking them.

**Type definitions**
```graphql
type Movie @node {
  title: String
  actors: [Actor!]!
    @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
}

type Actor @node {
  name: String
  age: Int
  born: DateTime
  movies: [Movie!]!
    @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
}

type ActedIn @relationshipProperties {
  screentime: Int
  character: String
}
```

**To Reproduce**

Populate the DB with the following data.

**Data**
```cypher
CREATE (m:Movie { title: "Terminator"})
CREATE (m)<-[:ACTED_IN { screentime: 60, character: "Terminator" }]-(arnold:Actor { name: "Arnold", age: 54, born: datetime('1980-07-02')})
CREATE (m)<-[:ACTED_IN { screentime: 120, character: "Sarah" }]-(:Actor {name: "Linda", age:37, born: datetime('2000-02-02')})
CREATE (m)<-[:ACTED_IN { screentime: 10, character: "Future Terminator" }]-(arnold)
```
Run the following query:
```graphql
query {
  movies {
    actorsAggregate {
      node {
        age {
          sum
        }
      }
    }
  }
}
```

**Expected behavior**
For a movie with actors connected via relationships, each unique actor should contribute only once to the aggregated sum of their ages. In the given scenario, although "Arnold" has two separate "ACTED_IN" relationships (due to different characters), his age should only be summed once alongside Linda's. Therefore, the expected aggregate age should be 91 (54 for Arnold + 37 for Linda), but the current behavior counts Arnold twice, leading to an incorrect sum.

**Current output**

```json
{
  "data": {
    "movies": [
      {
        "actorsAggregate": {
          "node": {
            "age": {
              "sum": 145
            }
          }
        }
      }
    ]
  }
}
```

**System (please complete the following information):**

* OS: [macOS]
* Version: [e.g. @neo4j/graphql@6.3.0]

[...]

PR #6018, opened 2025-02-21
Add count distinct on the Connection aggregation filter.

MacondoExpress:
# Description

Following the bug report https://github.com/neo4j/graphql/issues/6005, this PR adds tests and applies distinct on the count for nodes when defined explicitly with the count nodes syntax inside the Connection input object. The legacy version of aggregations continues to operate as before.


packages/graphql/src/translate/queryAST/factory/FilterFactory.ts
```diff
@@ -809,8 +810,20 @@ export class FilterFactory {
         return this.wrapMultipleFiltersInLogical(filterTruthy(filterASTs));
     }
 
-    private createCountFilter(operatorKey: string, value: unknown, attachedTo: "node" | "relationship"): CountFilter {
+    private createCountFilter(
+        operatorKey: string,
+        value: unknown,
+        attachedTo: "node" | "relationship",
+        useDeprecated = true
+    ): CountFilter {
```
angrykoala: With 4 parameters, considering changing this to an object input to give more context when calling it

packages/graphql/tests/integration/deprecations/generic-filtering/types/types-localdatetime-filter-deprecated.int.test.ts
```diff
@@ -84,7 +84,7 @@ describe("LocalDateTime - deprecated filters", () => {
         "should filter based on localDT comparison, for filter %s",
         async (filter) => {
             const futureId = generate({ readable: false });
-            const future = "2025-02-18T18:10:55.462Z".split("Z")[0];
+            const future = "6025-02-18T18:10:55.462Z".split("Z")[0];
```
angrykoala, line 87: ... well, this was hilarious

Setting a strict date for future is a bit of a bad practice, but I guess we don't really need to worry too much about y6k here

packages/graphql/tests/integration/issues/6005-filters.int.test.ts
```diff
@@ -0,0 +1,281 @@
...
+            CREATE (m)<-[:ACTED_IN { screentime: 120, character: "Another Character" }]-(:${Actor} {name: "Another actor", age: 37, born: datetime('2000-02-02')})
+            CREATE (m)<-[:ACTED_IN { screentime: 10, character: "Future Terminator" }]-(arnold)
+        `);
+    });
+
+    afterEach(async () => {
+        await testHelper.close();
+    });
+
+    test("should filter movies by actors count with duplicate results (deprecated syntax)", async () => {
```
angrykoala: We should move the deprecated syntax to the `deprecated` folder in a separate test file `6005.int.test`

packages/graphql/tests/integration/issues/6005-filters.int.test.ts
```diff
@@ -0,0 +1,281 @@
+/*
```
angrykoala: Per convention, this test should simply be called `6005.int.test.ts` and the name should only contain the url to the issue with the full context

packages/graphql/tests/integration/issues/6005-filters.int.test.ts
```diff
@@ -0,0 +1,281 @@
...
+ * limitations under the License.
+ */
+
+import type { UniqueType } from "../../utils/graphql-types";
+import { TestHelper } from "../../utils/tests-helper";
+
+/**
+ * For the following tests we assuming that the deprecated syntax keep the existing behavior while when using the new syntax the
+ * distinct is applied. This is applied only to count aggregation as for node aggregations second thoughts are needed.
+ **/
```
angrykoala: This context should be better placed in each test case, as part of the title.

packages/graphql/tests/tck/issues/6005-filters.test.ts
```diff
@@ -0,0 +1,343 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Neo4jGraphQL } from "../../../src";
+import { formatCypher, formatParams, translateQuery } from "../utils/tck-test-utils";
+
+describe("https://github.com/neo4j/graphql/issues/6005 filters (count only)", () => {
```
angrykoala: Same as the other, deprecated and new syntax should be split in 2 tests, one in deprecated folder and one here. 

The name of the file should be only the issue number, and the description the url per convention

angrykoala (changes requested):

packages/graphql/tests/integration/issues/6005-filters.int.test.ts
MacondoExpress: In this case, I put it in the same file intentionally to highlight the differences between the two, not only in the API but also in the behavior. I got the feeling that by moving in two separate files, a developer without the correct context could identify one of the behaviors as a regression or a bug.

packages/graphql/tests/integration/deprecations/generic-filtering/types/types-localdatetime-filter-deprecated.int.test.ts
MacondoExpress: Absolutely, this is a cherry-picked commit from the CI fix that was failing, I didn't refactor the test tho.

packages/graphql/tests/integration/issues/6005-filters.int.test.ts
MacondoExpress: This is because this tests is related only to filters, the test for 6005 should go on for the 6.x branch with the count fix projection

angrykoala: Still, that test should eventually be merged into the same file no?

Maybe a TODO or a comment is better suited than having a renamed file?

packages/graphql/tests/integration/issues/6005-filters.int.test.ts
angrykoala: I think that the same issue number being linked with the file in `deprecated` is good context on that.

In any case, deprecated behaviours are that, deprecated, regressions on that are tested in the deprecated folder. The new one is already in the correct place.

I'm just thinking that now that we finally start to have some conventions in our tests, it is not worth breaking those just for a quite edge case confusion that may arise with some deprecated behaviour. Breaking convention will generate confusion

PR #6024, opened 2025-02-25
Include deprecations for aggregation filters and changesets for the new aggregation filters

MacondoExpress:
# Description

As title this PR adds the missing deprecations for aggregation filters as well the changesets needed for aggregation filters in the aggregation-7 branch.


.changeset/gorgeous-goats-design.md
```diff
@@ -0,0 +1,25 @@
+---
+"@neo4j/graphql": minor
+---
+
+Following the changes of moving aggregations inside the connection fields, the aggregations filters are moved to the connection input field.
```
angrykoala: We probably don't need to mention "Following the changes of moving aggregations inside the connection fields"

Within the context of the full changelog we probably can just state what this change does:

Aggregation filters added to `aggregate` inputField inside `Connection` input fields. Old aggregate input fields deprecated.

.changeset/funny-tools-arrive.md
```diff
@@ -0,0 +1,15 @@
+---
+"@neo4j/graphql": patch
+---
+
+Following the changes of moving aggregations inside the connection fields, 
+the previous aggregations filters outside the connection filters are now deprecated.
+
+As part of the change, the flag `aggregationFiltersOutsideConnection` has been added to the excludeDeprecatedFields setting.
```
angrykoala: Same as with the other, no need to add the context of previous changes or "as part of this change". This will end up in the changelog with enough context

packages/graphql/src/schema/generation/where-input.ts
```diff
@@ -200,7 +202,7 @@ export function withSourceWhereInputType({
         userDefinedDirectivesOnTargetFields,
         features,
     });
-
+    // TODO: Likely this should be added only in case of relationshipAdapter.isFilterableByAggregate()
```
angrykoala, line 205: This TODO seems irrelevant now, as the check is done in line 215

.changeset/silent-phones-sort.md
```diff
@@ -0,0 +1,31 @@
+---
+"@neo4j/graphql": minor
+---
+
+Following the changes of moving aggregations inside the connection fields, the aggregation filter `count` is now supported on both nodes and relationships.
```
angrykoala:
```suggestion
The aggregation filter `count` now supports both, nodes and relationships.
```

angrykoala (approved):
Just changes in the changelog messages

.changeset/funny-tools-arrive.md
MacondoExpress: Good suggestion!

packages/graphql/src/schema/generation/where-input.ts
MacondoExpress: I think the line 215 is doing correctly but only for the deprecated filters, this should be valid also for the new aggregation filters. I'll test it in the following PR

PR #6040, opened 2025-03-04
Enable distinct for new aggregations

### LINKED ISSUE #6005: Aggregation results vary based on duplicate relationships between nodes

angrykoala:
# Description

This enable `DISTINCT` for aggregation attributes. This is only enabled for new aggregations, a new "DeprecatedAggregationAttributeField" has been added to legacy aggregations

### DIFF
--- a/packages/graphql/tests/integration/aggregations/field-level/field-level-aggregations.int.test.ts
+++ b/packages/graphql/tests/integration/aggregations/field-level/field-level-aggregations.int.test.ts
@@ -57,6 +57,7 @@ describe("Field Level Aggregations", () => {
             CREATE (m)<-[:ACTED_IN { screentime: 60, character: "Terminator" }]-(a1:${typeActor.name} { name: "Arnold", age: 54, born: datetime('1980-07-02')})
             CREATE (m)<-[:ACTED_IN { screentime: 50, character: "someone" }]-(a1)
             CREATE (m)<-[:ACTED_IN { screentime: 120, character: "Sarah" }]-(:${typeActor.name} {name: "Linda", age:37, born: datetime('2000-02-02')})
+            CREATE (m)<-[:ACTED_IN { screentime: 120, character: "Sarah" }]-(:${typeActor.name} {name: "John", age:37, born: datetime('2000-02-02')})
         `);
     });
 
@@ -89,7 +90,7 @@ describe("Field Level Aggregations", () => {
                     actorsConnection: {
                         aggregate: {
                             count: {
-                                nodes: 2,
+                                nodes: 3,
                             },
                         },
                     },
@@ -124,8 +125,8 @@ describe("Field Level Aggregations", () => {
                     actorsConnection: {
                         aggregate: {
                             count: {
-                                nodes: 2,
-                                edges: 3,
+                                nodes: 3,
+                                edges: 4,
                             },
                         },
                     },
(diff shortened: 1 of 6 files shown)


packages/graphql/tests/integration/aggregations/field-level/field-level-aggregations.int.test.ts
```diff
@@ -205,8 +205,8 @@ describe("Field Level Aggregations", () => {
                                     age: {
                                         max: 54,
                                         min: 37,
-                                        average: expect.closeTo(48.33),
-                                        sum: 145,
+                                        average: expect.closeTo(45.5),
+                                        sum: 91,
```
MacondoExpress: Can you add another actor with the same age to verify that same value properties are still represented? Even better if the test contains both `edge` and `node` on the same query. To be fair, I'm already somewhat confident this is already the case, but it will be good to avoid future regressions

PR #6048, opened 2025-03-07
Subscriptions as an opt-in feature

darrellwarde:
Subscriptions are now an opt-in feature which can be enabled by using the `@subscription` directive on either schema or type.

For example, to enable subscriptions for the whole schema (equivalent to before this breaking change):

```graphql
type Movie @node {
    title: String!
}

extend schema @subscription
```

To enable subscriptions just for the `Movie` type:

```graphql
type Movie @node @subscription {
    title: String!
}
```


packages/graphql/src/classes/Neo4jGraphQL.ts
```diff
@@ -503,7 +503,7 @@ class Neo4jGraphQL {
 
         const setup = async () => {
             const subscriptionsEngine = this.features?.subscriptionsEngine;
-            if (subscriptionsEngine) {
+            if (subscriptionsEngine && (await this.executableSchema)?.getSubscriptionType()) {
```
angrykoala: not too happy about having an await as part of a logical expression, maybe move it into a separate variable?

angrykoala (dismissed):

PR #6063, opened 2025-03-10
Remove nock as a dependency

angrykoala:
# Description

`nock` is no longer used in LTS and should be safe to remove


mjfwebb (approved):

angrykoala:
Scratch this change, nock is used as a peer dependency of `mock-jwks` for jwks tests

PR #6081, opened 2025-03-14
Fix missing auth rules for interfaces in aggregate fields

angrykoala:
# Description

Fix missing authentication rules for interfaces in aggregate fields in connections.


packages/graphql/tests/integration/interfaces/aggegations/filter-aggregation-interfaces-field-level-with-auth.int.test.ts
```diff
@@ -154,6 +154,31 @@ describe("Field-level filter interface query fields with authorization", () => {
         expect(queryResult.data).toBeNull();
     });
 
+    test("field level filter aggregation with missing role should fail", async () => {
```
MacondoExpress: I would move this file in the deprecated folder, and test the new API here!

PR #6091, opened 2025-03-14
7x: Typescript update

mjfwebb:
# Description

- Updates the old version of `@tsconfig/16` with an updated `@tsconfig/22` (matching the `engines` in `package.json` of `"node": ">=22.0.0"`)
- Updates tsconfig to retain the previous production build configuration which was essentially:
```json
        "module": "CommonJS",
        "target": "es2021",
        "moduleResolution": "node10"
```
- Updates Typescript to 5.8.2
- Addresses new errors raised during this update. Please see individual commits for errors raised.

## Complexity

Low

### DIFF
--- a/packages/graphql/src/schema-model/parser/parse-arguments.ts
+++ b/packages/graphql/src/schema-model/parser/parse-arguments.ts
@@ -18,7 +18,7 @@
  */
 
 import { inspect } from "@graphql-tools/utils";
-import type { Maybe } from "@graphql-tools/utils/";
+import type { Maybe } from "@graphql-tools/utils";
 import type { DirectiveNode, FieldNode, GraphQLDirective, GraphQLField } from "graphql";
 import { Kind, isNonNullType, print, valueFromAST } from "graphql";
 import type { ObjMap } from "graphql/jsutils/ObjMap";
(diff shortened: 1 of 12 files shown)


packages/graphql/src/schema-model/parser/parse-arguments.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import { inspect } from "@graphql-tools/utils";
-import type { Maybe } from "@graphql-tools/utils/typings/types";
+import type { Maybe } from "@graphql-tools/utils/";
```
darrellwarde: Weird trailing slash?

```suggestion
import type { Maybe } from "@graphql-tools/utils";
```

packages/graphql/src/tsconfig.production.json
```diff
@@ -2,7 +2,10 @@
     "extends": "../../../tsconfig.base.json",
     "compilerOptions": {
         "rootDir": ".",
-        "outDir": "../dist"
+        "outDir": "../dist",
+        "module": "CommonJS",
```
darrellwarde: OCD casing suggestion: https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-module-output-format

```suggestion
        "module": "commonjs",
```

darrellwarde (dismissed):

PR #6104, opened 2025-03-21
Refactor aggregation filters and add distinct when are defined inside connection

MacondoExpress:
# Description

This PR continues to fix the issues #6005  in case of filtering on Aggregations inside the Connection input. 

Distinct is now included for `node` filters inside the Connection input such as this:
```graphql
{
    movies(where: { 
        actorsConnection: {
            aggregate: {
                node: { 
                    age: { sum: { eq: 99 } } 
                }
            }
        }
    }) {
        title
    }
}
```
In this case, the aggregation sum on the actor's age counts each actor once, no matter how many times are connected to the movie
.

Note that multiple Subqueries are needed to compute for each specific aggregation. This is required as the distinct on the node target will make the aggregations on relationship properties invalid as they no longer have the correct cardinality.


packages/graphql/tests/integration/issues/6005.int.test.ts
```diff
@@ -62,7 +62,7 @@ describe("https://github.com/neo4j/graphql/issues/6005", () => {
         await testHelper.close();
     });
 
-    test("should filter movies by actors count with unique results", async () => {
+    test.only("should filter movies by actors count with unique results", async () => {
```
darrellwarde:
```suggestion
    test("should filter movies by actors count with unique results", async () => {
```

packages/graphql/tests/integration/issues/6005.int.test.ts
```diff
@@ -218,4 +218,76 @@ describe("https://github.com/neo4j/graphql/issues/6005", () => {
             ],
         });
     });
+
+    test("should return movies where the actors age sum is EQUAL to", async () => {
+        const age1 = 54;
+        const age2 = 37;
+        const age3 = 37;
+        // arnold should not counted twice
+        const sum = age1 + age2 + age3;
```
darrellwarde, line 223: Surely with the logic in this test, this would be the case?

packages/graphql/tests/integration/issues/6005.int.test.ts
```diff
@@ -218,4 +218,76 @@ describe("https://github.com/neo4j/graphql/issues/6005", () => {
...
+            ],
+        });
+    });
+
+    test("should return movies where the actors age avg is EQUAL to", async () => {
+        const age1 = 54;
+        const age2 = 37;
+        const age3 = 37;
+        // arnold should not counted twice
+        const avg = (age1 + age2 + age3) / 3;
```
darrellwarde, line 256: Same with this test, the age will be counted twice?

darrellwarde (dismissed):
It looks good - I think there's something to check out in the tests here but looks good for the most part!

packages/graphql/tests/integration/issues/6005.int.test.ts
MacondoExpress: Nice catch!

packages/graphql/tests/integration/issues/6005.int.test.ts
MacondoExpress: It is actually correct, it's just confusing because two of the actors have the same age. I've changed it, to remove the confusion!

PR #6159, opened 2025-04-02
Validates that interfaces are not partially implemented by types with node directive

angrykoala:
Fails schema validation if an interface is implemented by a type with `@node` but not all implemented types use `@node`. For example, the following is invalid:

```graphql
interface Person {
    name: String
}

type Director implements Person {
    name: String
}

type Actor implements Person @node {
    name: String
}
```


packages/graphql/src/schema/validation/custom-rules/valid-types/valid-object-type.ts
```diff
@@ -17,12 +17,19 @@
...
-import type { ASTVisitor, InterfaceTypeDefinitionNode, ObjectTypeDefinitionNode } from "graphql";
-import type { SDLValidationContext } from "graphql/validation/ValidationContext";
+import { type ASTVisitor, type InterfaceTypeDefinitionNode, type ObjectTypeDefinitionNode } from "graphql";
+import type { Neo4jValidationContext } from "../../Neo4jValidationContext";
 import { DocumentValidationError, assertValid, createGraphQLError } from "../utils/document-validation-error";
+import { typeIsANodeType } from "../utils/location-helpers/is-node-type";
 import type { ObjectOrInterfaceWithExtensions } from "../utils/path-parser";
 
-export function ValidObjectType(context: SDLValidationContext): ASTVisitor {
+export function ValidObjectType(context: Neo4jValidationContext): ASTVisitor {
```
MacondoExpress, line 26: I think is worth change the name of this function to reflect that is no longer related to ObjectType only

packages/graphql/src/schema/validation/Neo4jValidationContext.ts
```diff
@@ -93,3 +96,24 @@ function buildTypeMapWithExtensions(definitions: Readonly<DefinitionNode[]>): Ty
         return acc;
     }, {});
 }
+
+function buildInterfacesMap(
+    definitions: Readonly<DefinitionNode[]>,
+    typeMapWithExtensions: TypeMapWithExtensions
+): Record<string, Array<ObjectTypeDefinitionNode>> {
+    return definitions.reduce((acc, def): TypeMapWithExtensions => {
```
MacondoExpress, line 104:
```suggestion
    return definitions.reduce((acc, def): Record<string, Array<ObjectTypeDefinitionNode>> => {
```

PR #6165, opened 2025-04-03
Adds schema validation for unions with partially implemented nodes

angrykoala:
# Description

Fails schema validation if an union is composed of a type with `@node` but not all other types. For example, the following is invalid:

```graphql

type Director {
    name: String
}

type Actor @node {
    name: String
}
```


packages/graphql/src/schema/validation/custom-rules/valid-types/valid-union-type.ts
```diff
@@ -0,0 +1,70 @@
...
+    return {
+        UnionTypeDefinition(unionType: UnionTypeDefinitionNode) {
+            const { isValid, errorMsg } = assertValid(() => {
+                let isNodeInterface = false;
+                let hasNonNodeTypes = false;
+                for (const concreteType of unionType.types ?? []) {
+                    const concreteTypeFileName = concreteType.name.value;
+                    const type = typeMapWithExtensions[concreteTypeFileName];
+                    if (!type) {
+                        throw new Error("Invalid type");
```
MacondoExpress: I would raise a more helpful message, for instance by printing the type that was not found.

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-union-type.ts
```diff
@@ -0,0 +1,70 @@
...
+
+                        if (isConcreteTypeANode) {
+                            isNodeInterface = true;
+                        } else {
+                            hasNonNodeTypes = true;
+                        }
+                    }
+                }
+                if (isNodeInterface && hasNonNodeTypes) {
+                    throw new DocumentValidationError("Union needs to be fully implemented by `@node` types.", []);
```
MacondoExpress: This error raised is misleading as does not reflect the possibility of having a union fully implemented by no `@node` types.

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-union-type.ts
```diff
@@ -0,0 +1,70 @@
...
+export function ValidUnionType(context: Neo4jValidationContext): ASTVisitor {
+    const typeMapWithExtensions = context.typeMapWithExtensions;
+
+    if (!typeMapWithExtensions) {
+        throw new Error("No typeMapWithExtensions found in the validation context");
+    }
+    return {
+        UnionTypeDefinition(unionType: UnionTypeDefinitionNode) {
+            const { isValid, errorMsg } = assertValid(() => {
+                let isNodeInterface = false;
```
MacondoExpress:
```suggestion
                let hasNodeTypes = false;
```

packages/graphql/src/schema/validation/custom-rules/valid-types/valid-union-type.ts
angrykoala: Fair, although this is more of an assertion, not an error intended to reach the end user

PR #6153, opened 2025-04-01
Validates that the target of a relationship is a @node

angrykoala:
# Description

Fails schema validation if a field with `@relationship` targets a type without `@node`.

For example, the following schema will fail:

```graphql
type Movie @node {
    someActors: [Actor!]! @relationship(type: "ACTED_IN", direction: OUT)
}

type Actor {
    name: String
}
```

PR #6166, opened 2025-04-03
Validate relationship to an interface are implemented by nodes

angrykoala:
# Description

Implemente node validation (#6153) for relationship pointed to interfaces


packages/graphql/src/schema/validation/custom-rules/directives/relationship.ts
```diff
@@ -176,5 +180,18 @@ function validateRelationshipTarget(
                 []
             );
         }
+    } else if (relatedType.kind === Kind.INTERFACE_TYPE_DEFINITION) {
```
MacondoExpress, line 183: Do you want to add the union target validation here or in a follow-up PR?

angrykoala: Yes, let's keep this one only to interfaces

PR #6199, opened 2025-04-10
Support for non node types on cypher directives

angrykoala:
# Description

Add support for `@cypher` directive of fields targeting types that do not use the `@node` directive. For example:

```graphql
type Movie @node {
    title: String
    id: String!
    link: Link!
        @cypher(
            statement: """
            MATCH(l:${Link})
            WHERE l.movieId=this.id
            RETURN l {.name, .url} as link
            """
            columnName: "link"
        )
}

type Link {
    movieId: String!
    url: String!
    name: String!
}
```

This approach has some limitations, particularly with non-node interfaces and unions

### DIFF
--- a/packages/graphql/src/schema-model/generate-model.ts
+++ b/packages/graphql/src/schema-model/generate-model.ts
@@ -135,14 +135,9 @@ function getCypherTarget(schema: Neo4jGraphQLSchemaModel, attributeType: Attribu
     if (attributeType instanceof ObjectType) {
         const foundConcreteEntity = schema.getConcreteEntity(attributeType.name);
         if (!foundConcreteEntity) {
-            throw new Neo4jGraphQLSchemaValidationError(
-                `@cypher field must target type annotated with the @node directive${attributeType.name}, `
-            );
+            return undefined;
         }
         return schema.getConcreteEntity(attributeType.name);
-    }
-    if (attributeType instanceof InterfaceEntity || attributeType instanceof UnionEntity) {
-        throw new Error("@cypher field target cannot be an interface or an union");
     }
 }
 
--- a/packages/graphql/src/schema-model/generate-model.ts
+++ b/packages/graphql/src/schema-model/generate-model.ts
@@ -36,7 +36,7 @@ import { Neo4jGraphQLSchemaModel } from "./Neo4jGraphQLSchemaModel";
 import { Operation } from "./Operation";
 import type { Attribute } from "./attribute/Attribute";
 import type { AttributeType } from "./attribute/AttributeType";
-import { ListType, ObjectType } from "./attribute/AttributeType";
+import { InterfaceType, ListType, ObjectType } from "./attribute/AttributeType";
 import type { CompositeEntity } from "./entity/CompositeEntity";
 import { ConcreteEntity } from "./entity/ConcreteEntity";
 import type { Entity } from "./entity/Entity";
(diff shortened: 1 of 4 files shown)


packages/graphql/tests/integration/directives/cypher/cypher-with-non-node-types.int.test.ts
```diff
@@ -0,0 +1,316 @@
...
+                        },
+                    ],
+                },
+            ],
+        });
+    });
+
+    /**
+     * Interfaces without `@node` are not supported yet because there is no way to access "__resolveTree", as the information of the concrete type returned by the cypher field is not available
+     */
```
darrellwarde: What will happen in this case? Will we throw a validation error? Or just some weird indecipherable error?

packages/graphql/tests/integration/directives/cypher/cypher-with-non-node-types.int.test.ts
```diff
@@ -0,0 +1,316 @@
...
+                },
+            ],
+        });
+    });
+
+    test("query", async () => {
+        const Movie = testHelper.createUniqueType("Movie");
+
+        const typeDefs = /* GraphQL */ `
+            type Movie @node {
```
darrellwarde: `@node` shouldn't he here right?

```suggestion
            type Movie {
```

packages/graphql/tests/integration/directives/cypher/cypher-with-non-node-types.int.test.ts
```diff
@@ -0,0 +1,316 @@
...
+               WHERE l.movieId=this.id
+               RETURN l {.name, .url} as link
+               """, columnName: "link")
+            }
+
+            type ${Link} {
+                movieId: String!
+                url: String!
+                name: String!
+            }
```
darrellwarde, line 48: How about this kind of case where the Cypher might return a more complex object? 

```suggestion
            type ${Link} {
                movieId: String!
                url: String!
                name: String!
                somethingElse: SomethingElse!
            }
            
            type ${SomethingElse} {
               something: String!
            }
```

packages/graphql/src/schema-model/generate-model.ts
```diff
@@ -135,15 +135,10 @@ function getCypherTarget(schema: Neo4jGraphQLSchemaModel, attributeType: Attribu
...
-            throw new Neo4jGraphQLSchemaValidationError(
-                `@cypher field must target type annotated with the @node directive${attributeType.name}, `
-            );
+            return undefined;
         }
         return schema.getConcreteEntity(attributeType.name);
     }
-    if (attributeType instanceof InterfaceEntity || attributeType instanceof UnionEntity) {
-        throw new Error("@cypher field target cannot be an interface or an union");
-    }
```
darrellwarde: Maybe we should be keeping this given the skipped test?

darrellwarde (commented):
Some comments to think about?

packages/graphql/tests/integration/directives/cypher/cypher-with-non-node-types.int.test.ts
angrykoala: I'll add the test, but at that point is just GraphQL parsing the return object

packages/graphql/src/schema-model/generate-model.ts
angrykoala: This will fail on interfaces targeting nodes (which works properly now), this check is not executed on non node interfaces, keep in mind that those don't even reach the model

angrykoala: The skipped test throw an error that is not covered by any of these cases, I'll see if it is possible to validate that case here to throw a better error

darrellwarde (approved):

PR #6213, opened 2025-04-23
Update schema on subgraph schema performance tests

angrykoala:
# Description
Tests on the schema performance for subgraphs were not updated to v7.

This PR updates these and makes errors on schema generation visible


angrykoala:
The CI error is caused by the version in dev being run to compare. This PR will fix that issue

mjfwebb (approved):

PR #6276, opened 2025-05-08
Add jsonwebtoken dependency explicitly in package.json

angrykoala:
# Description

The dependency "jsonwebtoken" is used in the tests, yet this dependency was not declared in package.json. We relied on it being a dependency of a different package

### DIFF
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -59,6 +59,7 @@
         "is-uuid": "1.0.2",
         "jest": "29.7.0",
         "jest-extended": "4.0.2",
+        "jsonwebtoken": "9.0.2",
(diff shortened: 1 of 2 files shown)


packages/graphql/package.json
```diff
@@ -59,6 +59,7 @@
         "is-uuid": "1.0.2",
         "jest": "29.7.0",
         "jest-extended": "4.0.2",
+        "jsonwebtoken": "^9.0.2",
```
mjfwebb:
```suggestion
        "jsonwebtoken": "9.0.2",
```

angrykoala: Good shout

mjfwebb (approved):

PR #6252, opened 2025-04-30
Test using user and role for tests

darrellwarde:
# Description

This PR ensures that where possible, tests are run using a dedicated user for tests, with the minimal roles needed for execution. Code could be improved, just wanted to get something up before heading out!

PR #6306, opened 2025-05-16
Change test setup to use user with restricted permission

MacondoExpress:
Follow-up on the PR https://github.com/neo4j/graphql/pull/6252.
The scope of this PR is to create a restricted user with limited grants, and then impersonate that user when a GraphQL query is executed.


packages/graphql/tests/integration/issues/464.int.test.ts
```diff
@@ -154,42 +154,38 @@ describe("https://github.com/neo4j/graphql/issues/464", () => {
...
 
-            const books = await graphql({
-                schema: await neoSchema.getSchema(),
-                source: queryBooks,
-                contextValue: testHelper.getContextValue({ executionContext: session }),
-            });
+        const books = await graphql({
+            schema: await neoSchema.getSchema(),
+            source: queryBooks,
+            contextValue: await testHelper.getContextValue({ executionContext: session }),
```
angrykoala, line 185: Why is this await needed now? the signature of `getContextValue` has not changed for what I see

Not an issue, just curious

packages/graphql/jest.global-setup.js
```diff
@@ -48,12 +59,96 @@ module.exports = async function globalSetup() {
...
+    await session.run(cypherDropRole);
+}
+
+async function createUserAndRole(session) {
+    await session.run(cypherCreateUser);
+    await session.run(cypherCreateRole);
+    await session.run(cypherGrantRole);
+}
+
+/* Dummy Javascript copy of the utility function available at: ./src/utils/error-has-gql-status */
```
angrykoala: Is not really a Dummy copy right? Just a copy on Javascript of that one

packages/graphql/jest.global-setup.js
```diff
@@ -48,12 +59,96 @@ module.exports = async function globalSetup() {
...
+                    `\nJest /packages/graphql setup: Will NOT create a separate integration test user and role as the command is not supported in the current environment.`
+                );
+            } else {
+                throw error;
+            }
+        } finally {
+            if (session) {
+                await session.close();
+            }
+        }
```
angrykoala: I know that it wouldn't be exactly the same behaviour, so feel free to ignore this comment, but I feel this could be easier to read if the two try are merged, as the error handling is very similar:

```js
try {
  session = driver.session();
  await dropUserAndRole(session);
  session = driver.session();
  await createUserAndRole(session);

  for (const cypherGrant of readWriteGrants) {
      await session.run(cypherGrant);
  }
} catch (err){
           if (errorHasGQLStatus("50N42")) {
                console.log(
                    `\nJest /packages/graphql setup: Failure to drop test user/role, this is expected if the user/role does not exist. Error: ${error.message}`
                );
            else if (errorHasGQLStatus("42NFF")) {
                console.log(
                    `\nJest /packages/graphql setup: Will NOT create a separate integration test user and role as the command is not supported in the current environment.`
                );
            } else {
                throw error;
            }
} finally {
            if (session) {
                await session.close();
            }
}


```

packages/graphql/jest.global-setup.js
```diff
@@ -48,12 +59,96 @@ module.exports = async function globalSetup() {
...
+    }
+    if (driver) {
+        await driver.close();
     }
```
angrykoala: Should this be in `finally`?

angrykoala (dismissed):

packages/graphql/jest.global-setup.js
MacondoExpress: True!

packages/graphql/tests/integration/issues/464.int.test.ts
MacondoExpress: Ah no the signature is not changed, but it was wrong in the first place as the method is async

packages/graphql/jest.global-setup.js
MacondoExpress: good catch, this actually was already wrong as it was closing the driver only for the last try/catch. I didn't want to reorganise this file but we should. I'll do it.

packages/graphql/jest.global-setup.js
MacondoExpress: The difference in behaviour is actually needed in this case because when the procedure "DROP USER" is executed against a newly created database it fails because the user does not exists. The alternative would have been to query for the user and depending if the user exists or not executing the procedure to drop it.

PR #6535, opened 2025-07-30
Increase testing coverage for create authorization

angrykoala:
# Description

Testing on authorization mutations is lacking in some areas. This PR improves the coverage on authorization, explicitly on create mutations

Additionally, a helper tool `neo-expect` is added to improve the assertions over the neo4j database


packages/graphql/tests/integration/mutations/create/authorization/filter/create-authorization-filter.int.test.ts
```diff
@@ -0,0 +1,132 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { UniqueType } from "../../../../../utils/graphql-types";
+import { TestHelper } from "../../../../../utils/tests-helper";
+
+describe("create with authorization validate", () => {
```
MacondoExpress:
```suggestion
describe("create with authorization filter", () => {
```

packages/graphql/tests/integration/mutations/create/authorization/filter/create-authorization-filter.int.test.ts
```diff
@@ -0,0 +1,132 @@
...
+                },
+            },
+        });
+    });
+
+    afterEach(async () => {
+        await testHelper.close();
+    });
+
+    test("create and connect with authorization filter", async () => {
```
MacondoExpress, line 78: I think is worth testing with other operations as well. I'm thinking `create-> connect -> connect` `create -> connect ->update`

packages/graphql/tests/integration/mutations/create/authorization/validate/create-authorization-validate-interface.int.test.ts
```diff
@@ -0,0 +1,311 @@
...
+        Comment = testHelper.createUniqueType("Comment");
+
+        const typeDefs = /* GraphQL */ `
+            interface Content {
+                title: String!
+            }
+
+            type ${Post} implements Content @node {
+                title: String!
+                creatorId: ID!
```
MacondoExpress, line 43: I would remove `creatorId` it is not used in the authorization and it creates confusion, as well from the other tests

packages/graphql/tests/utils/neo-expect/expect-cypher.ts
```diff
@@ -0,0 +1,58 @@
...
+        const rawResults = result.records.map((r) => r.toObject());
+        const parsedResult = this.parseResult(rawResults);
+
+        expect(parsedResult).toEqual(expectation);
+    }
+
+    protected parseResult(rawResults: neo4j.RecordShape[]): Record<string, any>[] {
+        return rawResults.map((rawObject) => {
+            return Object.entries(rawObject).reduce((acc, [key, value]) => {
+                if (isNeoInt(value)) {
```
MacondoExpress, line 48: I think we need to handle all the conversion to non native JS types: https://neo4j.com/docs/javascript-manual/current/data-types
For integer, I believe this will fail when larger than 32 bit integer are returned such for BigInt

MacondoExpress (commented):
Left some comments

packages/graphql/tests/integration/mutations/create/authorization/validate/create-authorization-validate-interface.int.test.ts
angrykoala: It is used in the authorization directive below

```
@authorization(
    validate: [
        {
            operations: [CREATE, CREATE_RELATIONSHIP]
            where: { node: { creatorId: { eq: "$jwt.sub" } } }
        }
    ]
```

packages/graphql/tests/utils/neo-expect/expect-cypher.ts
angrykoala: Yeah, this is just a first iteration, there are plenty of cases that need to be ironed out

packages/graphql/tests/integration/mutations/create/authorization/filter/create-authorization-filter.int.test.ts
angrykoala: The operation `create->connect->update` is not valid (you can only connect from a connect)

The twice nested connect could be useful, I'll add it to validate and filter

packages/graphql/tests/integration/mutations/create/authorization/validate/create-authorization-validate-interface.int.test.ts
MacondoExpress: Ah right, I got confused, sorry for the noise!

packages/graphql/tests/utils/neo-expect/expect-cypher.ts
MacondoExpress: Maybe it's worth to raise a not implemented error when parsing something not yet handled, to avoid creating confusion to whoever will start using this helper. Or add these limitation in the surface APIs documentation of the helper so that the expectation are clear for new users of these.

PR #6693, opened 2025-09-23
Fix flaky integration test

angrykoala:
# Description

This test could return elements in a different order, updated expectations


mjfwebb (approved):
Nice!

PR #6747, opened 2025-10-15
update release workflows for OIDC publishing

angrykoala:
# Description

Once tested, this needs to be ported to all active release branches (3.x, lts, 4.x, 6.x)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,6 @@
-name: changesets
+# NOTE: This filename must be called release.yml. If changed, update the Trusted Publisher settings in npm
+
+name: Release
 
 on:
   push:
@@ -8,6 +10,9 @@ on:
       - "*.x"
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+permissions:
+  id-token: write # Required for OIDC
 
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -33,8 +38,13 @@ jobs:
         uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5
         with:
           node-version: lts/*
+          registry-url: https://registry.npmjs.org
           package-manager-cache: false
 
+      - name: Update npm # Need at least 11.5.1 for OIDC publishing
+        run: |
+          npm install -g npm@latest
```
mjfwebb: Nit:

```suggestion
        run: npm install -g npm@latest
```

mjfwebb (dismissed):

luffy1727 (approved):

.github/workflows/release.yml
```diff
@@ -1,4 +1,6 @@
-name: changesets
+# NOTE: This filename must be called release.yml. If changed, update the Trusted Publisher settings in npm
+
+name: Release
```
luffy1727, line 3: would this create a conflict since there exists a job jobs.release named Release as well. (Non blocking btw)

angrykoala: I think it is fine, as this is just used for reporting. The cypher-builder has the same setup and seems to be working fine :thinking:

luffy1727: lets ball

mjfwebb (approved):

PR #6760, opened 2025-10-20
Fix flaky test connect-union.int

angrykoala:
# Description

These tests were failing when the value `screenTime` was 0.

This PR changes all the random values in tests to be explicit values


keremgocen (dismissed):
LGTM:)

PR #6774, opened 2025-10-28
Update math operations to new translation layer

angrykoala:


packages/graphql/src/translate/queryAST/factory/Operations/DeleteFactory.ts
```diff
@@ -197,14 +197,6 @@ export class DeleteFactory {
...
             });
         }
-        if (isUnionEntity(target)) {
-            return this.createNestedDeleteOperationsForUnion({
-                deleteArg,
-                relationship,
-                target,
-                context,
-            });
-        }
```
a-alle, line 207: is this intended?

packages/graphql/src/translate/queryAST/ast/operations/UpdateOperation.ts
```diff
@@ -160,9 +160,7 @@ export class UpdateOperation extends Operation {
             matchClause,
             ...filterSubqueries,
             filtersWith,
-            ...mutationSubqueries.map((sq) =>
-                Cypher.utils.concat(new Cypher.With(nestedContext.target), new Cypher.Call(sq, [nestedContext.target]))
-            )
+            ...mutationSubqueries.map((sq) => Cypher.utils.concat(new Cypher.With("*"), new Cypher.Call(sq, "*")))
```
a-alle, line 163: is the "*" instead of nestedContext.target on purpose?

packages/graphql/src/translate/queryAST/factory/Operations/UpdateFactory.ts
```diff
@@ -183,6 +186,13 @@ export class UpdateFactory {
                         });
                     } else {
                         for (const op of Object.keys(targetInput[fieldName])) {
+                            const operations = Object.keys(targetInput[fieldName]);
+                            if (operations.length > 1) {
+                                const conflictingOperations = operations.map((op) => `[[${op}]]`);
+                                throw new GraphQLError(
+                                    `Conflicting modification of field ${fieldName}: ${conflictingOperations.join(", ")} on type ${target.name}`
+                                );
+                            }
```
a-alle: looks like this could be moved one line up, before the `for(const op of Object.keys(targetInput[fieldName]))` ?

packages/graphql/src/utils/find-conflicting-properties.ts
```diff
@@ -55,3 +59,30 @@ export function findConflictingProperties({
...
+    fields: string[],
+    entityOrRel: ConcreteEntityAdapter | RelationshipAdapter
+): Set<string> {
+    const fieldsByDbName = new Map<string, string[]>();
+
+    for (const rawField of fields) {
+        const { fieldName } = parseMutationField(rawField);
+        const dbName = entityOrRel.findAttribute(fieldName)?.databaseName;
+        if (dbName) {
+            const fields = fieldsByDbName.get(dbName) ?? [];
```
a-alle: is there a better name here so it doesn't shadow the fn argument? It's a bit confusing

a-alle: given that `fieldsByDbName` is not returned and only its `.values()` are used, can the set not be constructed from this initial loop?

a-alle (changes requested):

packages/graphql/src/translate/queryAST/ast/operations/UpdateOperation.ts
angrykoala: Yes, we need both, the target and the relationship, we can probably refine it further once all the tests are working

packages/graphql/src/translate/queryAST/factory/Operations/DeleteFactory.ts
angrykoala: The update operations handle unions before the delete happens, so we are already getting the concrete entities (or interfaces) once we reach the nested operation. We probably need to discuss on how to unify this behaviour in a more consistent way, right now it is tightly coupled with our input type

packages/graphql/src/translate/queryAST/factory/Operations/UpdateFactory.ts
angrykoala: Yeah, good catch, makes more sense outside of the loop

packages/graphql/src/utils/find-conflicting-properties.ts
angrykoala: We need to iterate through all the fields first, to get the duplicates. The set only contains those fields that have more than 1 duplicate element

packages/graphql/src/translate/queryAST/factory/Operations/DeleteFactory.ts
a-alle: ah yes, ok

packages/graphql/src/translate/queryAST/ast/operations/UpdateOperation.ts
a-alle: is a todo worth it?

a-alle (approved):

PR #6775, opened 2025-10-29
Update populated by new translation layer

angrykoala:


packages/graphql/src/translate/queryAST/factory/Operations/UpdateFactory.ts
```diff
@@ -193,12 +191,21 @@ export class UpdateFactory {
...
-                                attribute,
-                                targetInput[fieldName][op]
-                            );
+                            const operations = Object.keys(targetInput[fieldName]);
+                            if (operations.length > 1) {
+                                const conflictingOperations = operations.map((op) => `[[${op}]]`);
+                                throw new GraphQLError(
+                                    `Conflicting modification of field ${fieldName}: ${conflictingOperations.join(", ")} on type ${target.name}`
+                                );
+                            }
```
a-alle: out of this for-loop?

packages/graphql/src/translate/queryAST/utils/callback-bucket.ts
```diff
@@ -79,7 +79,13 @@ export class CallbackBucket {
                 const callbackFunction = callbacksList[cb.functionName];
                 if (callbackFunction) {
                     const paramValue = await callbackFunction(cb.parent, {}, this.context);
-                    cb.param.value = this.parseCallbackResult(paramValue, cb.type);
+                    if (paramValue === undefined) {
+                        cb.param.value = undefined;
+                    } else if (paramValue === null) {
+                        cb.param.value = null;
```
a-alle, line 85: How about sth like this?
```
if(!paramValue) {
  cb.param.value = paramValue 
}
```

packages/graphql/tests/integration/directives/populatedBy/populatedBy-relationship-properties.int.test.ts
```diff
@@ -3207,7 +3206,9 @@ describe("@populatedBy directive - Relationship properties", () => {
         test("should have access to parent in callback function for UPDATE", async () => {
             const testMovie = testHelper.createUniqueType("Movie");
             const testGenre = testHelper.createUniqueType("Genre");
-            const callback = (parent) => `${parent.title_SET}-slug`;
+            const callback = (parent) => {
```
a-alle: any reason for not keeping this as it was?
```
const callback = (parent) => `${parent.title_SET}-slug`;
```

a-alle (approved):

packages/graphql/src/translate/queryAST/factory/Operations/UpdateFactory.ts
angrykoala: whops, merging conflict incorrectly fixed :sweat_smile:

packages/graphql/src/translate/queryAST/utils/callback-bucket.ts
angrykoala: paramValue could be an empty string or 0, I'll rather make these cases explicitly tbh

packages/graphql/tests/integration/directives/populatedBy/populatedBy-relationship-properties.int.test.ts
angrykoala: not really, just had to add some logs to debug this, and left it as a multiline function, I'm not sure it makes much of a difference

PR #6794, opened 2025-11-03
Add ReleaseOps Aura Tests workflow

neo-tobias:
To run as a smoke test before a release of Neo4j, against the ReleaseOps Aura environment.

Complexity:

Low

### DIFF
--- a/.github/workflows/releaseops-aura-tests.yml
+++ b/.github/workflows/releaseops-aura-tests.yml
@@ -1,7 +1,7 @@
 name: "ReleaseOps Aura Tests"
-
+# Executes graphql tests in the ReleaseOps Aura environment as part of the Neo4j ReleaseOps pipeline. 
+# Triggered by ReleaseOps from TeamCity, to validate the new database version, before a release.
 on:
-  # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/releaseops-aura-tests.yml
+++ b/.github/workflows/releaseops-aura-tests.yml
@@ -1,0 +1,28 @@
+name: "ReleaseOps Aura Tests"
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+  pull_request_target:
+    branches:
+      - dev
+    paths:
+      - "**/CHANGELOG.md"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}-aura
+  cancel-in-progress: true
+
+jobs:
+  aura-professional-5-tests:
+    if: |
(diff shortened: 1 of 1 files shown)


neo-tobias:
URI: ${{ secrets.RELEASEOPS_AURA_PROFESSIONAL_5_URI }}
PASSWORD: ${{ secrets.RELEASEOPS_AURA_PROFESSIONAL_5_PASSWORD }}
needs to be setup to point to a db in
https://sre-devreleaseops.neo4j-dev.io/databases

.github/workflows/releaseops-aura-tests.yml
```diff
@@ -0,0 +1,28 @@
+name: "ReleaseOps Aura Tests"
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+  pull_request_target:
+    branches:
+      - dev
+    paths:
+      - "**/CHANGELOG.md"
```
Liam-Doodson: Is it possible to just have the `workflow_dispatch` as we already run these against AuraDB and have fairly high github action costs so don't want to trigger a lot of extra workflows?

```suggestion
```

.github/workflows/releaseops-aura-tests.yml
```diff
@@ -0,0 +1,28 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}-aura
+  cancel-in-progress: true
+
+jobs:
+  aura-professional-5-tests:
+    if: |
+      github.event_name != 'pull_request_target' ||
+      github.event.pull_request.user.login == 'neo4j-team-graphql' &&
+      startsWith(github.head_ref, 'changeset-release/')
```
Liam-Doodson: Not sure this would be needed if we got rid of the pull request target

```suggestion
```

.github/workflows/releaseops-aura-tests.yml
neo-tobias: yes sorry copy paste error, I only need workflow_dispatch

.github/workflows/releaseops-aura-tests.yml
neo-tobias: right will simplify

.github/workflows/releaseops-aura-tests.yml
```diff
@@ -0,0 +1,19 @@
+name: "ReleaseOps Aura Tests"
```
angrykoala, line 1: Could we add some comment with some context about this workflow?

The name itself is good, but it may be not enough context for future maintainers of these workflows to know what the intention is of these tests

neo-tobias: Yes added comments

PR #6802, opened 2025-11-05
Update nodejs in workflows

angrykoala:
# Description

Update node version in our workflows. 

Only nightly integration tests and reusable package tests are kept to Nodejs 20 for testing compatibility


mjfwebb (commented):
Do we know that all the actions support node 24 now?

mjfwebb (approved):

PR #6829, opened 2025-11-17
Fix disconnect operations

angrykoala:
# Description
Fix disconnect operations on new translation layer


packages/graphql/src/translate/queryAST/factory/Operations/DisconnectFactory.ts
```diff
@@ -101,6 +101,7 @@ export class DisconnectFactory {
         context: Neo4jGraphQLTranslationContext,
         callbackBucket: CallbackBucket
     ): CompositeDisconnectPartial {
+        console.log("Create composite Disconnect");
```
a-alle: 👀

a-alle (approved):

PR #6899, opened 2025-12-09
Adds populatedByOperation to context passed to populatedBy callback

### LINKED ISSUE #6605: Pass `operation` type to @populatedBy directive as fourth argument
Within the callback code of a `@populatedBy` directive I need to know wether the code is evaluated in context of a `CREATE` or `UDPATE` mutation. Depending on that, I need to return different values.

It would be great to have the `operation` type passed to the callback function as a fourth argument. Either as plain `string` or as `enum`, I don't care about that. Since the callback is called from within `translate-create.js` resp. `translate-update.js` it should be easy to pass the `operation` type to the call of `callbackBucket.resolveCallbacks()` and then to `await callbackFunction(cb.parent, {}, this.context, OPERATION_TYPE)`

I have tried to use the `@populatedBy` directive twice on a field with different callback function names and different operation types each, one for `CREATE` and one for `UPDATE` - but this fails since the schema validator allows this directive only once per field.

angrykoala:
# Description
Add `populatedByOperation` field in context passed to populatedBy callbacks. This allows to change the behaviour depending on whether the `populatedBy` callback is triggered by a create or update operation using `context.populatedByOperation`.

For example:

```js
const typeDefs = /* GraphQL */ `
    type Movie @node {
        title: String
        actors: [Person!]! @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
        modified: Boolean! @populatedBy(callback: "modified", operations: [CREATE, UPDATE])
    }
`;

    typeDefs,
    features: {
        populatedBy: {
            callbacks: {
                modified(parent, args, context) {
                    if (context.populatedByOperation === "CREATE") {
                        return false;
                    } else {
                        return true;
                    }
                },
            },
        },
    },
});

```

### DIFF
--- a/packages/graphql/tests/integration/directives/populatedBy/populatedBy-arguments.int.test.ts
+++ b/packages/graphql/tests/integration/directives/populatedBy/populatedBy-arguments.int.test.ts
@@ -258,7 +258,7 @@ describe("populatedBy arguments", () => {
             mutation {
                 updatePeople(
                     update: {
-                        actedIn: [{ update: { node: { title_SET: "Test" }, edge: { screenTime: { set: 100 } } } }]
+                        actedIn: [{ update: { node: { title: { set: "Test" } }, edge: { screenTime: { set: 100 } } } }]
                     }
                 ) {
                     people {
--- a/packages/graphql/tests/integration/directives/populatedBy/populatedBy-arguments.int.test.ts
+++ b/packages/graphql/tests/integration/directives/populatedBy/populatedBy-arguments.int.test.ts
@@ -273,7 +273,7 @@ describe("populatedBy arguments", () => {
 
         expect(nodeMockCallback).toHaveBeenCalledWith(
             {
-                title_SET: "Test",
+                title: { set: "Test" },
             },
             {},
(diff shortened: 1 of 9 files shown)


a-alle (dismissed):

packages/graphql/tests/integration/directives/populatedBy/populatedBy-arguments.int.test.ts
```diff
@@ -189,13 +203,91 @@ describe("populatedBy arguments", () => {
...
+            })
+        );
+    });
+
+    test("update relationships", async () => {
+        const query = /* GraphQL */ `
+            mutation {
+                updatePeople(
+                    update: {
+                        actedIn: [{ update: { node: { title_SET: "Test" }, edge: { screenTime: { set: 100 } } } }]
```
MacondoExpress:
```suggestion
                        actedIn: [{ update: { node: { title: { set: "Test" } }, edge: { screenTime: { set: 100 } } } }]
```

MacondoExpress (dismissed):

PR #6905, opened 2025-12-11
Remove unused util case-where

angrykoala:
# Description
This utility for generating cypher is no longer used


MacondoExpress (dismissed):

MacondoExpress:
The changes LGTM, 
The tests are failing, however, I think it's just the yarn.lock file that cause this

angrykoala:
@MacondoExpress 
The tests were failing before, I specifically updated the yarn.lock to see if that was the issue

PR #6976, opened 2026-01-14
Support for single element  (1 to *, 1 to 1) relationships (readonly)

angrykoala:
# Description

Basic re-introduction of single element relationships (1-1, 1-*). For example:

```graphql
type Movie @node {
    title: String!
    actor: Person @relationship(type: "ACTED_IN", direction: IN)
}

type Person @node {
    name: String!
}
```

1-1 relationships were removed in v7, due to several constraints, mainly in mutations, that were not enforceable without impacting performance severely. On top of that, certain legacy API decisions were inconsistent with many-to-many relationships.

This PR reintroduces 1-1 relationships as a readonly feature, follow-up PRs will reintroduce more features in a non-breaking way.

In this PR:

* Read operations for 1-1 relationships
  * The entity API will treat nested 1-* relationships as simple nested objects. Pagination and sorting operations will not be available
  * Connection API will maintain the same structure as many-to-many relationships (connection -> edges,) to maintain compatibility with Relay.
  * Aggregations (connection -> aggregation) not available for 1-1 relationships
  
  In follow up PRs, support for interfaces, unions, filtering and partial support for mutations will be added


a-alle (approved):
nice!

PR #6977, opened 2026-01-14
Vector directive validation

a-alle:
# Description

Adds a validator to check `@vector` is only used on object types that also have the `@node` directive set on them.
Also updates the invalid-directive-combination list for object types in regards to the `@relationshipProperties` which shouldn't be combined with any of the others except `@deprecated` and maybe `@shareable` (unsure about this one)

## Complexity

Complexity: Low

# Issue

> **Note**
>
>  

Closes 

# Checklist


packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
```diff
@@ -0,0 +1,69 @@
...
+            const { isValid, errorMsg } = assertValid(() => {
+                if (!isValidLocation) {
+                    throw new DocumentValidationError(
+                        `Directive "@${vectorDirective.name}" must be in a type with "@node"`,
+                        []
+                    );
+                }
+            });
+            const pathToNode = getPathToNode(path, ancestors);
+            console.log("pathToNode", pathToNode);
```
angrykoala:
```suggestion
```

packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
```diff
@@ -0,0 +1,69 @@
...
+            const isValidLocation = typeIsANodeType({ objectTypeDefinitionNode, typeMapWithExtensions });
+            const { isValid, errorMsg } = assertValid(() => {
+                if (!isValidLocation) {
+                    throw new DocumentValidationError(
+                        `Directive "@${vectorDirective.name}" must be in a type with "@node"`,
+                        []
+                    );
+                }
+            });
+            const pathToNode = getPathToNode(path, ancestors);
```
angrykoala, line 56: If this is only used for the error, consider moving this inside the if

angrykoala (changes requested):
This needs a changeset file

packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
a-alle: given the if is inside an assertion I'd keep the scopes separated. Plus it's consistent with the other validators the way it is now.

angrykoala (approved):

PR #6987, opened 2026-01-16
Allow the enabling/disabling connection queries

### LINKED ISSUE #6248: Option to disable `*Connection` operations
Currently, `read` and `aggregate` options in `query` are available to disable some operations. However, it is not possible to disable `Connection` operations whilst keeping the Entity read operations.

For example:

```graphql
type Movie @node {
    title: String!
}
```

Add an option to disable the operation `moviesConnection` from the generated queries

```diff
type Query {
  movies(limit: Int, offset: Int, sort: [MovieSort!], where: MovieWhere): [Movie!]!
-    moviesConnection(after: String, first: Int, sort: [MovieSort!], where: MovieWhere): MoviesConnection!
}
```

a-alle:
# Description

This PR allows configuring whether connection Query fields are available for types via a new `connection` argument added to the `@query` directive. This directive is available on object types, interface types and schema level, so the connection query fields can be targeted for all types in the schema easily, but also easily customisable on a case by case basis.
The default value of the `connection` argument is *`true`*, for backwards compatibility.
*Caveat*
Aggregations are no longer Query fields but fields inside the Connection query. Therefore, the new argument does impact the aggregation queries.
This implementation generates the following:
(consider defaults `@query(aggregate: false, connection: true, read: true)`)
* aggregate: true, connection: false -> connection field generated with only `aggregate` field
* aggregate: false, connection: true -> connection field generated without `aggregate` field, only `edges` field
* aggregate: false, connection: false -> connection field not generated
* read: true, connection: false -> only simple read field generated, connection field not generated
* read: false, connection: true -> connection field generated with `edges` field only
* read: false, connection: false -> no reads generated

## Complexity

Complexity: Low

# Issue

> **Note**
>
>  

Closes 

# Checklist


packages/graphql/tests/schema/issues/3541.test.ts
```diff
@@ -398,7 +398,12 @@ describe("Extending the schema in when using getSubgraphSchema", () => {
         const typeDefs = gql`
             extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@shareable"])
 
-            type Movie @query(read: false) @shareable @key(fields: "title") @key(fields: "id") @node {
+            type Movie
+                @query(read: false, connection: false)
```
angrykoala: this being an issue test, we should probably keep it as it was. The schema should not change anyway

.changeset/tiny-tips-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+"@neo4j/graphql": patch
+---
+
+Allow generation of connection query fields
```
angrykoala: Maybe this should have a bit more detail?

this implies that the connection fields were not generated before. Maybe an example of using the directive

packages/graphql/src/graphql/directives/query.ts
```diff
@@ -21,13 +21,19 @@ import { DirectiveLocation, GraphQLBoolean, GraphQLDirective, GraphQLNonNull } f
 
 export const queryDirective = new GraphQLDirective({
     name: "query",
-    description: "Instructs @neo4j/graphql to exclude read or aggregate operations from the query root type.",
+    description:
+        "Instructs @neo4j/graphql to exclude read, connection, or aggregate operations from the query root type.",
     args: {
         read: {
             description: "Disable/Enabled read operations from query root type",
```
angrykoala: Maybe we should update this description, to explicitly state that this include connection operations as well?

packages/graphql/src/schema/make-augmented-schema.ts
```diff
@@ -282,6 +282,7 @@ function makeAugmentedSchema({
             );
             const hasImplementedEntities = unionEntityAdapter.concreteEntities.length > 0;
 
+            // TODO: check why no connection query generated for unions - is this intentional?
```
angrykoala:
```suggestion
```

angrykoala (changes requested):
if read is set to false (without connections), connections must be removed. Otherwise this would be a breaking change

.changeset/tiny-tips-cross.md
```diff
@@ -0,0 +1,17 @@
+---
+"@neo4j/graphql": patch
+---
+
+Allow enabling/disabling of connection query fields on type by type basis as well as for the whole schema via a new `@query` directive argument `connection`. Default value of `connection` is the same as `read`, inheriting its default value of `true` if not provided.
+
+Examples:
```
angrykoala: A small code snippet of how to setup in graphql schema would be nice

.changeset/tiny-tips-cross.md
```diff
@@ -0,0 +1,17 @@
...
+
+- read: false -> only write operations generated
+- read: true -> all read operations generated, no aggregate field
+- read: false, connection: true -> only connection operation generated, no aggregate field
+- read: false, connection: true, aggregate: true -> only connection operation generated, with both edges and aggregate fields
+- connection: false, aggregate: true -> simple read generated, connection field generated with only aggregate field
+- read: false, aggregate: true -> connection field generated with aggregate field only
+- read: false, connection: true -> connection field generated with edges field only
+- aggregate: false, connection: false -> only simple read generated
+- read: true, connection: false -> only simple read field generated, connection field not generated
```
angrykoala: This detail is great, but probably fits better in the documentation than the changelog?

packages/graphql/src/schema-model/entity/model-adapters/InterfaceEntityAdapter.ts
```diff
@@ -104,6 +104,15 @@ export class InterfaceEntityAdapter {
...
     }
+    public isReadableFromConnectionRootQuery(schemaModel: Neo4jGraphQLSchemaModel): boolean {
+        if (this.annotations.query) {
+            if (this.annotations.query.connection !== undefined) {
+                return this.annotations.query.connection;
+            }
+            return this.annotations.query.read;
+        }
+        return schemaModel.annotations.query === undefined || schemaModel.annotations.query.connection === true;
+    }
```
angrykoala, line 115: **nitpick** Is there a way to avoid duplicating this method with the `ConcreteEntityAdapter`?

Don't worry about this if it is too much hassle

angrykoala (dismissed):
A couple of small comments, but happy to merge after those have been addressed

angrykoala (approved):

PR #7033, opened 2026-02-03
Add tests for relationship property errors raised for pop/push

### LINKED ISSUE #5766: Array methods do not raise an error on relationship properties.
**Describe the bug**
The array methodz `_PUSH`/ `_POP` do not raise an error when the array is not initialized; this happens only on relationship properties, while a proper error is raised on node properties.

**Type definitions**
```graphql
type Movie {
    title: String
    stuffs: [Int!]
    actors: Actor! @relationship(type: "ACTED_IN", direction: IN, properties: "ActedIn")
}

type ActedIn @relationshipProperties {
    stuffs: [Int!]
}

type Actor {
    name: String
    movies: Movie! @relationship(type: "ACTED_IN", direction: OUT, properties: "ActedIn")
}
```

**To Reproduce**
The below mutation does not raise an error:
```graphql
mutation MyMutation {
  updateMovies(update: { actors: { update: { edge: { stuffs_PUSH: 10 } } } }) {
    movies {
      actors {
        name
      }
    }
  }
}
```
The following will raise an error:
```graphql
mutation MyMutation {
  updateMovies(update: { stuffs_PUSH: 1 }) {
    movies {
      actors {
        name
      }
    }
  }
}
```
**Error**
```json
{
  "errors": [
    {
      "message": "Failed to invoke function `apoc.util.validatePredicate`: Caused by: java.lang.RuntimeException: Property stuffs cannot be NULL",
      "locations": [{ "line": 2, "column": 3 }],
      "path": ["updateMovies"]
    }
  ],
  "data": null
}
```

**Expected behavior**
A consistent behavior with an error that is raised, currently one mutation does not raise an error and at the same time, the property is not correctly mutated.

* OS: [ macOS]
* Version: [@neo4j/graphql@5.0.0]
* Node.js version: [22.0]

mjfwebb:
# Description

This PR adds tests showing that errors are properly raised when performing POP/PUSH operations on non-initialised relationship property arrays.

Additionally, this PR cleans up existing tests to use the latest conventions.

## Complexity

Low

# Issue

Closes #5766 

# Checklist


packages/graphql/tests/integration/mutations/update/array-methods/array-pop-and-push-errors.int.test.ts
```diff
@@ -171,4 +171,74 @@ describe("array-pop-and-push", () => {
...
+            `,
+            { id }
+        );
+
+        const gqlResult = await testHelper.executeGraphQL(query, {
+            variableValues: { id },
+        });
+
+        expect(gqlResult.errors).toBeDefined();
+        expect(
```
angrykoala: These two seem a bit redundant.

I think this can be rewritten as a single expectation, so avoid doing `.some` logic with a truthy:

```
expect(gqlResult.errors).toIncludeAllMembers([
    expect.objectContaining({
        message: expect.toInclude("stuffs cannot be NULL"),
    }),
]);
```

I'm not 100% sure which version is better though. Thoughts?

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-and-push-errors.int.test.ts
```diff
@@ -171,4 +171,74 @@ describe("array-pop-and-push", () => {
...
+
+        const query = `
+            mutation Mutation($id: ID) {
+                ${actor.operations.update}(where: { id_EQ: $id }, update: {
+                    actedIn: [
+                        {
+                            update: {
+                                edge: {
+                                    stuffs_PUSH: 10
+                                    morethings_POP: 1
```
angrykoala: This error doesn't seem to be returned / tested. Is that intentional?

As it is right now, this is only testing push errors I think, which makes the title misleading

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-and-push-errors.int.test.ts
```diff
@@ -171,4 +171,74 @@ describe("array-pop-and-push", () => {
...
+                    }
+                }
+            }
+        `;
+
+        await testHelper.executeCypher(
+            `
+                CREATE (a:${movie} {title: "The Matrix"}), (b:${actor} {id: $id, name: "Keanu"})
+                WITH a,b
+                CREATE (a)<-[actedIn: ACTED_IN { morethings: ["this", "that", "them"]}]-(b)
```
angrykoala: This Cypher is a bit wonky.

This could be done with a single `CREATE` statement. Plus we don't really need the return

```cypher
CREATE(:${movie} {title: "The Matrix"})<-[:ACTED_IN {morethings: [....]})-(:${actor} {id: $id, name: "Keanu"})
```

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-and-push-errors.int.test.ts
```diff
@@ -171,4 +171,74 @@ describe("array-pop-and-push", () => {
 
         expect(gqlResult.data).toBeUndefined();
     });
+
+    test("should throw an error when trying to pop and push from/to a non-existing array on relationship properties", async () => {
+        const movie = testHelper.createUniqueType("Movie");
```
angrykoala: **nitpick** If we name these `Movie` and `Actor`, when we put them in the cypher strings it **almost** look like idiomatic Cypher (if you squint hard enough):

```
MATCH(a:${Actor})
```

I'm aware the rest of this test file does not follow this convention, but most of the rest of the codebase does, I think it is worth it

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-errors.int.test.ts
```diff
@@ -292,4 +292,72 @@ describe("array-pop-errors", () => {
...
+
+        await testHelper.initNeo4jGraphQL({ typeDefs });
+
+        const id = generate({
+            charset: "alphabetic",
+        });
+
+        const query = /* GraphQL */ `
+            mutation Mutation($id: ID) {
+                ${actor.operations.update}(where: { id_EQ: $id }, update: {
```
angrykoala: `id_EQ` and `_POP` are deprecated, probably better to do it the new way: `id: { eq: $id}`

packages/graphql/tests/integration/mutations/update/array-methods/array-push-errors.int.test.ts
```diff
@@ -292,4 +292,72 @@ describe("array-push", () => {
 
         expect(gqlResult.data).toBeNull();
     });
+
+    test("should throw an error when trying to push to a non-existing array on relationship properties", async () => {
```
angrykoala, line 288: I'm not sure what the difference between this and the error in `pop-and-push` test is, both seem to test the same path of pushing to a non-existing array.

angrykoala (changes requested):
Thanks @mjfwebb these are good tests to add coverage to this case. A few comments on style and conventions for the testing

MacondoExpress:
Nice, This was fixed by the migration of the old reduce logic to the translation layer!

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-and-push-errors.int.test.ts
mjfwebb: Agreed. We have some cases where we do have the two expectations together

For example `packages/graphql/tests/integration/mutations/update/math.int.test.ts` contains 

```
expect(gqlResult.errors).toBeDefined();
expect(
    (gqlResult.errors as GraphQLError[]).some((el) => el.message.toLowerCase().includes(expectedError))
).toBeTruthy();
```

Could be nice to standardise this :+1: 

I'll change mine to your suggestion, and I think I'll probably make a follow up PR updating the other tests to follow the same, if you agree.

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-and-push-errors.int.test.ts
mjfwebb: Spot on. I actually wasn't sure how to test this, and I realised the error returned is only ever for the first operation which throws the error.

I meant to raise this up and ask for opinions about it. What's the intended behaviour here?

This only raises `Neo4jError: Failed to invoke function `apoc.util.validatePredicate`: Caused by: java.lang.RuntimeException: Property morethings cannot be NULL`

EDIT: I had the order wrong, morethings is the one that throws, so long as it isn't initialised in the creation cypher.

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-errors.int.test.ts
mjfwebb: Yeah, and the rest of the files. I'll update this PR's changes, but I'll also make a follow up PR for the other tests.

packages/graphql/tests/integration/mutations/update/array-methods/array-push-errors.int.test.ts
mjfwebb: Yeah, the pop-and-push one is the one in the wrong.

packages/graphql/tests/integration/mutations/update/array-methods/array-pop-errors.int.test.ts
mjfwebb: I changed my mind, I'll update it in this PR :sweat_smile:

angrykoala (approved):

PR #7316, opened 2026-06-18
fix: store Int @populatedBy callback values as Neo4j integers (#7315)

### LINKED ISSUE #7315: @populatedBy-directive stores `Int` values as `Float` inside the database
### Type definitions

```GraphQL
type Ticket {
  id: Int! @populatedBy(callback: "sequenceCallback", operations: [CREATE]) @settable(onUpdate: false)
  name: String
}
```

### Test data

```Shell
not necessary
```

### Steps to reproduce

start a server with this callback:

```javascript
const cb= async (root) => {
    return 1;
}

new Neo4jGraphQL({
    typeDefs,
    driver,
    features: {
        populatedBy: {
            callbacks: {
                sequenceCallback: cb
            }
        }
    }
})
```

execute this mutation:

```graphql
mutation {
  createTickets(input:[{
    name:"test"
  }]) {
    __typename
  }
}
```

execute this cypher in the database:

```cypher
MATCH (t:Ticket) RETURN apoc.meta.cypher.types(properties(t)) AS dataTypes
```

see the result:

```json
{
name:"STRING",
id:"FLOAT"
}
```

### What happened

The `@populatedBy` directive returns a number which is evaluated by `graphql\dist\translate\queryAST\utils\callback-bucket.js`
`callback-bucket.js` calls `cb.param.value = this.parseCallbackResult(paramValue, cb.type);` (line 34)
`parseCallbackResult()` calls `return graphql_1.GraphQLInt.parseValue(result);` (line 51) which returns a plain javascript number instead of a neo4j int object representation (`{"low","high"}`).
This value is passed to the cypher execution as cypher param and thus the value is stored as `Float` and not as `Int`.

Maybe `return graphql_1.GraphQLInt.parseValue(result);` has to be adapted to `return neo4j.int(graphql_1.GraphQLInt.parseValue(result));`?

### Expected behaviour

The returned value from the `@populatedBy` directive should be passed as neo4j int object representation to the cypher params to ensure that an `Int` is stored in the database.

### Version

7.5.3

### Database version

5.27.0 (aura)

### Relevant log output

```Shell

```

Liam-Doodson:
## Summary
Bug fix. `@populatedBy` callback values for `Int`-typed fields were persisted to Neo4j as `Float` instead of `Int`. The callback path returned a bare JS `number`, which the driver serialises as a 64-bit float, diverging from every other mutation-input path. Fixes #7315.

## Changes
- `parseCallbackResult` now wraps the `Int` case with `neo4j.int(...)`, mirroring the canonical conversion in `get-neo4j-resolve-tree.ts`. `GraphQLInt.parseValue` is retained, so 32-bit/integer validation is unchanged; only the `Int` branch is touched.
- Added `tests/integration/issues/7315.int.test.ts` asserting the **stored DB value** (via `executeCypher` + `neo4j.isInt`) — covering CREATE, UPDATE, `[Int!]` lists, and relationship properties, plus a `BigInt` regression guard. The bug was invisible to response-only assertions, so these inspect the database directly.
- Patch changeset for `@neo4j/graphql`.

## User-facing impact
`Int` fields populated by an `@populatedBy` callback are now stored as Neo4j integers (`neo4j.isInt` true) rather than floats. GraphQL responses are unchanged (the numeric value is still returned), so this is transparent at the API layer; it corrects the underlying property type in the database.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/graphql/tests/integration/issues/7315.int.test.ts
+++ b/packages/graphql/tests/integration/issues/7315.int.test.ts
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import neo4j from "neo4j-driver";
+import { TestHelper } from "../../utils/tests-helper";
+
+describe("https://github.com/neo4j/graphql/issues/7315", () => {
+    const testHelper = new TestHelper();
+
+    afterEach(async () => {
+        await testHelper.close();
+    });
+
+    test("Int @populatedBy callback value on CREATE is stored as a Neo4j integer, not a float", async () => {
+        const Movie = testHelper.createUniqueType("Movie");
+        const expected = 123456;
+
+        const callback = () => Promise.resolve(expected);
+
+        const typeDefs = /* GraphQL */ `
+            type ${Movie.name} @node {
+                id: ID
+                callback: Int! @populatedBy(operations: [CREATE], callback: "callback")
+            }
+        `;
+
+        await testHelper.initNeo4jGraphQL({
+            typeDefs,
+            features: {
+                populatedBy: {
+                    callbacks: {
+                        callback,
+                    },
+                },
+            },
+        });
+
+        const movieId = "movie_id";
+
+        const mutation = /* GraphQL */ `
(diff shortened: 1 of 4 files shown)


packages/graphql/tests/integration/issues/7315.int.test.ts
```diff
@@ -0,0 +1,382 @@
...
+            },
+        });
+
+        const movieId = "movie_id";
+
+        await testHelper.executeCypher(`CREATE (:${Movie.name} { id: $id })`, { id: movieId });
+
+        const mutation = /* GraphQL */ `
+            mutation {
+                ${Movie.operations.update}(where: { id_EQ: "${movieId}" }, update: { id_SET: "${movieId}" }) {
```
MacondoExpress: This syntax is deprecated to support a better migration path worth keeping updated:
```suggestion
                ${Movie.operations.update}(where: { id: { eq: "${movieId}" } }, update: { id: { set: "${movieId}"} }) {
```
And so for the rest of the file

MacondoExpress (dismissed):
Left a comment, but overall looks good

packages/graphql/tests/integration/issues/7315.int.test.ts
Liam-Doodson: Ah nice catch! This is the only place that deprecated syntax is used in the new tests right? 🔍

MacondoExpress (approved):

PR #7356, opened 2026-08-13
Jwt exp check on subscriptions

a-alle:
# Description

Implements exp claim check on JWTs  for subscription events

## Complexity

Complexity: Low

# Issue

> **Note**
>
>  Please link to the GitHub issue(s) in which the proposal for this work was discussed
>  
>  To link to multiple issues, use full syntax for each, for example `Closes #1, closes #2, closes #3`

Closes 

# Checklist

The following requirements should have been met (depending on the changes in the branch):

- [ ] Documentation has been updated
- [ ] TCK tests have been updated
- [ ] Integration tests have been updated
- [ ] Example applications have been updated
- [ ] New files have copyright header
- [ ] CLA (https://neo4j.com/developer/cla/) has been signed

PR #7358, opened 2026-08-14
Subscription authorization jwt exp check

a-alle:
# Description

Depends on #7356

Checking `@authentication` means calling checkAuthentication, if anything is wrong an error is thrown - fixed in 7356.
Checking `@subscriptionsAuthorization` means an _implicit_ authorization but filtering results instead of throwing error, this can only be done in the subscribe fn async iterator. We only update the auth context (aka verify the token) in the subscriptionWrapper, which wraps the resolve fn, which is too late bc subscriptionAuthorization will already have been called with the old auth token (so with a potentially expired token). This PR introduces an auth context update on every subscription event.

## Complexity

Complexity: Low

# Issue

> **Note**
>
>  

Closes 

# Checklist

### DIFF
--- a/packages/graphql/tests/e2e/subscriptions/authorization/authentication-expiry.e2e.test.ts
+++ b/packages/graphql/tests/e2e/subscriptions/authorization/authentication-expiry.e2e.test.ts
@@ -1,0 +1,405 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ */
+
+import type { Response } from "supertest";
+import supertest from "supertest";
+import { createBearerToken } from "../../../utils/create-bearer-token";
+import type { UniqueType } from "../../../utils/graphql-types";
+import { TestHelper } from "../../../utils/tests-helper";
+import type { TestGraphQLServer } from "../../setup/apollo-server";
+import { ApolloTestServer } from "../../setup/apollo-server";
+import { WebSocketTestClient } from "../../setup/ws-client";
+
+describe("Subscription authentication with expired token", () => {
+    const testHelper = new TestHelper({ cdc: true });
+    let typeMovie: UniqueType;
+    let jwtToken: string;
+    const secret = "secret";
+    let server: TestGraphQLServer;
+    let wsClient: WebSocketTestClient;
+
+    beforeAll(async () => {
+        jwtToken = createBearerToken(secret, { roles: ["admin"] });
+        await testHelper.assertCDCEnabled();
+
+        typeMovie = testHelper.createUniqueType("Movie");
+
+        const typeDefs = `
+            type ${typeMovie} @node @subscription {
+                title: String!
+            }
+
+            extend type ${typeMovie} @authentication(operations: [SUBSCRIBE, CREATE])
+            `;
+
+        const neoSchema = await testHelper.initNeo4jGraphQL({
+            typeDefs,
+            features: {
+                authorization: {
+                    key: secret,
+                },
+                subscriptions: await testHelper.getSubscriptionEngine(),
+            },
+        });
+        // eslint-disable-next-line @typescript-eslint/require-await
+        server = new ApolloTestServer(neoSchema, async ({ req }) => ({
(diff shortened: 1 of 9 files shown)


packages/graphql/tests/e2e/subscriptions/authorization/authentication-expiry.e2e.test.ts
```diff
@@ -0,0 +1,408 @@
...
+                        id: "user1",
+                    },
+                },
+            },
+        ]);
+        expect(wsClient.errors).toEqual([]);
+
+        await new Promise((resolve) => setTimeout(resolve, 11000)); // wait for 11s to expire the token
+
+        const userCreatedAfterExpiry = await createUser("user1", server);
```
Liam-Doodson, line 332: I think the behaviour here would be slightly clearer if the user has a different id

```suggestion
        const userCreatedAfterExpiry = await createUser("user2", server);
```

packages/graphql/tests/e2e/subscriptions/authorization/authentication-expiry.e2e.test.ts
```diff
@@ -0,0 +1,408 @@
...
+        expect(wsClient.events).toEqual([
+            {
+                [typeUser.operations.subscribe.created]: {
+                    [typeUser.operations.subscribe.payload.created]: {
+                        id: "user1",
+                    },
+                },
+            },
+        ]);
+        expect(wsClient.errors).toEqual([]);
```
Liam-Doodson, line 345: Not sure how easy this would be but I think the best UX would be if we return an error to the client when the jwt expires so they can react to it and re-create the subscription connection with a fresh jwt

As this is today, it's stopped returning the data but the client has no way of knowin

a-alle: after discussing this within the team the behaviour has been changed to not send any event at all if the jwt is expired (as opposed to sending an error)

packages/graphql/tests/e2e/subscriptions/authorization/authentication-expiry.e2e.test.ts
a-alle: this is on purpose as the authorization rule checks that the sub is "user1"

packages/graphql/tests/e2e/subscriptions/authorization/authentication-expiry.e2e.test.ts
```diff
@@ -0,0 +1,370 @@
...
+
+        await new Promise((resolve) => setTimeout(resolve, 2000)); // wait for 2s
+
+        const movieCreatedAfterExpiry = await createMovie("movie2", server);
+        expect(movieCreatedAfterExpiry.body.errors).toBeUndefined();
```
Liam-Doodson, line 99: This movie is also created before the expiry right? That's why it's in the events in the assert below right?

```suggestion
        const movieCreatedInTime2 = await createMovie("movie2", server);
        expect(movieCreatedInTime2.body.errors).toBeUndefined();
```

packages/graphql/tests/e2e/subscriptions/authorization/authentication-expiry.e2e.test.ts
```diff
@@ -0,0 +1,370 @@
...
+import type { Response } from "supertest";
+import supertest from "supertest";
+import { createBearerToken } from "../../../utils/create-bearer-token";
+import type { UniqueType } from "../../../utils/graphql-types";
+import { TestHelper } from "../../../utils/tests-helper";
+import type { TestGraphQLServer } from "../../setup/apollo-server";
+import { ApolloTestServer } from "../../setup/apollo-server";
+import { WebSocketTestClient } from "../../setup/ws-client";
+
+describe("Subscription authentication with expired token", () => {
```
Liam-Doodson, line 15: What would the behaviour be with these type defs if the client is created with a token that is already expired? I'm guessing the client would be rejected with an error?

Is is worth adding a test for this case?

PR #7367, opened 2026-08-18
Warnings should not appear for user's own deprecated fields which share the same type

risset:
# Description

When we validate `@relationship` directives, we check that there are no duplicate field type + relationship type + relationship direction combinations.

This should be relaxed if the duplicate relationship has the `@deprecated` annotation, since that behaviour would be more expected by users.

Example:

```gql
type Movie {
    title: String!
    actors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN) @deprecated
    newActors: [Actor!]! @relationship(type: "ACTED_IN", direction: IN)
}
```

## Complexity

Complexity: low

# Checklist

### DIFF
--- a/packages/graphql/src/schema/validation/custom-rules/directives/relationship.ts
+++ b/packages/graphql/src/schema/validation/custom-rules/directives/relationship.ts
@@ -10,6 +10,7 @@ import {
     type ObjectTypeDefinitionNode,
     type ObjectTypeExtensionNode,
 } from "graphql";
+import { DEPRECATED } from "../../../../constants";
 import {
     declareRelationshipDirective,
     relationshipDirective,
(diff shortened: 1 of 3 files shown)


MacondoExpress (approved):

packages/graphql/src/schema/validation/custom-rules/directives/relationship.ts
```diff
@@ -39,6 +40,12 @@ export function validateRelationshipDirective(context: Neo4jValidationContext):
             ).flatMap((extension) => extension.fields ?? []);
 
             [...(objectTypeDefinitionNode.fields ?? []), ...extensionsFields].forEach((field) => {
+                // don't count deprecated fields towards uniqueness
+                const isDeprecated = field.directives?.find((directive) => directive.name.value === DEPRECATED);
+                if (isDeprecated) {
+                    return;
```
MacondoExpress, line 46: I would log a warning here to give more visibility of the potential issues. I think there was some reason for this to be raised, I found https://github.com/neo4j/graphql/pull/3043  this, but I'm unsure on the whole implication. However, if the tests are passing then it looks good

risset: Ah that's some good context, reading that PR now for learning purposes :D

PR #7370, opened 2026-08-20
Validate vector provider config at schema construction

risset:
# Description

Currently, you can specify a provider in a `@vector` index that hasn't been configured in `features`, and this doesn't surface as an error until a request is made to the server.

This change ensures that when the schema is validated, missing provider configurations throw an error.
## Complexity

Complexity: Low

# Checklist

### DIFF
--- a/packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
+++ b/packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
@@ -3,7 +3,7 @@
  * Neo4j Sweden AB [http://neo4j.com]
  */
 
-import type { ASTVisitor, DirectiveNode, ObjectTypeDefinitionNode } from "graphql";
+import type { ASTVisitor, ObjectTypeDefinitionNode } from "graphql";
 import { vectorDirective } from "../../../../graphql/directives/vector";
 import type { VectorField } from "../../../../schema-model/annotation/VectorAnnotation";
 import { parseValueNode } from "../../../../schema-model/parser/parse-value-node";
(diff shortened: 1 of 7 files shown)


packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
```diff
@@ -58,12 +65,31 @@ export function validateVectorDirective(context: Neo4jValidationContext): ASTVis
     };
 }
 
-function assertMaxPhraseLengthIsValid(vectorDirectiveOnNode: DirectiveNode): void {
-    const indexesArg = vectorDirectiveOnNode.arguments?.find((argument) => argument.name.value === "indexes");
-    if (!indexesArg) {
-        return;
+// When a provider is specified by an index, ensure that it has a valid configuration
+function assertIndexProviderIsValid(indexes: VectorField[], vectors: Neo4jVectorSettings | undefined): void {
```
Liam-Doodson: I'm not sure the name `vectors` is super accurate. If I've read this correctly, this is actually the embedding provider settings right so something like `providerSettings` would be more accurate

I'm guessing you've gone with this naming based on the type name which I guess is also not super accurate?

Liam-Doodson: I think this comment on the naming probably applied to a few different places in this change

Liam-Doodson (commented):
Just added a bit of a nitpicky comment on the naming. I've also not approved as I think @a-alle or @MacondoExpress can probably give a better review than me on this repo

packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
```diff
@@ -41,7 +42,13 @@ export function validateVectorDirective(context: Neo4jValidationContext): ASTVis
                     );
                 }
                 for (const vectorDirectiveOnNode of vectorDirectivesOnNode) {
-                    assertMaxPhraseLengthIsValid(vectorDirectiveOnNode);
+                    const { indexes } = parseArguments<{ indexes: VectorField[] }>(
```
MacondoExpress, line 45: I would remap the `indexes` variable to a variable name such as `indexesArgument` to slightly provide more context to the reader

packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
MacondoExpress: +1 I would rename the `vectors` argument to `vectorSettings`

MacondoExpress (dismissed):
LGTM, 
I just raise the visibility that this "could" break "working" existing clients.
Clients may have working type definition with an unused invalid `@vector` directive. 
I'm more than happy with merging this, but worth have a conscious decision around this

Liam-Doodson:
I think it's an acceptable risk as it's at build time so it should be possible for users to spot and fix the invalid config before they deploy the change. Also just because they haven't hit the error before doesn't mean they won't in the future

packages/graphql/src/schema/validation/custom-rules/directives/vector.ts
risset: Agreed on the naming

MacondoExpress (approved):

PR #7386, opened 2026-09-15
Correlate ReleaseOps runs with invocation databases

bfeshti:
Accept a caller correlation identifier and invocation database ID for ReleaseOps Aura runs. Use the exact correlation as the run name and construct the target URI within devreleaseops, while preserving existing secret-based defaults for legacy calls.

Disable cancellation of running tests and serialize work per target database.


bfeshti:
@MacondoExpress did you have the chance to review this?

docs/releaseops/invocations.md
```diff
@@ -0,0 +1,15 @@
+# ReleaseOps invocations
```
Liam-Doodson: I think we're going to move the docs for this library back into this repo and a `docs` directory will probably be used for that so I think it would be better to just keep these docs with the workflow itself. Maybe just as an inline comment seeing as it's only a small description?

.github/tests/test_releaseops_inputs.py
```diff
@@ -0,0 +1,36 @@
+import os
```
Liam-Doodson: What's this used for? From what I can tell it's not called anywhere?

bfeshti: I added it for the input validation locally, including it here was not intentional, I will remove it

